PR #7, opened 2018-02-13
Feature/jenkins check copyright

r-englund:
Added a stage to the Jenkins pipeline that checks all copyright headers and fails if the year is wrong for some file. 
Based on the https://github.com/inviwo/inviwo/blob/master/tools/refactoring/update-copyright.py script

### DIFF
--- a/tools/refactoring/check-copyright.py
+++ b/tools/refactoring/check-copyright.py
@@ -2,8 +2,6 @@ import sys
 import os
 import re
 import codecs
-import colorama
-colorama.init()
 
 import refactoring # Note: refactoring.py need to be in the current working directory
 
--- a/tools/refactoring/check-copyright.py
+++ b/tools/refactoring/check-copyright.py
@@ -1,0 +1,70 @@
+import sys 
+import os
+import re
+import codecs
+import colorama
+colorama.init()
+
+import refactoring # Note: refactoring.py need to be in the current working directory
+
+try:
+	import colorama
+	colorama.init()
+	
+	def print_error(mess, **kwargs):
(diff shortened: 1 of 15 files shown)


tools/refactoring/check-copyright.py
```diff
@@ -0,0 +1,68 @@
...
+
+files = refactoring.find_files(paths, ['*'], excludes=excludespatterns)
+
+def replace(pattern, replacement) :
+	matches = refactoring.find_matches(files, pattern)
+	print(matches.size());
+
+
+copyright_regex = [
+	re.compile(r"(\s*[*#]\s+Copyright \(c\) 201\d-)201[1234567]( Inviwo Foundation\s*)") ,
```
ResearchDaniel, line 39: Is it possible to get the current year so that we do not  need to update the script each year?

ResearchDaniel (approved):
Does it check the whole repo or just committed code?
Easy to make it check just committed code in that case?
Possible to use current year to not need to update script every year?

r-englund:
> Possible to use current year to not need to update script every year? 

I thought about that, it would be easy to implement that. But that would mean every branch would fail in the begging of a new year. This script should be update the same time as the copy right year is updated. 

> Does it check the whole repo or just committed code?
> Easy to make it check just committed code in that case?

It checks the whole repo, it might be tricky to check only files that changed. Changed since when? Last commit, last common ancestor with master (how to handle master). It takes 6s to run, and since everything is correct at the moment, it can only report errors for new or modified files

tools/refactoring/check-copyright.py
r-englund: See comment on PR

Issue #14, opened 2018-02-15
utilgl::activateAndClearTarget does not set blend mode


martinfalk:
The blend mode is not set in `utilgl::activateAndClearTarget`.

This example affects the copy operation from the image inport due to alpha blending. That is, passing an image through this processor will affect the appearance of the input image. Copying, however, should "copy" the data without any blending.
```c++
void CameraWidget::process() {
    utilgl::BlendModeState blending(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA);

    // combine the previously rendered widget image with the input
    if (inport_.isConnected()) {
        utilgl::activateTargetAndCopySource(outport_, inport_);
    } else {
        utilgl::activateAndClearTarget(outport_, ImageType::ColorDepthPicking);
    }
   ...
```

petersteneteg:
I think we overuse these funtions, would it ot be faster in many cases to do the copy in the regular shaderpass? the nice thing about this functions is that it takes care of copying all layers etc, but would it not be possible to write some nice shader util functions that on can use instead and then avoid this extra shader pass?

r-englund:
Depends on the what you render. The fragment shader is only called for pixels that are being renderered. It wouldnt work for meshes, but for places where we draw a full screen quad that might work.  
We would still need to implement our own blending functions then, which might be slower than the OpenGL built in ones (hardware optimizations). Though might still be faster than the copy we do now.

petersteneteg:
Yes, this would only be relevant for the full quad of course..

martinfalk:
@petersteneteg I think the reason for the overuse is that people tend to look at existing processors, like the background processor, and see it implemented like that.
One problem copying the layers in the shaders will be the handling of no input vs. input since reading from an unbound texture is undefined.

r-englund:
From meeting: 
* Keep current way of binding and copying input image. Then we have the same way of binding and copying for both geometry and full-screen quad renderings
* Add assertion in the `activateTargetAndCopySource` to check if we have correct blending mode. (this will only check when running in debug or when forcing assertions)

ResearchDaniel:
@martinfalk still an issue?

martinfalk:
Neither `utilgl::activateTargetAndCopySource()` nor `ImageGL::copyRepresentationsTo()` (used by copy source) check the blend state.
[...]

PR #15, opened 2018-02-15
UserInterfaceGL: CameraWidget blend mode fix, #14

martinfalk:
When using the camera widget, the input image was modified when being copied to the outport due to alpha blending.


ResearchDaniel (approved):
Should we add an assert in activateTargetAndCopySource to make sure that no one does it by mistake?  (another PR)

martinfalk:
@ResearchDaniel I am not sure, because then we would need to query the current state of the blend function which seems to cause more overhead than a simple assert statement.

Suggest to move the discussion to #14

PR #19, opened 2018-02-22
Feedback: push docs to gh pages

r-englund:
**This PR is not for merging, its for feedback/discussion**

I did some experimenting with getting the master branch pipeline build to automatically update the doxygen pages on out github pages. 

The changes in the jenkinsfile adds a stage when building on specific branch. The stage  checksout the gh-pages, copied the generated doxygen files, add, commits and puhses to the gh-pages branch. 

This means we have would have the doxygen doc on out github pages that are allways up to date with master.   In addition, I have added a landing page that links to the latest docs and the docs for releases. (https://inviwo.github.io/inviwo/)

What we need to do before this could be merged is: 
* FIx a jenkins github user, now it looks like peter is pushing :) 
* Change  `if ( "${env.BRANCH_NAME}" == "feature/push-docs-to-gh-pages") {` to use master

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -176,7 +176,7 @@ node {
             // Mark as unstable, if we mark as failed, the report will not be published.
             currentBuild.result = 'UNSTABLE'
         }
-        stage('Publish') {
+        stage('Publish Regression Report') {
             publishHTML([
                 allowMissing: true,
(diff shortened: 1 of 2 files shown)


r-englund:
Moved to issue #85

PR #22, opened 2018-02-22
Doxygen: Added a separate target for generating previews

r-englund:
In the PR I have move the the generating of processor previews from DOXY-inviwo to its own target. 
This way, one can modify the doc in core and then "quickly" generate the html. Before, Doxy-inviwo had an dependency on the inviwo.exe project which made it impossible to build DOXY-inviwo whithout building the whole source.

### DIFF
--- a/tools/doxygen/doxygen.cmake
+++ b/tools/doxygen/doxygen.cmake
@@ -423,7 +423,7 @@ function(make_doxygen_target modules_var)
     add_custom_target("DOXY-generate-processor-previews"
         COMMAND inviwo --save-previews "${output_dir}/inviwo/html" --quit
         WORKING_DIRECTORY ${output_dir}
-        COMMENT "Generating Inviwo API documentation with Doxygen"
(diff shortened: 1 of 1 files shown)


tools/doxygen/doxygen.cmake
```diff
@@ -419,7 +419,19 @@ function(make_doxygen_target modules_var)
     set_target_properties("DOXY-Inviwo" PROPERTIES FOLDER "doc" EXCLUDE_FROM_ALL TRUE)
     add_dependencies("DOXY-ALL" "DOXY-Inviwo")
 
-    # Help, used for the help inside invowo
+
+    add_custom_target("DOXY-generate-processor-previews"
+        COMMAND inviwo --save-previews "${output_dir}/inviwo/html" --quit
+        WORKING_DIRECTORY ${output_dir}
+        COMMENT "Generating Inviwo API documentation with Doxygen"
```
ResearchDaniel: Comment should say
Generate preview images of processors to be used in Inviwo Doxygen API documentation.

ResearchDaniel (approved):
Just a minor comment

PR #23, opened 2018-02-22
Doxygen: Added option to add a custom build command to Doxy-inviwo that will automatically open the generated doxygen html

r-englund:
This is for convenience. Might need to be tested on Unix and Mac. 
Option is default to off so this is an opt in feature

### DIFF
--- a/tools/doxygen/doxygen.cmake
+++ b/tools/doxygen/doxygen.cmake
@@ -422,9 +422,18 @@ function(make_doxygen_target modules_var)
 
 
     if(${IVW_DOXYGEN_OPEN_HTML_AFTER_BUILD})
+        if(WIN32)
+            set(OPEN_COMMAND "start")
+        elseif(APPLE)
+            set(OPEN_COMMAND "open")
+        else()
+            set(OPEN_COMMAND "xdg-open")
+        endif()
+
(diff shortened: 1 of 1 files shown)


ResearchDaniel (approved):
Looks like it should work on all platforms as long as there is a default html-open executable

r-englund:
> Looks like it should work on all platforms as long as there is a default html-open executable 

I thought so as well. I know it works like that on windows, but not sure if it works like that on mac or linux.

ResearchDaniel:
https://tombevers.me/2013/02/03/opening-a-browser-or-files-from-the-terminal/

https://stackoverflow.com/questions/10006958/open-an-html-file-with-default-browser-using-bash-on-mac

martinfalk:
I think that would be `start` (Windows) and `open` or `xdg-open` (Linux / MacOS).

r-englund:
changing to
```
if(${IVW_DOXYGEN_OPEN_HTML_AFTER_BUILD})
        add_custom_command(TARGET DOXY-Inviwo 
            POST_BUILD
            COMMAND start 
            ARGS "${output_dir}/inviwo/html/index.html"
        )
    endif()
```
works on windows

martinfalk (approved):

PR #24, opened 2018-02-22
For Discussion: Move Wiki to doxygen

r-englund:
This Pull Request is for feedback/discusion rather than for merging. Making a PR instead of Issue to make it easier to look at code and results at the same time. 

The main idea is to move (and improve) the content of the wiki into the doxygen.
See http://130.236.145.152:8080/job/inviwo-pub/job/feature%252Fdoxy%252Freplace-wiki/Doxygen_Documentation/ for how it could look 

**Pros**
* All documentation in one place
* Easier to keep documentation update in sync with master ( When using wiki, the edits are public directly and when a PR impacts the documentation the master branch and the wiki will be out of sync until the PR is merged, or the opposite) 
* Easier management of documentation for different releases since the documentation can be included with the release which is not really possible with the wiki.   

**Cons**
* Wiki has a closer edit to result cycle (you can edit directly online with preview). Though: see footnote A

**Note: **
The current text in the .md files is not finished so don't  focus on that, more of the idea of using doxygen rather then the wiki. 

**Footnote A**
It will still be possible to edit the documentation online. But there are a few doxygen specfic "tag" in the .md files. See for example https://github.com/inviwo/inviwo/blob/feature/doxy/replace-wiki/docs/manual/manual.md


martinfalk:
One thing you forgot to mention is that doxygen parses markdown syntax. Meaning almost no changes need to be made and the `.md` files are reused and still usable.

I think that is a good idea.

r-englund:
moved to issue #86

PR #26, opened 2018-02-22
Core/Qt: Added python editor to property semantics

r-englund:


ResearchDaniel (approved):
Note for future PR: Make it possible to add semantics in a module so that we do not need to put them in core

PR #28, opened 2018-02-23
PythonQT: Remember last used code between sessions

r-englund:


modules/python3qt/pythoneditorwidget.cpp
```diff
@@ -198,9 +198,16 @@ PythonEditorWidget::PythonEditorWidget(QWidget* parent, InviwoApplication* app)
         settings.beginGroup(objectName());
         QString lastFile = settings.value("lastScript", "").toString();
         if (!lastFile.isEmpty()) loadFile(lastFile.toLocal8Bit().constData(), false);
```
martinfalk: should use block `{}` here for if statement

modules/python3qt/pythoneditorwidget.cpp
```diff
@@ -198,9 +198,16 @@ PythonEditorWidget::PythonEditorWidget(QWidget* parent, InviwoApplication* app)
...
-        
+        else {
+            QString src = settings.value("source", "").toString();
+            if(src.length()!=0){
+                pythonCode_->setPlainText(src);
+                script_.setSource(utilqt::fromQString(src)); // or use localQString?
```
martinfalk: what is the reason for the comment?

modules/python3qt/pythoneditorwidget.cpp
```diff
@@ -225,6 +232,7 @@ void PythonEditorWidget::closeEvent(QCloseEvent* event) {
     settings.beginGroup(objectName());
     settings.setValue("appendLog", appendLog_->isChecked());
     settings.setValue("lastScript", scriptFileName_.c_str());
+    settings.setValue("source", pythonCode_->toPlainText());
```
martinfalk: will this be the code edited last in the editor?

martinfalk (changes requested):

modules/python3qt/pythoneditorwidget.cpp
r-englund: Comment is because I forgot to remove it. Not sure which to use.

modules/python3qt/pythoneditorwidget.cpp
r-englund: This will save the source that currently are in the editor when the editor closes. 

Maybe we should also store the code upon run (prior to run). My though with the change is to recover code lost after either a crash or when you forget to save the code.

martinfalk: then prior run would be a good idea as well.
But: what happens if you editing a file, edit the script in Inviwo, run it (it will be stored), and then it crashes.
Upon a restart, the file will be loaded anyway.

modules/python3qt/pythoneditorwidget.cpp
martinfalk: preferably `utilqt::fromQString()` and `utilqt::toQString()`

martinfalk (approved):

r-englund:
@martinfalk: I have reviewed your changes and tried it on my machine, everything works/looks fine. (can't approve my own PR) 
I've added such that the code is remembered also after a run, see f1e63b4b5e1cf2f197dc212ec633160648dea801 (and ca32e464046b1e3145ef50e24abcbbdb8a0356d4 for code formatting)

PR #35, opened 2018-03-09
Base: VolumeSlice: Swizzeling fix (and code formating)

r-englund:


modules/base/processors/volumeslice.cpp
```diff
@@ -168,15 +165,19 @@ void VolumeSlice::process() {
                 auto layerdata = layerrep->getDataTyped();
 
                 switch (util::extent<T, 0>::value) {
```
r-englund, line 167: extent<char>::value (and other non vec types) return zero wile for vecX it return X as expected: Hence ppl expect it to return 1 for scalars

martinfalk (approved):
good catch

PR #38, opened 2018-03-13
Feature/decorated mesh

r-englund:
New interface for mesh using variadic templates, can replace a Basic/Simple mesh with almost no code changes. (In many cases no changes at all). 
Basic usage: 
```c++
using SimpleMesh = DecoratedMesh<buffertraits::PositionsBuffer, buffertraits::TexcoordBuffer, buffertraits::ColorsBuffer>;
SimpleMesh mesh;
mesh.addVertex(vec3(0.0f), vec3(0.0f), vec4(1,0,0,1) );
mesh.addVertex(vec3(1.0f), vec3(1.0f), vec4(0,1,0,1) );
```
OR
```c++
using ColoredMesh = DecoratedMesh<buffertraits::PositionsBuffer, buffertraits::ColorsBuffer>;
ColoredMesh mesh;
mesh.addVertex(vec3(0.0f), vec4(1,0,0,1) );
mesh.addVertex(vec3(1.0f), vec4(0,1,0,1) );
```

In old BasicMesh one could do the following to get a specific buffer: 
```c++
auto vertices = mesh->getVertecies();
auto colors  =mesh->getColors();
```

In the new mesh one will do: 
```c++
auto vertices = mesh->getTypedBuffer<buffertraits::PositionsBuffer>();
auto colors = mesh->getTypedBuffer<buffertraits::ColorsBuffer>()
```

In old Basic Mesh one could set a specific vertex value by `mesh->setVertexNormal(size_t, vec3)`, that would be `mesh->setVertex<buffertraits::NormalBuffer>(size_t, vec3)` in DecoratedMesh. 

More details on how to use the new mesh can be found at http://weber.itn.liu.se/~ricen76/decorated_mesh/group__decoratedmesh.html  (I host the doxygen myself for now since the jenkins machine is not reachable from outside the university ATM)


r-englund:
Note: There are some things left to clean up.
TODO:
- [ ] Move implementation and declaration of class functions 
- [ ] Document class methods 
- [ ] Naming (Is DecoratedMesh a good name) 

Reason why I haven't done that yet is that a want feedback before. So I don't do to much in vain.

petersteneteg:
You could do some conditional inheritance and crtp to add getColors getVertices etc. So you can get the same interface as for the old simple and basic mesh. So we can remove the old ones...

Maybe TypedMesh is more descriptive?

include/inviwo/core/datastructures/geometry/decoratedmesh.h
```diff
@@ -0,0 +1,432 @@
...
+    type t;
+    operator type() const { return t; }
+#endif
+};
+
+/**
+ * \ingroup decoratedmesh
+ * BufferTrait for Position buffers (glm::vec3)
+ */
+using PositionsBuffer = BufferTrait<float, 3, static_cast<int>(BufferType::PositionAttrib)>;
```
petersteneteg: "Buffer" seems redundant here...

include/inviwo/core/datastructures/geometry/decoratedmesh.h
```diff
@@ -0,0 +1,432 @@
...
+ * \see decoratedmesh
+ */
+template <typename T, unsigned DIM, int attrib, int location = attrib>
+struct BufferTrait {
+    using type = Vector<DIM, T>;
+    static inviwo::Mesh::BufferInfo bi() {
+        return {static_cast<inviwo::BufferType>(attrib), location};
+    }
+
+#if _MSC_VER <= 1900
```
petersteneteg: What is it that does not work in vs2015

r-englund: Should document this. But: 

I could not get the code for addVertex([custom parameters]) to compile. The following code in various variation refused to compile. I do not remember the exact error messages. 
```
template <typename T>
using TypeAlias = typename T::type;
...
void addVertex(TypeAlias<BufferTraits>... args) { addVertexImpl<0>(args...); }
```

include/inviwo/core/datastructures/geometry/decoratedmesh.h
r-englund: True, it was not in a namespace from the beginning

include/inviwo/core/datastructures/geometry/decoratedmesh.h
```diff
@@ -0,0 +1,432 @@
...
+ * \ingroup decoratedmesh
+ * BufferTrait for Position buffers (glm::vec3)
+ */
+using PositionsBuffer = BufferTrait<float, 3, static_cast<int>(BufferType::PositionAttrib)>;
+
+/**
+ * \ingroup decoratedmesh
+ * BufferTrait for Normal buffers (glm::vec3)
+ */
+using NormalBuffer = BufferTrait<float, 3, static_cast<int>(BufferType::NormalAttrib)>;
```
ResearchDaniel: Some are plural (PositionsBuffer, ColorsBuffer) and some are singular (NormalBuffer). We should stick to one. Suggest singular as Buffer could imply many anyway.

include/inviwo/core/datastructures/geometry/decoratedmesh.h
```diff
@@ -0,0 +1,432 @@
...
+    void copyConstrHelper() {}  // sink
+
+    template <unsigned I, typename T, typename... ARGS>
+    void copyConstrHelper() {
+        std::get<I>(buffers_).buffer_ =
+            std::static_pointer_cast<Buffer<typename T::type>>(Mesh::buffers_[I].second);
+        copyConstrHelper<I + 1, ARGS...>();
+    }
+
+private:
```
ResearchDaniel: second private declaration

include/inviwo/core/datastructures/geometry/decoratedmesh.h
```diff
@@ -0,0 +1,432 @@
...
+    using BufferTuple = std::tuple<BufferHolder<BufferTraits>...>;
+    BufferTuple buffers_;
+};
+
+/**
+ * \ingroup decoratedmesh
+ * Type definition of a DecoratedMesh useful for Spheres, consists of a vec3-buffer for position, a
+ * float-buffer for radii and vec4 for colors.
+ */
+using SphereMesh = DecoratedMesh<buffertraits::PositionsBuffer, buffertraits::RadiiBuffer,
```
ResearchDaniel: ColoredSphereMesh?

include/inviwo/core/datastructures/geometry/decoratedmesh.h
```diff
@@ -0,0 +1,432 @@
...
+    }
+
+    template <typename BT>
+    auto getTypedRAMRepresentation() {
+        return getTypedRAMRepresentation<
+            buffertraits::detail::get_index<BT, BufferTraits...>::value>();
+    }
+
+    template <unsigned I>
+    auto &getTypedDataContainer() {
```
ResearchDaniel: Should be getTypedEditableDataContainer since it uses getTypedEditableRAMRepresentation
Could alsoe be called getEditableTypedBufferContainer

include/inviwo/core/datastructures/geometry/decoratedmesh.h
```diff
@@ -0,0 +1,432 @@
...
+
+    IndexBufferRAM *addIndexBuffer(DrawType dt, ConnectivityType ct) {
+        auto indicesRam = std::make_shared<IndexBufferRAM>();
+        auto indices = std::make_shared<IndexBuffer>(indicesRam);
+        addIndicies(Mesh::MeshInfo(dt, ct), indices);
+        return indicesRam.get();
+    }
+
+    template <unsigned I>
+    auto getTypedBuffer() {
```
ResearchDaniel: Should it not return const auto&

ResearchDaniel (commented):
Sweet!

include/inviwo/core/datastructures/geometry/decoratedmesh.h
```diff
@@ -92,37 +91,92 @@ struct BufferTrait {
     type t;
     operator type() const { return t; }
 #endif
+
+    BufferTrait() {}
+    virtual ~BufferTrait() = default;
+
+    template <typename Mesh>
```
petersteneteg: No need for the template here..

include/inviwo/core/datastructures/geometry/decoratedmesh.h
```diff
@@ -208,8 +269,11 @@ using RadiiBuffer = BufferTrait<float, 1, static_cast<int>(BufferType::NumberOfB
  * \snippet modules/base/algorithm/meshutils.cpp Using Colored Mesh
  *
  */
+
+// DecoratedMesh<buffertraits::IndexBuffer,buffertraits::PositionBuffer> test;
+
 template <typename... BufferTraits>
-class DecoratedMesh : public Mesh {
+class DecoratedMesh : public Mesh, public BufferTraits... {
```
petersteneteg: Should probably assert that all BufferTraits are are derived from BufferTraits..

include/inviwo/core/datastructures/geometry/decoratedmesh.h
r-englund: For some reason it did not work without it on my machine (MSVC 17).

include/inviwo/core/datastructures/geometry/decoratedmesh.h
petersteneteg: I.e. ststic_assert

include/inviwo/core/datastructures/geometry/decoratedmesh.h
petersteneteg: It should, what error do you get?

r-englund: No error, it compiles fine. Problem is that neither of the constructors are called, the buffer_ pointer is  empty.

petersteneteg: Try casting *this to a Mesh

petersteneteg: When you call the constructor

r-englund: that worked

include/inviwo/core/datastructures/geometry/typedmesh.h
```diff
@@ -269,63 +291,82 @@ class RadiiBuffer
...
+        template <unsigned I, typename T2, typename... ARGS>
+        void set(T2 &t, ARGS... args) {
             std::get<I>(values) = t;
-            set<I+1>(args...);
+            set<I + 1>(args...);
         }
 
-        template<unsigned I,typename T2> void set(T2 &t){
+        template <unsigned I, typename T2>
+        void set(T2 &t) {
```
petersteneteg: Think values = VertexTuple{args...} should work...

petersteneteg:
One big problem with the Mesh data structure, is that we cant enforce the invariant that all buffer must be the same size. 
One ideally we would not like to expose the underlying containers as modifiable, i.e. it should not be possible to increase the size of only one buffer. But lor legacy reasons that might be hard to do..

But we could try and add iterator acces to the data instead of giving acces to the containers, and use that instead so we could at some point mabe depricate the direct container access?

Could use the zip iterator to give access to a row a a time for example.

petersteneteg:
Should also document that the user is responsible for making sure all buffers are the same size...

ResearchDaniel:
As long as the indices in the index buffer does not refer to any out of bounds you would also be ok. Agree that it is not intuitive if the buffers are if different sizes though.

petersteneteg:
@daniel true but even harder to maintain as an invariant...

martinfalk:
@petersteneteg regarding gcc version: thing is that we are using Ubuntu with long-term support at the moment... from 2016

r-englund:
> regarding gcc version: thing is that we are using Ubuntu with long-term support at the moment... from 2016 

Also, Jenkins should run gcc version that we state we are supporting.... Which is none when I now look for it.  Though I would argue that we should support at least the latest version off gcc available through apt-get in the latest LTS of Ubuntu.

r-englund:
Which is 5.4.0 https://packages.ubuntu.com/xenial/devel/gcc-5

martinfalk:
Recent gcc releases:
GCC 7.3 released [2018-01-25]
GCC 5.5 released [2017-10-10]
GCC 7.2 released [2017-08-14]
GCC 6.4 released [2017-07-04]

Could be of interest:  [install latest gcc on ubuntu](https://gist.github.com/application2000/73fd6f4bf1be6600a2cf9f56315a2d91)

petersteneteg:
Bionic Beaver (18.04 LTS) should be release quite soon i.e. April. Would be nice to get on the 6.* GCC branch at least, 5.* has some really annoying bugs...

include/inviwo/core/datastructures/geometry/basicmesh.h
```diff
@@ -29,81 +29,5 @@
...
-
-    std::shared_ptr<Buffer<vec3>> vertices_;
-    std::shared_ptr<Buffer<vec3>> texCoords_;
-    std::shared_ptr<Buffer<vec4>> colors_;
-    std::shared_ptr<Buffer<vec3>> normals_;
-};
-
-}  // namespace
-
+#include <inviwo/core/datastructures/geometry/typedmesh.h>
```
martinfalk, line 32: Is the idea to keep `BasicMesh` in the future?

martinfalk: Otherwise we should come up with a way to deprecate it.

include/inviwo/core/datastructures/geometry/typedmesh.h
```diff
@@ -0,0 +1,489 @@
...
+    : std::integral_constant<std::size_t, 1 + get_index<T, Ts...>::value> {};
+
+}  // namespace detail
+
+/**
+ * \ingroup typedmesh
+ * Struct used to tell TypedMesh what buffers to create.
+ * The first two template parameters, 'typename T, unsigned DIM' is used rather than
+ * ���Vector<DIM,T>��� or glm::TvecX due to compiler issues related to the clone function.
+ * Similarly, The template parameter ���int attrib��� needs can't be a strongly typed enum
```
martinfalk: not that easy to understand. Rephrase?

include/inviwo/core/datastructures/geometry/typedmesh.h
```diff
@@ -0,0 +1,489 @@
...
+ * (inviwo::BufferType) and hence it is an int.
+ *
+ * \see TypedMesh
+ * \see typedmesh
```
martinfalk: why both?

include/inviwo/core/datastructures/geometry/typedmesh.h
```diff
@@ -0,0 +1,489 @@
...
+template <typename T, unsigned DIM, int attrib, int location = attrib>
+class BufferTrait {
+public:
+    using type = Vector<DIM, T>;
+    static inviwo::Mesh::BufferInfo bi() {
+        return {static_cast<inviwo::BufferType>(attrib), location};
+    }
+
+#if _MSC_VER <= 1900
+    BufferTrait(type t) : t(t) {}
```
martinfalk, line 92: maybe add comment why this is necessary?

include/inviwo/core/datastructures/geometry/typedmesh.h
```diff
@@ -0,0 +1,489 @@
...
+    }
+};
+
+/**
+ * \ingroup typedmesh
+ * BufferTrait for Texture Coordinate buffers (glm::vec3)
+ */
+class TexcoordBuffer : public BufferTrait<float, 3, static_cast<int>(BufferType::TexcoordAttrib)> {
+public:
+    using Base = BufferTrait<float, 3, static_cast<int>(BufferType::TexcoordAttrib)>;
```
martinfalk, line 170: would it be possible to make the number of components a template argument? I don't see a point in always using vec3 for tex coords.

include/inviwo/core/datastructures/geometry/typedmesh.h
```diff
@@ -0,0 +1,489 @@
...
+class RadiiBuffer
+    : public BufferTrait<float, 1, static_cast<int>(BufferType::NumberOfBufferTypes)> {
+public:
+    using Base = BufferTrait<float, 1, static_cast<int>(BufferType::NumberOfBufferTypes)>;
+    using Base::Base;
+
+    std::shared_ptr<const Buffer<type>> getRadii() const { return Base::buffer_; }
+    std::shared_ptr<Buffer<type>> getEditableRadii() { return Base::buffer_; }
+
+    void setVertexRadius(size_t index, float radius){
```
martinfalk, line 219: rename to `setRadius` to be consistent

include/inviwo/core/datastructures/geometry/typedmesh.h
```diff
@@ -0,0 +1,489 @@
...
+ * # Simple Usage
+ *
+ * \code{.cpp}
+ * using MyMesh = TypedMesh<buffertraits::PositionsBuffer,buffertraits::ColorsBuffer>;
+ * MyMesh mesh;
+ * mesh.addVertex(vec3(0.0f), vec4(1,0,0,1) );
+ * mesh.addVertex(vec3(1.0f), vec4(0,1,0,1) );
+ * \endcode
+ *
+ * If texture coordinates is also needed for each vertex in the mesh then one could instead use:
```
martinfalk, line 246: _are_ also needed

include/inviwo/core/datastructures/geometry/typedmesh.h
```diff
@@ -0,0 +1,489 @@
...
+ *
+ * \code{.cpp}
+ * using MyMesh = TypedMesh<buffertraits::PositionsBuffer, buffertraits::TexcoordBuffer,
+ * buffertraits::ColorsBuffer>; MyMesh mesh; mesh.addVertex(vec3(0.0f), vec3(0.0f), vec4(1,0,0,1) );
+ * mesh.addVertex(vec3(1.0f), vec3(1.0f), vec4(0,1,0,1) );
+ * \endcode
+ *
+ * For meshes when more than a few vertices is added is more efficient to add all vertices to the
```
martinfalk, line 254: For meshes with more than a couple of vertices, it is more efficient to add all vertices to the ...

include/inviwo/core/datastructures/geometry/typedmesh.h
```diff
@@ -0,0 +1,489 @@
...
+ * using MyMesh = TypedMesh<PositionsBufferTrait,ColorsBufferTrait>;
+ * std::vector<MyMesh::Vertex> vertices;
+ * vertices.emplace_back(vec3(0.0f), vec4(1,0,0,1));
+ * vertices.emplace_back(vec3(1.0f), vec4(0,1,0,1));
+ * MyMesh mesh;
+ * mesh.addVertices(vertices);
+ * \endcode
+ *
+ * When creating meshes it is very common to also have Index buffers, in addition to the vertex
+ * buffers. To add a index buffer to the mesh you can use the function addIndexBuffer as demonstrate
```
martinfalk, line 268: demonstrated

include/inviwo/core/datastructures/geometry/typedmesh.h
```diff
@@ -0,0 +1,489 @@
...
+ *
+ * ## Creating camera frustum
+ * The following code snippet is another example where we create a camera frustum mesh for a given
+ * camera. It is the code used in meshutil::cameraFrustum
+ *
+ * \snippet modules/base/algorithm/meshutils.cpp Using Colored Mesh
+ *
+ */
+
+// TypedMesh<buffertraits::IndexBuffer,buffertraits::PositionBuffer> test;
```
martinfalk, line 296: remove?

include/inviwo/core/datastructures/geometry/typedmesh.h
```diff
@@ -0,0 +1,489 @@
...
+ * Type definition of a TypedMesh having positions(vec3), normals(vec3), texture
+ * coordinates(vec3) and colors(vec4). Example usage:
+ */
+using BasicMesh = TypedMesh<buffertraits::PositionsBuffer, buffertraits::NormalBuffer,
+                                buffertraits::TexcoordBuffer, buffertraits::ColorsBuffer>;
+
+
+
+
+using NewSimpleMesh = TypedMesh<buffertraits::PositionsBuffer,
```
martinfalk, line 485: `NewSimpleMesh` :)

martinfalk: not entirely sure whether this is a good name though. In particular since it will be used in a lot of places...

martinfalk (approved):
Nice work! Some minor comments

r-englund:
There is some documentation work that needs to be done before merging this. Closing the PR for now and reopen it when it is ready to be merged

include/inviwo/core/datastructures/geometry/typedmesh.h
r-englund: The first is the class the second is the documentation group

include/inviwo/core/datastructures/geometry/typedmesh.h
r-englund: done locally, will be pushed next commit

include/inviwo/core/datastructures/geometry/typedmesh.h
r-englund: Consitent with what? The other BufferBase-classes has setVertexPosition or setVertexNormal etc. That is to mimic the interface of basic mesh.

include/inviwo/core/datastructures/geometry/typedmesh.h
r-englund: This was for testing

include/inviwo/core/datastructures/geometry/typedmesh.h
martinfalk: OK, I see.

PR #43, opened 2018-03-20
PyBind fixes

PetiteViking:
Missing typename specifier caused compile error with GCC. Also, refactored typo in getDataFormat.


JochenJankowai:
Closing, got some things mixed up here

PR #47, opened 2018-03-20
BaseGL: Added spherical gradient background as option to Background processor

ResearchDaniel:


r-englund (changes requested):
The background regression test should be updated to include this background, it does include all other.

ResearchDaniel:
@r-englund Added test now

r-englund (approved):
Nice addition

modules/basegl/glsl/background.frag
```diff
@@ -74,6 +74,12 @@ vec4 linearGradientVertical(vec2 texCoord) {
     return mix(bgColor2, bgColor1, texCoord.y);
 }
 
+vec4 linearGradientSpherical(vec2 texCoord) {
+	// bgColor1: inner color in circle
+	// bgColor2: outer color
+    return mix(bgColor1, bgColor2, distance(vec2(0.5), texCoord));
```
r-englund: This distance will at most be sqrt(.5) (distance from corner pixel to center pixel). It will never be 1, hence, ´bgColor2´ will never be fully represented. Should we handle that?

PR #55, opened 2018-04-03
Feature/pythonscriptsource

martinfalk:
exposed DataMapper to inviwopy, extended pyvolume interface, and added PythonScriptSource processor

### DIFF
--- a/modules/python3/processors/pythonscriptsource.cpp
+++ b/modules/python3/processors/pythonscriptsource.cpp
@@ -1,0 +1,86 @@
+/*********************************************************************************
+ *
+ * Inviwo - Interactive Visualization Workshop
+ *
+ * Copyright (c) 2018 Inviwo Foundation
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ *********************************************************************************/
+
+#include <modules/python3/processors/pythonscriptsource.h>
+#include <modules/python3/python3module.h>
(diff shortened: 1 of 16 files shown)


modules/python3/processors/pythonscriptsource.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+void PythonScriptSource::process() {
+    if (script_.getSource().empty()) {
+        mesh_.setData(std::make_shared<Mesh>());
+        return;
+    }
+    script_.run({}, [&](pybind11::dict dict) {
+        if (dict.contains("mesh")) {
+            auto pyMesh = dict["mesh"];
+            auto mesh = std::shared_ptr<BasicMesh>(pyMesh.cast<BasicMesh*>());
```
ResearchDaniel: BasicMesh->Mesh?

modules/python3/processors/pythonscriptsource.h
```diff
@@ -0,0 +1,80 @@
...
+ * ### Properties
+ *   * __File Name__  file name of the python script
+ */
+
+/**
+ * \class PythonScriptSource
+ * \brief Loads a mesh and volume via a python script. The processor is invalidated
+ * as soon as the script changes on disk.
+ */
+class IVW_MODULE_PYTHON3_API PythonScriptSource : public Processor {
```
ResearchDaniel: Name sounds like it should output python script.
Possible to make it template based and separate Mesh and Volume?
Other procedural source processors are using Create, i.e. VolumeCreate. 
VolumePythonCreate?

modules/python3/processors/pythonscriptsource.cpp
```diff
@@ -0,0 +1,86 @@
...
+    "Python",                         // Category
+    CodeState::Experimental,          // Code state
+    "Python, CPU",                    // Tags
+};
+const ProcessorInfo PythonScriptSource::getProcessorInfo() const { return processorInfo_; }
+
+PythonScriptSource::PythonScriptSource()
+    : Processor()
+    , script_("")
+    , scriptFileName_("scriptFileName", "File Name", "")
```
ResearchDaniel: Would be great if it is possible to have a default file that shows how it should be used. Possible future PR

modules/python3/pythonscript.h
```diff
@@ -24,112 +24,127 @@
...
-        std::string filename_;
-        void* byteCode_;
-        bool isCompileNeeded_;
-    };
+    bool run(const VariableMap& extraLocalVariables = VariableMap(),
+             std::function<void(pybind11::dict)> callback = [](pybind11::dict dict) {});
+
+    bool run(std::function<void(pybind11::dict)> callback) { return run({}, callback); }
+
+    virtual void setFilename(const std::string& filename);
```
ResearchDaniel, line 95: Ideally, file location should not be needed at script level. Not related to this PR I guess

ResearchDaniel (dismissed):
Nice functionality! 
Have some comments on naming conventions we have used for other procedural source processors

martinfalk:
@petersteneteg suggests `PythonScriptProcessor` as a name candidate

Future work:
- make inports and outports of processor depending on python script
- use secondary script to set up ports
- provide example script

modules/python3/processors/pythonscriptsource.h
martinfalk: no, disagree with separating. No need for templating either. This processor is the first prototype for pulling in data via python, i.e. not creating it per se.

A more advanced version of the processor will determine its ports automatically.

modules/python3/processors/pythonscriptsource.cpp
martinfalk: sure, can do.

modules/python3/pythonscript.h
martinfalk: that was already there, added by @r-englund earlier.

modules/python3/processors/pythonscriptsource.cpp
martinfalk: inviwopy provides both `Mesh` and `BasicMesh`. But only `BasicMesh` provides the functionality for adding vertices like this (at least at the moment until the decorated mesh PR is approved and integrated into inviwopy):
```python
mesh = BasicMesh()
for p in pos :
    mesh.addVertex(vec3(p[0], p[1], p[2]), vec3(0), vec3(0), vec4(1, 0, 0, 1)
```

modules/python3/processors/pythonscriptsource.h
martinfalk: Regarding naming the processor, open for suggestions but they need to be generic.

ResearchDaniel:
Agree on PythonScriptProcessor. Especially if it will be generic.

petersteneteg:
Ideally we would want to be able to save them to the processor list just like the composites

modules/python3/pythonscript.h
martinfalk: I am a bit torn here. I can see use cases for both a string/script property as well as an external file.

modules/python3/processors/pythonscriptsource.cpp
martinfalk: added example workspace / script for generating a random volume

image

petersteneteg (approved):
We have a lot more work to do here, but think that can be an other PR...

Issue #51, opened 2018-03-26
Wrong blending in text rendering with Text Overlay processor


alexanderbock:
The text rendering does not set the transparency correctly or uses the wrong blend mode and does not produce a good bounding box when using italic fonts:

https://imgur.com/a/TJFMQ
![image](https://user-images.githubusercontent.com/9251300/38313079-9d605602-3823-11e8-9f1b-bad4c7eb5e87.png)

Reproduce:
1. Text Overlay processor into Canvas
2. Font Face -> OpenSans Italic
3. Font Size -> 48
4. Text ->  of -> ofa

martinfalk:
This looks like an issue with the glyph bounding box. Can't do much about that since we use FreeFont's interface to access the glyph. We query the respective letter and receive a glyph (in RAM) and a 2D extent.

martinfalk:
Text: `Lorem of`
![image](https://user-images.githubusercontent.com/9251300/38312714-bb445ae8-3822-11e8-9425-b7c0feee6d92.png)


The contents of the texture atlas look correct.
![image](https://user-images.githubusercontent.com/9251300/38312683-abcb2ba0-3822-11e8-8890-7d40ef1f8ebf.png)

martinfalk:
Blend function and depth function are set as
```c++
    utilgl::DepthFuncState depthFunc(GL_ALWAYS);
    utilgl::BlendModeState blending(GL_ONE, GL_ONE_MINUS_SRC_ALPHA);
```

martinfalk:
Rendering into a texture does not set the blending func, but the bbox is cropped, in case the `f` is at the end. 
@alexanderbock so it is all of the above
![image](https://user-images.githubusercontent.com/9251300/38313053-8ab48fc8-3823-11e8-8583-603907634638.png)

PR #61, opened 2018-04-06
Fix/various: improved font rendering / keys

martinfalk:
* updated bounding box calculations for font rendering. Glyphs extending over the text bounding box, e.g. for italic fonts, are no longer clipped (see #51)
* added properties to TextOverlayGL:
   + pixel offset for text positioning
   + line spacing
* added new key codes `IvwKey::Undefined` (for un-assigned shortcuts) and `IvwKey::Return` (was missing)

**Before / After (notice the "f")**
imageimage


petersteneteg (approved):
Nice

PR #62, opened 2018-04-09
Singleton bug

petersteneteg:
Potential fix for singleton bug


r-englund:
What was the bug? Is it that we get different instances in different DLLs? Similar issues as I had when introducing Pybind11?(https://github.com/inviwo/inviwo/commit/69b9efeb57d57cdf147e9d6b55ddc960f03d07ca)

petersteneteg:
Yes, this probably why you got a nullptr for InviwoApplication::getPtr
https://stackoverflow.com/questions/4924082/static-template-data-members-storage/4924318#4924318
has some details. According to that is should more or less never work. But I only ran in to problems when compiling with -fvisibility=hidden which pybind11 wants to do.

r-englund (approved):

Issue #34, opened 2018-03-07
Callback registration of ports


martinfalk:
Why is the interface for registering callbacks in a port so inconsistent?

- `onChange()` and `onInvalid()` support both, lambda expressions and member functions.
- `onConnect()` and `onDisconnect()` only support lambda expressions

r-englund:
Possible reason: 
onChange/onInvalid was added first, before we had c++11 and lambda support.  We later added support for c++11 and could add the lambda call back to those two methods (The old method was keept for backwards compatibility?). Finally we add the onConnect/Disconnect callbacks and there we only use the lambdas.

ResearchDaniel:
@martinfalk , @r-englund's reasoning is spot on. We did not use c++11 when the first functions  where written

r-englund:
So, what action do we take:
1. Status quo (keep as it is) 
2. add member functions to ´onConnect´ and ´onDisconnect´
3. remove member functions to ´onChange´ and ´onInvalid´

if option 3, my suggestion would be to not completly remove them now, but make it deprecated and remove it in a not to distance future

ResearchDaniel:
Vote: Depreceate old callbacks and remove them after some time

martinfalk:
but then also adding member functions to callbacks, i.e. `CallBackList::addMemberCallback()`, should be deprecated. Same for `removeMemberFunction()`

Does anyone know the runtime overhead of using a lambda expression calling a member func instead of calling the member func directly?

r-englund:
`CallBackList::addMemberFunction` actually creates a lambda, so my guess it that the lambda is faster

petersteneteg:
The lambda is usually inlined unless you put it in a std:: function which we usually do for callbacks like this...

martinfalk:
What about `BaseModuleCallback` in modulecallback.h?
There, only member functions are used.

petersteneteg:
The modulecallbacks need a rewrite, the concept is quite cool, but the implementation should be a bit more general I think.

martinfalk:
Deprecation warnings are in place. Member callbacks are still used in some places, e.g. SyntaxHighlighter. Requires RAII before they can be removed.

Closing this issue for now.

PR #67, opened 2018-04-10
Fix/deprecated callbacks, #34

martinfalk:
As discussed in #34, CallBacks to member functions are to be deprecated and will be removed shortly. In the mean time, the compiler should print deprecation warnings.

E.g.:
```c++
    template <typename T>
    [[deprecated("was declared deprecated. Use `onChange(std::function<void()>)` instead")]]
    const BaseCallBack* onChange(T* o, void (T::*m)());
```
will result in
```
openslidetovolume2.cpp(109): warning C4996: 'inviwo::Inport::onChange': was declared deprecated. Use `onChange(std::function<void()>)` instead
inviwo/core/ports/inport.h(179): note: see declaration of 'inviwo::Inport::onChange'
```

### DIFF
--- a/include/inviwo/core/ports/inport.h
+++ b/include/inviwo/core/ports/inport.h
@@ -185,19 +185,16 @@ const BaseCallBack* Inport::onChange(T* o, void (T::*m)()) {
 
 template <typename T>
 const BaseCallBack* Inport::onInvalid(T* o, void (T::*m)()) {
-    ivwDeprecatedMethod("const BaseCallBack* onInvalid(std::function<void()> lambda)");
     return onInvalidCallback_.addMemberFunction(o, m);
 }
 
 template <typename T>
 void Inport::removeOnChange(T* o) {
-    ivwDeprecatedMethod("void removeOnChange(const BaseCallBack* callback)");
     onChangeCallback_.removeMemberFunction(o);
 }
 
 template <typename T>
 void Inport::removeOnInvalid(T* o) {
-    ivwDeprecatedMethod("void removeOnInvalid(const BaseCallBack* callback)");
     onInvalidCallback_.removeMemberFunction(o);
 }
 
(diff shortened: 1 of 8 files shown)


martinfalk:
It might be possible, that the deprecation warnings are not shown with MSVC due to pre-compiled headers. Had some problems there where warning C4996 was suppressed.

ResearchDaniel (approved):
Looks good!

include/inviwo/core/ports/inport.h
```diff
@@ -181,16 +185,19 @@ const BaseCallBack* Inport::onChange(T* o, void (T::*m)()) {
 
 template <typename T>
 const BaseCallBack* Inport::onInvalid(T* o, void (T::*m)()) {
+    ivwDeprecatedMethod("const BaseCallBack* onInvalid(std::function<void()> lambda)");
```
petersteneteg: ivwDeprecatedMethod? why not [[deprecated]]?

modules/python3/pybindutils.h
```diff
@@ -33,8 +33,10 @@
 #include <modules/python3/python3moduledefine.h>
 #include <inviwo/core/common/inviwo.h>
 
+#include <warn/push>
```
petersteneteg: Might be nice to put a comment here on why we push/pop...

include/inviwo/core/ports/inport.h
petersteneteg: also should [[deprecated]] ivwDeprecatedMethod...

petersteneteg (changes requested):

include/inviwo/core/ports/inport.h
martinfalk: forgot to remove `ivwDeprecatedMethod`, `[[deprecated]]` is already in place.

martinfalk:
requested changes have been fixed

petersteneteg (approved):

Issue #69, opened 2018-04-11
Pre-compiled headers interfere with Compiler Warnings


martinfalk:
As long as pre-compiled headers are enabled, some warnings are suppressed. For example, `C4996: depreacted function/class` will not be shown despite not being explicitly disabled.

r-englund:
Even on rebuild?

martinfalk:
I built it from scratch. Cotire puts in some additional prefix headers
see [cotire docs](https://github.com/sakra/cotire/blob/master/MANUAL.md#the-prefix-header)

martinfalk:
the culprit was an unmatched `#pragma warning(push)` in `ext/ticpp/ticppapi.h`.

martinfalk:
There are more issues. The problem is that all, mostly external, headers are included in the precompiled headers without any warning guards. And the PCH headers are included before everything else in each source file thus the appearance of disabled warnings.

Warnings caused by:
- [ ] Qt: `conditional expression is const` (C4127)
- [ ] Eigen: `conditional expression is const` (C4127)
- [x] inviwopy: `class 'type' needs to have dll-interface to be used by clients of class` (C4251 despite being disabled globally, but not disabled for pybind11/inviwopy)

martinfalk:
At the moment, one can suppress the external warnigs by setting the CMake option for `COTIRE_ADDITIONAL_PREFIX_HEADER_IGNORE_PATH` to 
`<inviwo-dev>/modules/eigenutils/ext/eigen/Eigen;${QTDIR}/include`

martinfalk:
Decided to suppress C4127 for Inviwo. Closing this issue for now.

PR #72, opened 2018-04-11
Fix/warnings

martinfalk:
fixes issue with warnings being suppressed when PCH are enabled #69. Fixed various warnings.


petersteneteg (approved):
Nice work!

PR #75, opened 2018-04-13
Fix/center camera on geometry

ResearchDaniel:
- Made CameraProperty properties public instead of protected so that we can change and access min/max values
- Pressing "Center on geometry" in MeshRendererGL now also adjusts min/max values as well as near and far plane.
- Pressing "Compute near/far plane" in MeshRendererGL now also adjusts the min/max values.

Refactored the code to make it easy to include in other processors.


petersteneteg:
The camera property is very strange. It is very odd that is has a bunch of sub properties, but still has a camera that is supposed to hold all of the camera state.

include/inviwo/core/properties/cameraproperty.h
```diff
@@ -90,14 +90,22 @@ class IVW_CORE_API CameraProperty : public CompositeProperty, public TrackballOb
 
     void setAspectRatio(float aspectRatio);
     float getAspectRatio() const;
-
+    /**
+     * Locks processor network and sets given camera properties.
+     * @note Parameters will be capped by their min/max.
+     */
```
martinfalk: will it also unlock?

include/inviwo/core/properties/cameraproperty.h
```diff
@@ -90,14 +90,22 @@ class IVW_CORE_API CameraProperty : public CompositeProperty, public TrackballOb
...
     virtual float getNearPlaneDist() const override;
     virtual float getFarPlaneDist() const override;
 
     void setNearPlaneDist(float v);
     void setFarPlaneDist(float v);
+    /**
+     * Set near and far plane distance values and adjust their min/max ranges.
+     * Adjusts the min/max ranges of the properties to e.g. 0.1/10 times the given value
+     */
+    void setNearFarPlaneDist(float nearPlaneDist, float farPlaneDist, float minMaxRatio = 10.f);
```
martinfalk: any lock happening?

modules/base/algorithm/mesh/axisalignedboundingbox.cpp
```diff
@@ -0,0 +1,69 @@
...
+#include <modules/base/algorithm/mesh/axisalignedboundingbox.h>
+#include <modules/base/algorithm/dataminmax.h>
+
+namespace inviwo {
+
+namespace meshutil {
+
+std::pair<vec3, vec3> axisAlignedBoundingBox(
+    const std::vector<std::shared_ptr<const Mesh>>& meshes) {
+    vec3 worldMin(std::numeric_limits<float>::max());
```
martinfalk: this is incorrect if the given mesh is empty, i.e. has no vertex data.

modules/base/algorithm/mesh/axisalignedboundingbox.cpp
```diff
@@ -0,0 +1,69 @@
...
+        auto minmax = axisAlignedBoundingBox(*mesh);
+        worldMin = glm::min(worldMin, minmax.first);
+        worldMax = glm::max(worldMax, minmax.second);
+    }
+    return {worldMin, worldMax};
+}
+
+std::pair<vec3, vec3> axisAlignedBoundingBox(const Mesh& mesh) {
+    vec3 worldMin(std::numeric_limits<float>::max());
+    vec3 worldMax(std::numeric_limits<float>::lowest());
```
martinfalk, line 55: not correct for empty meshes

modules/base/algorithm/mesh/axisalignedboundingbox.h
```diff
@@ -0,0 +1,58 @@
...
+#define IVW_MESHAXISALIGNEDBOUNDINGBOX_H
+
+#include <modules/base/basemoduledefine.h>
+#include <inviwo/core/datastructures/geometry/mesh.h>
+
+namespace inviwo {
+
+namespace meshutil {
+/**
+ * \brief Compute bounding box of meshes in world space.
```
martinfalk: compute the axis-aligned bounding box ...

Or: The bounding box is aligned with the principal axes of world space.

modules/base/algorithm/mesh/meshcameraalgorithms.cpp
```diff
@@ -0,0 +1,80 @@
...
+                 glm::distance(camera.lookFrom_.getMinValue(), camera.lookTo_.get()));
+    auto furthestViewPoint = camera.getLookFrom() - cameraDir * maxViewLength;
+    // Project min/max points onto view ray and compute distances to the most zoomed out view point
+    farPlaneDist =
+        std::max(glm::distance(furthestViewPoint,
+                               cameraDir * glm::dot(worldSpaceBoundingBox.first, cameraDir)),
+                 glm::distance(furthestViewPoint,
+                               cameraDir * glm::dot(worldSpaceBoundingBox.second, cameraDir)));
+    // Try to have a reasonable precision in the depth buffer
+    nearPlaneDist = std::max(glm::epsilon<float>(), farNearRatio * farPlaneDist);
```
martinfalk: epsilon is a rather poor choice for a reasonable near clip value. This will most likely result in artifacts regarding the depth buffer. Suggesting a "larger" value, e.g. something larger than `1.0fe-4`.

martinfalk: this is also different from what the function docs state...

modules/base/algorithm/mesh/meshcameraalgorithms.h
```diff
@@ -0,0 +1,68 @@
...
+ * plane. The view directions considered are lookFrom min/max -> lookTo. Near plane is computed as
+ * max(1e^-6, farPlaneDistance * farNearRatio)
+ *
+ * @param meshes worldSpaceBoundingBox Min and max points of geometry
+ * @param camera Camera used as basis for computation
+ * @return Near and far plane distances.
+ */
+IVW_MODULE_BASE_API std::pair<float, float> computeNearFarPlanes(
+    std::pair<vec3, vec3> worldSpaceBoundingBox, const CameraProperty& camera,
+    float farNearRatio = 1.f / 10000.f);
```
martinfalk, line 65: is this a good ratio in practice?

martinfalk (dismissed):
good work. Minor changes necessary.

modules/base/algorithm/mesh/axisalignedboundingbox.cpp
ResearchDaniel: Use vec3(0) in that case or mesh offset for min/max?

modules/base/algorithm/mesh/axisalignedboundingbox.cpp
ResearchDaniel: True, we should check this for dataMinMax as well.

modules/base/algorithm/mesh/axisalignedboundingbox.cpp
ResearchDaniel: Or throw? Setting any value can be problematic when using it for multiple meshes, i.e. axisAlignedBoundingBox(
    const std::vector<std::shared_ptr<const Mesh>>& meshes)

modules/base/algorithm/mesh/meshcameraalgorithms.cpp
ResearchDaniel: We would like to keep it as small as possible to not mess with data on smaller scales. The farNearRatio should take care that the precision implications are limited.
Changed it to use same as docs. Perhaps even let user specify but use a default value?

modules/base/algorithm/mesh/meshcameraalgorithms.h
ResearchDaniel: Commonly used by game engines:
"The default Unity camera near and far clipping planes are 0.3 and 1000 respectively, so it's recommended to use a ratio similar to this (1:10000) "
https://answers.unity.com/questions/1193991/any-solution-to-floating-point-precision-issues-wh.html

modules/base/algorithm/mesh/axisalignedboundingbox.h
ResearchDaniel: Check

include/inviwo/core/properties/cameraproperty.h
ResearchDaniel: Adjusted

include/inviwo/core/properties/cameraproperty.h
ResearchDaniel: Adjusted comment

modules/base/algorithm/mesh/axisalignedboundingbox.cpp
martinfalk: either vec3(0) or throw, but throwing can be bad in for example in combination with automatic camera adjustment on inport changes (as happens in some processors).

I guess then we need to separate the implementation or rather use a separate compute function, which is used by both, single mesh and vector of mesh.

modules/base/algorithm/mesh/meshcameraalgorithms.cpp
martinfalk: The problem is with some datasets like the head, there the scene is suddenly larger than [1000, 1000, 1000]... Run quite often into problems with nearclip smaller than 0.1 while farclip is >100

modules/base/algorithm/mesh/meshcameraalgorithms.h
martinfalk: :+1:

modules/base/algorithm/mesh/meshcameraalgorithms.cpp
ResearchDaniel: Could not detect any artifacts when using the MRI head data set

martinfalk (approved):

PR #77, opened 2018-04-17
Assimp: Made AssimpReader look in both absolute and relative paths

ResearchDaniel:
Assimp: Made AssimpReader look in both absolute path and path relative to base path as done in many other readers.

### DIFF
--- a/modules/assimp/assimpreader.cpp
+++ b/modules/assimp/assimpreader.cpp
@@ -34,6 +34,7 @@
 #include <inviwo/core/datastructures/geometry/mesh.h>
 #include <inviwo/core/datastructures/buffer/bufferramprecision.h>
 #include <inviwo/core/io/datareaderexception.h>
+#include <inviwo/core/util/filesystem.h>
 
 #include <warn/push>
 #include <warn/ignore/all>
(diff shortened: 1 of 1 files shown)


r-englund (dismissed):
Maybe this should, in this or another PR, become a utility function?

ResearchDaniel:
@r-englund Hehe, had the same question to myself :)

petersteneteg:
Why do we do this? seems really strange. Should only be relevant in the special case that we are referring to data shipped with inviwo? And should not be needed even then. 
If we want to consider a relative path, then that should be relative to the workspace file if any.
Also this should only be done on relative paths?

petersteneteg (changes requested):
This seems like very questionable to be
is this documented behavior of our data readers?

ResearchDaniel:
@petersteneteg Ah, perhaps it is done when saving the workspace instead. I did not check, just compared to the other readers. I will check and remove the code from the other readers as well in that case.

ResearchDaniel:
FileProperty stores three paths in the workspace:
     1) Absolute
     2) Relative workspace
     3) Relative filesystem::getPath(PathType::Data)
FileProperty::deserialize looks for the file in the reverse order

Conclusion: Remove this kind of code in data readers

ResearchDaniel:
Replaced by #78

PR #78, opened 2018-04-17
Fix/reader search paths

ResearchDaniel:
Removed relative file path search from readers since it is taken care of by FileProperty, see PR #77


petersteneteg (approved):
Good

PR #88, opened 2018-04-20
Feature/tf refactoring

martinfalk:
introducing TFPrimitive, TFPrimitiveSet, and TransferFunctionEditorPrimitive
* TFPrimitive supersedes TransferFunctionDataPoint and IsoValue
* TFPrimitiveSet is used as common base class for both TF and isovalue collection
* based  TransferFunctionEditorControlPoints on abstract base class TransferFunctionEditorPrimitive
* cleaned up TransferFunctionEditor


martinfalk:
The grid of the TF editor is now also zero-based and zero is indicated by a slightly thicker line (if zero is within the value range).

ResearchDaniel (approved):
Well needed!
Relative/absolute needs some discussion but I can't see that it is used yet anyway?

modules/qtwidgets/properties/transferfunctioneditor.h
```diff
@@ -62,38 +63,35 @@ class IVW_MODULE_QTWIDGETS_API TransferFunctionEditor : public QGraphicsScene {
...
-    float getZoomRangeXMin() const;
-    void setZoomRangeXMin(float min);
-    float getZoomRangeXMax() const;
-    void setZoomRangeXMax(float max);
-    float getZoomRangeYMin() const;
-    void setZoomRangeYMin(float min);
 
-    float getZoomRangeYMax() const;
-    void setZoomRangeYMax(float max);
+    virtual void onControlPointAdded(TFPrimitive* p);
```
petersteneteg, line 67: Override?

include/inviwo/core/datastructures/tfprimitiveset.h
```diff
@@ -0,0 +1,182 @@
...
+
+    TFPrimitiveSetType getType() const;
+
+    size_t size() const;
+
+    TFPrimitive* operator[](size_t i);
+    const TFPrimitive* operator[](size_t i) const;
+
+    TFPrimitive* get(size_t i);
+    const TFPrimitive* get(size_t i) const;
```
petersteneteg: Would be nice with begin() and end() iters...

martinfalk:
The relative-absolute feature will come soon. The idea here is that a TF is self-contained. There will be a util function for converting a relative TF to an absolute and vice versa given a range. At the moment, the TF type is not used (or rather everything is normalized).

include/inviwo/core/datastructures/tfprimitiveset.h
martinfalk: hm, would that then be on the unsorted or sorted primitives?

martinfalk:
This PR was also intended to be as non-invasive as possible, i.e. refactoring existing functionality in preparation for future features like isovalues.

include/inviwo/core/datastructures/tfprimitiveset.h
martinfalk: choosing sorted for now.

modules/qtwidgets/properties/transferfunctioneditor.h
martinfalk: No, the editor is not yet an observer. Matching function names by "coincidence", I guess.

src/core/datastructures/tfprimitiveset.cpp
```diff
@@ -96,26 +96,52 @@ TFPrimitiveSet& TFPrimitiveSet::operator=(const TFPrimitiveSet& rhs) {
     return *this;
 }
 
+dvec2 TFPrimitiveSet::getRange() const {
+    switch (type_) {
+        case TFPrimitiveSetType::Absolute: {
+            if (sorted_.empty()) {
```
ResearchDaniel: This can become inconsistent. Why not use a DataMapper? getMinMax is perhaps what it should be called otherwise?

I think that we will need to be consistent in how we store data values in a TFPrimitiveSet. Either always normalized or always absolute. A DataMapper could be used to map the data into "absolute" values

ResearchDaniel (dismissed):

src/core/datastructures/tfprimitiveset.cpp
martinfalk: This function returns the "used" range of the TF which is [0,1] for relative and [first, last] for absolute. TF primitives are always stored as absolute so there is no need for a data mapper. Only mapping between a relative and absolute TF will require a datamapper.

martinfalk: This means that a relative TF will save its values as [0,1], which requires no mapping when the input data is changed. Similarly, an absolute TF also does not need a mapping when switching data. Using a datamapper, however, will introduce problems when using other data, etc.

martinfalk: Furthermore, we need a range in order to be able to create a TF preview

petersteneteg: I agree with Martin. We should not use any dataMapper here. 
The TF should not have a data range by it self. It is defined by the position of the first and last TF primitive.

ResearchDaniel: @petersteneteg We cannot define the data range using the primitives since the points are dynamic...
@martinfalk @petersteneteg  If they are always stored in absolute, why do we need to differentiate between normalized and absolute here? Should they not always be stored in normalized coordinates and instead mapped to a data range when requested?

petersteneteg (approved):

src/core/datastructures/tfprimitiveset.cpp
petersteneteg: We don't need or want a absolute data range for a absolute tf, there is no need for it.
We never store in normalized coordinates, but we have a enum to specify how to interpret the values. Either as absolute or as normalized.

martinfalk: @ResearchDaniel:
a) in case of Relative TF, `add()` performs a range check for [0,1] and throws an exception if outside
b) no, they should not always be normalized, that is the mess we are having right now. Assuming you have an absolute TF for a given data range (based in the current input volume). What shall be done when the input volume changes? Or worse, if there (for some reason) is no volume. Furthermore, how do you want to link two absolute TFs where they have different data_ranges?
c) The mapping from absolute to [0,1] should *only* happen during the visualization, i.e. in the shader. That's why it is called absolute.
d) the reason for differentiating becomes obvious when looking at the TF extent. Normalized always is within [0,1]. Absolute can be anything from float::lowest to float::max. Which means that you are able to define a TF from [-5,20] even though your current data range is [50,10000].

ResearchDaniel: @martinfalk As long as we are consistent. Is it not ambiguous to say that we always store absolute but allow relative? 
a) should be unnecessary if we always store absolute?
b) Always same when linked. 
The current implementation still adds normalized values but you say that they are stored in absolute, it is a bit ambiguous.
c) sure
d) true. But that think that is possible when using a DataMapper as well. Just allow any value to be added, but use the DataMapper range for displaying axes. 
If it is only a matter of interpretation then it should not be wrong to add values outside [0 1] in the current implementation?

PR #89, opened 2018-04-23
Fix/tfprimitives double

martinfalk:
Changing the position component of TF primitives from `float` to `double`. There is no reason of why one should use float here. The TF is only exposed to OpenGL via a texture, not the actual TF primitives.


petersteneteg (approved):
Finally! floats be gone!

src/core/datastructures/tfprimitiveset.cpp
```diff
@@ -146,12 +146,24 @@ std::vector<TFPrimitiveData> TFPrimitiveSet::getUnsorted() const {
...
+    result.second.reserve(sorted_.size());
+    for (auto& v : sorted_) {
+        result.first.push_back(static_cast<float>(v->getPosition()));
+        result.second.push_back(v->getColor());
+    }
+
+    return result;
+}
+
+std::pair<std::vector<float>, std::vector<vec4>> TFPrimitiveSet::getVectorsf() const {
```
ResearchDaniel, line 161: Future PR: We could return a data frame or something that allows it to be extended to multiple attributes. Access could be by column name instead, position, color and so on. Perhaps we could even use a data frame for storage in the TF now that I think of it.

ResearchDaniel (approved):
Nice!

PR #95, opened 2018-04-27
Cotire update

petersteneteg:


martinfalk (approved):
About time :-) Nice work with the cleanup.

PR #108, opened 2018-05-08
Adding DiscreteData Module

AnkeAnke:
**State**
- The basic setup is done, there is at least one example for abstract classes
- Has worked with one project (KTHInviwo, Percolation)
- Tests should be extended
- Lots of places to discuss, please comment!
-- Iterators & cache
-- Connection iteration data
-- DataSet with const Channels only?

**What environment has the code been compiled and tested on?** 
- Windows
- VC 2017


modules/discretedata/channels/analyticchannel.h
```diff
@@ -0,0 +1,122 @@
...
+                    const std::string& name, GridPrimitive definedOn = GridPrimitive::Vertex)
+        : DataChannel<T>(numComponents, name, definedOn)
+        , numElements_(numElements)
+        , dataFunction_(dataFunction) {}
+
+    virtual ~AnalyticChannel() = default;
+
+    // Methods
+public:
+    ind getNumElements() const override { return numElements_; }
```
petersteneteg: We usually use size for the number of elements, to be consistant with stl etc. makes writing generic code easier...

modules/discretedata/channels/analyticchannel.h
```diff
@@ -0,0 +1,122 @@
...
+    where the destination memory is pre-allocated.
+    Indices are linear.
+
+    @author Anke Friederici and Tino Weinkauf
+*/
+template <typename T>
+class AnalyticChannel : public DataChannel<T> {
+    // Types
+public:
+    typedef typename std::function<void(T*, ind)> Function;
```
petersteneteg: Having the dimension of vec undefined does not seem very typesafe

modules/discretedata/channels/analyticchannel.h
```diff
@@ -0,0 +1,122 @@
...
+    ivwAssert(index >= 0 && index < numElements_, "Index out of bounds: " << index);
+
+    dataFunction_(dest, index);
+}
+
+/*--------------------------------------------------------------*
+*  Exported Types                                               *
+*--------------------------------------------------------------*/
+
+typedef AnalyticChannel<float>      AnalyticChannelFloat;
```
petersteneteg: We prefer the more readable using x = y

modules/discretedata/channels/bufferchannel.h
```diff
@@ -0,0 +1,199 @@
...
+                  GridPrimitive definedOn = GridPrimitive::Vertex)
+        : DataChannel<T>(numComponents, name, definedOn), buffer_(numElements * numComponents) {}
+
+    /** \brief Direct construction
+    *   @param data Raw data, copy values
+    *   @param numComponents Size of vector at each position
+    *   @param name Name associated with the channel
+    *   @param definedOn GridPrimitive the data is defined on, default: 0D vertices
+    */
+    BufferChannel(std::vector<T>& data, ind numComponents, const std::string& name,
```
petersteneteg: const vector& ?

modules/discretedata/channels/bufferchannel.h
```diff
@@ -0,0 +1,199 @@
...
+    BufferChannel(T* data, ind numElements, ind numComponents, const std::string& name,
+                  GridPrimitive definedOn = GridPrimitive::Vertex)
+        : DataChannel<T>(numComponents, name, definedOn)
+        , buffer_(data, data + numElements * numComponents) {}
+
+protected:
+
+    // Methods
+public:
+    virtual ind getNumElements() const override;
```
petersteneteg: size?

modules/discretedata/channels/bufferchannel.h
```diff
@@ -0,0 +1,199 @@
...
+
+    /** \brief Indexed point access, constant
+    *   @param index Linear point index
+    *   @return Pointer to data, size T[NumComponents]
+    */
+    const T* operator[](ind index) const { return get(index); }
```
petersteneteg: I don't like these raw pointers to array without a explict size.

modules/discretedata/channels/bufferchannel.h
```diff
@@ -0,0 +1,199 @@
...
+void BufferChannel<T>::fill(T* const dest, const ind index) const {
+    ivwAssert(index >= 0 && index < getNumElements(), "Index out of bounds: " << index);
+
+    memcpy(dest, buffer_.data() + (index * numComponents_), sizeof(T) * numComponents_);
+}
+
+/*--------------------------------------------------------------*
+*  Exported Types                                               *
+*--------------------------------------------------------------*/
+
```
petersteneteg: Using

modules/discretedata/channels/bufferchannel.h
```diff
@@ -0,0 +1,199 @@
...
+    ivwAssert(index >= 0 && index < getNumElements(), "Index out of bounds: " << index);
+
+    return buffer_.data() + index * numComponents_;
+}
+
+template <typename T>
+void BufferChannel<T>::fill(T* const dest, const ind index) const {
+    ivwAssert(index >= 0 && index < getNumElements(), "Index out of bounds: " << index);
+
+    memcpy(dest, buffer_.data() + (index * numComponents_), sizeof(T) * numComponents_);
```
petersteneteg: This limmits T to trivial types, should maybe be noted, or just use std::copy, will always work and will be optimized to memcpy for trivial types

modules/discretedata/channels/channeliterator.h
```diff
@@ -0,0 +1,107 @@
...
+    /** Decrement randomly */
+    ChannelIterator operator-(ind offset) { return ChannelIterator(Parent, Index - offset); }
+
+    /** Decrement randomly */
+    ChannelIterator operator-=(ind offset) { Index -= offset; }
+
+    // Members
+protected:
+    /** Index to the current element */
+    ind Index;
```
petersteneteg, line 117: Lowercase members

modules/discretedata/channels/channeliterator.h
```diff
@@ -0,0 +1,107 @@
...
+public:
+    ChannelIterator(DataChannel<T>* parent, ind index)
+        : Parent(parent), Index(index), Data(nullptr), DataIndex(-1) {}
+
+    ChannelIterator() : Parent(nullptr), Index(-1), Data(nullptr), DataIndex(-2) {}
+
+    ~ChannelIterator() { delete[] Data; }
+
+    /** Dereference to get data */
+    T* operator*();
```
petersteneteg: How do I get the number of components?

petersteneteg: Should return a vec<N>& I think using some vec type, glm, std::arrray, span, or some other

petersteneteg: We should probably discuss if you shoukd use glm or some oher vec type soon, it will probably influence the design alot

modules/discretedata/channels/channeliterator.h
```diff
@@ -0,0 +1,107 @@
...
+
+    //*** Bidirectional Iteration ***\\
+
+    /** Walk forward */
+    ChannelIterator& operator++() {
```
petersteneteg: Post increment?

modules/discretedata/channels/channeliterator.h
```diff
@@ -0,0 +1,107 @@
...
+namespace dd {
+
+/** \struct ChannelIterator
+*   Generalized iterator over any DataChannel.
+*   Handles data deletion if necessary - runtime check whether needed.
+*/
+template <typename T>
+class ChannelIterator {
+public:
+    ChannelIterator(DataChannel<T>* parent, ind index)
```
petersteneteg: should define iteractor category tag types. see http://en.cppreference.com/w/cpp/iterator/iterator_traits

modules/discretedata/channels/channeliterator.inl
```diff
@@ -0,0 +1,55 @@
...
+*********************************************************************************/
+
+#pragma once
+
+template <typename T>
+T* inviwo::dd::ChannelIterator<T>::operator*() {
+    ivwAssert(Parent, "No channel to iterate is set.");
+
+    // If our parent is a buffer, return a direct data pointer.
+    if (dynamic_cast<BufferChannel<T>*>(Parent))
```
petersteneteg: it's probably a big perf hit to do a dynamic cast on every iterator dereference. This functionality probably has to be in the channel not in the iterator.

modules/discretedata/channels/channeliterator.inl
```diff
@@ -0,0 +1,55 @@
...
+        return dynamic_cast<BufferChannel<T>*>(Parent)->get(Index);
+
+    // No buffer, evaluate analytics.
+    // Is the data up to date?
+    if (DataIndex != Index) {
```
petersteneteg: So in this case the channel is analytical and hence const? right, so there is really no reason to provide a non const iterator since it can't be modified anyway. And if it cant be modified you could just return the "T" by value not having to go through all this trouble of have a local buffer holding the value so you can point to it.

We will probably need to provide a ConstChannelIterator and a ChannelIterator. Not even sure if the mutable one is needed since all the data is usually immutable anyway? at least the const one seems the more important one.

modules/discretedata/channels/datachannel.cpp
```diff
@@ -0,0 +1,78 @@
...
+#include <inviwo/core/metadata/metadataowner.h>
+
+#include "datachannel.h"
+
+namespace inviwo {
+namespace dd {
+
+Channel::Channel(ind numComponents, const std::string& name, DataFormatId format,
+                 GridPrimitive prim)
+    : numComponents_(numComponents) {
```
petersteneteg: In my thinking we only use the meta data for "extra" information not usually in the data structure,  I would rather store name, format, primitive as regular class members.

modules/discretedata/channels/datachannel.cpp
```diff
@@ -0,0 +1,78 @@
...
+    // Create and set name.
+    StringMetaData* nameData = createMetaData<StringMetaData>("Name");
+    *nameData = name;
+
+    // Create and set format identifier.
+    IntMetaData* formatData = createMetaData<IntMetaData>("DataFormatId");
+    *formatData = (int)format;
```
petersteneteg, line 47: One should never use c-style cast in c++

modules/discretedata/channels/datachannel.h
```diff
@@ -0,0 +1,171 @@
...
+namespace dd {
+
+/** Discretedata index type **/
+typedef signed long long ind;
+
+/** Mapping structure name to respective dimension.
+*   Assign channels to any dimensions this way.
+*   If these do not suffice, cast the respective short.
+*/
+enum GridPrimitive : char {
```
petersteneteg: use enum classs

modules/discretedata/channels/datachannel.h
```diff
@@ -0,0 +1,171 @@
...
+
+#include <discretedata/discretedatamoduledefine.h>
+#include <inviwo/core/common/inviwo.h>
+#include <inviwo/core/metadata/metadataowner.h>
+
+namespace inviwo {
+namespace dd {
+
+/** Discretedata index type **/
+typedef signed long long ind;
```
petersteneteg, line 39: use using

modules/discretedata/channels/datachannel.h
```diff
@@ -0,0 +1,171 @@
...
+private:
+    /** Sets the "GridPrimitiveType" meta data
+    *   Should be constant, only DataSet is allowed to write.
+    */
+    void setGridPrimitiveType(GridPrimitive);
+
+    // Attributes
+public:
+    virtual ind getNumElements() const = 0;
+    ind getNumComponents() const { return numComponents_; }
```
petersteneteg: the dataformat also has the number of components...

modules/discretedata/channels/datachannel.h
```diff
@@ -0,0 +1,171 @@
...
+
+private:
+    /** Sets the "GridPrimitiveType" meta data
+    *   Should be constant, only DataSet is allowed to write.
+    */
+    void setGridPrimitiveType(GridPrimitive);
+
+    // Attributes
+public:
+    virtual ind getNumElements() const = 0;
```
petersteneteg: use size

modules/discretedata/channels/datachannel.h
```diff
@@ -0,0 +1,171 @@
...
+    virtual ~DataChannel() = default;
+
+    // Methods
+public:
+    /** \brief Indexed point access, copy data
+    *   Thread safe.
+    *   @param dest Position to write to, expect T[NumComponents]
+    *   @param index Linear point index
+    */
+    virtual void fill(T* dest, ind index) const = 0;
```
petersteneteg: Imho this kind of interface with T* is nice very safe, should really be avoided.

modules/discretedata/channels/datachannel.h
```diff
@@ -0,0 +1,171 @@
...
+
+    /** \brief Indexed point access, copy data
+    *   Thread safe.
+    *   @param dest Position to write to, expect T[NumComponents]
+    *   @param index Linear point index
+    */
+    void operator()(T* dest, ind index) const { fill(dest, index); }
```
petersteneteg: Same as above

modules/discretedata/channels/datachannel.h
```diff
@@ -0,0 +1,171 @@
...
+
+    ChannelIterator<T> begin();
+
+    ChannelIterator<T> end();
```
petersteneteg: should also have const iterators

modules/discretedata/channels/datachannel.h
```diff
@@ -0,0 +1,171 @@
...
+
+    /** Sets the "Name" meta data */
+    void setName(const std::string&);
+
+    /** Returns the "GridPrimitiveType" meta data */
+    GridPrimitive getGridPrimitiveType() const;
+
+    /** Returns the "DataFormatId" meta data */
+    DataFormatId getDataFormatId() const;
+
```
petersteneteg, line 100: you might want to add support for dispatching on the dataformat here similar to VolumeRAM for example

modules/discretedata/channels/datachannel.inl
```diff
@@ -0,0 +1,72 @@
...
+#include <inviwo/core/metadata/metadata.h>
+
+namespace inviwo {
+namespace dd {
+
+template <typename T>
+DataChannel<T>::DataChannel(ind numComponents, const std::string& name, GridPrimitive definedOn)
+    : Channel(numComponents, name, DataFormatId::NotSpecialized, definedOn) {
+    // Switch all types.
+    if (std::is_same<T, f16>::value) setDataFormatId(DataFormatId::Float16);
```
petersteneteg: you can just use ```DataFormat<T>::id()```

modules/discretedata/connectivity/connectioniterator.h
```diff
@@ -0,0 +1,154 @@
...
+
+class ConnectionRange;
+
+/** \class ConnectionIterator
+*   Iterates over one GridPrimitive type in a Connectivity.
+*/
+class IVW_MODULE_DISCRETEDATA_API ConnectionIterator {
+    friend ConnectionIterator operator+(ind, ConnectionIterator&);
+    friend ConnectionIterator operator-(ind, ConnectionIterator&);
+
```
petersteneteg, line 50: missing iterator tag types

modules/discretedata/connectivity/connectioniterator.h
```diff
@@ -0,0 +1,154 @@
...
+    ~ConnectionIterator() = default;
+
+    /** Dereference to 'get data' */
+    ElementIterator operator*() const;
+
+    //*** Bidirectional Iteration ***\\
+
+    /** Walk forward */
+    ConnectionIterator& operator++() {
```
petersteneteg, line 67: missing post increment

modules/discretedata/connectivity/connectivity.h
```diff
@@ -0,0 +1,128 @@
...
+    /** \brief Range of all elements to iterate over
+    *   @param dim Dimension to return the elements of
+    */
+    ElementRange all(GridPrimitive dim) const;
+
+    /** \brief Get the measure (i.e., length, area, volume...) of an element
+    *   @param dim Dimension of element (edge, face, volume...)
+    *   @Param index Index of respective element type
+    */
+    virtual double getPrimitiveMeasure(GridPrimitive, ind) const { return -1; }
```
ResearchDaniel: should be abstract?
ind?

modules/discretedata/connectivity/structuredgrid.cpp
```diff
@@ -0,0 +1,264 @@
...
+    return neighbors;
+}
+}
+
+std::vector<ind> StructuredGrid::getConnections(ind idxLin, GridPrimitive from,
+                                                GridPrimitive to) const {
+    if (from == to && from == gridDimension_) {
+        // Linear Index to nD Cell Index.
+        ind idxCutoff = idxLin;
+        std::vector<ind> index(numCellsPerDimension_.size(), -1);
```
ResearchDaniel: Any reason to initialize to -1?
reserve might be faster...

modules/discretedata/connectivity/structuredgrid.cpp
```diff
@@ -0,0 +1,264 @@
...
+            break;
+        case DataFormatId::UInt32:
+            measure = computeHexVolume<glm::u32>(index);
+            break;
+        case DataFormatId::UInt64:
+            measure = computeHexVolume<glm::u64>(index);
+            break;
+
+        default:
+            LogWarn("Data type not supported. Edit " << __FILE__);
```
ResearchDaniel: Probably better to throw exception

modules/discretedata/connectivity/structuredgrid.h
```diff
@@ -0,0 +1,110 @@
...
+namespace dd {
+
+class ElementIterator;
+
+/** \class StructuredGrid
+    \brief A curvilinear grid in nD
+
+    @author Anke Friederici and Tino Weinkauf
+*/
+class IVW_MODULE_DISCRETEDATA_API StructuredGrid : public Connectivity {
```
ResearchDaniel, line 45: SpatialData?
https://github.com/inviwo/inviwo/blob/69cca9b6ecd58037bda0ed9e6f53d02f189f19a7/include/inviwo/core/datastructures/spatialdata.h

ResearchDaniel (commented):
Looks good! Some questions and comments

modules/discretedata/connectivity/connectivity.h
AnkeAnke: Is not abstract so far as it is nice to have a volume computation given, but not absolutely necessary (hence, standard implementation with invalid output).

ind is a typedef for signed long long, which is used for all indices, iteration variables etc consistently in the project. We decided on signed (faster?, possibility to return 'invalid value' as in this case, consistent type should prevent unnecessary casting). The typedef is in datachannel.h.

modules/discretedata/connectivity/structuredgrid.cpp
AnkeAnke: This way we have he possibility to check invalid values. I made use of it in the loose implementation of unstructured data, for example (not uploaded yet). Also, this iterates over dimensions. The highest I've ever needed is 7, so I expect <100 dimensions, paired with it happening only once each new grid is initialized, I would prefer comfort over speed here.

modules/discretedata/connectivity/structuredgrid.cpp
AnkeAnke: Good point.
On another hand I am still wondering whether we should try to keep as much as possible of the module `noexcept` for performance, maybe even explicitly. Though I'm not sure how far that can go with stl bits everywhere. So far I've relied on asserts mostly.
Either way, this code will change to use Peter's dispatch functionality soon.

modules/discretedata/connectivity/structuredgrid.h
AnkeAnke: We had a look into it, but mainly two reasons against it:
- We find the model, world transforms more confusing than anything else. Though I guess it might be beneficial to have them, but did not yet find a scenario where I need them.
- It is possible to have a 1D structured grid (a line) in a 3D space. I think that can not be represented by SpatialData?

I'm very open to integrate them if there is a benefit. I don't know where they are used so far.
Any thoughts?

modules/discretedata/connectivity/connectioniterator.h
```diff
@@ -0,0 +1,154 @@
...
+    /** List of neighborhood indices */
+    std::shared_ptr<const std::vector<ind>> connection_;
+};
+
+class ConnectionRange {
+public:
+    ConnectionRange(ind fromIndex, GridPrimitive fromDim, GridPrimitive toDim,
+                    const Connectivity* parent);
+
+    ConnectionIterator begin() { return ConnectionIterator(parent_, toDimension_, connections_, 0); }
```
petersteneteg, line 142: Const?

modules/discretedata/connectivity/structuredgrid.cpp
ResearchDaniel: The loop afterwards, `index[dim] = idxCutoff % dimSize;`, overwrites all values anyway?
No biggie though

modules/discretedata/connectivity/structuredgrid.h
AnkeAnke: Also, the grid so far is independent of the point dimension. So this is probably more relevant for the DataSet - might actually be a good place to save the transformations.

modules/discretedata/connectivity/structuredgrid.cpp
ResearchDaniel: I would try and see if it has any performance impact in that case. This case might bite you in the *** later :D

modules/discretedata/connectivity/structuredgrid.h
ResearchDaniel: Was also thinking that it might be more suitable in the DataSet. 
The SpatialData is used to place the data in world space. I guess the vertices must be given in world space right now?

modules/discretedata/connectivity/connectivity.h
```diff
@@ -0,0 +1,128 @@
...
+    Connectivity(GridPrimitive gridDimension)
+        : gridDimension_(gridDimension)
+        , numGridPrimitives_(gridDimension + 1, -1) {
+    }  // Initialize sizes with -1. Overwrite when known.
+    virtual ~Connectivity() = default;
+
+    // Methods
+public:
+    /** \brief Returns the maximal dimension of the grid */
+    GridPrimitive getDimension() const { return gridDimension_; }
```
petersteneteg, line 79: I find dimension to be a bit confusing here, it that the best name? would Rank, or Dimensionality be better? Or is this the standard name in the field? I think dimension is often confused with NumElements?

modules/discretedata/connectivity/connectivity.h
```diff
@@ -0,0 +1,128 @@
...
+
+    /** \brief Return the number of elements of the given type
+    *   @param elementType Type to get number of
+    *   @return Number of elements
+    */
+    virtual ind getNumElements(GridPrimitive elementType) const;
```
petersteneteg, line 85: would prefer size

modules/discretedata/connectivity/connectivity.h
```diff
@@ -0,0 +1,128 @@
...
+
+    /** \brief Get the map from one element to another
+    *   E.g. cell to its vertices, vertex to its neighbors, vertex to connected faces
+    *   @param index Index of element in dimension from
+    *   @param from Dimension the index lives in
+    *   @param to Dimension the result lives in
+    *   @return All connected indices in dimension to
+    */
+    virtual std::vector<ind> getConnections(ind index, GridPrimitive from,
```
petersteneteg: are there any limitation of the combination of to and from, or is all permutations valid?

modules/discretedata/connectivity/elementiterator.h
```diff
@@ -0,0 +1,143 @@
...
+
+class ConnectionRange;
+
+/** \class ElementIterator
+*   Iterates over one GridPrimitive type in a Connectivity.
+*/
+class IVW_MODULE_DISCRETEDATA_API ElementIterator {
+    friend ElementIterator operator+(ind, ElementIterator&);
+    friend ElementIterator operator-(ind, ElementIterator&);
+
```
petersteneteg, line 48: iterator tag types missing

modules/discretedata/connectivity/elementiterator.h
```diff
@@ -0,0 +1,143 @@
...
+    ~ElementIterator() = default;
+
+    /** Dereference to 'get data' */
+    ElementIterator operator*() const;
+
+    //*** Bidirectional Iteration ***\\
+
+    /** Walk forward */
+    ElementIterator& operator++() {
```
petersteneteg, line 63: post increment missing

modules/discretedata/connectivity/elementiterator.h
```diff
@@ -0,0 +1,143 @@
...
+public:
+    ElementIterator(const Connectivity* parent, GridPrimitive dimension, ind index = 0)
+        : index_(index), parent_(parent), dimension_(dimension) {}
+
+    ElementIterator() : index_(-1), parent_(nullptr), dimension_(GridPrimitive(-1)) {}
+
+    ~ElementIterator() = default;
+
+    /** Dereference to 'get data' */
+    ElementIterator operator*() const;
```
petersteneteg, line 58: return index here?

modules/discretedata/connectivity/elementiterator.h
```diff
@@ -0,0 +1,143 @@
...
+    // Members
+protected:
+    /** Index to the current element */
+    ind index_;
+
+    /** Pointer to Connectivity iterated through - Does not delete */
+    const Connectivity* parent_;
+
+    /** GridPrimitive type iterated over (0D vertices etc) */
+    const GridPrimitive dimension_;
```
petersteneteg, line 130: Question it the GridPrimitive the same as the Polytope (https://en.wikipedia.org/wiki/Polytope)? I think it would be nice if we could find a different naming scheme here then using dimension, since it's meaning is a bit ambiguous

petersteneteg: or is a Simplex type? https://en.wikipedia.org/wiki/Simplex there are so many names for these things :)

modules/discretedata/connectivity/connectivity.h
```diff
@@ -0,0 +1,128 @@
...
+protected:
+    /** Highest dimension of GridPrimitives */
+    GridPrimitive gridDimension_;
+
+    /** Saves the known number of primitves */
+    mutable std::vector<ind> numGridPrimitives_;
+
+public:
+    /** Vertex positions */
+    std::shared_ptr<const Channel> vertices_;
```
petersteneteg: Does this have to be in the base class? should the derived classes not define this them self?

modules/discretedata/connectivity/structuredgrid.cpp
```diff
@@ -0,0 +1,264 @@
...
+#include "discretedata/connectivity/elementiterator.h"
+
+namespace inviwo {
+namespace dd {
+
+StructuredGrid::StructuredGrid(GridPrimitive gridDimension, const std::vector<ind>& numCellsPerDim)
+    : Connectivity(gridDimension), numCellsPerDimension_(numCellsPerDim) {
+    ivwAssert(numCellsPerDim.size() == gridDimension,
+              "Grid dimension should match cell dimension.");
+    numCellsPerDimension_ = std::vector<ind>(numCellsPerDim);
```
petersteneteg, line 41: again?

modules/discretedata/connectivity/structuredgrid.h
```diff
@@ -0,0 +1,110 @@
...
+    double computeHexVolume(ind index) const;
+
+protected:
+    std::vector<ind> numCellsPerDimension_;
+};
+
+template <typename T>
+double StructuredGrid::computeHexVolume(ind index) const {
+    // Work with respective type
+    std::shared_ptr<const DataChannel<T>> doubleVertices =
```
petersteneteg: not used?

modules/discretedata/connectivity/structuredgrid.cpp
```diff
@@ -0,0 +1,264 @@
...
+    ind numCells = 1;
+    ind numVerts = 1;
+    for (ind dim = GridPrimitive::Vertex; dim < gridDimension; ++dim) {
+        numCells *= numCellsPerDimension_[dim];
+        numVerts *= numCellsPerDimension_[dim] + 1;
+    }
+
+    numGridPrimitives_[gridDimension] = numCells;
+    numGridPrimitives_[GridPrimitive::Vertex] = numVerts;
+}
```
petersteneteg, line 52: Don't use need to initialize the Connectivity::vertices_ member? or where is that done?

modules/discretedata/connectivity/structuredgrid.cpp
```diff
@@ -0,0 +1,264 @@
...
+double StructuredGrid::getPrimitiveMeasure(GridPrimitive dim, ind index) const {
+    if (!this->vertices_) return -1;
+
+    // Only implemented 3D bodies so far.
+    if (dim != GridPrimitive::Volume) return -1;
+
+    double measure = -1;
+
+    // TODO: Make this kind of code obsolete or at least compact.
+    switch (vertices_->getDataFormatId()) {
```
petersteneteg: here you could use the dispatch thing

modules/discretedata/connectivity/structuredgrid.cpp
AnkeAnke: Ooooh. Right. Well, yes.

modules/discretedata/dataset.h
```diff
@@ -0,0 +1,209 @@
...
+#include <inviwo/core/common/inviwo.h>
+#include <inviwo/core/datastructures/datatraits.h>
+
+#include "channels/bufferchannel.h"
+#include "connectivity/connectivity.h"
+
+namespace inviwo {
+namespace dd {
+
+typedef std::shared_ptr<Channel> SharedChannel;
```
petersteneteg, line 43: From my experience, i think it's better not to typedef the shared_ptr like this, the user really needs to know that it is a std::shared_ptr to be able to create is using for example std::make_shared. And also for how to think about it with respect to lifetime etc. so I would prefer to use std::shared_ptr<Channel> std::shared_ptr<const Channel> in the interface instead of the typedefs.

modules/discretedata/dataset.h
```diff
@@ -0,0 +1,209 @@
...
+typedef std::shared_ptr<const Channel> SharedConstChannel;
+
+struct ChannelCompare {
+    bool operator()(const std::pair<std::string, GridPrimitive>& u,
+                    const std::pair<std::string, GridPrimitive>& v) const {
+        return (u.second < v.second) || u.first.compare(v.first) < 0;
+    }
+};
+// Map used for storing and querying channels by name and GridPrimitive type.
+typedef std::map<
```
petersteneteg, line 53: define inside of DataChannelMap?

modules/discretedata/dataset.h
```diff
@@ -0,0 +1,209 @@
...
+    /** Default copy shares Channels */
+    DataChannelMap(const DataChannelMap& copy) = default;
+
+    // Methods
+public:
+    /** Add a new channel to the set
+    *   @param channel Pointer to data, takes memory ownership
+    *   @return Shared pointer for further handling
+    */
+    SharedChannel addChannel(Channel* channel);
```
petersteneteg, line 100: In an new think like this, I don't think we need to supply the raw pointer interface. the user should always use std::make_shared to create the channels anyway. So I think this overload can be removed.

modules/discretedata/dataset.h
```diff
@@ -0,0 +1,209 @@
...
+
+    /** Add a new channel to the set
+    *   @param channel Shared pointer to data, remains valid
+    */
+    void addChannel(SharedConstChannel channel);
+
+    /** Returns the first channel from an unordered list.
+    */
+    SharedConstChannel getFirstChannel() const;
```
petersteneteg, line 109: What's the use case for getting the first one? is that a special one?

modules/discretedata/dataset.h
```diff
@@ -0,0 +1,209 @@
...
+    *   Swaps position in vector
+    *   @param channel Shared pointer to data
+    *   @return Successfull - channel was saved in the set indeed
+    */
+    bool removeChannel(SharedConstChannel channel);
+
+    /** Number of channels currently held
+    */
+    ind numChannels() const { return ChannelSet.size(); }
+
```
petersteneteg, line 156: can I ask for all the existing names?

modules/discretedata/dataset.h
```diff
@@ -0,0 +1,209 @@
...
+    std::shared_ptr<const DataChannel<T>> getChannel(
+        const std::string& name, GridPrimitive definedOn = GridPrimitive::Vertex) const;
+
+    /** Returns the specified channel if it is in the desired format, returns first instance found
+    *   @param name Unique name of requested channel
+    *   @param definedOn GridPrimitive type the channel is defined on, default 0D vertices
+    */
+    template <typename T>
+    std::shared_ptr<const DataChannel<T>> getChannel(
+        const char* name, GridPrimitive definedOn = GridPrimitive::Vertex) const;
```
petersteneteg: don't think we need overloads for const char*

modules/discretedata/dataset.h
```diff
@@ -0,0 +1,209 @@
...
+
+    /** Default copy shares Channels and Connectivity */
+    DataSet(const DataSet& copy) = default;
+
+    // Methods
+public:
+    /// Returns a const typed shared pointer to the grid, if casting is possible.
+    template <typename T>
+    std::shared_ptr<const T> getGrid() const {
+        return std::dynamic_pointer_cast<const T, const Connectivity>(Grid);
```
petersteneteg: you don't need the second template arg to dynamic_pointer_cast it is deduced

modules/discretedata/dataset.h
```diff
@@ -0,0 +1,209 @@
...
+*/
+class IVW_MODULE_DISCRETEDATA_API DataSet {
+
+    // Construction / Deconstruction
+public:
+    DataSet() = default;
+    virtual ~DataSet() = default;
+
+    /** Default copy shares Channels and Connectivity */
+    DataSet(const DataSet& copy) = default;
```
petersteneteg: what about assignment and move?

modules/discretedata/util.h
```diff
@@ -0,0 +1,69 @@
...
+#include "discretedata/connectivity/connectivity.h"
+
+namespace inviwo {
+namespace dd {
+namespace util {
+
+/** See std::hash_combine implementation from Boost
+*/
+template <class T>
+inline void CombineHash(std::size_t& hash, const T& value) {
```
petersteneteg: think we have this in inviwo-core also...

modules/discretedata/connectivity/structuredgrid.cpp
AnkeAnke: Yes, probably a very premature opt. thought. I'll add in dispatch and see how that handles it.

petersteneteg (changes requested):
Really nice work!! there are many minor issues, and some interface things that I think we might need to discuss. But a very nice start I think!

modules/discretedata/connectivity/structuredgrid.cpp
petersteneteg: I would not worry about the cost of exceptions here at all...

modules/discretedata/connectivity/structuredgrid.h
AnkeAnke: Well, there is no way to display it so far... ;) Currently there is only one space.
We would not scale positions etc. into a unit cube or similar but keep it in the original space. So I'd say we have the data in model space, deriving from spatial entity would add a world space. 

When two data sets are in different scales/units, the relative scaling should probably be done via the model space matrix. But we'd then have to be careful about scaling the data (e.g., velocity should always be in the same space as the positions, while temperature would not be scaled). Difficult topic.

petersteneteg: The spatial model could be implements as a spatial position channel if needed.

petersteneteg:
Btw you should probably add the DiscreteData module to the jenkinsfile (about here https://github.com/inviwo/inviwo/blob/b068ac59c57658a38ec4c27ed10bc9a08c0065d1/Jenkinsfile#L80-L99) so it gets compiled in jenkins.

modules/discretedata/channels/analyticchannel.h
AnkeAnke: I'd rather stay with this.
We decided to have ind (signed long long) everywhere to keep consistent in our code. Negative values as invalid marker, and the agreement that signed is the way to go.
Now, we could go with size_t in this one case, but that is not obvious to the outside user and then there is a mixed interface again.

modules/discretedata/channels/analyticchannel.h
AnkeAnke: I'll have to think and look through this again, implement numComponents_ as a template arg and find out whether it works. I know we had a reason against it, but I'm not sure what it was and whether it is still valid ;)

petersteneteg: One way would also be to just let T be what ever type you want, Float, double, glm::vec3, std::array<float, 17>, myGreatTensorType, or what ever, instead of just float or double like now. That way you don’t have to add the number of components as a template argument at all it would rather be built into T.

modules/discretedata/channels/analyticchannel.h
petersteneteg: ind is fine with me, I was more referring to the name of the function

AnkeAnke: Misunderstood you there.
As we have two different sizes (components and elements) I like it better this explicit. If there was only one I'd fully agree.

modules/discretedata/channels/analyticchannel.h
AnkeAnke: For the current setup that would work. But then having more channel types (like unsteay, hich just adds a 1.0 at the end, or another compound type) would be more tedious again.

modules/discretedata/channels/bufferchannel.h
AnkeAnke: Aye. 12bd8db10050f840bac722f260cab0376ef31a25

modules/discretedata/dataset.h
AnkeAnke: Currently the map is public anyways. But a list of available channels would be a better solution, I agree.

AnkeAnke: Added in e9b5b392941ed6bef3a2c9ed9c938f68114dae9a

AnkeAnke: Added in e9b5b392941ed6bef3a2c9ed9c938f68114dae9a

modules/discretedata/channels/analyticchannel.h
petersteneteg: But then if you would let T represent the whole type not just the element type, there would only be one size :)

modules/discretedata/channels/analyticchannel.h
petersteneteg: Not sure I agree here I believe that leaving all the detail up to the type T would give you the most freedom in this case. doing something like adding a 1.0 would be trivial in that case. Maybe you can expand more on what the problem would be?

modules/discretedata/dataset.h
AnkeAnke: I like to be able to just throw in a "name" without explicitly generating a std::string.

You talked about these basic_string_views, I'll look into it.

petersteneteg: Don't think you need to worry about that performance impact. Would say that is premature optimisation. But you still find it important, then from the api point of view you should probably have that overload for all of the functions that takes string arguments, like the non template version and getAsBuffer...
About std::string_view since it is c++17 we don't use it in core so far. since we only require c++14. But since this is a completely new module and not a mandatory one for running inviwo... I think we could require c++17 in here if you want to. Would be kind of a waste not using the lastest standard when doing this from scratch...

modules/discretedata/channels/analyticchannel.h
AnkeAnke: Scenario: I'm integrating inertial particles. You have a 3D position, 3D velocity, and a 1D 1.0 at the end. I'd like to have a
`UnsteadyChannel<CompoundChannel<BufferChannel<float>, BufferChannel<float>>>`. All have type float, and the overall size would be 7. Representing this with Vec3 in there could be cumbersome.

modules/discretedata/dataset.h
AnkeAnke: I remember there was some reason. Compiling it with actually working code right now.

modules/discretedata/channels/analyticchannel.h
petersteneteg: Interesting, I think I understand what you want to achieve. But what would happen if for example the positions where in doubles and the velocity in floats? how would this api handle that?

I would to something like
```
CompoundChannel<BufferChannel<dvec3>, BufferChannel<vec3>, AnalyticalChannel<float>>
```
and instad of returning setting a pointer to a float. I would return a std::tuple<dvec3, vec3, float> or maybe std::tuple<dvec3&,  vec3&, float> if the data should be modifiable

modules/discretedata/util.h
AnkeAnke: Any clue where I should look...?

petersteneteg: sure
https://github.com/inviwo/inviwo/blob/a385ae059c75338576940b2a2f866a75a6dbc865/include/inviwo/core/util/stdextensions.h#L669-L677

modules/discretedata/dataset.h
AnkeAnke: Actually, the error was the missing `const`. Works fine now. Removed.

modules/discretedata/channels/analyticchannel.h
AnkeAnke: Your tuple could not be processed like a math vector though. How does one normalize a vector of float and int? If different types of data are needed, they can be stored in different channels. 

The advantage that remains is the ability to store Tensors without the need to reinterpret a large vector.

We'll be discussing whether numComponents should be a template argument on Monday.

modules/discretedata/dataset.h
AnkeAnke: All should be default. Would you prefer to have all 5 there explicitly or none?

modules/discretedata/dataset.h
AnkeAnke: Removed. 414bc9f9dd1e522163234430364c04fae58d4a48

modules/discretedata/channels/analyticchannel.h
AnkeAnke: GLM does not have a template-sized vector or matrix type as far as I see. So I'll include some other vectors.

petersteneteg: It does not at the our version, but in the latest master the dimension is a template param. We should update at some point. In the mean time we have used a template alias 
https://github.com/inviwo/inviwo/blob/3764b5095d98707c8897eac6b16645df9419af5e/include/inviwo/core/util/glm.h#L743-L747
Our does not work beyond N = 4 though...

petersteneteg:
I have now updated to the latest GLM verison....

modules/discretedata/channels/analyticchannel.h
AnkeAnke: It's size now.

modules/discretedata/dataset.h
petersteneteg: loop up the guideline regarding the "role of five" or the "rule of zero"

petersteneteg (approved):

petersteneteg:
Great work!, Looking forward to start playing with it!

Issue #110, opened 2018-05-09
PCH broken after latest changes


martinfalk:
Compilation times have increased dramatically due to problems with PCH. 
E.g. the precompiled header for QtWidgets only includes a single file.

martinfalk:
It appear that the Qt include paths are not yet set.
Fix in qtwidgets/CMakeLists.txt:
set qt dependencies before calling cotire
```
# Create module
ivw_create_module(NO_PCH ${SOURCE_FILES} ${MOCED_FILES} ${HEADER_FILES})
target_link_libraries(inviwo-module-qtwidgets PRIVATE Qt5::Core Qt5::Gui Qt5::Widgets)

ivw_compile_optimize_on_target(inviwo-module-qtwidgets)
```

martinfalk:
With this change, the PCH header of qtwidget contains 42 headers from core and inviwoapplicationqt.h.

martinfalk:
@petersteneteg comments?

martinfalk:
When using MSVC, all cotire logs now contain the following warning
`cl : Command line warning D9025 : overriding '/W3' with '/W4'`

martinfalk:
Applied fix described above. Might need some testing.

PR #111, opened 2018-05-11
QtWidgets, OpenGLQt, Python3Qt, AnimationQt: fixed broken PCH, #110

martinfalk:


modules/animationqt/CMakeLists.txt
```diff
@@ -38,6 +38,9 @@ qt5_add_resources(QRC_FILE "${CMAKE_CURRENT_SOURCE_DIR}/resources/animation.qrc"
 
 #--------------------------------------------------------------------
 # Create module
-ivw_create_module(${SOURCE_FILES} ${HEADER_FILES} ${QRC_FILE})
+ivw_create_module(NO_PCH ${SOURCE_FILES} ${HEADER_FILES} ${QRC_FILE})
 target_link_libraries(inviwo-module-animationqt PRIVATE Qt5::Core Qt5::Gui Qt5::Widgets)
+
+ivw_compile_optimize_on_target(inviwo-module-eigenutils)
```
r-englund: eigenutils? guess it should be animationqt?

modules/animationqt/CMakeLists.txt
```diff
@@ -38,6 +38,9 @@ qt5_add_resources(QRC_FILE "${CMAKE_CURRENT_SOURCE_DIR}/resources/animation.qrc"
 
 #--------------------------------------------------------------------
 # Create module
-ivw_create_module(${SOURCE_FILES} ${HEADER_FILES} ${QRC_FILE})
+ivw_create_module(NO_PCH ${SOURCE_FILES} ${HEADER_FILES} ${QRC_FILE})
```
r-englund, line 41: what does the NO_PCH means? Does it mean we tell the ivw_create_module function to not set PCH, but we still we have it since we enable it a few lines bellow this?

r-englund (approved):
See comments!

modules/animationqt/CMakeLists.txt
martinfalk: yes, that is correct. It means that cotire is not run immediately when creating the module. One has to run cotire explicitly afterwards. This was necessary to be able to exclude certain paths.
Also allows us to disable PCH for a particular project/module

modules/animationqt/CMakeLists.txt
martinfalk: good catch

PR #132, opened 2018-05-25
submodule check

petersteneteg:
CMake: added a check and error messages to see that submodules have been cloned

### DIFF
--- a/cmake/verifysubmodules.cmake
+++ b/cmake/verifysubmodules.cmake
@@ -1,0 +1,49 @@
+#################################################################################
+#
+# Inviwo - Interactive Visualization Workshop
+#
+# Copyright (c) 2013-2018 Inviwo Foundation
+# All rights reserved.
+# 
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met: 
+# 
+# 1. Redistributions of source code must retain the above copyright notice, this
+# list of conditions and the following disclaimer. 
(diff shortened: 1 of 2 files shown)


cmake/verifysubmodules.cmake
```diff
@@ -0,0 +1,49 @@
...
+	file(READ ${modules_file} contents)
+    string(REPLACE "\n" ";" lines "${contents}")
+    foreach(line ${lines})
+        #^\s*path\s*=\s*(\S+)$*
+        string(REGEX MATCH "^[ \t]*path[ \t]*=[ \t]*([^ \t]*)[ \t]*$" found_item ${line})
+        if(CMAKE_MATCH_1)
+			file(GLOB dircontents RELATIVE ${dir}/${CMAKE_MATCH_1} ${dir}/${CMAKE_MATCH_1}/[^.]*)
+			list(LENGTH dircontents len)
+			if(${len} EQUAL 0)
+				message(FATAL_ERROR "Inviwo found an emtpy submodule at '${CMAKE_MATCH_1}'. "
```
ResearchDaniel, line 43: Can we try to execute this command and if it fails print error?

ResearchDaniel (approved):
Possible to run the command even?

cmake/verifysubmodules.cmake
petersteneteg: I rather have the user do this manually

PR #139, opened 2018-05-30
Mesh converter

petersteneteg:
Base: Added a mesh converter processor to convert a mesh into lines or points.

### DIFF
--- a/modules/base/algorithm/mesh/meshconverter.h
+++ b/modules/base/algorithm/mesh/meshconverter.h
@@ -1,0 +1,49 @@
+/*********************************************************************************
+ *
+ * Inviwo - Interactive Visualization Workshop
+ *
+ * Copyright (c) 2018 Inviwo Foundation
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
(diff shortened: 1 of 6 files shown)


modules/base/algorithm/mesh/meshconverter.h
```diff
@@ -0,0 +1,49 @@
...
+#include <modules/base/basemoduledefine.h>
+#include <inviwo/core/common/inviwo.h>
+
+#include <inviwo/core/datastructures/geometry/mesh.h>
+
+namespace inviwo {
+
+namespace meshutil {
+
+std::unique_ptr<Mesh> toPointMesh(const Mesh& mesh);
```
martinfalk: comments of what the function do would be nice, like copying only position and color buffer of the input mesh. Not doing anything for point meshes when calling `toLineMesh`

modules/base/processors/meshconverterprocessor.cpp
```diff
@@ -0,0 +1,86 @@
...
+};
+const ProcessorInfo MeshConverterProcessor::getProcessorInfo() const { return processorInfo_; }
+
+MeshConverterProcessor::MeshConverterProcessor()
+    : Processor()
+    , inport_("inport")
+    , outport_("outport")
+    , type_{"type",
+            "Type",
+            {{"lines", "To Lines", Type::ToLines}, {"points", "To Points", Type::ToPoints}},
```
martinfalk, line 51: why not points first, then lines?

martinfalk (approved):

modules/base/processors/meshconverterprocessor.cpp
petersteneteg: lines are more fun :)

martinfalk (approved):

PR #144, opened 2018-05-31
Fix/various

martinfalk:


martinfalk:
* improved TypedMesh (removed special cases for VS2015)
* added unittests for TypedMesh

petersteneteg (approved):

PR #145, opened 2018-05-31
Feature/various

r-englund:
2 commits that I need for my Haptics project: 
* using std::move to move objects into  the vector while dematerialisering  (objects created using factories) 
* Support for using quaternions as properties


martinfalk (approved):

petersteneteg:
You should really register the actual properties also, not just the widgets...

r-englund:
Yea, I missed that, fixed in #146

PR #148, opened 2018-06-04
Core/QtWidgets: Improved Raw volume data reader.

ResearchDaniel:
- Fixed bug where 12-bit data would not load
- Added data range, value range and units
- Renamed DataReaderDialog to VolumeDataReaderDialog


modules/qtwidgets/rawdatareaderdialogqt.cpp
```diff
@@ -43,27 +44,103 @@ namespace inviwo {
...
+                dataRangeMax_->setValue(4095);
+                valueRangeMin_->setValue(-1000);
+                valueRangeMax_->setValue(3000);
+                break;
+            case 4:  // UINT
+            case 5:  // INT
+            case 6:  // FLOAT
+            case 7:  // DOUBLE
+				dataRangeMin_->setValue(0);
+				dataRangeMax_->setValue(1);
```
r-englund: Defaulting to 0 and 1 for INT and UINT?

modules/qtwidgets/rawdatareaderdialogqt.cpp
```diff
@@ -43,27 +44,103 @@ namespace inviwo {
...
+	valueRangeMax_ = new QDoubleSpinBox();
+	valueRangeMax_->setRange(-std::numeric_limits<double>::max(), std::numeric_limits<double>::max());
+	QLabel* unit = new QLabel("Unit (m/s, HU, W)");
+	valueUnit_ = new  QLineEdit();				///< Unit, i.e. Hounsfield/absorption/W.
+
+	auto rowCount = 0;
+    dataTypeLayout->addWidget(bitDepthLabel, rowCount, 0);
+    dataTypeLayout->addWidget(bitDepth_, rowCount++, 2);
+
+	dataTypeLayout->addWidget(new QLabel("Min"), rowCount, 1);
```
r-englund: are you using tabs instead of spaces here?

modules/qtwidgets/rawdatareaderdialogqt.cpp
```diff
@@ -148,22 +225,27 @@ RawDataReaderDialogQt::RawDataReaderDialogQt() {
     mainLayout->addWidget(readOptionsBox, 5, 0, 3, 2);
     mainLayout->addWidget(buttonBox,      8, 1);
     setLayout(mainLayout);
+	// Set default to USHORT. Will Update default values of ranges 
+	bitDepth_->setCurrentIndex(2);
```
r-englund: Why default to USHORT? I guess most of you data is USHORT, most of mine isn't. So why ushort :)

modules/qtwidgets/rawdatareaderdialogqt.cpp
```diff
@@ -43,27 +44,103 @@ namespace inviwo {
...
     bitDepth_->addItem("UCHAR");
     bitDepth_->addItem("CHAR");
     bitDepth_->addItem("USHORT");
-    bitDepth_->addItem("USHORT_12");
+    bitDepth_->addItem("USHORT (12-bits)");
     bitDepth_->addItem("UINT");
     bitDepth_->addItem("INT");
     bitDepth_->addItem("FLOAT");
     bitDepth_->addItem("DOUBLE");
+	connect(bitDepth_, static_cast<void (QComboBox::*)(int)>(&QComboBox::currentIndexChanged), this, [=](int index) {
```
r-englund: Consider moving the code in the lambda to a member function since it makes this function quite long. Having it as a member function you can also call it instead of `bitDepth_->setCurrentIndex(2);` to set the default values.

modules/qtwidgets/rawdatareaderdialogqt.cpp
petersteneteg: That will not work well....

modules/qtwidgets/rawdatareaderdialogqt.cpp
r-englund: There are many places in this PR that github  show tabs where we want spaces.

modules/qtwidgets/rawdatareaderdialogqt.cpp
ResearchDaniel: Will change

modules/qtwidgets/rawdatareaderdialogqt.cpp
ResearchDaniel: Wonder if VS has changed my setting. Will fix

modules/qtwidgets/rawdatareaderdialogqt.cpp
ResearchDaniel: Better than uchar :) 
We have ok initial settings for USHORT. Others are a bit cumbersome since we don't have any good guess on initial data mapper values.

modules/qtwidgets/rawdatareaderdialogqt.cpp
ResearchDaniel: WIll do

ResearchDaniel:
@r-englund @petersteneteg I think that I've addressed all comments now

r-englund (approved):

petersteneteg (approved):

PR #155, opened 2018-06-08
Fix/volume metadata

martinfalk:


ResearchDaniel (approved):
Could add documentation in Volume.h saying that filename is stored in metadata for volumes stemming from files

martinfalk:
Good point. Will add that.

PR #166, opened 2018-06-15
Core: ForEach: Change order of template arugment to support iterating over containers where value_type is not defined

r-englund:
Core: ForEach: Change order of template arugment to support iterating over containers where value_type is not defined


include/inviwo/core/util/foreach.h
```diff
@@ -40,19 +40,18 @@
...
-        template <typename Callback, typename IT>
-        void foreach_helper(std::false_type, IT a, IT b, Callback callback, size_t = 0) {
-            std::for_each(a, b, callback);
-        }
-
-        template <typename Callback, typename IT>
-        void foreach_helper(std::true_type, IT a, IT b, Callback callback, size_t start = 0) {
-            std::for_each(a, b, [&](auto v) { callback(v, start++); });
-        }
+namespace detail {
```
r-englund, line 43: This part is just clang format (Everything in namespace detail)

petersteneteg (approved):
The dispatch pool lambda should probably capture it's closure by value and not reference....

PR #172, opened 2018-06-18
OpenGL: ClockGL: Fix for non-empty test in start()

r-englund:


modules/opengl/clockgl.cpp
```diff
@@ -50,7 +50,7 @@ void ClockGL::start() {
 
     if (!queries_.empty() && queries_.back().state == State::Started) {
         glQueryCounter(queries_.back().startId(), GL_TIMESTAMP);
-    } else if (queries_.empty() && queries_.front().state == State::Unused) {
+    } else if (!queries_.empty() && queries_.front().state == State::Unused) {
         std::rotate(queries_.begin(), queries_.begin() + 1, queries_.end());
         glQueryCounter(queries_.back().startId(), GL_TIMESTAMP);
```
r-englund, line 55: Shouldn't this code also set the state to started?

martinfalk: yes, definitely.

martinfalk (approved):
good catch.

PR #173, opened 2018-06-19
PlottingGL: Made it possible to select witch columns to show in the scatterplot matrix

r-englund:
Before this commit the scatterplot matrix would use all columns in a dataframe. Now the user can select which columns to include.

### DIFF
--- a/modules/plottinggl/processors/scatterplotmatrixprocessor.cpp
+++ b/modules/plottinggl/processors/scatterplotmatrixprocessor.cpp
@@ -80,10 +80,10 @@ ScatterPlotMatrixProcessor::ScatterPlotMatrixProcessor()
                           auto p =
                               ivec2(me->posNormalized() * dvec2(static_cast<double>(numParams_)));
                           if (p.x == p.y) {
-                              color_.setSelectedValue(id2id_[p.x]);
+                              color_.setSelectedValue(visibleIDToColumnID_[p.x]);
                           } else {
-                              selectedX_.setSelectedValue(id2id_[p.x]);
-                              selectedY_.setSelectedValue(id2id_[p.y]);
+                              selectedX_.setSelectedValue(visibleIDToColumnID_[p.x]);
+                              selectedY_.setSelectedValue(visibleIDToColumnID_[p.y]);
                           }
                       }
 
@@ -289,7 +289,7 @@ void ScatterPlotMatrixProcessor::createScatterPlots() {
         for (auto x = dataFrame.begin(); x != dataFrame.end(); ++x) {
             a++;
             if (!isIncluded(*x)) continue;
-            id2id_[numParams_++] = a;
+            visibleIDToColumnID_[numParams_++] = a;
 
             for (auto y = x + 1; y != dataFrame.end(); ++y) {
                 if (!isIncluded(*y)) continue;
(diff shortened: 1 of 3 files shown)


modules/plottinggl/processors/scatterplotmatrixprocessor.cpp
```diff
@@ -68,26 +69,27 @@ ScatterPlotMatrixProcessor::ScatterPlotMatrixProcessor()
...
-
-          },
-          MouseButton::Left, MouseState::Press)
+    , mouseEvent_("mouseEvent", "Mouse Event",
+                  [&](Event *e) {
+                      if (auto me = dynamic_cast<MouseEvent *>(e)) {
+                          auto p =
+                              ivec2(me->posNormalized() * dvec2(static_cast<double>(numParams_)));
+                          if (p.x == p.y) {
+                              color_.setSelectedValue(visibleIDToColumnID_[p.x]);
```
martinfalk, line 84: are you guaranteed that p.x/p.y is always inbetween [0, `visibleIDToColumnID_.size()`)?

modules/plottinggl/processors/scatterplotmatrixprocessor.cpp
```diff
@@ -68,26 +69,27 @@ ScatterPlotMatrixProcessor::ScatterPlotMatrixProcessor()
...
+                          } else {
+                              selectedX_.setSelectedValue(visibleIDToColumnID_[p.x]);
+                              selectedY_.setSelectedValue(visibleIDToColumnID_[p.y]);
+                          }
+                      }
+
+                  },
+                  MouseButton::Left, MouseState::Press)
```
martinfalk, line 92: why press and not release?

modules/plottinggl/processors/scatterplotmatrixprocessor.cpp
```diff
@@ -341,14 +356,34 @@ void ScatterPlotMatrixProcessor::createLabels() {
...
+        for (auto x = dataFrame.begin(); x != dataFrame.end(); ++x) {
+            if (!isIncluded(*x)) continue;
             auto tex = util::createTextTexture(textRenderer_, (*x)->getHeader(), fontSize_.get(),
                                                fontColor_.get());
             labelsTextures_.push_back(tex);
         }
     }
 }
 
+bool ScatterPlotMatrixProcessor::isIncluded(std::shared_ptr<Column> col) {
```
martinfalk, line 368: I don't like that `isIncluded()` has side effects, i.e. creates a property if it doesn't exist yet.
But what happens if I connect a different DataFrame to the inport which has a column with the same name as the old DataFrame, which you had disabled. Then this particular column would not be shown, correct? Is this what we want?

martinfalk (approved):

modules/plottinggl/processors/scatterplotmatrixprocessor.cpp
r-englund: Yes, as long as the code in other places in the processors doesnt change

modules/plottinggl/processors/scatterplotmatrixprocessor.cpp
r-englund: It was like that before, did not think of chaning it.

modules/plottinggl/processors/scatterplotmatrixprocessor.cpp
r-englund: Well, I guess you can argue for both cases. Its tricky to know if it is a completely new dataframe, or a modified version of the previous one. Hence, the code is considering a modified one. In my usecase I need it to remember it state even when columns is added/removed. 

The fact that it is created if it does not exists could be improved. In the PCP-processor I do it a bit smarter. This is solution that I quickly implemented to be able to hide cols. Got tired of copying it between computers so I made this PR.

PR #175, opened 2018-06-21
Feature/doxygen warnings

r-englund:
By adding `INCLUDE_PATH           = ${IVW_ROOT_DIR}/ext` to doxygen.cmake and modifying  `EXCLUDE_PATTERNS` to exclude `*/modules/*/ext/*` rather than `*/ext/*` we will get rid of a few hundred doxygen warnings related to `#include <warn/push>` etc. 
Furthermore, This PR addresses various doxygen warnings that we had in many places in our code.

### DIFF
--- a/include/inviwo/core/common/version.h
+++ b/include/inviwo/core/common/version.h
@@ -63,7 +63,7 @@ public:
     /**
      * \brief Parses the version. Defaults to version 1.0.0.0
      *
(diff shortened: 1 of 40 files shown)


r-englund:
BTW, on my computer the numbers of warnings goes from 500-ish to 28

petersteneteg (approved):
Nice work, great to see someone care about our doxygen....

PR #182, opened 2018-07-04
Indirect iterator

petersteneteg:
Example:
```
std::vector<std::unique_ptr<int>> vec;
for (int i = 0; i < 5; ++i) {
    vec.push_back(std::make_unique<int>(i));
}

auto b = util::makeIndirectIterator(vec.begin());
auto e = util::makeIndirectIterator(vec.end());
for (; b != e; ++b) {
    std::cout << *b << " ";
}
```
Note no need to do the extra deref of the unique_ptr since that is handled by the indirect Iterator
*b will be a reference to int.

The use case if to container types that stores items using a vector of pointers, but want to expose an iterator directly to the item not to the pointer...

Also some bug fixes for other iterators, operator* has to be const for an iterator to be standard conforming...

### DIFF
--- a/include/inviwo/core/util/indirectiterator.h
+++ b/include/inviwo/core/util/indirectiterator.h
@@ -172,15 +172,15 @@ struct IndirectIterator {
     }
 
     template <typename I = Iter, typename = require_t<std::random_access_iterator_tag, I>>
-    reference operator[](difference_type i) {
+    reference operator[](difference_type i) const {
         return *iterator_[i];
     }
 
-    reference operator*() { 
+    reference operator*() const { 
         return **iterator_; 
     }
 
-    pointer operator->() { return detail_indirect::asPointer<is_const>::get(*(iterator_.operator->())); }
+    pointer operator->() const { return detail_indirect::asPointer<is_const>::get(*(iterator_.operator->())); }
 
     const Iter& base() const {return iterator_;}
     Iter& base() {return iterator_;}
--- a/include/inviwo/core/util/indirectiterator.h
+++ b/include/inviwo/core/util/indirectiterator.h
@@ -32,6 +32,7 @@
 
 #include <iterator>
 #include <type_traits>
+#include <memory>
 
 namespace inviwo {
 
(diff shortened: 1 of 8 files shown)


include/inviwo/core/util/indirectiterator.h
```diff
@@ -0,0 +1,225 @@
...
+    static constexpr const T* get(const std::unique_ptr<T, D>& p) {
+        return p.get();
+    }
+    template <typename T>
+    static constexpr const T* get(const std::shared_ptr<T>& p) {
+        return p.get();
+    }
+};
+template <>
+struct asPointer<false> {
```
ResearchDaniel, line 91: Difference to `asPointer<true>`?

include/inviwo/core/util/indirectiterator.h
```diff
@@ -0,0 +1,225 @@
...
+    template <typename T>
+    static constexpr T* get(const std::shared_ptr<T>& p) {
+        return p.get();
+    }
+};
+
+}  // namespace detail_indirect
+
+template <typename Iter, bool PropagateConst = true>
+struct IndirectIterator {
```
ResearchDaniel: Could add PR comment as documentation here to explain what Indirect refers to.

include/inviwo/core/util/stdextensions.h
```diff
@@ -375,6 +375,19 @@ inline iter_range<Iter> as_range(std::pair<Iter, Iter> const& x) {
     return iter_range<Iter>(x);
 }
 
+template <class Cont>
```
ResearchDaniel: Cont as in Container I guess? Would be more clear if spelled out

ResearchDaniel (approved):

include/inviwo/core/util/indirectiterator.h
petersteneteg: The bool is for const or not

include/inviwo/core/util/indirectiterator.h
petersteneteg: Will do

PR #183, opened 2018-07-05
Rawreader

petersteneteg:
Made the raw reader remember it's state

### DIFF
--- a/modules/qtwidgets/rawdatareaderdialogqt.h
+++ b/modules/qtwidgets/rawdatareaderdialogqt.h
@@ -56,12 +56,14 @@ public:
     virtual dvec3 getSpacing() const override;
     virtual bool getEndianess() const override;
     virtual DataMapper getDataMapper() const override;
+    virtual size_t getDataOffset() const override;
 
     virtual void setFormat(const DataFormatBase* format) override;
     virtual void setDimensions(uvec3 dim) override;
     virtual void setSpacing(dvec3 spacing) override;
     virtual void setEndianess(bool endian) override;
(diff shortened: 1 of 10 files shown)


modules/qtwidgets/rawdatareaderdialogqt.h
```diff
@@ -74,9 +80,11 @@ class IVW_MODULE_QTWIDGETS_API RawDataReaderDialogQt : public VolumeDataReaderDi
     QLineEdit* spaceY_;
     QLineEdit* spaceZ_;
 
+    /*
     QSpinBox* timeSteps_;
     QSpinBox* headerOffset_;
```
weinkauf: Offset is super important for a raw reader! Why do we remove this?

petersteneteg: Because there was only the widget thing, the value was never used anywhere... I left the code in there so we can add support for it

weinkauf: Understood. I see it now.

core\io\rawvolumereader.cpp, Line 129 uses an offset of "0u" and not the value from the dialog. It should be easy to fix.

Shall I create a commit on this branch that fixes this?

petersteneteg: Sure

martinfalk (approved):

petersteneteg:
note inviwo-dev #1373

petersteneteg:
will merge this now, since it been more then a month... but any new work in it's own PR...

PR #184, opened 2018-07-07
Animation updates

petersteneteg:
Still needs more work in the gui, but Im on vacation now...


petersteneteg:
see https://github.com/inviwo/inviwo/pull/214

PR #193, opened 2018-07-26
README: Need full path url to image to work in doxygen as well

r-englund:


r-englund:
There are more issues, will create a new PR

Issue #192, opened 2018-07-26
Issues with CollapsibleGroupBoxWidgetQt


martinfalk:
`CollapsibleGroupBoxWidgetQt::onDidProperty()` is broken for two reasons when a property does not have a corresponding property widget.
1) newly added widgets are inserted into the vector holding the property widgets. This does not account for potentially "missing" widgets, i.e. the order of the property widgets might not correspond to the order in the property owner. In case the widget does not exist, nothing is added to the property widget vector!
2) `QGridLayout` does **not** support inserting rows into the layout. Adding a widget to an already occupied cell results in both widgets showing up on top of each other.

For 1) see
https://github.com/inviwo/inviwo/blob/e4007c17c286099a3d395cf6cd492ef32d3a849f/modules/qtwidgets/properties/collapsiblegroupboxwidgetqt.cpp#L324-L330

martinfalk:
This also affects `CollapsibleGroupBoxWidgetQt::onWillRemoveProperty()` since the index is used to directly look up widgets.
https://github.com/inviwo/inviwo/blob/e4007c17c286099a3d395cf6cd492ef32d3a849f/modules/qtwidgets/properties/collapsiblegroupboxwidgetqt.cpp#L339-L346

petersteneteg:
In the animation bransch we reorder widget by just removing add adding all off them. Still seems very fast...

petersteneteg:
Maybe we need a dummy for when the widgets are missing

martinfalk:
Working on a solution which creates a new grid layout when inserting new properties. When Properties are added at the end, they are simply appended.

martinfalk:
Also adding nullptr to the property widget vector so property and property widget vectors are in sync.

PR #195, opened 2018-07-26
Feature/listproperty

martinfalk:
* revised and improved the ListProperty by @xeTaiz  (ListProperty was extracted in commit 5b0e9ed)
* added a PropertyWidget for the ListProperty
* extended PropertyOwner slightly to remove properties based on an index
* fixes regarding CollapsibleGroupBoxWidgetQt, #192

A `ListProperty` holds a number of "prefab" objects (unique ptr to properties) which are used to instantiate new list entries. The property widget features small "x" buttons for removing individual elements (if removal is enabled) while a "+" button next to the property label adds new elements (if adding is enabeld). In case multiple prefabs exist, a dropdown menu is shown when pressing the "+" button.
image

Alternatively, items can be added/removed via the context menu of the list property.
image

### Usage
```c++
// using a single prefab object and at most 10 elements
ListProperty listProperty("myListProperty", "My ListProperty",
    std::make_unique<BoolProperty>("boolProp", "BoolProperty", true), 10);

// multiple prefab objects
ListProperty listProperty("myListProperty", "My List Property", 
    []() {
        std::vector<std::unique_ptr<Property>> v;
        v.emplace_back(std::make_unique<IntProperty>("template1", "Template 1", 5, 0, 10));
        v.emplace_back(std::make_unique<IntProperty>("template2", "Template 2", 2, 0, 99));
        return v;
    }());
```

This also works when using different types of properties as prefab objects:
```c++
ListProperty listProperty("myListProperty", "My List Property", 
    []() {
        std::vector<std::unique_ptr<Property>> v;
        v.emplace_back(std::make_unique<BoolProperty>("boolProperty1", "Boolean Flag", true));
        v.emplace_back(std::make_unique<TransferFunctionProperty>("customTF1", "Transfer Function"));
        v.emplace_back(std::make_unique<IntProperty>("template1", "Template 1", 5, 0, 10));
        return v;
    }());
```

Prefabs can be added later on as well using `ListProperty::addPrefab(std::unique_ptr<Property>&& p)`.

### TODOs
- [ ] undo after removing subproperties messes up ordering since only the deleted property will be restored and added at the end.
- [x] linking needs improving
- [ ] support minimum number of list elements


include/inviwo/core/properties/listproperty.h
```diff
@@ -0,0 +1,207 @@
...
+/**
+ * \class ListProperty
+ * \brief A property that has specified sub-properties which can be added using the graphical user
+ * interface.
+ *
+ * Represents a list of properties. Properties can be added by using the prefab objects registered
+ * with the list property. The prefab objects serve as templates for instatiating new list entries.
+ *
+ * The UI flags (ListPropertyUIFlags) determine whether the widget will allow to add and/or remove
+ * list entries. The number of list elements is limited by setting maxNumberOfElements. A value of 0
```
petersteneteg, line 63: A min number might also be useful, in many cases we always want atleast one I think?

include/inviwo/core/properties/listproperty.h
```diff
@@ -0,0 +1,207 @@
...
+     */
+    size_t getPrefabCount() const;
+
+    /**
+     * \brief add a new prefab object \p p to be used as template when instantiating new list
+     * elements
+     *
+     * @param p  prefab object
+     */
+    void addPrefab(std::unique_ptr<Property>&& p);
```
petersteneteg: You don't need the && here, can just pass it by value...

include/inviwo/core/properties/listproperty.h
```diff
@@ -0,0 +1,207 @@
...
+
+    InviwoPropertyInfo();
+
+    ListProperty(std::string identifier, const std::string& displayName,
+                 size_t maxNumberOfElements = 0,
+                 ListPropertyUIFlags uiFlags = ListPropertyUIFlag::Add | ListPropertyUIFlag::Remove,
+                 InvalidationLevel invalidationLevel = InvalidationLevel::InvalidResources,
+                 PropertySemantics semantics = PropertySemantics::Default);
+    ListProperty(std::string identifier, const std::string& displayName,
+                 std::vector<std::unique_ptr<Property>>&& prefabs, size_t maxNumberOfElements = 0,
```
petersteneteg: No need for && just pass by value...

modules/qtwidgets/properties/collapsiblegroupboxwidgetqt.cpp
```diff
@@ -393,6 +346,25 @@ void CollapsibleGroupBoxWidgetQt::onSetVisible(Property* property, bool visible)
     if (property == property_) {
         PropertyWidgetQt::onSetVisible(property, visible);
     }
+
+    if (isChildRemovable()) {
+        for (auto&& elem : util::zip(properties_, propertyWidgets_)) {
+            if (get<0>(elem) == property) {
```
petersteneteg: You can use elem.first() or elem.second()

modules/qtwidgets/properties/collapsiblegroupboxwidgetqt.cpp
```diff
@@ -412,4 +384,196 @@ const std::vector<PropertyWidgetQt*>& CollapsibleGroupBoxWidgetQt::getPropertyWi
...
+        addPropertyWidget(propertyWidgetGroupLayout_, propertyWidgetGroupLayout_->rowCount(),
+                          propertyWidget);
+    } else {
+        // recreate property widget grid layout from scratch since QGridLayout does not support
+        // inserting rows.
+        //
+        // Assigning another widget to an already occupied grid cell is possible, but the
+        // widgets are put on top of each other.
+        auto layoutWidget = createPropertyLayoutWidget();
+        auto layout = static_cast<QGridLayout*>(layoutWidget->layout());
```
petersteneteg, line 529: This does seems complicated, could one not just remove all widgets from the layout and then add them back in the right order?

modules/qtwidgets/properties/listpropertywidgetqt.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+namespace inviwo {
+
+ListPropertyWidgetQt::ListPropertyWidgetQt(ListProperty* property)
+    : CompositePropertyWidgetQt(property), listProperty_(property) {
+
+    setShowIfEmpty(true);
+
+    // inject "add" button in the header of the composite property for adding list elements
+    if (auto headerlayout = dynamic_cast<QHBoxLayout*>(layout()->itemAt(0)->layout())) {
```
petersteneteg: Should this not be an assert rather them a if?

modules/qtwidgets/properties/listpropertywidgetqt.cpp
```diff
@@ -0,0 +1,188 @@
...
+                    }
+                }
+            }
+        });
+
+        if (listProperty_) {
+            addItemButton_->setVisible((listProperty_->getUIFlags() & ListPropertyUIFlag::Add) ==
+                                       ListPropertyUIFlag::Add);
+
+            // listProperty_->PropertyOwnerObservable::addObserver(this);
```
petersteneteg: Remove?

modules/qtwidgets/properties/listpropertywidgetqt.cpp
```diff
@@ -0,0 +1,188 @@
...
+        }
+
+        headerlayout->insertWidget(headerlayout->count() - 2, addItemButton_);
+    }
+}
+
+void ListPropertyWidgetQt::updateFromProperty() {
+    CompositePropertyWidgetQt::updateFromProperty();
+
+    if (listProperty_) {
```
petersteneteg, line 103: Can listproperty be null? If so why?

modules/qtwidgets/properties/listpropertywidgetqt.cpp
```diff
@@ -0,0 +1,188 @@
...
+    }
+}
+
+bool ListPropertyWidgetQt::isChildRemovable() const {
+    return ((listProperty_->getUIFlags() & ListPropertyUIFlag::Remove) ==
+            ListPropertyUIFlag::Remove);
+}
+
+std::unique_ptr<QMenu> ListPropertyWidgetQt::getContextMenu() {
+    auto menu = PropertyWidgetQt::getContextMenu();
```
petersteneteg: CompositePropertyWidgetQt?

modules/qtwidgets/properties/listpropertywidgetqt.cpp
```diff
@@ -0,0 +1,188 @@
...
+                listProperty_->clearInitiatingWidget();
+            });
+        }
+
+        menu->addSeparator();
+    }
+
+    return menu;
+}
+/*
```
petersteneteg: Remove?

src/core/properties/listproperty.cpp
```diff
@@ -0,0 +1,240 @@
...
+void ListProperty::set(const ListProperty* src) {
+    NetworkLock l(this);
+    // check for matching prefab types first
+    if (getPrefabIDs() == src->getPrefabIDs()) {
+        // TODO: should we sync/consider the UI flags here as well?
+
+        maxNumElements_ = src->maxNumElements_;
+
+        while (size()) {
+            removeProperty(static_cast<size_t>(0));
```
petersteneteg: What are you doing here?

petersteneteg: Only need to clone it the types are different, the default should be jpto just call set on the sub properties

src/core/properties/listproperty.cpp
```diff
@@ -0,0 +1,240 @@
...
+    }
+}
+
+void ListProperty::setMaxNumberOfElements(size_t n) {
+    maxNumElements_ = n;
+    if (n > 0) {  // n = 0 means no limit
+        // remove superfluous list items
+        const bool removedElements = size() > n;
+        while (size() > n) {
+            removeProperty(size() - 1);
```
petersteneteg, line 157: Are these properties deleted?

petersteneteg (changes requested):

include/inviwo/core/properties/listproperty.h
martinfalk: How would that one then be initialized? In particular if multiple prefabs exist.
And how should this be affected by `clear()`?

modules/qtwidgets/properties/collapsiblegroupboxwidgetqt.cpp
martinfalk: then one should potentially update the examples for `util::zip` as well...

modules/qtwidgets/properties/collapsiblegroupboxwidgetqt.cpp
martinfalk: I have tried that. The problem with just exchanging the layout is that the moment you reparent a layout, all its child widgets no longer have a parent. Thus, each property widget is suddenly a floating window. The second issue with `Q*Layout` is that the docs specifically state that a layout needs to be parented to a widget **prior** adding any child items. This means I have to create a second container widget plus layout first (`layoutWidget`) before I can start moving the items. Moving is done implicitly by calling `addWidget()` on the new layout. This will remove the widget from all former layouts.

image

modules/qtwidgets/properties/listpropertywidgetqt.cpp
martinfalk: I wanted to be not as restrictive. But I guess it could be an assertion.

modules/qtwidgets/properties/listpropertywidgetqt.cpp
martinfalk: Will do, forgot that one.

modules/qtwidgets/properties/listpropertywidgetqt.cpp
martinfalk: I don't know, but I have seen these checks in multiple places, i.e. property widgets.

modules/qtwidgets/properties/listpropertywidgetqt.cpp
martinfalk: Hm, will think about that. Could be that CompositePropertyWidgetQt at some point might add stuff that we are not interested in.

modules/qtwidgets/properties/listpropertywidgetqt.cpp
martinfalk: will do.

include/inviwo/core/properties/listproperty.h
petersteneteg: Would need to pass also the sub properties in the constructor..., might get to complicated...

src/core/properties/listproperty.cpp
martinfalk: This was the first draft for linking. Open for suggestions.
But when cloning only non-matching properties, PropertyOwner needs functions for inserting properties, i.e. `insertProperty()`, so the order is similar.

src/core/properties/listproperty.cpp
martinfalk: This calls the regular `removeProperty()` function which deletes owned properties (returns nullptr), or removes them from the owner and returns a pointer.

modules/qtwidgets/properties/collapsiblegroupboxwidgetqt.cpp
petersteneteg: True

modules/qtwidgets/properties/collapsiblegroupboxwidgetqt.cpp
petersteneteg: But why move them, just remove all of them then addd them back, you only ever need one layout? Or am I missing something?

petersteneteg: We do
        for (auto w : widgets) {
            keyframesLayout_->removeWidget(w.second);
        }
        for (auto w : widgets) {
            keyframesLayout_->addWidget(w.second);
        }

In the animation stuff, and that seems to work?

modules/qtwidgets/properties/listpropertywidgetqt.cpp
petersteneteg: That way at least you get an issue when you change the compositepropertywidget...

modules/qtwidgets/properties/listpropertywidgetqt.cpp
petersteneteg: Document that in that case, the default is to call the parent...

src/core/properties/listproperty.cpp
petersteneteg: You can fall back on cloning all if one is different... linking appens alot and should be as cheap as possible...

src/core/properties/listproperty.cpp
petersteneteg: Ok

modules/qtwidgets/properties/collapsiblegroupboxwidgetqt.cpp
martinfalk: That needs testing. There might be side effects like widgets being temporarily unparented. In addition, the widgets here might have another widget on the same row (for removal). But the widget does not know about it.

modules/qtwidgets/properties/listpropertywidgetqt.cpp
martinfalk: Will check CompositePropertyWidgetQt

src/core/properties/listproperty.cpp
martinfalk: Cloning also has severe downsides. I tried it with TF properties in the list. It horribly failed, due to missing callbacks, wrong VolumePort, etc.

martinfalk: I think callbacks are a general problem when "cloning" properties. @r-englund mentioned something similar.

modules/qtwidgets/properties/collapsiblegroupboxwidgetqt.cpp
martinfalk: `addWidget()` for a QGridLayout (http://doc.qt.io/qt-5/qgridlayout.html#addWidget) requires at least a row and column argument. Since QGridLayout does not support insertion and/or updating the grid rows, we would have to append the widgets always after the last row. Thereby increasing the row count every single time by n when we update the layout.

martinfalk: The only difference to the solution used right now would be that `util::OnScopeExit freeOldLayout` would not be required.

modules/qtwidgets/properties/listpropertywidgetqt.cpp
martinfalk: Settling for an `inviwo::Exception` since an assertion might be easily missed.

include/inviwo/core/properties/listproperty.h
martinfalk: Will not fix for now.

include/inviwo/core/properties/listproperty.h
martinfalk: fixed

include/inviwo/core/properties/listproperty.h
martinfalk: fixed

modules/qtwidgets/properties/collapsiblegroupboxwidgetqt.cpp
martinfalk: fixed and docs extended

modules/qtwidgets/properties/listpropertywidgetqt.cpp
martinfalk: fixed

src/core/properties/listproperty.cpp
martinfalk: List elements are now matched based on their class identifier. Missing elements are cloned while superfluous list entries are deleted.

modules/qtwidgets/properties/listpropertywidgetqt.cpp
petersteneteg: How do you miss an assertion?!? It will terminate the app

petersteneteg (approved):

modules/qtwidgets/properties/listpropertywidgetqt.cpp
martinfalk: Of course it will do that. If assertions are enabled, i.e. either compiling debug or force-enabling them.

martinfalk: could change it to an assertion if you want.

PR #200, opened 2018-07-30
Core: fixed TF linking issue, closes #199

### LINKED ISSUE #199: TF Linking broken
Scenario: Link two transfer functions. Create a new TF point in one of the TFs. Manipulate the same TF point in the other TF. This will not affect the second TF!

martinfalk:


r-englund (approved):

src/core/datastructures/tfprimitive.cpp
```diff
@@ -37,6 +37,10 @@ TFPrimitive::TFPrimitive(double pos, const vec4& color) : data_({pos, color}) {}
 
 TFPrimitive::TFPrimitive(const TFPrimitiveData& data) : data_(data) {}
 
+TFPrimitive::TFPrimitive(const TFPrimitive& rhs) : data_(rhs.data_) {}
```
petersteneteg, line 40: Is this not equal to default? Maybe comment on why in that case?

petersteneteg (commented):
?

src/core/datastructures/tfprimitive.cpp
martinfalk: necessary because the constructor of Observable is not called!

martinfalk: see #199

petersteneteg: There should really be a comment about that...

martinfalk: added comment in 63f25e5

PR #201, opened 2018-07-31
Feature/tfselector

martinfalk:
* new processor `Transfer Function Selector` using a ListProperty of TFs. Allows creating multiple TF presets and then select one as output.
* fixed TF preview when the TF does not contain any primitives (widget was showing a black to white gradient)

### DIFF
--- a/modules/base/processors/tfselector.h
+++ b/modules/base/processors/tfselector.h
@@ -1,0 +1,106 @@
+/*********************************************************************************
+ *
+ * Inviwo - Interactive Visualization Workshop
+ *
+ * Copyright (c) 2018 Inviwo Foundation
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ *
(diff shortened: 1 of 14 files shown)


modules/base/processors/tfselector.h
```diff
@@ -0,0 +1,106 @@
...
+#include <inviwo/core/properties/listproperty.h>
+#include <inviwo/core/properties/propertyownerobserver.h>
+#include <inviwo/core/ports/imageport.h>
+
+namespace inviwo {
+
+/** \docpage{org.inviwo.TFSelector, TFSelector}
+ * ![](org.inviwo.TFSelector.png?classIdentifier=org.inviwo.TFSelector)
+ * allows to select a transfer function from a number of presets which can be added and modified by
+ * the user. The image is passed through without any modifications.
```
ResearchDaniel, line 50: Having an image as input/output is a hack. Should be mentioned that? Does it work without any ports? 
How about encapsulating this into a composite property so that it can be reused?

ResearchDaniel (approved):

modules/base/processors/tfselector.h
martinfalk: The port is necessary due to the use of event properties in order to cycle through the various TFs using the keyboard. That is the processor needs to be connected to the canvas.

martinfalk: Will consider creating a composite property.

petersteneteg: I also think the setup with the processor is a bit of a hack, should probably be a composite and not a processor, but then we don’t what it in every processor with a tf... do we need some other way of handling this kind of things?

martinfalk: I disagree. A composite would definitely come in handy, but in order to provide the TF selection functionality for all other processors featuring a TF (in particular with keyboard shortcuts)

Issue #204, opened 2018-08-10
util::KeepTrueWhileInScope is broken


martinfalk:
When using `util::KeepTrueWhileInScope`, the bool value is set to `true`. But in the destructor the bool value is set to `false` independent of the previous state.

https://github.com/inviwo/inviwo/blob/2e1c41a9d1209dfa5fb9ea91928167eca1381754/include/inviwo/core/util/raiiutils.h#L40-L47

r-englund:
This is what is was designed to do. To set the value to true when created and to false when destroyed.

martinfalk:
yeah, but what if it is used on the same variable twice in the same block? e.g. #203

petersteneteg:
I would tend to agree with r-englund here. Seems unwise to use it in a nested fashion. But I don't have a strong opinion either way. But changing this behaviour need quite heavy testing, since we use it quite a lot, especially in the serialisation. And we have used it with this behaviour for many years...

martinfalk:
Unwise yes. But you cannot always guarantee that this does not happen when using callbacks.

PR #205, opened 2018-08-10
Fix/various

martinfalk:


include/inviwo/core/util/raiiutils.h
```diff
@@ -39,11 +39,18 @@ namespace util {
 
 class IVW_CORE_API KeepTrueWhileInScope {
 public:
-    KeepTrueWhileInScope(bool* b) : variable_(b) { if(variable_) (*variable_) = true; }
-    ~KeepTrueWhileInScope() { if(variable_) (*variable_) = false; }
```
petersteneteg: Are we sure we don’t depend on this behavior anywhere?

petersteneteg (approved):

include/inviwo/core/util/raiiutils.h
martinfalk: Cannot guarantee that. People might have a different interpretation of what `KeepTrueWhileInScope` should be doing (see comment by @r-englund in #204)

martinfalk: Still wondering though why the canvas was shown in the first place. Maybe something within Qt changed slightly and the visibility changes now trigger signals. Who knows ...

petersteneteg: I would be very careful with this, might break serialisation in hard to notice ways...

martinfalk: And it might be incorrect in many others as well.

PR #217, opened 2018-08-30
VolumeSlice: Added synchronization from world position to slice index…

Alboin:
… and plane position.

### DIFF
--- a/modules/basegl/processors/volumeslicegl.cpp
+++ b/modules/basegl/processors/volumeslicegl.cpp
@@ -250,6 +250,7 @@ VolumeSliceGL::VolumeSliceGL()
 
     worldPosition_.setReadOnly(true);
     addProperty(worldPosition_);
+    worldPosition_.onChange([this]() { updateFromWorldPosition(); });
 
     addProperty(handleInteractionEvents_);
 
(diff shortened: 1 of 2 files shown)


modules/basegl/processors/volumeslicegl.cpp
```diff
@@ -736,6 +737,31 @@ void VolumeSliceGL::rotationModeChange() {
...
+        const ivec3 indexPos(ivec3(worldToIndex * vec4(worldPosition_.get(), 1.0f)) + ivec3(1));
+
+        NetworkLock lock(this);
+        sliceX_.set(indexPos.x);
+        sliceY_.set(indexPos.y);
+        sliceZ_.set(indexPos.z);
+
+        planePosition_.set(texturePos);
+        invalidateMesh();
+    }
```
petersteneteg, line 772: Nice, Since this on updates both slize* and placePosistion I think it would be more constant if sliceChange and positionChange also updated worldPosistion instead of renderPositionIndicator as it is now...

petersteneteg (changes requested):
Nice, would like a minor tweak to make it more consistant

ResearchDaniel (approved):
Sweet!

petersteneteg (approved):

PR #218, opened 2018-08-31
Misc

petersteneteg:


modules/animationqt/animationviewqt.cpp
```diff
@@ -42,6 +42,7 @@
 #include <QGridLayout>
 #include <QScrollBar>
 #include <QGraphicsScene>
+#include <QStyleOption>
```
martinfalk, line 45: probably don't need this?

modules/qtwidgets/properties/syntaxhighlighter.cpp
```diff
@@ -64,8 +64,8 @@ template <>
 void SyntaxHighligther::loadConfig<None>() {
     auto ivec4toQtColor = [](const ivec4& i) { return QColor(i.r, i.g, i.b, i.a); };
     auto settings = InviwoApplication::getPtr()->getSettingsByType<QtWidgetsSettings>();
-    QColor textColor = ivec4toQtColor(settings->glslTextColor_.get());
-    backgroundColor_ = ivec4toQtColor(settings->glslBackgroundColor_.get());
+    QColor textColor = ivec4toQtColor(settings->pyTextColor_.get());
+    backgroundColor_ = ivec4toQtColor(settings->pyBGColor_.get());
```
martinfalk, line 68: Do you know why this is called `pyTextColor_`? Isn't this a global color for the syntax highlighter?

martinfalk (approved):

modules/qtwidgets/properties/syntaxhighlighter.cpp
petersteneteg: There is a py and a glsl version...

modules/animationqt/animationviewqt.cpp
petersteneteg: True

PR #223, opened 2018-09-04
Base: Made information and basis properties read-only in VolumeSequenceSource

ResearchDaniel:
Made information and basis properties read-only in VolumeSequenceSource and disabled their usage since they where applied to all values no matter if you had changed them or not. it did not make sense to override these settings for all volumes. Still kept for information purposes.


petersteneteg (approved):
Think we should discuss what 4D data structures that we want in the meeting...

PR #228, opened 2018-09-11
Core: Fix division by zero when camera animation is enabled

ResearchDaniel:


src/core/interaction/trackball.cpp
```diff
@@ -866,10 +866,17 @@ void inviwo::Trackball::animate() {
             const glm::quat identity(1.0f, 0.0f, 0.0f, 0.0f);
             const float t = 0.1f;
             const float dot = glm::dot(lastRot_, identity);
-            const float theta = std::acos(dot);
-            const float sintheta = std::sin(theta);
-            lastRot_ = lastRot_ * (std::sin((1.0f - t) * theta) / sintheta) +
-                       identity * (std::sin(t * theta) / sintheta);
+			if (dot != 0) {
```
petersteneteg: should this not be != 1?

ResearchDaniel: You are right, fixed in new commit. I could not reproduce the problem either but have added the fix anyway

petersteneteg (approved):

petersteneteg (approved):

PR #234, opened 2018-09-13
Exception stack trace

petersteneteg:
Added a stacktrace to our exceptions, and the ability to log and view and ignore any unchought inviwo exception...

A potential problem is the overhead of creating the stacktrace in our exception

image


martinfalk (approved):
Might be a bit heavy on both memory and time needed to create the stack trace. Need an option for disabling/enabling it!
Maybe in CMake or some global options...

petersteneteg:
replaced by https://github.com/inviwo/inviwo/pull/250

Issue #236, opened 2018-09-14
Image Layers are not initialized


martinfalk:
Since ImageOutports always have some dummy data (8 by 8 pixel), the following network shows random artifacts despite the image source processor does not have any data set to its outport.

Here, the filename of ImageSource is not set.

![image](https://user-images.githubusercontent.com/9251300/45541250-da84f680-b80e-11e8-8052-922683fc87ba.png)

martinfalk:
Should a color layer of an `Image` always be initialized?

martinfalk:
Same applies when inspecting an unset ImageOutport with the PortInspector

![image](https://user-images.githubusercontent.com/9251300/45546523-d14f5600-b81d-11e8-8e79-719ef73293d8.png)

ResearchDaniel:
I could reproduce it on Mac as well. Would setting the outport data to null in ImageSource resolve the issue?

martinfalk:
Might solve this. But at the moment, the process function is only evaluated once you provide an existing file (connected to isReady()).

Need to test.

ResearchDaniel:
Have a fix for it

martinfalk:
Setting the image output of ImageSource to nullptr does the trick.
In `ImageSource::ImageSource()`:
```
    auto func = static_cast<void (ImageOutport::*)(std::shared_ptr<Image>)>(&ImageOutport::setData);
    (outport_.*(func))(nullptr);
```

martinfalk:
When creating an `Image` from a single layer, e.g. a color image loaded from disk as done in `Image Source`, the other layers (picking, depth) are not initialized.

Issue #233, opened 2018-09-12
ImageScaling processor broken


martinfalk:
The `ImageScaling` processor interferes with resizing events even when disabled. Input is not passed through when "Enabled" property is set to `disabled`.

PR #243, opened 2018-09-14
Base: Set ImageSource outport to null by default and if no file has been specified

ResearchDaniel:
…Closes issue #236, #233

### DIFF
--- a/modules/base/processors/imagesource.cpp
+++ b/modules/base/processors/imagesource.cpp
@@ -53,8 +53,7 @@ ImageSource::ImageSource(InviwoApplication* app, const std::string& file)
     , file_("imageFileName", "File name", file, "image")
     , imageDimension_("imageDimension_", "Dimension", ivec2(0), ivec2(0), ivec2(10000), ivec2(1),
                       InvalidationLevel::Valid, PropertySemantics("Text")) {
-    // Remove default data created by ImageOutport
-    outport_.setData(static_cast<Image*>(nullptr));
+
     addPort(outport_);
 
     auto rf = app_->getDataReaderFactory();
@@ -66,16 +65,13 @@ ImageSource::ImageSource(InviwoApplication* app, const std::string& file)
 
     imageDimension_.setReadOnly(true);
     addProperty(imageDimension_);
-    
+
     isReady_.setUpdate([this]() { return filesystem::fileExists(file_.get()); });
     file_.onChange([this]() { isReady_.update(); });
 }
 
 void ImageSource::process() {
-    if (file_.get().empty()) {
-        outport_.setData(static_cast<Image*>(nullptr));
-        return;
-    }
+    if (file_.get().empty()) return;
 
     std::string ext = filesystem::getFileExtension(file_.get());
     auto factory = app_->getDataReaderFactory();
--- a/modules/base/processors/imagesource.cpp
+++ b/modules/base/processors/imagesource.cpp
@@ -53,7 +53,8 @@ ImageSource::ImageSource(InviwoApplication* app, const std::string& file)
     , file_("imageFileName", "File name", file, "image")
     , imageDimension_("imageDimension_", "Dimension", ivec2(0), ivec2(0), ivec2(10000), ivec2(1),
                       InvalidationLevel::Valid, PropertySemantics("Text")) {
-
+    // Remove default data created by ImageOutport
+    outport_.setData(static_cast<Image*>(nullptr));
     addPort(outport_);
 
     auto rf = app_->getDataReaderFactory();
@@ -65,13 +66,16 @@ ImageSource::ImageSource(InviwoApplication* app, const std::string& file)
 
     imageDimension_.setReadOnly(true);
     addProperty(imageDimension_);
-
+    
     isReady_.setUpdate([this]() { return filesystem::fileExists(file_.get()); });
     file_.onChange([this]() { isReady_.update(); });
 }
 
 void ImageSource::process() {
-    if (file_.get().empty()) return;
+    if (file_.get().empty()) {
+        outport_.setData(static_cast<Image*>(nullptr));
+        return;
+    }
 
     std::string ext = filesystem::getFileExtension(file_.get());
     auto factory = app_->getDataReaderFactory();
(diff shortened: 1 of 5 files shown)


modules/basegl/processors/imageprocessing/imagescaling.cpp
```diff
@@ -74,16 +75,14 @@ ImageScaling::ImageScaling()
...
-    };
+    auto triggerRescale = [this]() { resizeInports(); };
 
     enabled_.onChange(triggerRescale);
     scalingFactor_.onChange([this]() {
-        customFactor_.setVisible(scalingFactor_.get() < 0.0);
-        ResizeEvent e(calcInputImageSize());
-        inport_.propagateEvent(&e);
+        if (enabled_) {
+            customFactor_.setVisible(scalingFactor_.get() < 0.0);
```
ResearchDaniel, line 80: <= 0?

modules/basegl/processors/imageprocessing/imagescaling.cpp
```diff
@@ -101,37 +100,48 @@ void ImageScaling::process() {
...
-    if (outport_.hasData()) {
-        size = outport_.getDimensions();
-    }
-    if (enabled_.get()) {
-        const double factor =
-            (scalingFactor_.get() < 0.0) ? customFactor_.get() : scalingFactor_.get();
-        size = size2_t(dvec2(size) / factor);
+    if (!enabled_) return lastValidOutputSize_;
+
+    const double factor = (scalingFactor_.get() < 0.0) ? customFactor_.get() : scalingFactor_.get();
```
ResearchDaniel, line 125: <= 0.0
Otherwise divide by 0 could occur on next line

modules/basegl/processors/imageprocessing/imagescaling.cpp
martinfalk: Cannot happen, `scalingFactor_` is an OptionProperty with the following options:
```
{"one-eighth", "12.5%", 0.125},
                         {"one-fourth", "25%", 0.25},
                         {"one-half", "50%", 0.5},
                         {"three-quarters", "75%", 0.75},
                         {"one", "100%", 1.0},
                         {"three-halves", "150%", 1.5},
                         {"two", "200%", 2.0},
                         {"four", "400%", 4.0},
                         {"eight", "800%", 8.0},
                         {"custom", "Custom", -1.0},
```

modules/basegl/processors/imageprocessing/imagescaling.cpp
martinfalk: see comment above

modules/base/processors/imagesource.cpp
```diff
@@ -65,13 +66,16 @@ ImageSource::ImageSource(InviwoApplication* app, const std::string& file)
...
-
+    
     isReady_.setUpdate([this]() { return filesystem::fileExists(file_.get()); });
     file_.onChange([this]() { isReady_.update(); });
 }
 
 void ImageSource::process() {
-    if (file_.get().empty()) return;
+    if (file_.get().empty()) {
+        outport_.setData(static_cast<Image*>(nullptr));
```
petersteneteg: Will this code ever run? Process should never run if the file is empty...

modules/base/processors/imagesource.cpp
```diff
@@ -53,7 +53,8 @@ ImageSource::ImageSource(InviwoApplication* app, const std::string& file)
     , file_("imageFileName", "File name", file, "image")
     , imageDimension_("imageDimension_", "Dimension", ivec2(0), ivec2(0), ivec2(10000), ivec2(1),
                       InvalidationLevel::Valid, PropertySemantics("Text")) {
-
+    // Remove default data created by ImageOutport
+    outport_.setData(static_cast<Image*>(nullptr));
```
petersteneteg: should this be the default then?

modules/base/processors/imagesource.cpp
ResearchDaniel: I guess the isReady will prevent the process function from running unless the file exists? Not sure what happens after deserializing the network though

modules/base/processors/imagesource.cpp
ResearchDaniel: That would probably have quite an impact on existing code so it would need to be tested. Separate issue and PR IMHO

martinfalk: reverted it for now. Checking for `isReady()` instead of using `hasData()` in background does the trick since this incorporates the fact whether the connected processor has been validated.

modules/basegl/processors/imageprocessing/imagescaling.cpp
```diff
@@ -74,16 +75,14 @@ ImageScaling::ImageScaling()
 
     customFactor_.setVisible(false);
 
-    auto triggerRescale = [this]() {
-        ResizeEvent e(calcInputImageSize());
-        inport_.propagateEvent(&e);
-    };
+    auto triggerRescale = [this]() { resizeInports(); };
```
petersteneteg: Remove lambda

modules/basegl/processors/imageprocessing/imagescaling.cpp
```diff
@@ -101,37 +100,48 @@ void ImageScaling::process() {
...
-        if (inport_.isConnected()) {
-            ResizeEvent e(calcInputImageSize());
-            inport_.propagateEvent(&e);
-        }
-    } else {
-        bool used = event->hasBeenUsed();
-        if (event->shouldPropagateTo(&inport_, this, source)) {
-            inport_.propagateEvent(event);
-            used |= event->hasBeenUsed();
+        if (enabled_) {
```
petersteneteg: Move to outer it

modules/basegl/processors/imageprocessing/imagescaling.cpp
```diff
@@ -101,37 +100,48 @@ void ImageScaling::process() {
 
 void ImageScaling::propagateEvent(Event* event, Outport* source) {
     if (event->hasVisitedProcessor(this)) return;
-    event->markAsVisited(this);
 
-    invokeEvent(event);
-    if (event->hasBeenUsed()) return;
+    if ((event->hash() == ResizeEvent::chash())) {
```
petersteneteg: Extra ()

petersteneteg (approved):

modules/basegl/processors/imageprocessing/imagescaling.cpp
martinfalk: but the size of the resize event must be cached independent of whether the processor is enabled or not.

petersteneteg (approved):

PR #252, opened 2018-09-20
QtWidgets: updated color property widget

martinfalk:
Revamped color property widget which allows to edit colors directly. Supports floating point range `[0,1]`, int range `[0,255]`, and hex color codes (`#RGB`, `#RGBA`, `#RRGGBB`, `#RRGGBBAA`).
Invalid input is indicated by red border and changes discarded if either <Esc> is pressed or the widget looses focus.

image


martinfalk:
Contains fix for tab order of property widgets.
* consistent tab order (composite does not get tab focus unless it features a checkbox)
* selecting a slider will focus the corresponding spinbox
* fixes tab order when changing property semantics

martinfalk:
TF dialog now features also the new `ColorLineEdit` meaning that one can enter colors either as hex, uint8, or [0,1] double
image

petersteneteg (approved):
Great!

PR #254, opened 2018-09-24
Cmake: Made it possible to specify external modules and projects with backslash

ResearchDaniel:
inviwomodulepaths.h would be incorrect if external module paths were specified with \path\to\folder


cmake/globalconfig.cmake
```diff
@@ -99,11 +99,11 @@ set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_CURRENT_LIST_DIR}/modules")
 
 #--------------------------------------------------------------------
 # Add parameter for paths to external modules
-set(IVW_EXTERNAL_MODULES "" CACHE STRING "Paths to directory containing external modules")
+set(IVW_EXTERNAL_MODULES "" CACHE STRING "Semicolon (;) separated paths to directories containing external modules")
 
 #--------------------------------------------------------------------
 # Add parameter for paths to external projects
-set(IVW_EXTERNAL_PROJECTS "" CACHE STRING "Paths to directory containing external projects")
+set(IVW_EXTERNAL_PROJECTS "" CACHE STRING "Semicolon (;) separated paths to directories containing cmake config file (.cmake/CMakeLists.txt file)")
```
petersteneteg: What do you mean by .cmake/?

cmake/globalmacros.cmake
```diff
@@ -91,6 +91,7 @@ endfunction()
 function(ivw_generate_module_paths_header)
     set(dirs "")
     foreach(dir ${IVW_MODULE_DIR} ${IVW_EXTERNAL_MODULES})
+		file(TO_CMAKE_PATH "${dir}" dir)
```
petersteneteg: Indentation seems off...

cmake/globalconfig.cmake
petersteneteg: Should probably not call CMakeLists.txt files Cmake config file as that might be confused with FooConfig.cmake files...

petersteneteg (approved):

cmake/globalconfig.cmake
ResearchDaniel: I'll adjust and say that add_subdirectory will be called

cmake/globalmacros.cmake
ResearchDaniel: Fixed

PR #257, opened 2018-09-25
Core: Trackball animation fixes.

ResearchDaniel:
Two errors sneaked in in the last fix. 
1. abs(dot) should be less than 1 to allow rotations.
2. lastRot_ = identity was always set.

Also fixed a crash occuring when changing system setting "followObjectDuringRotation"

### DIFF
--- a/src/core/interaction/trackball.cpp
+++ b/src/core/interaction/trackball.cpp
@@ -869,7 +869,7 @@ void inviwo::Trackball::animate() {
             const float t = 0.1f;
             const float dot = glm::dot(lastRot_, identity);
 			// Avoid division by zero when sin( n*pi ) == 0.
-			// Occurs when acos(+-1) == 0/pi (90 degreees)  
+			// Occurs when acos(+-1) == 0/pi (0/180 degreees)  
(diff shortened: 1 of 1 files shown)


src/core/interaction/trackball.cpp
```diff
@@ -157,7 +157,8 @@ Trackball::Trackball(TrackballObject* object)
 
     auto systemSettings = InviwoApplication::getPtr()->getSettingsByType<SystemSettings>();
     followObjectDuringRotation_ = systemSettings->followObjectDuringRotation_.get();
-    systemSettings->followObjectDuringRotation_.onChange([systemSettings,this]() {
+    systemSettings->followObjectDuringRotation_.onChange([this]() {
```
petersteneteg, line 160: What was wrong with the old code?

petersteneteg (approved):

src/core/interaction/trackball.cpp
ResearchDaniel: Nothing wrong with this one but the other one took the system settings pointer by reference. I think that it is not too good to store a pointer either so I changed to get it each time.

PR #258, opened 2018-09-25
Feature/imagelayer

martinfalk:
* new processor for copying an arbitrary layer of the input image to a color layer
* some minor fixes

image


petersteneteg (approved):

modules/basegl/processors/imageprocessing/imagelayer.cpp
```diff
@@ -0,0 +1,120 @@
...
+#include <modules/opengl/texture/textureutils.h>
+#include <modules/opengl/texture/textureunit.h>
+#include <modules/opengl/image/imagegl.h>
+
+namespace inviwo {
+
+// The Class Identifier has to be globally unique. Use a reverse DNS naming scheme
+const ProcessorInfo ImageLayer::processorInfo_{
+    "org.inviwo.ImageLayer",  // Class identifier
+    "Image Layer",            // Display name
```
ResearchDaniel, line 43: Image Layer To Color?

The naming of the processor is a bit ambigous, it can be also be interpreted as being an image layer.
Perhaps the processor should be called ImageLayerToColor? or ImageLayerToColorLayerConverter

ResearchDaniel (dismissed):
Nice, just an issue with naming

modules/basegl/processors/imageprocessing/imagelayer.cpp
petersteneteg: Think image ImageLayer is fine

martinfalk: That is inherent with our design! At some point it was decided that an Image has at least one color layer plus depth and picking at all times. If you consider this, there is no ambiguity.

ResearchDaniel: But the processor converts a layer to the color layer. Not sure if we had some conclusion on this but I think that they should be verbs?

ResearchDaniel: And/or be descriptive in what it does

PR #261, opened 2018-09-26
Fix/discretdata

MutterOberin:
I experienced a compiler error on the new discreet data. 

C2678: binary '==' : no operator found which takes a left-hand operand of type (or there is no acceptable conversion)


modules/discretedata/channels/datachannel.h
```diff
@@ -43,11 +43,11 @@ typedef signed long long ind;
 *   If these do not suffice, cast the respective short.
 */
 enum class GridPrimitive : ind {
-    Undef = -1,
-    Vertex = 0,
-    Edge = 1,
-    Face = 2,
-    Volume = 3,
+    Undef       = -1,
```
petersteneteg, line 46: does clang-format give you this formatting? We always format everything with clang-format.
If you for some reason really don't want clang-format formating for a code section, that section as to be marked with
//clang-format off
section
//clang-format on

modules/discretedata/connectivity/cell.h
```diff
@@ -35,40 +35,40 @@ namespace discretedata {
...
-    TRIANGLE_STRIP = 6,
-    POLYGON = 7,
-    PIXEL = 8,
-    QUAD = 9,
-    TETRA = 10,
-    VOXEL = 11,
-    HEXAHEDRON = 12,
-    WEDGE = 13,
-    PYRAMID = 14,
+    EMPTY_CELL       = 0,
```
petersteneteg, line 38: See comment above...
Unrelated to this we really should not use capital letter here. We _only_ use capital letters for MACROS

modules/discretedata/connectivity/periodicgrid.cpp
```diff
@@ -37,7 +37,7 @@ namespace discretedata {
 
 PeriodicGrid::PeriodicGrid(GridPrimitive gridDimension, const std::vector<ind>& numCellsPerDim, const std::vector<bool>& isDimPeriodic)
     : StructuredGrid(gridDimension, numCellsPerDim), isDimPeriodic_(isDimPeriodic) {
-    assert(numCellsPerDim.size() == gridDimension &&
+    assert(numCellsPerDim.size() == (size_t)gridDimension &&
```
petersteneteg: Never use c-style casts.

modules/glfw/canvasglfw.cpp
```diff
@@ -186,6 +186,7 @@ void CanvasGLFW::character(GLFWwindow* window, unsigned int character) {
     // Needed for text input
     auto thisCanvas = getCanvasGLFW(window);
     // Convert UTF32 character
+
```
petersteneteg, line 189: This should probably not be in the PR...

petersteneteg (changes requested):

modules/discretedata/channels/datachannel.h
MutterOberin: I personally like enums aligned. But of course it has to fit the rest, in terms of code style. We can revert that, if it is a conflict with the rest of the code base.

modules/discretedata/connectivity/cell.h
MutterOberin: Isn't that already in? I mean this should not be part of the code review. I just pops up because of the format change.

modules/discretedata/connectivity/periodicgrid.cpp
MutterOberin: Should it be std::static_cast<size_t>(gridDimension); ?

modules/glfw/canvasglfw.cpp
MutterOberin: This one was fixed on my machine and on master so it was resulting in a conflict. I merged them, this is why the commit is on the branch.

modules/discretedata/connectivity/periodicgrid.cpp
```diff
@@ -37,7 +37,7 @@ namespace discretedata {
 
 PeriodicGrid::PeriodicGrid(GridPrimitive gridDimension, const std::vector<ind>& numCellsPerDim, const std::vector<bool>& isDimPeriodic)
     : StructuredGrid(gridDimension, numCellsPerDim), isDimPeriodic_(isDimPeriodic) {
-    assert(numCellsPerDim.size() == (size_t)gridDimension &&
+    assert(numCellsPerDim.size() == static_cast<size_t>(gridDimension) &&
```
MutterOberin, line 40: Corrected the cast

MutterOberin (commented):
Corrected the c-style cast

modules/discretedata/channels/datachannel.h
petersteneteg: We are kind of strict with the use of clang-format,

petersteneteg: But I can fix this later on. Have a bunch of minor things I will change...

modules/discretedata/connectivity/cell.h
petersteneteg: Yes, I can change this...

modules/glfw/canvasglfw.cpp
petersteneteg: Generally we never merge master into a feature branch. You would instead rebase the feature branch on top of the latest master.

petersteneteg (approved):

PR #270, opened 2018-10-03
WebBrowser: Modified the webbrowser processor so that it asks for a l…

Alboin:
…ocal file per default instead of a webpage.

**Updated feature / Fixes** 

A new property to choose whether to load a local file through file browsing or enter a webpage url.

Also updated the example workspace since the processor was updated.

@ResearchDaniel  I now added my local path to the example html-file, will this work on other machines as well? Or will the path to my local file cause problems? If that is the case, how can the path be made relative to the workspace? 

- Windows 10
- Visual Studio 15.7.6
- QT Creator 4.6.1 (Community)
- CMake 3.11.2
- Python 3.6 (32-bit)

### DIFF
--- a/modules/webbrowser/processors/webbrowserprocessor.cpp
+++ b/modules/webbrowser/processors/webbrowserprocessor.cpp
@@ -58,12 +58,16 @@ WebBrowserProcessor::WebBrowserProcessor()
     // Output from CEF is 8-bits per channel
     , background_("background")
     , outport_("webpage", DataVec4UInt8::get())
+    , fileName_("fileName", "HTML file", "")
     , url_("URL", "URL", "http://www.inviwo.org")
     , reload_("reload", "Reload")
     , addPropertyGroup_("addProperty", "Add property to synchronize")
     , type_("property", "Property")
     , propertyHtmlId_("propertyHtmlId", "Html id")
     , add_("add", "Add")
+    , sourceType_("sourceType", "Source",
+                  {{"localFile", "Local File", SourceType::LocalFile},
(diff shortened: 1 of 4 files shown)


modules/webbrowser/processors/webbrowserprocessor.cpp
```diff
@@ -166,24 +196,34 @@ WebBrowserProcessor::~WebBrowserProcessor() {
...
+        if (prop == &sourceType_ || prop == &fileName_ || prop == &url_ || prop == &reload_ ||
+            prop == &addPropertyGroup_ || prop == &type_ || prop == &propertyHtmlId_ ||
+            prop == &add_) {
             continue;
         }
         browserClient_->propertyCefSynchronizer_->startSynchronize(prop, prop->getIdentifier());
     }
     // Must reload page to connect property with Frame, see PropertyCefSynchronizer::OnLoadEnd
-    browser_->GetMainFrame()->LoadURL(url_.get());
+    if (sourceType_.get() == SourceType::LocalFile) {
```
ResearchDaniel: Getting the URL  could be moved to a separate function since it is repeated a couple of times

ResearchDaniel (approved):
Nice!
Just a minor refactoring comment

ResearchDaniel:
@Alboin FileProperty takes care of relative file path so it should work on other machines as well

modules/webbrowser/processors/webbrowserprocessor.cpp
Alboin: Agree! Done!

PR #273, opened 2018-10-11
Core: Added property conversion between FileProperty and StringProperty.

Alboin:
- Windows 10
- Visual Studio 15.7.6 
- QT Community 4.6.1
- CMake 3.11.2
- Python 3.6 (32-bit)

### DIFF
--- a/src/core/common/inviwocore.cpp
+++ b/src/core/common/inviwocore.cpp
@@ -332,6 +332,9 @@ InviwoCore::InviwoCore(InviwoApplication* app)
     using Vec3s = std::tuple<vec3, dvec3, ivec3, size3_t>;
     using Vec4s = std::tuple<vec4, dvec4, ivec4, size4_t>;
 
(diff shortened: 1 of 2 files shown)


src/core/common/inviwocore.cpp
```diff
@@ -332,6 +332,9 @@ InviwoCore::InviwoCore(InviwoApplication* app)
     using Vec3s = std::tuple<vec3, dvec3, ivec3, size3_t>;
     using Vec4s = std::tuple<vec4, dvec4, ivec4, size4_t>;
 
+    registerPropertyConverter(std::move(std::make_unique<FileToStringConverter>()));
```
petersteneteg: no need to cast to R-value with std::move since the return value of make_unique already is a R-value by definition.

petersteneteg (approved):
Nice, I was missing those converters just the other day!

PR #275, opened 2018-10-25
Fix/webbrowser visitor

Alboin:
See individual commits for what has been implemented.

- Windows 10
- Visual Studio 15.7.6
- QT Creator 4.6.1 (Community)
- CMake 3.11.2
- Python 3.6 (32-bit)

### DIFF
--- a/modules/webbrowser/properties/propertywidgetcef.cpp
+++ b/modules/webbrowser/properties/propertywidgetcef.cpp
@@ -49,7 +49,7 @@ public:
             auto start = domString.find("<!--");
             auto stop = domString.find("-->");
             if (start != std::string::npos)
-                domString = domString.substr(0, start) + domString.substr(stop + 3, domString.length());
+                domString.erase(start, stop - start + 3);
             else
                 break;
         }
--- a/modules/webbrowser/properties/propertywidgetcef.cpp
+++ b/modules/webbrowser/properties/propertywidgetcef.cpp
@@ -42,12 +42,22 @@ public:
         : CefStringVisitor(), htmlId_(htmlId), widget_(widget), frame_(frame){};
 
     void Visit(const CefString& string) OVERRIDE {
-        std::string domString_ = string;
+        std::string domString = string;
+
+        // Remove all the html comments to avoid finding element id's in the comments.
+        while (true) {
+            auto start = domString.find("<!--");
+            auto stop = domString.find("-->");
+            if (start != std::string::npos)
+                domString = domString.substr(0, start) + domString.substr(stop + 3, domString.length());
+            else
+                break;
+        }
 
         // Remove any occurences of " or ' from the domString_ to remove possible variations on html
         // id declarations.
-        domString_.erase(std::remove(domString_.begin(), domString_.end(), '"'), domString_.end());
-        domString_.erase(std::remove(domString_.begin(), domString_.end(), '\''), domString_.end());
+        domString.erase(std::remove(domString.begin(), domString.end(), '"'), domString.end());
+        domString.erase(std::remove(domString.begin(), domString.end(), '\''), domString.end());
 
         std::stringstream ss1;
         ss1 << "id:" << htmlId_;
@@ -55,8 +65,8 @@ public:
         ss2 << "id=" << htmlId_;
 
         // If the widget's html-id is in the given frame's DOM-document, set it's frame.
-        if (domString_.find(ss1.str()) != std::string::npos ||
-            domString_.find(ss2.str()) != std::string::npos) {
+        if (domString.find(ss1.str()) != std::string::npos ||
+            domString.find(ss2.str()) != std::string::npos) {
             widget_->frame_ = frame_;
             widget_->updateFromProperty();
         }
--- a/modules/webbrowser/properties/propertywidgetcef.cpp
+++ b/modules/webbrowser/properties/propertywidgetcef.cpp
@@ -33,6 +33,8 @@
 
 namespace inviwo {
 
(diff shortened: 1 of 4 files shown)


modules/webbrowser/properties/propertywidgetcef.cpp
```diff
@@ -33,6 +33,42 @@
...
+
+        // If the widget's html-id is in the given frame's DOM-document, set it's frame.
+        if (domString_.find(ss1.str()) != std::string::npos ||
+            domString_.find(ss2.str()) != std::string::npos) {
+            widget_->setFrame(frame_);
+            widget_->updateFromProperty();
+        }
+    };
+
+    std::string domString_;
```
ResearchDaniel: Unused?

modules/webbrowser/properties/propertywidgetcef.h
```diff
@@ -82,6 +82,11 @@ class IVW_MODULE_WEBBROWSER_API PropertyWidgetCEF : public PropertyWidget, publi
      * Set frame containing html item.
      */
     void setFrame(CefRefPtr<CefFrame> frame);
```
ResearchDaniel: setFrame should only be accessed by CefDOMSearchId? (friend)

modules/webbrowser/properties/propertywidgetcef.h
```diff
@@ -82,6 +82,11 @@ class IVW_MODULE_WEBBROWSER_API PropertyWidgetCEF : public PropertyWidget, publi
...
+
+    /*
+     * Check if widget is part of a given frame.
```
ResearchDaniel: Suggestion:
Checks if frame contains the html id of the widget and sets it if it is.
CefFrame is required for communication between Inviwo and the web browser.

modules/webbrowser/properties/propertywidgetcef.cpp
```diff
@@ -43,7 +79,13 @@ void PropertyWidgetCEF::setFrame(CefRefPtr<CefFrame> frame) {
     frame_ = frame;
```
ResearchDaniel: This should call setIfPartOfFrame?

modules/webbrowser/properties/propertywidgetcef.cpp
```diff
@@ -33,6 +33,42 @@
 
 namespace inviwo {
 
+class CefDOMSearchId : public CefStringVisitor {
```
ResearchDaniel, line 38: Add overall description of what it does and why it is needed, i.e. 
Checks if widget html id exist in the frame and sets it if it does. 
@note Cannot use CefDOMVisitor since it requires the renderer process.

modules/webbrowser/properties/propertywidgetcef.cpp
```diff
@@ -33,6 +33,42 @@
...
+    CefDOMSearchId(const std::string& htmlId, PropertyWidgetCEF* widget,
+                   const CefRefPtr<CefFrame> frame)
+        : CefStringVisitor(), htmlId_(htmlId), widget_(widget), frame_(frame){};
+
+    void Visit(const CefString& string) OVERRIDE {
+        domString_ = string;
+
+        // Remove any occurences of " or ' from the domString_ to remove possible variations on html
+        // id declarations.
+        domString_.erase(std::remove(domString_.begin(), domString_.end(), '"'), domString_.end());
```
ResearchDaniel: How does the search cope with commented code?

ResearchDaniel (changes requested):
Nice, some minor comments

modules/webbrowser/properties/propertywidgetcef.cpp
Alboin: If someone would comment "id=buttonId" or such, the visitor will act as if that is an actual html element.

modules/webbrowser/properties/propertywidgetcef.cpp
Alboin: No, but it does not need to be a public variable. Moved it to another scope.

modules/webbrowser/properties/propertywidgetcef.h
Alboin: Added CefDOMSearchId as a friend class and moved the function to protected.

modules/webbrowser/properties/propertywidgetcef.h
Alboin: You mean to change the comment right? Changed it to your suggestion.

modules/webbrowser/properties/propertywidgetcef.cpp
ResearchDaniel: Would either fix it or add it as a limitation in the documentation

modules/webbrowser/properties/propertywidgetcef.cpp
Alboin: Changed so that setFrame() calls setIfPartOfFrame instead of the previous above.

widget->setFrame(frame) is then called istead of widget->setIfPartOfFrame(frame) from the class that synchronizes all widgets (propertycefsynchronizer.cpp).

modules/webbrowser/properties/propertywidgetcef.cpp
Alboin: Added a loop to remove any html comments from the DOM-string in this commit: https://github.com/inviwo/inviwo/pull/275/commits/3b37bda616c4204362f9a5ceed5d19a2742a60c6

ResearchDaniel (approved):
Had one minor comment that you can fix if you want to

Alboin:
@ResearchDaniel Done!

PR #278, opened 2018-10-31
Webbrowser fixes

ResearchDaniel:
- Windows: Prevent console windows from popping up
- Mac/linux: Removed libcef_lib since it seem to be windows only

### DIFF
--- a/modules/webbrowser/properties/propertycefsynchronizer.cpp
+++ b/modules/webbrowser/properties/propertycefsynchronizer.cpp
@@ -33,6 +33,13 @@
 #include <modules/webbrowser/properties/ordinalpropertywidgetcef.h>
 #include <modules/webbrowser/properties/stringpropertywidgetcef.h>
 
+#include <modules/webbrowser/webbrowsermodule.h>
+
+#include <warn/push>
+#include <warn/ignore/all>
+#include "include/cef_parser.h"
+#include <warn/pop>
+
 namespace inviwo {
 PropertyCefSynchronizer::PropertyCefSynchronizer() {
     registerPropertyWidget<BoolPropertyWidgetCEF, BoolProperty>(PropertySemantics("Default"));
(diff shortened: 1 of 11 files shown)


modules/webbrowser/properties/propertycefsynchronizer.cpp
```diff
@@ -33,6 +33,13 @@
 #include <modules/webbrowser/properties/ordinalpropertywidgetcef.h>
 #include <modules/webbrowser/properties/stringpropertywidgetcef.h>
 
+#include <modules/webbrowser/webbrowsermodule.h>
+
+#include <warn/push>
+#include <warn/ignore/all>
+#include "include/cef_parser.h"
```
petersteneteg, line 40: generally "include" would not be in the header path. Do you have the correct include paths setup?

modules/webbrowser/webbrowsermodule.cpp
```diff
@@ -162,4 +168,70 @@ WebBrowserModule::~WebBrowserModule() {
...
 }
+    
+std::string WebBrowserModule::getDataURI(const std::string& data, const std::string& mime_type) {
+    return "data:" + mime_type + ";base64," +
+    CefURIEncode(CefBase64Encode(data.data(), data.size()), false)
+    .ToString();
+}
+    
+std::string WebBrowserModule::getCefErrorString(cef_errorcode_t code) {
+#define CASE(code) \
```
petersteneteg: why use a define for this small snippet?

modules/webbrowser/webbrowsermodule.h
```diff
@@ -34,6 +34,17 @@
 #include <inviwo/core/common/inviwomodule.h>
 #include <inviwo/core/util/timer.h>
 
+
+#include <warn/push>
+#include <warn/ignore/all>
+
+#include "include/internal/cef_types.h"
```
petersteneteg, line 41: same strange path...

petersteneteg (approved):
cant't really say I understand this, but it looks more or less reasonable

modules/webbrowser/properties/propertycefsynchronizer.cpp
ResearchDaniel: Use same as CEF examples, so yes :)

modules/webbrowser/webbrowsermodule.cpp
ResearchDaniel: Better suggestion?

ResearchDaniel:
@petersteneteg PR became a bit more involved since I wanted to make the new CEF version work on Mac as well. Have now verified that it also works on Windows

modules/webbrowser/webbrowsermodule.cpp
petersteneteg: just write out the code is's only a few chars more...

PR #285, opened 2018-11-07
UserInterfaceGL/Webbrowser: Mac fixes

ResearchDaniel:
UserInterfaceGL/Webbrowser: Mac fixes: Array initialization according to GLSL spec (see section 4.1.4) and added missing link libraries for CEF.

### DIFF
--- a/modules/plottinggl/glsl/legend.frag
+++ b/modules/plottinggl/glsl/legend.frag
@@ -75,8 +75,8 @@ void main() {
     // blend in the checkerboard as background to the TF depending on its opacity
     vec4 finalColor = over(checkerBoard(centeredPos), colorTF);
 
-    // set border flag iff the fragment coord is within the border
-    bool border = borderWidth > 0 && any(greaterThan(abs(centeredPos), outputDim * 0.5));
+    // set border flag if the fragment coord is within the border
+    float border = borderWidth > 0 && any(greaterThan(abs(centeredPos), outputDim * 0.5)) ? 1.f : 0.f;
     FragData0 =  mix(finalColor, color, border);
 
     // no depth input, reset depth to largest value
--- a/modules/plottinggl/glsl/legend.frag
+++ b/modules/plottinggl/glsl/legend.frag
@@ -70,14 +70,14 @@ void main() {
 
     // increase alpha to allow better visibility by 1 - (1 - a)^4 and then add "backgroundAlpha" to
     // set alpha to 1 if no background is wanted
-    colorTF.a = mix(1.0f - pow(1.0f - colorTF.a, 4.0f), 1.0, float(backgroundStyle));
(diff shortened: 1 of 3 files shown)


modules/plottinggl/glsl/legend.frag
```diff
@@ -75,8 +75,8 @@ void main() {
     // blend in the checkerboard as background to the TF depending on its opacity
     vec4 finalColor = over(checkerBoard(centeredPos), colorTF);
 
-    // set border flag iff the fragment coord is within the border
-    bool border = borderWidth > 0 && any(greaterThan(abs(centeredPos), outputDim * 0.5));
+    // set border flag if the fragment coord is within the border
+    float border = borderWidth > 0 && any(greaterThan(abs(centeredPos), outputDim * 0.5)) ? 1.f : 0.f;
```
martinfalk: You should not use `f` as float specifier in GLSL shaders!

any particular reason for using float and not int? Could you test
```suggestion
    int border = borderWidth > 0 && any(greaterThan(abs(centeredPos), outputDim * 0.5)) ? 1 : 0;
```

modules/userinterfacegl/glsl/renderui.frag
```diff
@@ -33,7 +33,7 @@ uniform sampler2DArray arrayTexSampler;
 // texture indices for normal, pressed, checked, 
 // halo normal, halo pressed, halo checked
 // border normal, border pressed, border checked
-uniform int arrayTexMap[9] = { 0, 0, 0, 0, 0, 0, 0, 0, 0 };
+uniform int arrayTexMap[9] = int[]( 0, 0, 0, 0, 0, 0, 0, 0, 0 );
```
martinfalk, line 36: this one is quite interesting. I think we need the `{}` initialization in other shaders, too...

martinfalk (changes requested):

modules/plottinggl/glsl/legend.frag
petersteneteg: This is the wrong fix... it should be
FragData0 =  mix(finalColor, color, bvec4(border));

petersteneteg (dismissed):

modules/plottinggl/glsl/legend.frag
ResearchDaniel: > You should not use `f` as float specifier in GLSL shaders!

Okay, then I'll remove the ones six lines above too :)

ResearchDaniel: > This is the wrong fix... it should be
> FragData0 = mix(finalColor, color, bvec4(border));

Seem to work too

ResearchDaniel:
Made suggested changes

modules/userinterfacegl/glsl/renderui.frag
ResearchDaniel: Did a quick check in the plottinggl module but did not find any other. Could exist elsewhere though

martinfalk (approved):

PR #286, opened 2018-11-07
BaseGL: Modified the volumecombiner to not normalize the incoming vol…

Alboin:
…umes before performing calculations on them, since this will cause problems with volumes containing negative values.

### DIFF
--- a/modules/basegl/processors/volumeprocessing/volumecombiner.cpp
+++ b/modules/basegl/processors/volumeprocessing/volumecombiner.cpp
@@ -169,13 +169,13 @@ void VolumeCombiner::buildShader(const std::string& eqn) {
             ss << "    vec4 " << vol << ";\n";
             ss << "    if (all(greaterThanEqual(" << coord << ", vec3(0))) &&"
                << " all(lessThanEqual(" << coord << ", vec3(1)))) {\n";
-            ss << "        " << vol << " = getNormalizedVoxel(" << v << ", " << vp << ", " << coord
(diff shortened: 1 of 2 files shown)


modules/basegl/processors/volumeprocessing/volumecombiner.cpp
```diff
@@ -169,13 +169,13 @@ void VolumeCombiner::buildShader(const std::string& eqn) {
             ss << "    vec4 " << vol << ";\n";
             ss << "    if (all(greaterThanEqual(" << coord << ", vec3(0))) &&"
                << " all(lessThanEqual(" << coord << ", vec3(1)))) {\n";
-            ss << "        " << vol << " = getNormalizedVoxel(" << v << ", " << vp << ", " << coord
+            ss << "        " << vol << " = getVoxel(" << v << ", " << vp << ", " << coord
```
petersteneteg: Not sure about this, I mean getVoxel will also do "normalisation" as OpenGL does it... 
we also have getSignNormalizedVoxel. Think this should probably be exposed as an option. Or someone has to convince me which one is the "correct" one...

petersteneteg (changes requested):

modules/basegl/processors/volumeprocessing/volumecombiner.cpp
Alboin: Done!

petersteneteg (approved):

PR #287, opened 2018-11-07
Plotting: Updated the dataframeexporter so that it works also when th…

Alboin:
…e user enter a filepath in string format instead of only when using the file dialouge.

### DIFF
--- a/modules/plotting/processors/dataframeexporter.cpp
+++ b/modules/plotting/processors/dataframeexporter.cpp
@@ -94,9 +94,13 @@ void DataFrameExporter::exportNow() {
         LogWarn("File already exists: " << exportFile_);
         return;
     }
-    if (exportFile_.getSelectedExtension() == exportFile_.getNameFilters()[0]) {
+    // convert the fileEnding to all lower case, for string comparison a few lines down.
+    std::string fileEnding = filesystem::getFileExtension(exportFile_.get());
+    std::transform(fileEnding.begin(), fileEnding.end(), fileEnding.begin(), ::tolower);
+
+    if (fileEnding == "csv") {
         exportAsCSV(separateVectorTypesIntoColumns_);
-    } else if (exportFile_.getSelectedExtension() == exportFile_.getNameFilters()[1]) {
+    } else if (fileEnding == "xml") {
         exportAsXML();
     } else {
         LogWarn("Unsupported file type: " << exportFile_);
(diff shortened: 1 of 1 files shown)


ResearchDaniel (approved):
Looks good to me. @martinfalk ?

modules/plotting/processors/dataframeexporter.cpp
```diff
@@ -94,9 +94,13 @@ void DataFrameExporter::exportNow() {
         LogWarn("File already exists: " << exportFile_);
         return;
     }
-    if (exportFile_.getSelectedExtension() == exportFile_.getNameFilters()[0]) {
+    // convert the fileEnding to all lower case, for string comparison a few lines down.
+    std::string fileEnding = filesystem::getFileExtension(exportFile_.get());
+    std::transform(fileEnding.begin(), fileEnding.end(), fileEnding.begin(), ::tolower);
```
petersteneteg: this line will generate a warning since ::tolower returns an int not a char

One can use iCaseCmp from stringconversion.h to do case insensitive compare

modules/plotting/processors/dataframeexporter.cpp
```diff
@@ -94,9 +94,13 @@ void DataFrameExporter::exportNow() {
...
+
+    if (fileEnding == "csv") {
         exportAsCSV(separateVectorTypesIntoColumns_);
-    } else if (exportFile_.getSelectedExtension() == exportFile_.getNameFilters()[1]) {
+    } else if (fileEnding == "xml") {
```
petersteneteg: there is a slight semantic difference with here, exportFile_.getSelectedExtension() can return different readers for the same extension, the given solution will not handle that. but since there is only csv and xml that should not matter here

petersteneteg (changes requested):

modules/plotting/processors/dataframeexporter.cpp
petersteneteg: @martinfalk the real issue is probably around here
https://github.com/inviwo/inviwo/blob/63c9853e4ddfa696ebffcfa2b23f6e82f3a39394/modules/qtwidgets/properties/filepropertywidgetqt.cpp#L83

petersteneteg: we should probably set the selected item there to...

Alboin: Sooo... No changes are needed here?

petersteneteg: Martin is working on it...

martinfalk:
Superseded by #288.

PR #288, opened 2018-11-08
Fix/csvexport

martinfalk:
Various fixes and improvements for CSV export.

### DIFF
--- a/include/inviwo/core/util/fileextension.h
+++ b/include/inviwo/core/util/fileextension.h
@@ -24,7 +24,7 @@
  * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
  * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
  * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
- * 
+ *
  *********************************************************************************/
 
 #ifndef IVW_FILEEXTENSION_H
@@ -44,24 +44,35 @@ public:
     FileExtension(std::string extension, std::string description);
     virtual ~FileExtension() = default;
 
-    /** 
+    /**
      * \brief extracts a FileExtension object from a string. This function assumes
      * that the extension is given within the right most parentheses.
-     * 
+     *
      * @param str Input string  formed like "Text files (*.txt)".
      * @return FileExtension object created from the information given in the input string.
      */
-    static FileExtension createFileExtensionFromString(const std::string &str);
+    static FileExtension createFileExtensionFromString(const std::string& str);
     std::string toString() const;
 
     bool empty() const;
+    bool matchesAll() const;
+
+    /**
+     * \brief checks whether the given string is matched by this FileExtension
+     * Extensions are matched case insensitive while an empty FileExtension, i.e. '*', will match
+     * all strings.
+     *
+     * @param str   string to be tested
+     * @return true if the string is matched by the FileExtension
+     */
+    bool matches(const std::string& str) const;
 
     virtual void serialize(Serializer& s) const override;
     virtual void deserialize(Deserializer& d) override;
 
     static FileExtension all();
 
-    std::string extension_; ///< File extension in lower case letters.
+    std::string extension_;  ///< File extension in lower case letters.
     std::string description_;
 };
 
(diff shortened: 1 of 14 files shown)


include/inviwo/core/util/fileextension.h
```diff
@@ -44,56 +44,68 @@ class IVW_CORE_API FileExtension : public Serializable {
...
+     *
      * @param str Input string  formed like "Text files (*.txt)".
      * @return FileExtension object created from the information given in the input string.
      */
-    static FileExtension createFileExtensionFromString(const std::string &str);
+    static FileExtension createFileExtensionFromString(const std::string& str);
     std::string toString() const;
 
     bool empty() const;
+    bool matchesAll() const;
```
ResearchDaniel, line 64: What does this one do? Same as *? Could add a comment

include/inviwo/core/util/fileextension.h
```diff
@@ -44,56 +44,68 @@ class IVW_CORE_API FileExtension : public Serializable {
...
+     *
+     * @param str   string to be tested
+     * @return true if the string is matched by the FileExtension
+     */
+    bool matches(const std::string& str) const;
 
     virtual void serialize(Serializer& s) const override;
     virtual void deserialize(Deserializer& d) override;
 
     static FileExtension all();
```
ResearchDaniel, line 79: Not part of this PR but any would probably make more sense :)

modules/plotting/processors/dataframeexporter.cpp
```diff
@@ -94,12 +100,16 @@ void DataFrameExporter::exportNow() {
         LogWarn("File already exists: " << exportFile_);
         return;
     }
-    if (exportFile_.getSelectedExtension() == exportFile_.getNameFilters()[0]) {
-        exportAsCSV(separateVectorTypesIntoColumns_);
-    } else if (exportFile_.getSelectedExtension() == exportFile_.getNameFilters()[1]) {
+    if (exportFile_.getSelectedExtension() == xmlExtension_) {
```
ResearchDaniel, line 103: What if the file dialog has not been used, i.e. set via scripting. Should it be:
`exportFile_.getSelectedExtension() == xmlExtension_ || filesystem::getFileExtension(exportFile_) == xmlExtension_`

src/core/util/fileextension.cpp
```diff
@@ -88,6 +88,19 @@ std::string FileExtension::toString() const {
 
 bool FileExtension::empty() const { return extension_.empty() && description_.empty(); }
 
+bool FileExtension::matchesAll() const { return extension_ == "*"; }
+
+bool FileExtension::matches(const std::string& str) const {
+    if (empty()) {
+        return false;
+    }
+    if (extension_.empty()) {
```
ResearchDaniel, line 97: or matchesAll()?

ResearchDaniel (changes requested):
Might be some errors. 
Consider renaming FileExtension::all() to FileExtension::any()  (and in that case matchesAny)

include/inviwo/core/util/fileextension.h
petersteneteg: the filter is usually called 
>   "*", "All Files"

not "Any Files" hence I think ```all``` makes more sense...

include/inviwo/core/util/fileextension.h
martinfalk: this one returns true, if the extension only contains an asterisk, i.e. the `FileExtension` matches all files. 
```c++
bool FileExtension::matchesAll() const { return extension_ == "*"; }
```
Will add a comment.

include/inviwo/core/util/fileextension.h
martinfalk: agree with Peter. I had a look at various file dialogs in different applications.

modules/plotting/processors/dataframeexporter.cpp
martinfalk: Good point. I will move the extension checking from the widget into the fileproperty itself.

src/core/util/fileextension.cpp
martinfalk: not necessarily, someone, e.g. @ResearchDaniel, might come along and decide to add a custom `FileExtension("*", "Matches any file")` to the property. Then the description of the FileExtension would not match `::all()`

ResearchDaniel (approved):

PR #297, opened 2018-11-14
CMake: better handling of default modules

petersteneteg:
Added an option to control if a module should be on by default, and remove the old global setting.
To enable the module by default add the following to the module `depends.cmake` file:
```
    set(EnableByDefault ON)
```


ResearchDaniel (approved):
No biggie, but I think that it would make more sense if the function would be called in the same cmake file as the module is created. 
`ivw_enable_module_by_default(module_name ON)`

petersteneteg:
@ResearchDaniel yes that would be nice, but by then the module is already enabled rendering the option mute...

PR #320, opened 2018-11-30
Editor: Added missing parameter for lambda in consolewidget

petersteneteg:


petersteneteg:
for some strange reason the other PR did not manage to build in jenkins

PR #321, opened 2018-11-30
Feature/cosmetic

martinfalk:
various cosmetic changes like updating workspaces and visualizers to the latest module versions.
* modified volume raycasting visualizer
* updated example workspaces (`stream_tubes` now based on `Integral Lines to Mesh` processor)

Stream Tubes:

Hurricane Isabel:


modules/opengl/src/debugmessages.cpp
```diff
@@ -192,9 +192,11 @@ void handleOpenGLDebugMode(Canvas::ContextID context) {
     auto mode = openglSettings->debugMessages_.getSelectedValue();
     auto severity = openglSettings->debugSeverity_.getSelectedValue();
     setOpenGLDebugMode(mode, severity);
-    logDebugMode(mode, severity, context);
+    if (mode != debug::Mode::Off) {
+        logDebugMode(mode, severity, context);
```
ResearchDaniel, line 196: Should not logDebugMode handle off mode?

ResearchDaniel (approved):
One minor comment

modules/opengl/src/debugmessages.cpp
martinfalk: This is for the initial message when a canvas is created. The reasoning here is that you don't want to see this message every time a canvas is created (applies to canvas processor, port inspectors, etc.)

martinfalk: On the other hand, you might want to see the status change from "debug" to "off".

PR #324, opened 2018-12-05
OpenGL: Enforce specification of vertex shader since it otherwise res…

### LINKED ISSUE #322: Mac: INVALID_OPERATION when using ImageGLProcessor
Processors based on ImageGLProcessor, for example Image Grayscale, produces invalid operation OpenGL error on line 83:
`utilgl::activateTargetAndCopySource(outport_, inport_, ImageType::ColorOnly);`
which traces to 
line 180 in imagegl.cpp:
`    target->activateBuffer(ImageType::AllLayers);
`
Output from processor is invalid after the error has occurred.

**Steps to reproduce**
Open a workspace including an image source, grayscale and canvas on Mac.

**Environment information**
* Operating System:
Mac OSX High Sierra 10.13.6
Intel Iris Graphics 550

ResearchDaniel:
…ults in undefined behavior according to OpenGL specifications. Closes #322

### DIFF
--- a/modules/basegl/src/processors/imageprocessing/imageglprocessor.cpp
+++ b/modules/basegl/src/processors/imageprocessing/imageglprocessor.cpp
@@ -31,6 +31,7 @@
 #include <modules/opengl/image/imagegl.h>
 #include <modules/opengl/shader/shader.h>
 #include <modules/opengl/texture/textureutils.h>
+#include <modules/opengl/shader/shaderresource.h>
 #include <modules/opengl/shader/shaderutils.h>
 #include <modules/opengl/buffer/framebufferobject.h>
 
@@ -47,7 +48,15 @@ ImageGLProcessor::ImageGLProcessor(std::shared_ptr<const ShaderResource> fragmen
     , dataFormat_(nullptr)
     , swizzleMask_(swizzlemasks::rgba)
     , internalInvalid_(false)
-    , shader_({{ShaderType::Vertex, utilgl::findShaderResource("img_identity.vert")}, {ShaderType::Fragment, fragmentShader}},
+    , shader_({
+        {ShaderType::Vertex, std::make_shared<StringShaderResource>("org.inviwo.ImageGLProcessor.vert",
+        // singleDrawImagePlaneRect uses mesh with vertices and tex coords that do not need transformation
+        "out vec3 texCoord_;"
+        "void main() {"
+        "    texCoord_ = in_TexCoord;"
+        "    gl_Position = in_Vertex;"
+        "}")
+    }, {ShaderType::Fragment, fragmentShader}},
               buildShader ? Shader::Build::Yes : Shader::Build::No) {
     addPort(inport_);
     addPort(outport_);
--- a/modules/basegl/src/processors/imageprocessing/imageglprocessor.cpp
+++ b/modules/basegl/src/processors/imageprocessing/imageglprocessor.cpp
@@ -48,15 +48,16 @@ ImageGLProcessor::ImageGLProcessor(std::shared_ptr<const ShaderResource> fragmen
     , dataFormat_(nullptr)
     , swizzleMask_(swizzlemasks::rgba)
     , internalInvalid_(false)
-    , shader_({
-        {ShaderType::Vertex, std::make_shared<StringShaderResource>("org.inviwo.ImageGLProcessor.vert",
-        // singleDrawImagePlaneRect uses mesh with vertices and tex coords that do not need transformation
-        "out vec3 texCoord_;"
-        "void main() {"
-        "    texCoord_ = in_TexCoord;"
-        "    gl_Position = in_Vertex;"
-        "}")
-    }, {ShaderType::Fragment, fragmentShader}},
+    , shader_({{ShaderType::Vertex, std::make_shared<StringShaderResource>(
+                                        "org.inviwo.ImageGLProcessor.vert",
+                                        // singleDrawImagePlaneRect uses mesh with vertices and tex
+                                        // coords that do not need transformation
+                                        "out vec3 texCoord_;"
+                                        "void main() {"
+                                        "    texCoord_ = in_TexCoord;"
+                                        "    gl_Position = in_Vertex;"
+                                        "}")},
+               {ShaderType::Fragment, fragmentShader}},
               buildShader ? Shader::Build::Yes : Shader::Build::No) {
     addPort(inport_);
     addPort(outport_);
--- a/modules/basegl/src/processors/imageprocessing/imageglprocessor.cpp
+++ b/modules/basegl/src/processors/imageprocessing/imageglprocessor.cpp
@@ -47,7 +47,7 @@ ImageGLProcessor::ImageGLProcessor(std::shared_ptr<const ShaderResource> fragmen
     , dataFormat_(nullptr)
     , swizzleMask_(swizzlemasks::rgba)
     , internalInvalid_(false)
-    , shader_({{ShaderType::Fragment, fragmentShader}},
+    , shader_({{ShaderType::Vertex, utilgl::findShaderResource("img_identity.vert")}, {ShaderType::Fragment, fragmentShader}},
               buildShader ? Shader::Build::Yes : Shader::Build::No) {
     addPort(inport_);
     addPort(outport_);
--- a/modules/basegl/src/processors/imageprocessing/imageglprocessor.cpp
+++ b/modules/basegl/src/processors/imageprocessing/imageglprocessor.cpp
@@ -33,6 +33,7 @@
 #include <modules/opengl/texture/textureutils.h>
 #include <modules/opengl/shader/shaderresource.h>
 #include <modules/opengl/shader/shaderutils.h>
+#include <modules/opengl/shader/standardshaders.h>
 #include <modules/opengl/buffer/framebufferobject.h>
 
 namespace inviwo {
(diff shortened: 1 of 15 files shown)


modules/opengl/src/shader/shader.cpp
```diff
@@ -38,6 +38,9 @@ namespace inviwo {
 
 Shader::Shader(const std::vector<std::pair<ShaderType, std::string>> &items, Build buildShader)
     : id_{glCreateProgram()}, warningLevel_{UniformWarning::Ignore} {
+    if (std::find_if(std::begin(items), std::end(items), [](const auto& item) { return item.first == ShaderType::Vertex;}) == std::end(items)) {
+        throw Exception("Vertex shader required, provide for example img_identity.vert");
```
petersteneteg: this will leak the program....

modules/basegl/src/processors/imageprocessing/imageglprocessor.cpp
```diff
@@ -47,7 +47,7 @@ ImageGLProcessor::ImageGLProcessor(std::shared_ptr<const ShaderResource> fragmen
     , dataFormat_(nullptr)
     , swizzleMask_(swizzlemasks::rgba)
     , internalInvalid_(false)
-    , shader_({{ShaderType::Fragment, fragmentShader}},
+    , shader_({{ShaderType::Vertex, utilgl::findShaderResource("img_identity.vert")}, {ShaderType::Fragment, fragmentShader}},
```
petersteneteg: would be nicer to just use a StringShaderResource here...

modules/opengl/src/shader/shader.cpp
ResearchDaniel: Good catch. Wrapped the id in a scoped variable

modules/basegl/src/processors/imageprocessing/imageglprocessor.cpp
ResearchDaniel: Replaced it with a StringShaderResource.

modules/opengl/src/shader/shader.cpp
martinfalk: This won't work as intended! In principle one can define multiple vertex shaders. But they must contain exactly one main function when combined... I'd assume there will be a GLSL link error due to missing main.

modules/opengl/src/shader/shader.cpp
```diff
@@ -97,11 +106,11 @@ Shader::Shader(const Shader &rhs) : id_{glCreateProgram()}, warningLevel_{rhs.wa
     ShaderManager::getPtr()->registerShader(this);
 }
 
-Shader::Shader(Shader &&rhs) : id_{rhs.id_}, ready_(rhs.ready_), warningLevel_{rhs.warningLevel_} {
+Shader::Shader(Shader &&rhs) : ready_(rhs.ready_), warningLevel_{rhs.warningLevel_} {
 
     ShaderManager::getPtr()->unregisterShader(&rhs);
-
-    rhs.id_ = 0;
+    prog_ = rhs.prog_;
```
martinfalk: there is a leak here! `this->prog_` is already initialized and has a program. Either you delete that program first or swap `prog_` with `rhs.prog_`

modules/opengl/src/shader/shader.cpp
```diff
@@ -118,9 +127,9 @@ Shader::Shader(Shader &&rhs) : id_{rhs.id_}, ready_(rhs.ready_), warningLevel_{r
 Shader &Shader::operator=(const Shader &that) {
     if (this != &that) {
         bool needRegister = false;
-        if (id_ == 0) {
+        if (prog_.id == 0) {
```
martinfalk: why would `prog_.id` ever be zero?

modules/opengl/src/shader/shader.cpp
martinfalk: What about `ShaderManager::getPtr()->unregisterShader(this);` ?

See destructor and move assignment.

martinfalk (changes requested):

modules/opengl/src/shader/shader.cpp
ResearchDaniel: I think that would be a different error. Here we just check that one did not miss providing a vertex shader. I guess we could add a search for a main function but it would make it a bit more complicated and a bit out of scope for the PR?

modules/opengl/src/shader/shader.cpp
ResearchDaniel: This is the same behavior as before. I assumed that  `ShaderManager::getPtr()->unregisterShader(&rhs);` above deleted the program. Will double check

ResearchDaniel: So I checked and it does not seem remove the program in that way. Will adjust the ScopedShaderProgram

modules/opengl/src/shader/shader.cpp
ResearchDaniel: Unlikely, but could be error during creation or moved beforehand. Not part of the PR at least :)

modules/opengl/src/shader/shader.cpp
ResearchDaniel: Fixed it

ResearchDaniel: Not necessary in move constructor

ResearchDaniel:
@petersteneteg @martinfalk I think that I adjusted all the comments

modules/opengl/src/shader/shader.cpp
```diff
@@ -97,11 +112,11 @@ Shader::Shader(const Shader &rhs) : id_{glCreateProgram()}, warningLevel_{rhs.wa
     ShaderManager::getPtr()->registerShader(this);
 }
 
-Shader::Shader(Shader &&rhs) : id_{rhs.id_}, ready_(rhs.ready_), warningLevel_{rhs.warningLevel_} {
+Shader::Shader(Shader &&rhs)
+    : prog_{rhs.prog_.id}, ready_(rhs.ready_), warningLevel_{rhs.warningLevel_} {
```
petersteneteg: This should be done by the wapper class

modules/opengl/src/shader/shader.cpp
```diff
@@ -118,9 +133,9 @@ Shader::Shader(Shader &&rhs) : id_{rhs.id_}, ready_(rhs.ready_), warningLevel_{r
 Shader &Shader::operator=(const Shader &that) {
     if (this != &that) {
         bool needRegister = false;
-        if (id_ == 0) {
+        if (!prog_.created()) {
             needRegister = true;
-            id_ = glCreateProgram();
+            prog_.id = glCreateProgram();
```
petersteneteg: Same here, this is leaking encapsulation...

modules/opengl/src/shader/shader.cpp
ResearchDaniel: Not possible since we need to deregister the rhs shader as far as I understood. This applies in a couple of situations where it is not possible to do something without knowing about the shader (assignment). I therefore tried to keep the ScopedShaderProgram as light-weight as possible

modules/opengl/src/shader/shader.cpp
ResearchDaniel: Same reasoning here, the ScopedShaderProgram is just responsible for scoping, not other parts since I could not find a good way of moving all responsibilities to it...

petersteneteg: That kind of shared responsibility does not look like a very attractive pattern...

ResearchDaniel: Agree, but I think it would require quite some changes otherwise. Handing over a reference to the shader would possibly resolve it but that is also quite ugly. Happy to take suggestions on how to resolve it in a better way.

modules/opengl/src/shader/shader.cpp
martinfalk: yes, I think so, too.

martinfalk (approved):

modules/opengl/src/shader/shader.cpp
petersteneteg: Let me have a look on this...

petersteneteg: Easiest way would have been to just check for the vertex shader, before ever creating the program..

petersteneteg (dismissed):
This need more work

modules/opengl/include/modules/opengl/shader/standardshaders.h
```diff
@@ -0,0 +1,48 @@
...
+#include <inviwo/core/common/inviwo.h>
+
+#include <modules/opengl/shader/shaderresource.h>
+#include <modules/opengl/shader/shadertype.h>
+
+namespace inviwo {
+
+namespace utilgl {
+
+IVW_MODULE_OPENGL_API std::pair<ShaderType, std::shared_ptr<const ShaderResource>> imgQuadVert();
```
ResearchDaniel: Should document what this one does. Provide the shader as documentation?
Would also rename it to imgQuadVertexShader(), possibly even adding create, i.e., createImgQuadVertexShader

modules/opengl/include/modules/opengl/shader/standardshaders.h
```diff
@@ -0,0 +1,48 @@
...
+
+IVW_MODULE_OPENGL_API std::pair<ShaderType, std::shared_ptr<const ShaderResource>>
+imgIdentityVert();
```
ResearchDaniel, line 45: Same here, document what it does and possibly rename

modules/opengl/src/shader/standardshaders.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+#include <modules/opengl/shader/standardshaders.h>
+
+namespace inviwo {
+
+namespace utilgl {
+std::pair<ShaderType, std::shared_ptr<const ShaderResource>> imgQuadVert() {
+    static const std::shared_ptr<const ShaderResource> vert =
+        std::make_shared<StringShaderResource>("image_quad.vert", R"(
+out vec3 texCoord_;
```
ResearchDaniel, line 40: indent missing?

ResearchDaniel (commented):
Looks good. I would add some comments about the shaders and maybe rename them to be a bit clearer

modules/opengl/src/shader/standardshaders.cpp
petersteneteg: No, using raw strings, here so no indent...

modules/opengl/include/modules/opengl/shader/standardshaders.h
petersteneteg: So its only a string shader resource, not a shaderobject, or even a shader. Hence it never creates anything, just returns the string data...

petersteneteg: the do the same thing as the img_identity.vert and img_texturequad.vert... which has not further explanation.

modules/opengl/include/modules/opengl/shader/standardshaders.h
petersteneteg: added short desc

modules/opengl/include/modules/opengl/shader/standardshaders.h
ResearchDaniel: Would just be nice to be able to understand what they do without going to the source. This could be done by just copying in the shader code in the documentation. No major issue

PR #334, opened 2018-12-13
Feature/png read write

r-englund:
I wrote a reader and writer to load PNG using libpng directly rahter than going through cimg. 
With CImg we only read and write 8 bit even if the pixels are 16. This new module supports 16 bit as well as 8. 

For reading PNG libPNG takes care of converting palettes to RGB/RGBA when needed. Hence, we always get RGB/RGBA Alpha images.
We also support GRAY and GRAY_ALPHA images and outputs them as 1 respectively 2 channel layers with the appropriate swizzle masks.  

When writing float/double images to disk we have to convert them to 16 bit ints (since png only support integers, not floats). This is done right now to using the inviwo function glm_convert_normalized wich gives the same behavior as before. 

**Testing**
For testing a wrote a regression test that uses  images from the [PngSuite](http://www.schaik.com/pngsuite/).  ~160 images of various formats. In our "regular" regression test we have used images generated by inviwo as references, for this regresion test I use the original images as reference. This though does not work well with the imagecompare.py used by the regression test since it requires the image of being of same mode (for example we can't compare a RGB with a Palette)


modules/png/src/pngwriter.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+#include <inviwo/png/pngwriter.h>
+
+#include <inviwo/core/datastructures/image/layer.h>
+#include <inviwo/core/datastructures/image/layerram.h>
+#include <inviwo/core/datastructures/image/layerramprecision.h>
+
+#include <png.h>
+
+#pragma optimize("", off)
```
petersteneteg: debug?

src/core/datastructures/image/layerram.cpp
```diff
@@ -42,7 +42,7 @@ LayerRAM::LayerRAM(size2_t dimensions, LayerType type, const DataFormatBase* for
 bool LayerRAM::copyRepresentationsTo(LayerRepresentation* targetLayerRam) const {
     static DataWriterType<Layer>* layerWriter_ = InviwoApplication::getPtr()
                                                      ->getDataWriterFactory()
-                                                     ->getWriterForTypeAndExtension<Layer>("png")
+                                                     ->getWriterForTypeAndExtension<Layer>("bmp")
```
petersteneteg, line 48: This is a horrible hack here... might be nice to add a comment explaining what is going on

tools/ivwpy/regression/imagecompare.py
```diff
@@ -44,11 +44,28 @@ def __init__(self, testImage, refImage, enhance = 10):
         self.numberOfDifferentPixels = None
         self.maxDifference = None
 
+        channels = len(self.testImage.getbands());
+
+
+        # Inviwo don't save PNG using a pallete, if test image is a pallete we
+        # need to convert it to RGB or RGBA
+        if self.refImage.mode == 'P':
```
petersteneteg: what about 1/2 channels?

modules/png/include/inviwo/png/pngreader.h
```diff
@@ -0,0 +1,65 @@
...
+#include <inviwo/core/datastructures/diskrepresentation.h>
+#include <inviwo/core/datastructures/image/layerdisk.h>
+
+namespace inviwo {
+
+    class IVW_MODULE_PNG_API PNGLayerReaderException : public Exception {
+public:
+        PNGLayerReaderException(const std::string& message = "",
+                                ExceptionContext context = ExceptionContext());
+    virtual ~PNGLayerReaderException() throw() {}
```
petersteneteg: use noexcept!

src/core/datastructures/image/layerram.cpp
r-englund: The whole thing with using a DataWriterType<Layer> for resizing/copying is very hacky, but I will add a comment.

modules/png/src/pngwriter.cpp
r-englund: Yes, will remove

tools/ivwpy/regression/imagecompare.py
r-englund: In a PNG image a palette is always representing RGB colors. Alpha can be added as an additional channel. 
Anyway, I changed this a bit, see a9bb185b115f5ba319cf72b1e957ad9dcf88de32

tools/ivwpy/regression/app.py
```diff
@@ -55,14 +55,14 @@ def findModuleTest(path):
...
 	return tests
 
 class App:
-	def __init__(self, 
-		         appPath, 
-				 moduleTestPaths, 
+	def __init__(self,
+		         appPath,
+				 moduleTestPaths,
 				 outputDir,
```
r-englund: Hrm, need to check my space vs tabs settings in my editor

src/core/datastructures/image/layerram.cpp
petersteneteg: Exactly...

modules/png/readme.md
```diff
@@ -0,0 +1,3 @@
+# png Module
+
+Description of the png module
```
ResearchDaniel: Suggestion: Enables support for reading PNG image files.

modules/png/src/pngmodule.cpp
```diff
@@ -0,0 +1,48 @@
...
+#include <inviwo/png/pngmodule.h>
+#include <inviwo/png/pngreader.h>
+#include <inviwo/png/pngwriter.h>
+#include <inviwo/png/pngutils.h>
+
+namespace inviwo {
+
+pngModule::pngModule(InviwoApplication* app) : InviwoModule(app, "png") {
+    
+    LogInfo("Using LibPNG Version " << pngutil::getLibPNGVesrion());
```
ResearchDaniel, line 39: Don't think that it is necessary to log this info.

modules/png/src/pngreader.cpp
```diff
@@ -0,0 +1,173 @@
...
+    switch (color_type) {
+        case PNG_COLOR_TYPE_GRAY_ALPHA:
+            swizzleMask = swizzlemasks::luminanceAlpha;
+            channels = 2;
+            break;
+        case PNG_COLOR_TYPE_GRAY:
+            swizzleMask = swizzlemasks::luminance;
+            channels = 1;
+            break;
+        case PNG_COLOR_TYPE_PALETTE:
```
ResearchDaniel, line 110: [[fallthrough]];

modules/png/src/pngwriter.cpp
```diff
@@ -0,0 +1,198 @@
...
+        png_set_swap(png_ptr);
+
+        auto pixels = ram->getDataTyped();
+
+        using T2 = typename util::same_extent<T, glm::uint16>::type;
+        std::vector<T2> newData(size.x * size.y);
+        for (int i = 0; i < size.x * size.y; i++) {
+            const static T zero(0);
+            const static T one(1);
+            newData[i] = util::glm_convert_normalized<T2>(glm::clamp(pixels[i], zero, one));
```
ResearchDaniel, line 116: Should this not use min/max of the layer data for normalization?

ResearchDaniel (changes requested):

modules/png/src/pngwriter.cpp
r-englund: This gives the same behaviour as CImg and will produce the same images as is shown in the canvas. This is used when saving depth textures as pngs, and for the depth textures in thumbnails. 

If we had a datarange/value ranges as we have for volumes we could consider using it here

modules/png/src/pngmodule.cpp
r-englund: This log message was in the CImg module before, I just moved it here. The CImg Module has similar log messages for tiff and jpg libs.

ResearchDaniel (approved):

PR #337, opened 2018-12-20
Webbrowser: Prevent error page from showing when deserializing

ResearchDaniel:


modules/webbrowser/src/properties/propertycefsynchronizer.cpp
```diff
@@ -80,8 +85,6 @@ void PropertyCefSynchronizer::OnLoadError(CefRefPtr<CefBrowser> browser, CefRefP
         ss << "<br/>Description: " << errorText.ToString();
     }
 
-    ss << "</body></html>";
```
martinfalk: any particular reason why the closing tags are removed?

martinfalk (approved):

modules/webbrowser/src/properties/propertycefsynchronizer.cpp
ResearchDaniel: Hmm, looks like a mistake. Should be reverted. Good catch

PR #341, opened 2019-01-03
Set style override to suppress GTK warning

PetiteViking:
Got the warning "QApplication: invalid style override passed, ignoring it." when starting Inviwo on Ubuntu. See https://forum.qt.io/topic/75398/qt-5-8-0-qapplication-invalid-style-override-passed-ignoring-it/2

### DIFF
--- a/apps/inviwo/inviwo.cpp
+++ b/apps/inviwo/inviwo.cpp
@@ -55,6 +55,7 @@ int main(int argc, char** argv) {
     inviwo::LogCentral::init(&logger);
     auto logCounter = std::make_shared<inviwo::LogErrorCounter>();
     logger.registerLogger(logCounter);
+    qputenv("QT_STYLE_OVERRIDE","");
     inviwo::InviwoApplicationQt inviwoApp(argc, argv, "Inviwo");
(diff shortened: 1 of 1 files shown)


apps/inviwo/inviwo.cpp
```diff
@@ -55,6 +55,7 @@ int main(int argc, char** argv) {
     inviwo::LogCentral::init(&logger);
     auto logCounter = std::make_shared<inviwo::LogErrorCounter>();
     logger.registerLogger(logCounter);
+    qputenv("QT_STYLE_OVERRIDE","");
```
petersteneteg: I would add the link/comment here also, Should this only happen on Linux?

petersteneteg (changes requested):

apps/inviwo/inviwo.cpp
JochenJankowai: As far as I can see, this problem is related to GTK which is a widget toolkit used by several desktop environments for Linux-based systems. It is primarily targeted for X11 and Wayland but can technically run on macOS and Windows. I think that the latter is a virtually non-existent use case though. So yes, I believe this should only happen on Linux. Should I add the corresponding macro as well?

petersteneteg (approved):

PR #343, opened 2019-01-08
Global clang format

petersteneteg:
To enable format checking on jenkins...


tools/refactoring/format.py
```diff
@@ -0,0 +1,24 @@
...
+
+import refactoring # Note: refactoring.py need to be in the current working directory
+
+paths = [
+	"C:/Users/petst55/Work/Inviwo/Inviwo-dev-mirror"
+]
+
+excludespatterns = ["*/ext/*", "*/templates/*", "*/tools/codegen/*" , "*moc_*", "*cmake*"];
+
+files = refactoring.find_files(paths, ['*.h', '*.hpp' '*.cpp'], excludes=excludespatterns)
```
r-englund: Missing a comma between hpp and cpp so it will only run on .h files

r-englund (approved):
Script only runs on .h files (not .cpp and .hpp), see comment

PR #345, opened 2019-01-08
Plotting: Added tooltips to the parallel coordinates plot

r-englund:
Added tooltip to the PCP when hovering a line and when moving an axis-handle

image

### DIFF
--- a/modules/plotting/include/modules/plotting/datastructures/dataframeutil.h
+++ b/modules/plotting/include/modules/plotting/datastructures/dataframeutil.h
@@ -52,6 +52,9 @@ std::shared_ptr<plot::DataFrame> IVW_MODULE_PLOTTING_API
 combineDataFrames(std::vector<std::shared_ptr<plot::DataFrame>> histogramTimeDataFrame,
                   bool skipIndexColumn = false, std::string skipcol = "index");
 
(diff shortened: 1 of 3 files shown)


martinfalk (approved):

modules/plotting/include/modules/plotting/datastructures/dataframeutil.h
```diff
@@ -52,6 +52,9 @@ std::shared_ptr<plot::DataFrame> IVW_MODULE_PLOTTING_API
 combineDataFrames(std::vector<std::shared_ptr<plot::DataFrame>> histogramTimeDataFrame,
                   bool skipIndexColumn = false, std::string skipcol = "index");
 
+std::string IVW_MODULE_PLOTTING_API createToolTipForRow(const plot::DataFrame &dataframe,
```
ResearchDaniel, line 55: Would be nice if one could specify which columns to show

ResearchDaniel (approved):
Just one feature request :)

modules/plotting/include/modules/plotting/datastructures/dataframeutil.h
r-englund: I agree, and that is something I probably will implement later when I need it, in another PR. I added the util-function so that it would be easier to do. 
Will also add tooltip to the scatterplot in the future.

PR #350, opened 2019-01-09
Core/Base: Fix for basis/world-matrix property ranges in the volumeinformation processor

r-englund:
The min/max/inc values for the matrixPorperties in VolumeInformationProcessor was a diagonal matrix and hence could not display a rotation matrix (or any non-diagonal matrix) correctly.

### DIFF
--- a/include/inviwo/core/util/glm.h
+++ b/include/inviwo/core/util/glm.h
@@ -812,7 +812,7 @@ struct almostEqual<glm::mat<C, R, T, Q>> {
 
 }  // namespace detail
 
-template <typename M, typename T = M::value_type>
+template <typename M, typename T = typename M::value_type>
 M filledMat(T v) {
     return M{T(0)} + v;
 }
--- a/include/inviwo/core/util/glm.h
+++ b/include/inviwo/core/util/glm.h
@@ -812,6 +812,11 @@ struct almostEqual<glm::mat<C, R, T, Q>> {
 
 }  // namespace detail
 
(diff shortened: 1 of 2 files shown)


include/inviwo/core/util/glm.h
```diff
@@ -812,6 +812,11 @@ struct almostEqual<glm::mat<C, R, T, Q>> {
 
 }  // namespace detail
 
+template <typename M, typename T = M::value_type>
+M filledMat(T v) {
```
petersteneteg, line 816: Would it work with just fill? Or filled?

petersteneteg (approved):

include/inviwo/core/util/glm.h
r-englund: I'm fine with either. My though with including the Mat part is to indicate that the function is aimed for matrices (since it is not needed for vectors), ´util::filled<T>(v)´ could look like it also works on other datatypes as well (std::vector, images etc).

r-englund: I guess, technically, the function works for any datatype that has a single value constructor and operator+ taking a M::value_type parameters.

petersteneteg: for generic code using both vec and mat, it can be nice I think... might add a comment about intended usage?

petersteneteg: I would call it just filled

r-englund:
Rebased, see #353

PR #354, opened 2019-01-14
PlottingGL: Fixed issue where tooltip same text was displayed in othe…

ResearchDaniel:
…r parts of the screen. Made it possible to deselect an item in parallel coordinates


petersteneteg (approved):

petersteneteg:
@r-englund since we made it public, so one can get to the actual range value

ResearchDaniel:
> Why remove the underscores?

@r-englund  Made the two properties public.

PR #364, opened 2019-01-18
Feature/workspaces

martinfalk:
* updated workspaces to latest version and to include preview images.
* added commandline arg for updating only workspace in a specific directory


tools/changelog-to-html.py
```diff
@@ -164,6 +159,17 @@ def makeCmdParser():
 def main(args):
 	args = makeCmdParser();
 
+	if len(missing_modules)>0: 
+		print_error("Warning: Could not generate HTML changelog. Missing python modules:")
+		for k,v in missing_modules.items():
+			print_error("    {:20s} {}".format(k,v))	
+		print_info("    To install run: 'pip install {}'".format(" ".join(missing_modules.keys())))
```
ResearchDaniel: I would still argue for
Python -m pip install 
since pip install requires the script path to be in the PATH on windows
https://stackoverflow.com/questions/34247930/pip-not-working-in-python-3-5-on-windows

include/inviwo/core/network/workspaceutils.h
```diff
@@ -41,6 +41,8 @@ namespace util {
 
 void IVW_CORE_API updateWorkspaces(InviwoApplication* app);
 
+void IVW_CORE_API updateWorkspacesInPath(InviwoApplication* app, const std::string &path);
```
ResearchDaniel: could also be 
`void IVW_CORE_API updateWorkspaces(InviwoApplication* app, const std::string &inPath)`

modules/python3/src/python3module.cpp
```diff
@@ -50,7 +50,7 @@ Python3Module::Python3Module(InviwoApplication* app)
     : InviwoModule(app, "Python3")
     , pythonInterpreter_(util::make_unique<PythonInterpreter>(this))
     , pythonScriptArg_("p", "pythonScript", "Specify a python script to run at startup", false, "",
-                       "Path to the file containing the script")
+                       "python script")
```
ResearchDaniel, line 53: Still a file?

src/qt/editor/inviwomainwindow.cpp
```diff
@@ -130,7 +130,14 @@ InviwoMainWindow::InviwoMainWindow(InviwoApplicationQt* app)
                              "path"}
     , openData_{"d", "data", "Try and open a data file", false, "", "file name"}
     , updateWorkspaces_{"", "update-workspaces",
-                        "Go through and update all workspaces to the latest version"}
+                        "Update workspaces of all modules to the latest version "
+                        "(located in '<module>/data/workspaces/*')"}
```
ResearchDaniel, line 134: module->modules

ResearchDaniel (changes requested):
Great, some minor comments

modules/python3/src/python3module.cpp
martinfalk: `"python script"` refers to the short part of the argument, not its description!
Cf.
```
inviwo.exe  [-p <python script>]
                                        [--update-workspaces-in-path

-p <python script>,  --pythonScript <python script>
     Specify a python script to run at startup
```

src/qt/editor/inviwomainwindow.cpp
martinfalk: don't agree. This refers to the respective module path. You maybe refer to `modules/<module>/data/...`?

include/inviwo/core/network/workspaceutils.h
martinfalk: any reason why?

tools/changelog-to-html.py
martinfalk: fixed

src/qt/editor/inviwomainwindow.cpp
ResearchDaniel: Yes, you are correct

include/inviwo/core/network/workspaceutils.h
ResearchDaniel: More common to overload function than to prepend argument to its name.

martinfalk:
PR rebased into #366

PR #370, opened 2019-01-24
Feature/various

martinfalk:
various bug fixes


src/qt/editor/toolsmenu.cpp
```diff
@@ -80,7 +80,7 @@ void createProcessorDocMenu(InviwoApplication *app, QMenu *docsMenu) {
                 const auto processor = factory->create(pfo->getClassIdentifier());
                 const auto &inports = processor->getInports();
                 const auto &outports = processor->getOutports();
-                const auto &properties = processor->getProperties();
```
r-englund, line 83: Well, this will include all properties, but not the hierarchy. eg:
* Raycasting properties
* TF
* SamplingRate
* ...
* Lighting Property
* Ambiend
* Diffuse 
* ... 

### where one might want something like this

* Raycasting properties
* * TF
* * SamplingRate
* * ...
* Lighting Property
* * Ambiend
* * Diffuse 
* * ...

martinfalk: So far, the indentation is hardly used (except for VolumeSlice) for composites. But I've stumbled upon this ('noiseprocessor.h') where a second level is used to indicate options of an optionproperty (and that pattern appears several times):
```c++
/**
 * ### Properties
 *   * __size__ Size of the output image.
 *   * __type__ Witch type of noise to generate.
 *   * __range__ The min/max values of the output values (default: [0 1]).
 *   * __Perlin Noise:__
 *     + __persistence__ Controls the sharpnes in Perlin noise
 *     + __levels__ Numbers of levels used in the generation.
 *   * __Poisson Disk Sampling:__
 *     + __poissonDotsAlongX__ Average number of points along the x-axis.
 *     + __poissonMaxPoints__ Maximum number of output points (total).
 */
```

modules/basegl/src/processors/sphererenderer.cpp
```diff
@@ -69,7 +69,8 @@ SphereRenderer::SphereRenderer()
     , forceColor_("forceColor", "Force Color", false, InvalidationLevel::InvalidResources)
     , defaultColor_("defaultColor", "Default Color", vec4(0.7f, 0.7f, 0.7f, 1.0f), vec4(0.0f),
                     vec4(1.0f))
-    , useMetaColor_("useMetaColor", "Use meta color mapping", false)
+    , useMetaColor_("useMetaColor", "Use meta color mapping", false,
```
petersteneteg: need this in the cube and tube renderer too...

petersteneteg (approved):

PR #371, opened 2019-01-28
Feature/parallel coordinates delayed brushing

Alboin:
The added option gives a huge performance boost to the parallel coordinates plot when the brushing is linked to heavy computations such as computations on volume sequences.

- Windows 10
- Visual Studio 15.7.6
- QT Community 4.6.1
- CMake 3.11.2
- Python 3.6 (32-bit)


modules/plottinggl/include/modules/plottinggl/processors/parallelcoordinates/parallelcoordinates.h
```diff
@@ -109,11 +109,15 @@ class IVW_MODULE_PLOTTINGGL_API ParallelCoordinates : public Processor {
                     const std::vector<ParallelCoordinatesAxisSettingsProperty *> &enabledAxis);
 
     void updateBrushing();
+    std::unordered_set<size_t> brushedID;
```
ResearchDaniel: We use an underscore for protected/private member variables, i.e. brushedIDs_

modules/plottinggl/include/modules/plottinggl/processors/parallelcoordinates/parallelcoordinates.h
```diff
@@ -109,11 +109,15 @@ class IVW_MODULE_PLOTTINGGL_API ParallelCoordinates : public Processor {
...
 
     DataInport<DataFrame> dataFrame_;
     BrushingAndLinkingInport brushingAndLinking_;
     ImageOutport outport_;
 
+    BoolProperty brushOnlyWhenMouseRelease_;
+    bool mouseReleased = true;
```
ResearchDaniel: mouseReleased_
The implementation also support touch :)
I think that this variable could be removed with some refactoring. See later comment

modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
```diff
@@ -746,6 +753,10 @@ void ParallelCoordinates::handlePicked(PickingEvent *p) {
         axisVector_[axisID]->moveHandle(upper, newY);
         p->markAsUsed();
     }
+    if (p->getPressState() == PickingPressState::Release) {
+        mouseReleased = true;
+        updateBrushing();
+    }
 }
 
 void ParallelCoordinates::updateBrushing() {
```
ResearchDaniel, line 762: updatedBrushing could take PickingState as input. Would remove mouseReleased variable.

modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
```diff
@@ -571,7 +575,7 @@ void ParallelCoordinates::drawLines(size2_t size) {
     lineShader_.setUniform("selected", 0);
     lineShader_.setUniform("filtered", 0);
     for (size_t i = 0; i < numLines; i++) {
-        if (brushingAndLinking_.isFiltered(indexCol[i])) {
+        if (brushedID.find(indexCol[i]) != brushedID.end()) {
```
ResearchDaniel: How will this affect filtering from other components? Should it not be brushingAndLinking_.isFiltered(indexCol[i]) || brushedID.find(indexCol[i]) != brushedID.end() ?

ResearchDaniel (changes requested):
Some comments

r-englund:
Haven't had time to look into it in detail. But it feels unnecessary to add a new index set, since one already exists in the BrushingAndLinkingInport. I wounder if we could move this to the BrushingAndLinkingInport instead. Like a pause for event propagation, when the PCP processor gets a mouse down event it pauses the the events sent from BrushingAndLinkingInport and when it releases it "unpause" it and the BrushingAndLinkingInport would send any (should be enough with only the latest of them) filter/selection event done while paused.

Alboin:
As I mentioned verbally there is some problems with the delayed update of brushing filter events. When the feature is activated in the parallel coordinates plot it does not fire a mouserelease-event properly (which will update the brushing) when the mouse is released over another plot or such. 

But @ResearchDaniel told me that this might be fixed with https://github.com/inviwo/inviwo/commit/6c99ee306ee4e8f33ebae13e6da110e64c06bc69.

Alboin:
> As I mentioned verbally there is some problems with the delayed update of brushing filter events. When the feature is activated in the parallel coordinates plot it does not fire a mouserelease-event properly (which will update the brushing) when the mouse is released over another plot or such.
> 
> But @ResearchDaniel told me that this might be fixed with [6c99ee3](https://github.com/inviwo/inviwo/commit/6c99ee306ee4e8f33ebae13e6da110e64c06bc69).

Great idea! I'd tried merging the commit onto this branch but the problem persists...

Alboin:
> > As I mentioned verbally there is some problems with the delayed update of brushing filter events. When the feature is activated in the parallel coordinates plot it does not fire a mouserelease-event properly (which will update the brushing) when the mouse is released over another plot or such.
> > But @ResearchDaniel told me that this might be fixed with [6c99ee3](https://github.com/inviwo/inviwo/commit/6c99ee306ee4e8f33ebae13e6da110e64c06bc69).
> 
> Great idea! I'd tried merging the commit onto this branch but the problem persists...

Well at least you tried @Alboin 👍 better luck next time!

Maybe we can merge this PR onto master anyway and fix this issue later on?

modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
```diff
@@ -55,6 +55,7 @@ class IVW_MODULE_BRUSHINGANDLINKING_API BrushingAndLinkingManager {
     void remove(const BrushingAndLinkingInport* src);
 
     bool isFiltered(size_t idx) const;
+    bool isFilteredByOther(size_t idx, const inviwo::BrushingAndLinkingInport* src) const;
```
ResearchDaniel, line 58: I think that the BrushingAndLinkingManager should be responsible for pausing/resuming the filtering instead of the brushing and linking ports. 
Then we would not need this extra function?

modules/brushingandlinking/include/modules/brushingandlinking/ports/brushingandlinkingports.h
```diff
@@ -55,13 +55,20 @@ class IVW_MODULE_BRUSHINGANDLINKING_API BrushingAndLinkingInport
     bool isFiltered(size_t idx) const;
     bool isSelected(size_t idx) const;
 
+    void pauseFilterEvents();
```
ResearchDaniel, line 58: Add documentation on the purpose and what happens after calling these functions

modules/brushingandlinking/src/ports/brushingandlinkingports.cpp
```diff
@@ -44,8 +44,10 @@ BrushingAndLinkingInport::BrushingAndLinkingInport(std::string identifier)
 void BrushingAndLinkingInport::sendFilterEvent(const std::unordered_set<size_t> &indices) {
     if (filterCache_.size() == 0 && indices.size() == 0) return;
     filterCache_ = indices;
-    FilteringEvent event(this, filterCache_);
-    getProcessor()->propagateEvent(&event, nullptr);
+    if (!filterEventsPaused_) {
```
ResearchDaniel, line 47: This could be moved to the BrushingAndLinkingManager?

ResearchDaniel (changes requested):
After discussion: How about adding an "isInteracting" state to the BrushingAndLinkingManager. Compute intensive processors connected to brushing and linking would check the state and optionally recompute when interaction ends.

modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
petersteneteg: how does the manager know which event producer that is paused? this seems like a local state, why should we manage it globally?

petersteneteg:
> After discussion: How about adding an "isInteracting" state to the BrushingAndLinkingManager. Compute intensive processors connected to brushing and linking would check the state and optionally recompute when interaction ends.

This is a quite different implementation all together. I will requite all other consumers to be updated to check the "isInteracting" flag and then skip updating. The current solution does not need and changes to any of the other consumers

Alboin:
> 
> This is a quite different implementation all together. I will requite all other consumers to be updated to check the "isInteracting" flag and then skip updating. The current solution does not need and changes to any of the other consumers

The drawback with "current solution" is that when pausing the brushing for one processor, it pauses on all others as well. With Daniels approach one could select to pause the brushing on only the more computationally heavy processors. A linked scatterplot could then be updated in realtime with the PCP but a volume computation be configured to wait until end of interaction to update, as an example.

petersteneteg:
> > This is a quite different implementation all together. I will requite all other consumers to be updated to check the "isInteracting" flag and then skip updating. The current solution does not need and changes to any of the other consumers
> 
> The drawback with "current solution" is that when pausing the brushing for one processor, it pauses on all others as well. With Daniels approach one could select to pause the brushing on only the more computationally heavy processors. A linked scatterplot could then be updated in realtime with the PCP but a volume computation be configured to wait until end of interaction to update, as an example.

That is a good point. Given that all other processors also support it....

petersteneteg:
I plan to have a larger discussion about the brushing soonish. This could be further discussed then,...

martinfalk:
@Alboin 
> As I mentioned verbally there is some problems with the delayed update of brushing filter events. When the feature is activated in the parallel coordinates plot it does not fire a mouserelease-event properly (which will update the brushing) when the mouse is released over another plot or such.
> But @ResearchDaniel told me that this might be fixed with [6c99ee3](https://github.com/inviwo/inviwo/commit/6c99ee306ee4e8f33ebae13e6da110e64c06bc69).

The issue you are referring to is twofold. Mouse release events within the same canvas are now propagated/fired accordingly as in 6c99ee3. However, using picking within an `ImageLayout` is not quite correct. At the moment, picking events are __only__ forwarded to the view under the mouse. Meaning when starting to interact in one view and releasing the mouse button in another view, the first view will **not** be notified. This needs to be fixed.

Alboin:
Any progress on this @petersteneteg, @martinfalk, @ResearchDaniel ? It would really improve the performance of the brushing and computations on volume sequences that I use, so I'm eager to have it implemented in one way or another :)

ResearchDaniel:
So, I have seen other systems where they instead set a time interval before new filtering events can be sent. That might be a better approach?

Alboin:
That sounds confusing. Then the filtering event would most of the time be sent either before you're finished filtering or afterwards, with an added delay. Instead of sending an event exactly when you're done filtering.

ResearchDaniel:
I guess you can "force" send it when releasing the button.

Alboin:
But why having an interval in the first place then? To make fewer updates along the way?

Alboin:
Pling plong! 🔔⏳

ResearchDaniel:
This PR has practically been rendered useless thorugh proper use of background threads :)

Close it? @Alboin @petersteneteg

Alboin:
Sure!  👍

PR #375, opened 2019-01-29
Fix/inputselector property update

Alboin:
**Updated feature / Fixes** 
The inputselector processor previously rebuilt the OptionPropertyInt-list, with which selection of input was made, at every input_.onChange(). Through this commit, the list is only rebuilt when the connections to the processor inport is changed.

Problems arose with the previous implementation when synchronizing the OptionPropertyInt to other processors, as every change on input_ updated all other processors synchronized to this.

- Windows 10
- Visual Studio 15.7.6
- QT Community 4.6.1
- CMake 3.11.2
- Python 3.6 (32-bit)


ResearchDaniel (changes requested):
Input selection looks good but there are other changes for the brushing and linking included

Alboin:
Will remove this branch and make a better one 🌟

Alboin:
Branch replaced by: https://github.com/inviwo/inviwo/pull/378

PR #376, opened 2019-01-29
WebBrowser: Last character in created json message was missing.

Alboin:
- Windows 10
- Visual Studio 15.7.6
- QT Community 4.6.1
- CMake 3.11.2
- Python 3.6 (32-bit)

### DIFF
--- a/modules/webbrowser/src/properties/propertycefsynchronizer.cpp
+++ b/modules/webbrowser/src/properties/propertycefsynchronizer.cpp
@@ -114,7 +114,7 @@ bool PropertyCefSynchronizer::OnQuery(CefRefPtr<CefBrowser> browser, CefRefPtr<C
         if (nextValPos == std::string::npos) {
             LogWarn("Missing enclosing } in: " + requestStr);
         } else {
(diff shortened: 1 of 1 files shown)


modules/webbrowser/src/properties/propertycefsynchronizer.cpp
```diff
@@ -114,7 +114,7 @@ bool PropertyCefSynchronizer::OnQuery(CefRefPtr<CefBrowser> browser, CefRefPtr<C
         if (nextValPos == std::string::npos) {
             LogWarn("Missing enclosing } in: " + requestStr);
         } else {
-            auto message = requestStr.substr(0, nextValPos);
+            auto message = requestStr.substr(0, nextValPos + 1);
```
petersteneteg, line 117: you should really use our json library to do the json parsing... look in ext/json...

ResearchDaniel: Agree, but that would be another PR

ResearchDaniel (approved):

PR #380, opened 2019-01-31
Base: InputSelector: Fix for updatePorts being called multiple times while deserialzing

r-englund:
An inports onConnect function is called several times while deserializing a network. Once per connected outport, first time it is called it only has one connection, second time two connections and so on. This lead to losing which option was selected when serializing and always the first connected port would be used.

### DIFF
--- a/modules/base/include/modules/base/processors/inputselector.h
+++ b/modules/base/include/modules/base/processors/inputselector.h
@@ -80,6 +80,7 @@ private:
     OptionPropertyInt selectedPort_;
 
     void updateOptions();
+    bool updatedNedded_ = true;
 };
 
 template <typename Inport, typename Outport>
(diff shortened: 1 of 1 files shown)


petersteneteg (approved):
Looks like something that we maybe should solve in the deserialization? Would that be possible?

modules/base/include/modules/base/processors/inputselector.h
```diff
@@ -113,9 +114,12 @@ InputSelector<Inport, Outport>::InputSelector()
 
 template <typename Inport, typename Outport>
 void InputSelector<Inport, Outport>::updateOptions() {
+    std::string selectedID;
+    if(selectedPort_.size() != 0){
+        selectedID = selectedPort_.getSelectedIdentifier();
+    }
```
martinfalk: `TemplateOptionProperty::replaceOptions` should be used. It takes care of maintaining the selected item, if it continues to exist.

martinfalk (changes requested):

modules/base/include/modules/base/processors/inputselector.h
```diff
@@ -114,28 +114,20 @@ InputSelector<Inport, Outport>::InputSelector()
...
-    if(selectedPort_.size() != 0){
-        selectedID = selectedPort_.getSelectedIdentifier();
-    }
-    selectedPort_.clearOptions();
-    
-    int idx = 0;
+    std::vector<OptionPropertyIntOption> options;
+
     for (auto port : inport_.getConnectedOutports()) {
         auto id = port->getProcessor()->getIdentifier();
```
petersteneteg, line 120: Should this be identifier or displayname?

martinfalk: Suggesting `getDisplayName()` since that one will be more informative. See #381

martinfalk (approved):

PR #387, opened 2019-02-05
Fix/utf8 bom

ResearchDaniel:
Fixes issue where the BOM-character in csv-files saved by Excel where interpreted as real data. We now ignore the BOM when reading CSV and text files.


petersteneteg (approved):
👍

PR #388, opened 2019-02-05
Fix/numberedit

martinfalk:
* new widget `NumerLineEdit` based on a Qt SpinBox with the following improvements
   - needs less space (QSpinBox reserves space for lower/upper limit including full precision making it really wide for arbitrary precision)
   - if the number does not fit, it is rounded. If still too long, it is shown in scientific notation
   - full number shown during editing
   - accepts scientific notation, i.e. `1.434e-4` (also for integer values, assuming positive exponent)
* proper alignment of sliders and spinboxes in property list
* `NumberLineEdit` replaces all `CustomDoubleSpinBoxQt` widgets for ordinal and range properties
* `Spherical` and `SpinBoxSpherical` semantics have wrap-around enabled for phi and theta

**Before:**
image

**After:**
image

**While editing**
image

### DIFF
--- a/modules/qtwidgets/src/numberlineedit.cpp
+++ b/modules/qtwidgets/src/numberlineedit.cpp
@@ -37,13 +37,14 @@
 #include <QDoubleValidator>
 #include <QEvent>
 #include <QLocale>
+#include <QStyleOptionSpinBox>
+#include <QStyle>
+#include <QApplication>
 #include <warn/pop>
 
 #include <numeric>
 #include <unordered_map>
 #include <array>
-
-#pragma optimize("", off)
 
 namespace inviwo {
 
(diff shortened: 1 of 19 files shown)


modules/qtwidgets/src/numberlineedit.cpp
```diff
@@ -0,0 +1,279 @@
...
+#include <QDoubleValidator>
+#include <QEvent>
+#include <QLocale>
+#include <warn/pop>
+
+#include <numeric>
+#include <unordered_map>
+#include <array>
+
+#pragma optimize("", off)
```
petersteneteg: remove

petersteneteg (approved):

PR #408, opened 2019-02-11
PlottingGL: Added button to reset the PCP handle positions.

Alboin:
* Windows 10
* Visual Studio 15.7.6
* QT Community 4.6.1
* CMake 3.11.2
* Python 3.6 (32-bit)

### DIFF
--- a/modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
+++ b/modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
@@ -295,7 +295,7 @@ ParallelCoordinates::ParallelCoordinates()
     resetHandlePositions_.onChange([&]() {
         for (auto axis : axisVector_) {
             axis->moveHandle(true, std::numeric_limits<double>::max());
-            axis->moveHandle(false, std::numeric_limits<double>::min());
+            axis->moveHandle(false, std::numeric_limits<double>::lowest());
         }
     });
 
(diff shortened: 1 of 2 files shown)


modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
```diff
@@ -288,6 +292,13 @@ ParallelCoordinates::ParallelCoordinates()
     });
     selectedColorAxis_.onChange([&]() { recreateLines_ = true; });
 
+    resetHandlePositions_.onChange([&]() {
+        for (auto axis : axisVector_) {
+            axis->moveHandle(true, std::numeric_limits<double>::max());
+            axis->moveHandle(false, std::numeric_limits<double>::min());
```
petersteneteg: Probably wany lowest not min...

Alboin: Does it make any difference? As far as I understand I send the mouse Y position as arguments, which should never be negative anyways, right?

Alboin: But since the handles can be more or less outside of the window perhaps you're right and it would be better to use lowest...

I'll change it to lowest!

r-englund (approved):

r-englund:
Nice, I have thought of adding this my self for a while but kept postponing it :)

PR #411, opened 2019-02-12
Fix/pcp various fixes

Alboin:
Added a slider for controlling the alpha of the filtered lines in the PCP can be used to create a nicer looking plot.

Old plot:
pcful
(Notice how the non-filtered lines are parlty covered and occluded by the filtered lines)

New plot:
pcfin
(The non-filtered lines are much easier to distinct here. Since the filtered lines are in this case partly transparent, areas with many lines will appear brighter. I don't know if this is a wanted or unwanted behavior.)

Also added an option to enable different line color when hovering:
pchover
(The color is based on the color of picked lines)

* Windows 10
* Visual Studio 15.7.6
* QT Community 4.6.1
* CMake 3.11.2
* Python 3.6 (32-bit)

### DIFF
--- a/modules/plottinggl/glsl/pcp_lines.frag
+++ b/modules/plottinggl/glsl/pcp_lines.frag
@@ -4,6 +4,7 @@ in float lfalloffAlpha;
 
 uniform bool additiveBlend;
 uniform float alpha;
+uniform float filteredAlpha;
 uniform float lineWidth;
 
 uniform int selected;
@@ -27,8 +28,10 @@ void main() {
         if(subtractiveBelnding == 1){
             res.rgb = 1-res.rgb;
         }
-        if (filtered == 1) 
+        if (filtered == 1) {
              res.xyz = mix(res.xyz, filterColor.xyz, filterIntensity);
+			 res.a = filteredAlpha;
+		}
         if (additiveBlend) {
             res.a *= alpha * pow(lfalloffAlpha, falllofPower);
         }
--- a/modules/plottinggl/glsl/pcp_lines.frag
+++ b/modules/plottinggl/glsl/pcp_lines.frag
@@ -32,7 +32,6 @@ void main() {
         }
         if (filtered == 1) {
             res = mix(res, filterColor, filterIntensity);
-            res.a = filterColor.w;
         }
         if (additiveBlend) {
             res.a *= alpha * pow(lfalloffAlpha, falllofPower);
--- a/modules/plottinggl/glsl/pcp_lines.frag
+++ b/modules/plottinggl/glsl/pcp_lines.frag
@@ -24,22 +24,20 @@ uniform sampler2D tfSelection;
 void main() {
     vec4 res = vec4(1);
     if (selected > 0) {
-        res = texture(tfSelection, vec2(ltexCoord.y,0.5f));
+        res = texture(tfSelection, vec2(ltexCoord.y, 0.5f));
     } else {
-        res = texture(tf, vec2(ltexCoord.x,0.5f));
-        if(subtractiveBelnding == 1){
-            res.rgb = 1-res.rgb;
+        res = texture(tf, vec2(ltexCoord.x, 0.5f));
+        if (subtractiveBelnding == 1) {
+            res.rgb = 1 - res.rgb;
         }
         if (filtered == 1) {
-             res.xyz = mix(res.xyz, filterColor.xyz, filterIntensity);
-			 res.a = filteredAlpha;
-		}
+            res.xyz = mix(res.xyz, filterColor.xyz, filterIntensity);
+            res.a = filteredAlpha;
+        }
         if (additiveBlend) {
             res.a *= alpha * pow(lfalloffAlpha, falllofPower);
         }
-		if (hovered == 1)
-			res.xyz = texture(tfSelection, vec2(ltexCoord.y,0.5f)).xyz;
-
+        if (hovered == 1) res.xyz = texture(tfSelection, vec2(ltexCoord.y, 0.5f)).xyz;
     }
 
     PickingData = vec4(lpickColor, 1.0);
--- a/modules/plottinggl/glsl/pcp_lines.frag
+++ b/modules/plottinggl/glsl/pcp_lines.frag
@@ -10,7 +10,7 @@ uniform float lineWidth;
 uniform int selected;
 uniform vec4 selectedColor = vec4(1, 0, 0, 1);
 
-uniform int hovered;
+uniform int hovering;
 
 uniform int filtered;
 uniform int subtractiveBelnding;
@@ -31,13 +31,13 @@ void main() {
             res.rgb = 1 - res.rgb;
         }
         if (filtered == 1) {
-            res.xyz = mix(res.xyz, filterColor.xyz, filterIntensity);
-            res.a = filteredAlpha;
+            res = mix(res, filterColor, filterIntensity);
+            res.a = filterColor.w;
         }
         if (additiveBlend) {
             res.a *= alpha * pow(lfalloffAlpha, falllofPower);
         }
-        if (hovered == 1) res.xyz = texture(tfSelection, vec2(ltexCoord.y, 0.5f)).xyz;
+        if (hovering == 1) res.xyz = texture(tfSelection, vec2(ltexCoord.y, 0.5f)).xyz;
     }
 
     PickingData = vec4(lpickColor, 1.0);
--- a/modules/plottinggl/glsl/pcp_lines.frag
+++ b/modules/plottinggl/glsl/pcp_lines.frag
@@ -10,6 +10,8 @@ uniform float lineWidth;
 uniform int selected;
 uniform vec4 selectedColor = vec4(1, 0, 0, 1);
 
+uniform int hovered;
+
 uniform int filtered;
 uniform int subtractiveBelnding;
 uniform vec4 filterColor;
@@ -35,6 +37,8 @@ void main() {
         if (additiveBlend) {
             res.a *= alpha * pow(lfalloffAlpha, falllofPower);
         }
+		if (hovered == 1)
+			res.xyz = texture(tfSelection, vec2(ltexCoord.y,0.5f)).xyz;
 
     }
 
--- a/modules/plottinggl/glsl/pcp_lines.frag
+++ b/modules/plottinggl/glsl/pcp_lines.frag
@@ -32,6 +32,7 @@ void main() {
         }
         if (filtered == 1) {
             res = mix(res, filterColor, filterIntensity);
+            res.a = filterColor.w;
         }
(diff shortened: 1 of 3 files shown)


modules/plottinggl/glsl/pcp_lines.frag
```diff
@@ -27,11 +30,15 @@ void main() {
...
-        if (filtered == 1) 
+        if (filtered == 1) {
              res.xyz = mix(res.xyz, filterColor.xyz, filterIntensity);
+			 res.a = filteredAlpha;
+		}
         if (additiveBlend) {
             res.a *= alpha * pow(lfalloffAlpha, falllofPower);
         }
+		if (hovered == 1)
+			res.xyz = texture(tfSelection, vec2(ltexCoord.y,0.5f)).xyz;
```
r-englund: Formating in this shader is a bit of, not sure if clang-format runs on .frag files. But atleast change tabs to spaces

modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
```diff
@@ -96,6 +97,7 @@ ParallelCoordinates::ParallelCoordinates()
 
     , blendMode_("blendMode", "Blend Mode")
     , alpha_("alpha", "Alpha", 0.9f)
+    , filteredAlpha_("filteredAlpha", "Filtered Alpha", 0.5)
```
r-englund: append an f to the 0.5 to avoid warning

modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
```diff
@@ -610,6 +618,12 @@ void ParallelCoordinates::drawLines(size2_t size) {
         drawObject.draw(i);
     }
 
+    lineShader_.setUniform("hovered", 1);
+    lineShader_.setUniform("selected", 0);
+    lineShader_.setUniform("filtered", 0);
+    if (!brushingAndLinking_.isFiltered(hoveredLine_) && hoveredLine_ != -1)
```
r-englund: if you change the order in this if statement we do not need to do the lookup when hoveredLine_ == -1
```
if (hoveredLine_ != -1 && !brushingAndLinking_.isFiltered(hoveredLine_)){
   ...
}
```

r-englund (approved):
Nice, see some comments

petersteneteg:
generally we try never merge master into a feature branch, but you should rather rebase the feature branch onto master in case there are merge issues...

modules/plottinggl/glsl/pcp_lines.frag
```diff
@@ -4,11 +4,14 @@ in float lfalloffAlpha;
 
 uniform bool additiveBlend;
 uniform float alpha;
+uniform float filteredAlpha;
```
ResearchDaniel, line 7: Why not use filteredColor.w?

modules/plottinggl/glsl/pcp_lines.frag
```diff
@@ -4,11 +4,14 @@ in float lfalloffAlpha;
...
 uniform float lineWidth;
 
 uniform int selected;
 uniform vec4 selectedColor = vec4(1, 0, 0, 1);
 
+uniform int hovered;
```
ResearchDaniel: ->hovering
My thought: is selected, is filtered, is hovering

modules/plottinggl/glsl/pcp_lines.frag
```diff
@@ -21,18 +24,20 @@ uniform sampler2D tfSelection;
...
     } else {
-        res = texture(tf, vec2(ltexCoord.x,0.5f));
-        if(subtractiveBelnding == 1){
-            res.rgb = 1-res.rgb;
+        res = texture(tf, vec2(ltexCoord.x, 0.5f));
+        if (subtractiveBelnding == 1) {
+            res.rgb = 1 - res.rgb;
+        }
+        if (filtered == 1) {
+            res.xyz = mix(res.xyz, filterColor.xyz, filterIntensity);
```
ResearchDaniel: res = mix(res, filterColor, filterIntensity);

modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
```diff
@@ -224,6 +228,8 @@ ParallelCoordinates::ParallelCoordinates()
         // plus 1 to avoid text at the directly at the borders
     });
 
+    enableHoverColor_.onChange([&]() { invalidate(InvalidationLevel::InvalidOutput); });
```
ResearchDaniel: I don't think this is necessary. It is the default behaviour

modules/plottinggl/glsl/pcp_lines.frag
```diff
@@ -21,18 +24,20 @@ uniform sampler2D tfSelection;
...
+            res.a = filteredAlpha;
         }
-        if (filtered == 1) 
-             res.xyz = mix(res.xyz, filterColor.xyz, filterIntensity);
         if (additiveBlend) {
             res.a *= alpha * pow(lfalloffAlpha, falllofPower);
         }
-
+        if (hovered == 1) res.xyz = texture(tfSelection, vec2(ltexCoord.y, 0.5f)).xyz;
```
ResearchDaniel: This is the same as the selection color?

modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
```diff
@@ -86,6 +86,7 @@ ParallelCoordinates::ParallelCoordinates()
                            vec4(.5f, .5f, .5f, 1))
     , tf_("tf", "Line Color")
     , tfSelection_("tfSelection", "Selection Color")
+    , enableHoverColor_("enableHoverColor", "Enable Hover Color", false)
```
ResearchDaniel: On by default?
How is the hover color determined?

ResearchDaniel (changes requested):
Nice additions. Some comments and questions

modules/plottinggl/glsl/pcp_lines.frag
Alboin: Yeah that would work as well. But it's a little bit cumbersome to change the alpha of filteredColor in the GUI compared to a separate alpha-slider. Perhaps a solution would to keep the slider but synchronize it with filteredColor.w and then use that in the shader?

modules/plottinggl/glsl/pcp_lines.frag
Alboin: Yes. Could have a separate TF to set a separate hover color instead, would that be better?

Note that even though hover and selected now have the same color the selected has less color falloff or wider linewidth or something that makes them look different from each other.

modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
Alboin: Sure.

The hover color is the same as the selected color:
> Note that even though hover and selected now have the same color the selected has less color falloff or wider linewidth or something that makes them look different from each other.

ResearchDaniel:
Is there a need to adjust it often? A good default setting would be best. If that does not work out of the box, could we render the filtered lines first? That might give good results

ResearchDaniel:
No need for a separate TF for hover, only a single color needed?

ResearchDaniel:
Ok, as long as it is visibly different. Less to specify is better!

Alboin:
> No need for a separate TF for hover, only a single color needed?

Yeah one would think that. But the selection color is set through a TF so I thought the hover color would be set in the same way, if it were to be set separately. I guess there is some cases where a TF would be nice but a single color would definitely be easier to interact with.

> Ok, as long as it is visibly different. Less to specify is better!

I will let it be as it is for now then.

Alboin:
> Is there a need to adjust it often? A good default setting would be best. If that does not work out of the box, could we render the filtered lines first? That might give good results

During the creation of a workspace it could be changed quite often to experiment with different settings, and it's easier to experiment with two sliders (_filterIntensity__ and _filterAlpha__) than one slider and one dialogue window where you have to type in numbers on the keyboard. 

With 440883d the filtered lines are rendered first. This improves the look somewhat and also the picking, since some of the filtered lines would previously block the picking of non-filtered lines.

modules/plottinggl/glsl/pcp_lines.frag
```diff
@@ -31,13 +31,13 @@ void main() {
             res.rgb = 1 - res.rgb;
         }
         if (filtered == 1) {
-            res.xyz = mix(res.xyz, filterColor.xyz, filterIntensity);
-            res.a = filteredAlpha;
+            res = mix(res, filterColor, filterIntensity);
+            res.a = filterColor.w;
```
ResearchDaniel, line 35: No need to assign it afterwards (otherwise the mix above with all components won’t make a difference)

Can’t we simply assign a filtered color instead of using mix with a filter intensity?

ResearchDaniel (approved):
Looks good now but would be nice if we could simplify it even further by removing the filter-color mixing.

modules/plottinggl/glsl/pcp_lines.frag
Alboin: > No need to assign it afterwards (otherwise the mix above with all components won’t make a difference)

True, removed that row!

> Can’t we simply assign a filtered color instead of using mix with a filter intensity?

We sure can but it won't give the same result. The advantage of mixing the two colors is that you can get a hint of the filtered lines original color if you want. Another way could be to just modify the saturation of the filtered lines.

Alboin: >> No need to assign it afterwards (otherwise the mix above with all components won’t make a difference)

> True, removed that row!

Alright, seems we were both wrong. The row is needed or else alpha value of the filtered lines will be:
`res.w = mix(unfilteredAlpha, filteredAlpha, filterIntensity)`
which is not the same as
`res.w = filteredAlpha`

ResearchDaniel: That’s why I think we could assign a color. Saturation sounds like a good idea though :)

Alboin: I don't really understand what you mean with the assigned color. Isn't that what were doing right now?

ResearchDaniel: Now we are mixing the two colors. I was thinking of simply assigning the color instead:
res = filteredColor;

Alboin: > The advantage of mixing the two colors is that you can get a hint of the filtered lines original color if you want.

I think it's still a nice feature to be able to see the original colors of the filtered lines.

PR #418, opened 2019-02-14
Feature/pcp axis picking

Alboin:
I have added the ability to pick columns in the Parallel Coordinates Plot which are then sent to the brushing inport. Some properties had to be added such as axis width (hard to pick an axis that is 1 pixel in width), axis hover color and axis picking color.

Perhaps a BoolProperty should be added as well to the PCP to enable/disable axis picking/hovering, as it might not be a useful functionality for every user.

* Windows 10
* Visual Studio 15.7.6
* QT Community 4.6.1
* CMake 3.11.2
* Python 3.6 (32-bit)


modules/plottinggl/glsl/pcp_axis.vert
```diff
@@ -2,6 +2,11 @@
...
 
-void main(){
-    gl_Position = vec4(getPosWithSpacing(vec2(x,in_Vertex.y)),0,1);
-} 
\ No newline at end of file
+out vec3 pickColor;
+out vec2 texCoord;
+
+void main() {
+    pickColor = in_Normal.rgb;
```
ResearchDaniel: Normal?

modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
```diff
@@ -81,6 +80,8 @@ ParallelCoordinates::ParallelCoordinates()
     , axisProperties_("axisProps_", "Axis")
     , colors_("colors", "Colors")
     , axisColor_("axisColor", "Axis Color", vec4(.6f, .6f, .6f, 1))
+    , axisHoverColor_("axisHoverColor", "Axis Hover Color", vec4(.3f, .3f, .3f, 1))
+    , axisSelectedColor_("axisSelectedColor", "Axis Selected Color", vec4(.8f, .0f, .0f, 1))
```
ResearchDaniel: Red color as default? How about a brighter color than the other axis

modules/plottinggl/include/modules/plottinggl/processors/parallelcoordinates/parallelcoordinates.h
```diff
@@ -137,6 +140,8 @@ class IVW_MODULE_PLOTTINGGL_API ParallelCoordinates : public Processor {
     FloatProperty falllofPower_;
     FloatProperty lineWidth_;
     FloatProperty selectedLineWidth_;
+    FloatProperty axisWidth_;
+    FloatProperty selectedAxisWidth_;
```
ResearchDaniel: Would it be possible to just have a thicker axis for picking , i.e. not visually? Then we could skip these properties and use a heuristic/hard coded values instead. 
Less is more :)

ResearchDaniel (approved):
Looks good to me! If possible, it would be good with some “automatic” variables so that the user does not need to specify them

modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
Alboin: Maybe fully white for selected, light gray for hover and dark gray by default?
pcp

modules/plottinggl/include/modules/plottinggl/processors/parallelcoordinates/parallelcoordinates.h
Alboin: We could. I just haven't figured it out in a nice way just yet ;)

Do we want the picking axis to be wider by a ratio or by a fix set of pixels?

modules/plottinggl/glsl/pcp_axis.vert
Alboin: I don't know. Copied it from the shader for the PCP lines.

modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
ResearchDaniel: Probably not fully white. A bit brighter than hover.

modules/plottinggl/include/modules/plottinggl/processors/parallelcoordinates/parallelcoordinates.h
ResearchDaniel: Can you check if the fragment Coord is more than one pixel from the center of the line and draw transparent color in that case?

Probably want it to be at least 3 pixels wide? If the axis is wider there is no need to extend the picking.

modules/plottinggl/glsl/pcp_axis.vert
ResearchDaniel: You are setting the pickColor uniform in cpp code. Should not use the normal

modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
Alboin: image

Better? 🥇

Alboin: The selected axis is also a little bit wider than the rest.

Alboin:
Commit 254fe6b changes the lines of the PCP to (in my opinion) look better throughout the plot, except for at the start/end/joints of every line. They're now rendered equally wide, when previously the lines with a steep slope would be rendered thinner than those with no slope.

Old lines:
image

New lines:
image

Alboin:
@petersteneteg Started to look at changing to the newer way of using shaders as you spoke about but the amount of work it would take for **me** would not be worth the results :)

So if you want to implement it you're welcome but for the GutBrainAxis-project it's not that important, the above commits will work just as fine.

PR #421, opened 2019-02-18
Webbrowser: Process function is not called until page is loaded. 

ResearchDaniel:
Process function is not called until page is loaded. Moved responsibilities in from PropertyCefSynchronizer to WebBrowserClient.


ResearchDaniel:
@Alboin ping please review

Alboin (approved):
Don't really understand the details of it, but I trust you do.

PR #422, opened 2019-02-19
Two Axis Valuator Trackball

xeTaiz:
Hi there!
As I mentioned in the last meeting I was looking into other trackball interaction techniques. Several papers suggested the two axis valuator. This essentially maps horizontal mouse moves to rotations around camera_up and vertical moves to rotations around camera_right.
I'm putting this up here to test and get some feedback, whether you agree that this implementation feels better than the previous implementation. I have only replaced the trackball in the MeshRenderProcessorGL. The tst.ivw workspace that is included here let's you test it immediately. You can also compare by using any other processor that has a trackball attached (VolumeRaycaster or so).

I have only overriden rotate and zoom (triggered from mouse wheel) of the old trackball, so the panning, touch input etc should all work as before. Also note the fixUp property (I think its true by default in the tst.ivw) which should be pretty useful for some renders. You can also limit the vertical angle if fixUp is true using another property.

Let me know what you think. If you approve I'll make sure it is integrated nicely in all processors that have trackballs.

Greetings,
Dome


petersteneteg:
Nice, will give it a try... do you have any suggestions for how we. Could use multiple trackballs at the same time? I would for example like to have a FPS style trackall in some cases...

xeTaiz:
Since the different trackballs and the fps cam will mostly differ in the rotate method, we could just add a dropdown to choose which rotate function to use. Ideally we can make the button inputs / mouse bindings consistent for all methods then.
I will look into it. Actually I should have a FPS cam implementation in the inviwo trackball lying around from the lionman project (which also works pretty nicely for touch if i remember correctly)

ResearchDaniel:
@xeTaiz That would be great. We have discussed that we should have different camera interaction methods for some time :)

ResearchDaniel:
It seem to have some bugs (left canvas is the existing trackball and the right one is the two-axis):
http://weber.itn.liu.se/~danjo37/trackballs/trackballs.html

I have disabled the "Follow object during rotation" in the system settings in the beginning of the movie to make it more similar to how the two-axis implementation behaves. I turn it on about half-way into the movie. 


One would probably need some hints on the different types of interaction depending on where on the screen one clicks. For anyone who wants to see illustrations on how the two-axis works compared to others I can recommend this link: https://people.cs.uct.ac.za/~jgain/publications/rotationcontrollers.pdf

xeTaiz:
So I changed the two axis valuator implementation again, because it was pretty bad before imho. Especially the part that restricts vertical movements to not flip the camera when it lies on wolrd up.

I also added a very simple first person cam and a switcher. Further I started rebinding a lot of the things in order to make it somewhat consistent. I have settled with the following:
Left mouse moves camera as before
mouse wheel for zooming (alt: + and - buttons)
WASD/RF for camera movement (R,F for up and down)
shift + WASD for the panning
the buttons for rotation were changed from AWSD -> HJKL (maybe the arrow buttons are better here? to actually match the "up, down, .." labels).

@ResearchDaniel  I'm not entirely sure what the problem with "Follow Object During Rotation" (FODR) is in the video. The FODR flag should only be used by the ArcBall method and should not affect the two axis valuator. This makes sure that the cursor is on the same spot on the object after moving right? I don't think that is something that we want to use in the two axis valuator, as it would necessarily result in the same method as the ArcBall + FODR? 

Is there any reason why the FODR option is in the system settings instead of the trackball settings? Is it used for anything else? 

Edit: Note that I have not tested any of those on a touch screen yet.

ResearchDaniel:
Note, I did not try the changes yet as I'm traveling but thought that I would comment anyway.
> So I changed the two axis valuator implementation again, because it was pretty bad before imho. Especially the part that restricts vertical movements to not flip the camera when it lies on wolrd up.

Does it fix the issues in the video? For example the transition between the arcball-area and the outer areas in the screen.

> I also added a very simple first person cam and a switcher.

That sounds great! It is easier if one PR deals with one feature. Perhaps we can make a separate pull request for this change?

> @ResearchDaniel I'm not entirely sure what the problem with "Follow Object During Rotation" (FODR) is in the video.

There is no problem with FODR :) I wanted to show the issues with Two-axis and how it behaved compared to our current setup.

> The FODR flag should only be used by the ArcBall method and should not affect the two axis valuator.

I think that it could be used in both cases. The two-axis interaction is only different when you interact with the outer parts of the screen?

> Is there any reason why the FODR option is in the system settings instead of the trackball settings? Is it used for anything else?

I guess no good reason, would make sense to have it in the trackball settings...

petersteneteg:
Think it makes sense to have it all in one PR! Since most of it is in the trackball anyway. 
Just a not about the setting, having it in the settings changes all trackballs, having a option in a trackball would only affect that trackball...

xeTaiz:
> Just a not about the setting, having it in the settings changes all trackballs, having a option in a trackball would only affect that trackball...

We could just link the property and add it to the automatic linking thingy to get around this, right?

xeTaiz:
Also, I have not implemented any roll rotations in the Two Axis Valuator (TAV). I think most TAV implementations map interactions outside the trackball circle to roll. I find that rather confusing to be honest. Do you have any suggestions where to put it?

I was thinking of a behavior similar to blender, where you can press a button to start rotations around one axis only and confirm with a mouse click.

xeTaiz:
I removed the FODR from system settings and added it as another trackball method call "Object follows Finger". Also finally checked it on the touch screen, works perfectly fine for me.

petersteneteg (changes requested):
I Think this looks good. My only comment it that it needs to run through clang format. We don't allow lines over 100 chars, and we never have more then one declaration per line. If special formatting is needed one has to mark that with a //clang-format on/off section since I regularly auto format all our code.

xeTaiz:
Okay I clang-formatted the files and changed the default trackball to the old implementation.
Also renamed the Object follows finger to Object follows Cursor..

Note that there are still some confusing properties that only work for some trackball methods. I will make sure to hide them appropriately once the other PR (#458) that enables a nicer api for that is merged. I will make this part of a larger "Refactor of prominent processors" PR or so.

petersteneteg (approved):

src/core/interaction/trackball.cpp
```diff
@@ -380,13 +473,105 @@ void Trackball::rotate(Event* event) {
...
+mat4 Trackball::roll(const float radians) const {
+    return glm::translate(getLookFrom())  // to origin
+           * glm::rotate(radians, glm::normalize(getLookTo() - getLookFrom())) *
+           glm::translate(-getLookFrom());  // translate back
+}
+
+/* \brief Moves camera along -cam_right */
+void Trackball::moveLeft(Event* event) {
+    const vec3 right = getLookRight();
+    setLookFrom(getLookFrom() - movementSpeed_.get() * right);
```
ResearchDaniel: use setLook to avoid multiple property invalidations

src/core/interaction/trackball.cpp
```diff
@@ -380,13 +473,105 @@ void Trackball::rotate(Event* event) {
...
+    setLookTo(getLookTo() - movementSpeed_.get() * right);
+}
+
+/* \brief Moves camera along cam_right */
+void Trackball::moveRight(Event* event) {
+    const vec3 right = getLookRight();
+    setLookFrom(getLookFrom() + movementSpeed_.get() * right);
```
ResearchDaniel: use setLook to avoid multiple property invalidations

src/core/interaction/trackball.cpp
```diff
@@ -380,13 +473,105 @@ void Trackball::rotate(Event* event) {
...
+    setLookTo(getLookTo() + movementSpeed_.get() * right);
+}
+
+/* \brief Moves camera along cam_up */
+void Trackball::moveUp(Event* event) {
+    const vec3 up = getLookUp();
+    setLookFrom(getLookFrom() + movementSpeed_.get() * up);
```
ResearchDaniel: use setLook to avoid multiple property invalidations

src/core/interaction/trackball.cpp
```diff
@@ -380,13 +473,105 @@ void Trackball::rotate(Event* event) {
...
+    setLookTo(getLookTo() + movementSpeed_.get() * up);
+}
+
+/* \brief Moves camera along -cam_up */
+void Trackball::moveDown(Event* event) {
+    const vec3 up = getLookUp();
+    setLookFrom(getLookFrom() - movementSpeed_.get() * up);
```
ResearchDaniel: use setLook to avoid multiple property invalidations

src/core/interaction/trackball.cpp
```diff
@@ -380,13 +473,105 @@ void Trackball::rotate(Event* event) {
...
+    setLookTo(getLookTo() - movementSpeed_.get() * up);
+}
+
+/* \brief Moves camera along view_dir */
+void Trackball::moveForward(Event* event) {
+    const vec3 viewDir = glm::normalize(getLookTo() - getLookFrom());
+    setLookFrom(getLookFrom() + movementSpeed_.get() * viewDir);
```
ResearchDaniel: use setLook to avoid multiple property invalidations

src/core/interaction/trackball.cpp
```diff
@@ -380,13 +473,105 @@ void Trackball::rotate(Event* event) {
...
+    setLookTo(getLookTo() + movementSpeed_.get() * viewDir);
+}
+
+/* \brief Moves camera along -view_dir */
+void Trackball::moveBackward(Event* event) {
+    const vec3 viewDir = glm::normalize(getLookTo() - getLookFrom());
+    setLookFrom(getLookFrom() - movementSpeed_.get() * viewDir);
```
ResearchDaniel: use setLook to avoid multiple property invalidations

src/core/interaction/trackball.cpp
```diff
@@ -833,6 +1022,22 @@ void Trackball::panDown(Event* event) {
...
+ * Uses the step zoom functions and triggers one
+ * step per mouse wheel tick. Horizontal scrolling is disregarded
+ */
+void Trackball::zoomWheel(Event* event) {
+    auto wheelEvent = static_cast<WheelEvent*>(event);
+    int steps = static_cast<int>(wheelEvent->delta().y);
+
+    for (size_t i = 0; i < abs(steps); ++i)
+        if (steps > 0)
+            zoomIn(event);
```
ResearchDaniel: This will cause one property invalidation per iteration. Better to refactor zoomIn/stepZoom to take number of steps=

ResearchDaniel (changes requested):
Found some slightly intricate issues with property invalidation that should not be too difficult to sort out. 
Since the trackball is used to control a CameraProperty we need to be careful not to cause multiple invalidations. An invalidation will occur when setting the properties of the CameraProperty (setLookFrom etc.)

ResearchDaniel (approved):

ResearchDaniel:
Right click (zoom) still does not work for me

ResearchDaniel (approved):

PR #426, opened 2019-02-20
BrushingAndLinking: Ensure that selection is common to all connected …

ResearchDaniel:
…processors. In other words: Selection should be the same in all connected plots. Before, each processor had its own selection set, which is not desirable for selection.

This behavior can be seen if connecting two parallel coordinate plots to the same brushing and linking.

### DIFF
--- a/modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
+++ b/modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
@@ -48,26 +48,26 @@ public:
                               InvalidationLevel validationLevel = InvalidationLevel::InvalidOutput);
     virtual ~BrushingAndLinkingManager();
     /* 
-     * Return number of selected rows.
+     * Return the number of selected items/rows.
      */
     size_t getNumberOfSelected() const;
     /* 
-     * Return number of filtered rows.
+     * Return the number of filtered items/rows, i.e. the number of items/rows that should not be displayed.
      */
     size_t getNumberOfFiltered() const;
 
     void remove(const BrushingAndLinkingInport* src);
 
-    bool isFiltered(size_t row) const;
-    bool isSelected(size_t row) const;
+    bool isFiltered(size_t idx) const;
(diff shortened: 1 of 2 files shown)


Alboin (approved):
Hurray!

modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
```diff
@@ -40,46 +40,49 @@ class BrushingAndLinkingInport;
...
+ * \brief Manages row filtering, row selection and column selection from multiple sources.
  */
 class IVW_MODULE_BRUSHINGANDLINKING_API BrushingAndLinkingManager {
 public:
     BrushingAndLinkingManager(Processor* p,
                               InvalidationLevel validationLevel = InvalidationLevel::InvalidOutput);
     virtual ~BrushingAndLinkingManager();
-
+    /* 
+     * Return number of selected rows.
```
petersteneteg: This has nothing to do with rows, this is not only used for data frames. Brushing and linking should know nothing of columns and rows

ResearchDaniel: In some sense yes, but we now have the possibility to select columns as well. I can revert these semantic changes

ResearchDaniel: How about:
Return number of selected items/rows

ResearchDaniel: fixed

petersteneteg: Can keep the columns for now, but we will have to generalize that at some point are remove the hardkoded names...

ResearchDaniel: Ok, then I'll merge these changes. Scatterplots need to take selection into account as well, future PR

PR #429, opened 2019-02-22
Plotting: Changed the Option to DataFrameColumn conversion so that if…

Alboin:
… an index outside the range is selected, the nearest index is selected.

* Windows 10
* Visual Studio 15.7.6
* QT Community 4.6.1
* CMake 3.11.2
* Python 3.6 (32-bit)

### DIFF
--- a/modules/plotting/include/modules/plotting/properties/optionconverter.h
+++ b/modules/plotting/include/modules/plotting/properties/optionconverter.h
@@ -44,10 +44,7 @@ protected:
 protected:
     virtual void convertimpl(const plot::DataFrameColumnProperty *src,
                              OptionProperty *dst) const override {
-        if (src->getSelectedIndex() > dst->getValues().size())
-            dst->setSelectedIndex(dst->getValues().size() - 1);
-        else
-            dst->setSelectedIndex(src->getSelectedIndex());
+        dst->setSelectedIndex(glm::clamp<size_t>(0, dst->size() - 1, src->getSelectedIndex()));
     }
 };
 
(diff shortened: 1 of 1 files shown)


modules/plotting/include/modules/plotting/properties/optionconverter.h
```diff
@@ -44,7 +44,10 @@ class DataFrameColumnToOptionConverter
 protected:
     virtual void convertimpl(const plot::DataFrameColumnProperty *src,
                              OptionProperty *dst) const override {
-        dst->setSelectedIndex(src->getSelectedIndex());
+        if (src->getSelectedIndex() > dst->getValues().size())
```
ResearchDaniel: `dst->setSelectedIndex(glm::clamp(0, dst->size()-1, src->getSelectedIndex())`

ResearchDaniel (changes requested):

modules/plotting/include/modules/plotting/properties/optionconverter.h
Alboin: Nice! But I get an error when trying it:
image
Both for giving glm::clamp the template argument dst->size() and also src->getSelectedIndex()

ResearchDaniel: That was just from the top of my head :) I guess clamp is not needed since it is size_t. How about something like
`dst->setSelectedIndex(std::min(src->getSelectedIndex(), dst->size() - 1));`

petersteneteg: Try
dst->setSelectedIndex(glm::clamp<size_t>(0, dst->size()-1, src->getSelectedIndex())

Alboin:
Ping @ResearchDaniel

ResearchDaniel (approved):

PR #432, opened 2019-02-25
Feature/qt fixes

martinfalk:


modules/qtwidgets/include/modules/qtwidgets/inviwoqtutils.h
```diff
@@ -176,12 +176,19 @@ IVW_MODULE_QTWIDGETS_API QMenu* getMenu(std::string menuName, bool createIfNotFo
...
+ * @param format   image file format
+ * @param quality  image quality [0,100] (0 - small compressed, 100 - large uncompressed),
+ *                 -1 uses default settings
+ * @return base64 string of the corresponding image
+ *
+ * \see QImage::save()
  */
-IVW_MODULE_QTWIDGETS_API std::string toBase64(const QImage& image);
+IVW_MODULE_QTWIDGETS_API std::string toBase64(const QImage& image,
+                                              const std::string& format = "PNG", int quality = -1);
```
ResearchDaniel, line 191: Should format be jpeg by default?

ResearchDaniel (approved):

modules/qtwidgets/include/modules/qtwidgets/inviwoqtutils.h
martinfalk: Don't think so. We use the base64 encoding also for the port inspectors. There we should definitely use some lossless format :-)

martinfalk:
added a "Windows" menu which lists all processor widgets.
image

petersteneteg (approved):

PR #436, opened 2019-02-27
High dpi support

petersteneteg:
img_0301


r-englund:
Looks good on Windows 10 Full HD Monitor, though: 

This thing is really itchy on the eyes :) 
bild

The python editor has a white box (QLineEdit?) in the log-window when opened 
bild

Smart link drop down in linkeditor looks strange (this might not be new): 
bild

martinfalk:
@r-englund the smart link button in the LinkEditor was never fixed so far ;-)

petersteneteg:
Think the white line is a selected line or something like that

martinfalk:
Is it just me or is the font used in title bars of InviwoDockWidget a bit too heavy, i.e. too bold?

`font-weight: 500;`
image

vs. `font-weight: 400;` (regular font)
image

martinfalk:
top `font-weight: 500;`
bottom `font-weight: 600;`

image

martinfalk:
DockWidget icons on a non-high-dpi system look a bit flimsy.
image

modules/qtwidgets/src/inviwoqtutils.cpp
```diff
@@ -589,6 +589,19 @@ QString windowTitleHelper(const QString& title, const QWidget* widget) {
     return cap;
 }
 
+int refSpacePx(const QWidget* w) { return emToPx(w, refSpaceEm()); }
+
+QSize emToPx(const QWidget* w, QSizeF ems) {
+    return QSize{emToPx(w, ems.width()), emToPx(w, ems.height())};
+}
+
+int emToPx(const QWidget* w, double em) { return emToPx(w->fontMetrics(), em); }
```
martinfalk: need to ensure that the widget is polished before accessing the font metrics.
`w->ensurePolished();` 

see https://doc.qt.io/qt-5/qwidget.html#ensurePolished

modules/qtwidgets/src/properties/propertywidgetqt.cpp
```diff
@@ -64,6 +64,13 @@ const int PropertyWidgetQt::minimumWidth = 250;
 const int PropertyWidgetQt::spacing = 7;
```
martinfalk, line 64: what is the reason for a `minimumWidth` of 250? Can't we reduce this to maybe 220 or even 200?

martinfalk: (defined in line 63)

resources/svgicons/dock-sticky.svg
```diff
@@ -0,0 +1,74 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
martinfalk, line 1: Too thin of a border in unscaled version.

image

resources/svgicons/dock-docked.svg
```diff
@@ -0,0 +1,78 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
martinfalk, line 1: Border too thin
image

martinfalk (changes requested):
Nice work. And so shiny!

r-englund:
> 
> 
> Is it just me or is the font used in title bars of InviwoDockWidget a bit too heavy, i.e. too bold?
> 

Yes, I thought so to

xeTaiz:
Fyi, I just built this on Linux to make sure it works there as well. Looks pretty good on 1440p!

modules/qtwidgets/src/properties/propertywidgetqt.cpp
petersteneteg: Think a lot of our widgets will look very cramped then, but we can test...

petersteneteg:
> > Is it just me or is the font used in title bars of InviwoDockWidget a bit too heavy, i.e. too bold?
> 
> Yes, I thought so to

Sure, think I like the regular one better

modules/qtwidgets/src/inviwoqtutils.cpp
petersteneteg: Was considering that at some point but forgot to add it since it sems to work anyway...

petersteneteg:
> Fyi, I just built this on Linux to make sure it works there as well. Looks pretty good on 1440p!

Great 👍

martinfalk:
Updated dockwidget icons
image

martinfalk (approved):

modules/qtwidgets/src/properties/propertywidgetqt.cpp
martinfalk: I've put in 200. There is no difference since you can squeeze the property list only down to 330px. (Win 10, Qt 5.12, no dpi scaling).

PR #442, opened 2019-02-28
Feature/image importexport

martinfalk:
Various fixes regarding loading / saving images
* png export: >16bit integer formats converted to uint16, signed int formats converted to unsigned.
* fixed tiff header reader
* file property uses correct file extension


petersteneteg (approved):
Good, nice bugfixes

PR #445, opened 2019-03-05
Feature/entryexit

martinfalk:
Added a new entry exit point processor for arbitrary meshes (meshes are assumed to be closed). This allows for arbitrary bounding geometry besides a simple box. The vertex positions are mapped to texture coordinates of a particular volume by using the World To Data transformation of the volume.
In case of non-convex meshes, the entry points will correspond to the front-most surface and exit points to the back-most one.

**Spheres**
image

**or why not a cow** (images on the left correspond to entry and exit points)


martinfalk:
The network for the spherical cut-out.
image

modules/basegl/include/modules/basegl/processors/geometryentryexitpoints.h
```diff
@@ -0,0 +1,91 @@
...
+ *
+ * ### Outports
+ *   * __entry__ The first hit point in texture coordinates [0,1]
+ *   * __exit__  The last hit point in texture coordinates [0,1]
+ *
+ * ### Properties
+ *   * __Camera__   Scene camera
+ */
+
+/**
```
petersteneteg: Remove

modules/basegl/include/modules/basegl/processors/geometryentryexitpoints.h
```diff
@@ -0,0 +1,91 @@
...
+ * \brief VERY_BRIEFLY_DESCRIBE_THE_PROCESSOR
```
ResearchDaniel: Remove

modules/opengl/include/modules/opengl/openglutils.h
```diff
@@ -196,6 +196,22 @@ struct IVW_MODULE_OPENGL_API ClearColor {
     vec4 oldColor_;
 };
 
+struct IVW_MODULE_OPENGL_API ClearDepth {
```
ResearchDaniel, line 258: Document what this one does :)

modules/basegl/include/modules/basegl/processors/entryexitpointsprocessor.h
```diff
@@ -49,16 +47,17 @@ namespace inviwo {
 /** \docpage{org.inviwo.EntryExitPoints, Entry exit points}
  * ![](org.inviwo.EntryExitPoints.png?classIdentifier=org.inviwo.EntryExitPoints)
  * Computes the entry and exit points of a triangle mesh from the camera position in texture space.
- * The output color will be zero if no intersection is found, otherwise .
+ * The output color will be zero if no intersection is found, otherwise.
+ *
  * ### Inports
- *   * __MeshInport__ The mesh to intersect.
+ *   * __geometry__     The input geometry used for determining entry and exit points
```
ResearchDaniel: __Mesh__ The input mesh ...

modules/basegl/include/modules/basegl/processors/geometryentryexitpoints.h
```diff
@@ -0,0 +1,91 @@
...
+ * DESCRIBE_THE_PROCESSOR_FROM_A_DEVELOPER_PERSPECTIVE
+ */
+class IVW_MODULE_BASEGL_API GeometryEntryExitPoints : public Processor {
```
ResearchDaniel, line 64: Should this not be MeshEntryExitPoints?

ResearchDaniel (approved):
Nice! 
In some sense this makes the "old" entry-exit computation obsolete?
In that case we could update the other workspaces to use the new way in a future PR.

petersteneteg (approved):

modules/basegl/include/modules/basegl/processors/geometryentryexitpoints.h
martinfalk: that's the big question. We use geometry and mesh somewhat interchangeably for some reason...

modules/basegl/include/modules/basegl/processors/entryexitpointsprocessor.h
martinfalk: in other processors it is `geometry`

modules/basegl/include/modules/basegl/processors/geometryentryexitpoints.h
ResearchDaniel: I think that we started with Geometry as a general term for triangular meshes, point clouds, procedural geometry, etc. 
Then, we more or less only supported Meshes so we renamed the geometry to Mesh. Hence, we do not have a geometry class anymore.
Unfortunately, the renaming did not consider all documentation...

PR #450, opened 2019-03-07
WebBrowser: Added dataframe inport that is sent as json to javascript

Alboin:


modules/webbrowser/src/processors/webbrowserprocessor.cpp
```diff
@@ -239,6 +245,29 @@ void WebBrowserProcessor::OnLoadingStateChange(CefRefPtr<CefBrowser> browser, bo
 void WebBrowserProcessor::process() {
     // Vertical flip of CEF output image
     cefToInviwoImageConverter_.convert(renderHandler_->getTexture2D(), outport_, &background_);
+
+    // If any dataframes are connected, send their input in json to individual javascript functions
+    if (dataFrames_.isConnected() && dataFrames_.hasData()) {
+        auto frame = browser_->GetMainFrame();
+        size_t dataframeNr = 0;
+        for (auto dataFrame : dataFrames_.getVectorData()) {
+            using json = nlohmann::json;
```
ResearchDaniel, line 254: Converting a DataFrame to json format should be moved to a utility function. Idealy in the Plotting module.

modules/webbrowser/src/processors/webbrowserprocessor.cpp
```diff
@@ -239,6 +245,29 @@ void WebBrowserProcessor::OnLoadingStateChange(CefRefPtr<CefBrowser> browser, bo
 void WebBrowserProcessor::process() {
     // Vertical flip of CEF output image
     cefToInviwoImageConverter_.convert(renderHandler_->getTexture2D(), outport_, &background_);
+
+    // If any dataframes are connected, send their input in json to individual javascript functions
+    if (dataFrames_.isConnected() && dataFrames_.hasData()) {
```
ResearchDaniel: This will send data no matter if it changed or not. Could be quite slow

ResearchDaniel (changes requested):
I'm not sure if this is the right approach. This is a special case and probably needs better integration with brushing and linking. 
If we add a data frame as input, why not all other data types as well?

Probably better to separate this functionality into a separate processor

PR #453, opened 2019-03-14
BrushingAndLinking: Fixed a bug where the column selection would not …

Alboin:
…run if the local selection cache was empty, but the remote set with selected indices was not. This lead to that if an index was selected from another processor the isColumnSelected(index) would return true, but when a sendColumnSelectionEvent({ }) was called to deselect it nothing happened.

Perhaps this should be done for sendSelectionEvent() and sendFilterEvent() as well?

### DIFF
--- a/modules/brushingandlinking/src/ports/brushingandlinkingports.cpp
+++ b/modules/brushingandlinking/src/ports/brushingandlinkingports.cpp
@@ -50,10 +50,12 @@ void BrushingAndLinkingInport::sendFilterEvent(const std::unordered_set<size_t> 
 
 void BrushingAndLinkingInport::sendSelectionEvent(const std::unordered_set<size_t> &indices) {
     bool noRemoteSelections = false;
-    if(isConnected() && hasData())
+    if (isConnected() && hasData()) {
         noRemoteSelections = getData()->getSelectedIndices().empty();
-    if (selectionColumnCache_.empty() && indices.empty() && noRemoteSelections) return;
-    if (selectionCache_.size() == 0 && indices.size() == 0) return;
+    }
+    if (selectionCache_.empty() && indices.empty() && noRemoteSelections) {
+        return;
+    }
     selectionCache_ = indices;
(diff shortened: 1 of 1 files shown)


modules/brushingandlinking/src/ports/brushingandlinkingports.cpp
```diff
@@ -56,7 +56,8 @@ void BrushingAndLinkingInport::sendSelectionEvent(const std::unordered_set<size_
 }
 
 void BrushingAndLinkingInport::sendColumnSelectionEvent(const std::unordered_set<size_t> &indices) {
-    if (selectionColumnCache_.size() == 0 && indices.size() == 0) return;
+    auto remoteSelections = getData()->getSelectedColumns();
+    if (selectionColumnCache_.size() == 0 && indices.size() == 0 && remoteSelections.size() == 0) return;
```
r-englund: use '.empty()' instead of '.size() == 0'

r-englund (changes requested):
Since the manager is using an IndexList for the filter events we do not need this fix for the FilterEvents, but for the other events we would need this fix.

ResearchDaniel (approved):
Looks good to me. You probably want to incorporate @r-englund suggestions

ResearchDaniel:
Also, if it means that there are bugs in the other ones too, then fix them as well :)

r-englund (approved):

petersteneteg (changes requested):
This should be reformatted, the if needs a pair of braces. We don't allow ifs with a single hanging line in out style guide.

Why is selection Cache and column selection caches mixed?

Alboin:
> This should be reformatted, the if needs a pair of braces. We don't allow ifs with a single hanging line in out style guide.

Didn't know that! It's been updated with braces now :)
> 
> Why is selection Cache and column selection caches mixed?

Good question! Cleared up the mix-up.

petersteneteg (approved):

PR #458, opened 2019-03-18
Beautify Property api

xeTaiz:
Heyho!
I was making some changes to the `Property` api in oder to get some of the most written code a bit cleaner. I made the following changes

- `PropertyOwner::addProperty()` now returns a reference to the property, allowing to append further modifications (as e.g. whats following) 
- `PropertyOwner::addProperties()` was added, so you can add multiple properties in one line, to not produce a hundred lines wasting space for trivial stuff.
- `Property::visibilityDependsOn()` was added to give easy control over the visibility state of a property.

What you can do now is things like this (from `SimpleLightingProperty`):

`
addProperty(specularColor_).visibilityDependsOn(shadingMode_, [](const OptionPropertyInt& opt){
        return opt.get() == ShadingMode::Phong;
    });
`
So this will add the `specularColor_` to the processor and make its visibility dependent on the `shadingMode_` in a way defined by the lambda that's following.
This lambda must return a boolean (otherwise it won't compile) and the visibility is always set dependent on **one** property. Internally this registers an onChange callback on the property you pass it (in this case `shadingMode_`). This is not meant to control visibility using super complicated logic, but rather to give an easy way to hide properties when not needed, so that the property list is kept clean.

Once this is done and merged I will also go over some of the most prominent processors and adapt this API so that people see it exists and to clean up the old code a bit.

### DIFF
--- a/include/inviwo/core/properties/property.h
+++ b/include/inviwo/core/properties/property.h
@@ -301,6 +301,25 @@ public:
         });
     }
 
+    /* \brief sets readonly depending another property `prop`, according to `callback`
+     * @param prop is the property on which the readonly state depends
+     * @param callback is a function that outputs a readonly boolean value. The function gets `prop` as parameter
+     *
+     * Checks the expression in `callback` every time `prop` is changed and sets the
+     * readonly state accordingly. Note that this registers an onChange callback on `prop`, which
+     * might result in poor performance when `prop` is a very frequently changed property.
+     */
+    template<typename P, typename DecisionFunc>
+    const BaseCallBack* readonlyDependsOn(P& prop, DecisionFunc callback) {
+        typename std::result_of<DecisionFunc(P&)>::type b = true;
+        static_assert(std::is_same<decltype(b), bool>::value, "The readonly callback must return a boolean!");
+        static_assert(std::is_base_of<Property, P>::value, "P must be a Property!");
+        return prop.onChange([callback, &prop, this](){
+            bool readonly = callback(prop);
+            this->setReadOnly(readonly);
+        });
+    }
+
     virtual Document getDescription() const;
 
     template <typename T, typename U>
--- a/include/inviwo/core/properties/property.h
+++ b/include/inviwo/core/properties/property.h
@@ -299,6 +299,7 @@ public:
             bool visible = callback(prop);
             this->setVisible(visible);
         });
+        this->setVisible(callback(prop));
     }
 
     /* \brief sets readonly depending another property `prop`, according to `callback`
@@ -318,6 +319,7 @@ public:
             bool readonly = callback(prop);
             this->setReadOnly(readonly);
         });
+        this->setReadOnly(callback(prop));
     }
 
     virtual Document getDescription() const;
--- a/include/inviwo/core/properties/property.h
+++ b/include/inviwo/core/properties/property.h
@@ -44,6 +44,7 @@
 #include <inviwo/core/util/introspection.h>
 
 #include <functional>
+#include <type_traits>
 
 namespace inviwo {
 
@@ -268,7 +269,7 @@ public:
     [[deprecated("was declared deprecated. Use `onChange(std::function<void()>)` instead")]]
     const BaseCallBack* onChange(T* object, void (T::*method)());
     template <typename T>
-    [[deprecated("was declared deprecated. Use `removeOnChange(const BaseCallBack*)` instead")]] 
+    [[deprecated("was declared deprecated. Use `removeOnChange(const BaseCallBack*)` instead")]]
     void removeOnChange(T* object);
     // clang-format on
 
@@ -280,6 +281,25 @@ public:
 
     virtual void setVisible(bool val);
     virtual bool getVisible();
+
+    /* \brief sets visibility depending another property `prop`, according to `callback`
+     * @param prop is the property on which the visibility depends
+     * @param callback is a function that outputs a visibility boolean value. The function gets `prop` as parameter
+     *
+     * Checks the expression in `callback` every time `prop` is changed and sets the
+     * visibility accordingly. Note that this registers an onChange callback on `prop`, which
+     * might result in poor performance when `prop` is a very frequently changed property.
+     */
+    template<typename P, typename DecisionFunc>
+    const BaseCallBack* visibilityDependsOn(P& prop, DecisionFunc callback) {
+        typename std::result_of<DecisionFunc(P&)>::type b = true;
+        static_assert(std::is_same<decltype(b), bool>::value, "The visibility callback must return a boolean!");
+        static_assert(std::is_base_of<Property, P>::value, "P must be a Property!");
+        return prop.onChange([callback, &prop, this](){
+            bool visible = callback(prop);
+            this->setVisible(visible);
+        });
+    }
 
     virtual Document getDescription() const;
 
@@ -335,7 +355,7 @@ private:
 
 // clang-format off
 template <typename T>
-[[deprecated("was declared deprecated. Use `onChange(std::function<void()>)` instead")]] 
+[[deprecated("was declared deprecated. Use `onChange(std::function<void()>)` instead")]]
 const BaseCallBack* Property::onChange(T* o, void (T::*m)()) {
     return onChangeCallback_.addLambdaCallback([o, m]() {
         if (m) (*o.*m)();
@@ -343,7 +363,7 @@ const BaseCallBack* Property::onChange(T* o, void (T::*m)()) {
 }
 
 template <typename T>
-[[deprecated("was declared deprecated. Use `removeOnChange(const BaseCallBack*)` instead")]] 
+[[deprecated("was declared deprecated. Use `removeOnChange(const BaseCallBack*)` instead")]]
 void Property::removeOnChange(T* o) {
     onChangeCallback_.removeMemberFunction(o);
 }
--- a/include/inviwo/core/properties/propertyowner.h
+++ b/include/inviwo/core/properties/propertyowner.h
@@ -56,8 +56,8 @@ public:
     PropertyOwner& operator=(const PropertyOwner& that);
     virtual ~PropertyOwner() = default;
 
-    virtual Property& addProperty(Property* property, bool owner = true);
-    virtual Property& addProperty(Property& property);
+    virtual void addProperty(Property* property, bool owner = true);
+    virtual void addProperty(Property& property);
 
     template <typename... Ts>
     void addProperties(Ts&... properties);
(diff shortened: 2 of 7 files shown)


xeTaiz:
The same as for `visibilityDependsOn` is coming in for the `readonly` state in a moment

xeTaiz:
Actually I just found another issue with this, so don't merge yet

xeTaiz:
For example this is the `SimpleLightingProperty` contructor, hiding light colors automatically:
```
addProperties(shadingMode_, lightPosition_, referenceFrame_);
    addProperty(ambientColor_).visibilityDependsOn(shadingMode_, [](const OptionPropertyInt& opt){
        return opt.get() == ShadingMode::Ambient ||
               opt.get() == ShadingMode::BlinnPhong ||
               opt.get() == ShadingMode::Phong;
    });
    addProperty(diffuseColor_).visibilityDependsOn(shadingMode_, [](const OptionPropertyInt& opt){
        return opt.get() == ShadingMode::Diffuse ||
               opt.get() == ShadingMode::BlinnPhong ||
               opt.get() == ShadingMode::Phong;
    });
    auto specCallback = [](const OptionPropertyInt& opt){
        return opt.get() == ShadingMode::Specular ||
               opt.get() == ShadingMode::BlinnPhong ||
               opt.get() == ShadingMode::Phong;
    };
    addProperty(specularColor_).visibilityDependsOn(shadingMode_, specCallback);
    addProperty(specularExponent_).visibilityDependsOn(shadingMode_, specCallback);
    addProperties(applyLightAttenuation_, lightAttenuation_);
```

xeTaiz:
Since this is somehwat related:
Would you find it useful to change other `Property` methods as well so that they return `Property&`? 
For something like 

```
shadingMode_
    .addOption("none", "No Shading", ShadingMode::None)
    .addOption("ambient", "Ambient", ShadingMode::Ambient)
    .addOption("diffuse", "Diffuse", ShadingMode::Diffuse)
    .addOption("specular", "Specular", ShadingMode::Specular)
    .addOption("blinnphong", "Blinn Phong", ShadingMode::BlinnPhong)
    .addOption("phong", "Phong", ShadingMode::Phong)
    .setSelectedValue(ShadingMode::Phong)
    .setCurrentStateAsDefault();
```

or 

```
addProperty(ambientColor_)
    .setSemantics(PropertySemantics::Color)
    .visibilityDependsOn(shadingMode_, [](const OptionPropertyInt& opt){
        return opt.get() == ShadingMode::Ambient ||
               opt.get() == ShadingMode::BlinnPhong ||
               opt.get() == ShadingMode::Phong;
    });
```

include/inviwo/core/properties/propertyowner.h
```diff
@@ -56,8 +56,11 @@ class IVW_CORE_API PropertyOwner : public PropertyOwnerObservable,
     PropertyOwner& operator=(const PropertyOwner& that);
     virtual ~PropertyOwner() = default;
 
-    virtual void addProperty(Property* property, bool owner = true);
-    virtual void addProperty(Property& property);
+    virtual Property& addProperty(Property* property, bool owner = true);
+    virtual Property& addProperty(Property& property);
+
+    template <typename... Ts>
+    void addProperties(Ts&... properties);
```
r-englund, line 63: :+1: 
Would be nice with a similar method for adding ports

xeTaiz: The port also requires a name. I'm not sure if it's easily doable to make the argument list inerleaved or whatever...
Also I think you usually don't have so many ports that it's horribly verbose code if you don't have such function (compared to properties).
Any suggestions on how a nice interface should then look like for ports? 
`addPorts(imInport1, "DIffuse Tex", imInport2, "Whatever");` ??

r-englund: The name is for port groups which is only used for ImagePorts (as far as I now). Maybe one can only add to one port group at a time (or the default) like:
```c++
addPorts("GroupName",port1,port2,port3,...);
// or just
addPorts(port1,port2,port3,...); // Default group name used
```
Anyway, I think that would go into another PR

xeTaiz: That seems reasonable. I will collect some more feedback on my other proposals and then split the whole thing into PRs accordingly ;)

xeTaiz:
As I have explored the possible methods where returning `Property&` instead of void makes sense I figured it would fit in nicely for the `visibilityDependsOn()` function as well. Currently this returns the `BaseCallBack*` that the inner `onChange()` returns. I initially thought it might be a good idea to be able to capture this callback somewhere, like it is possible for the `onChange()`. However I would actually change this to `Property&` as well, since I think that these callbacks controlling visibility will probably not be needed anywhere else, right? Probably the benefit of being able to chain it like above is higher. Afterall that's a pretty boring callback that does not functionally do anything...

petersteneteg:
The basecallback pointer is only needed to remove the callback. Do one ever need to remove such callback?

include/inviwo/core/properties/propertyowner.h
```diff
@@ -167,6 +170,23 @@ std::vector<T*> PropertyOwner::getPropertiesByType(bool recursiveSearch /* = fal
...
+    inline void addPropertyHelper(PropertyOwner& owner) {
+    }
+    template <typename... Ts>
+    void addPropertyHelper(PropertyOwner& owner, Property& p, Ts&... props) {
+        owner.addProperty(p);
+        addPropertyHelper(owner, props...);
+    }
+
+}
+
```
petersteneteg, line 181: Needs formatting

petersteneteg (approved):

xeTaiz:
I have looked more into the chaining stuff. 
### There seem to arise a few issues:
Every custom property that is defined in anyones projects will probably have to make some changes whenever one of those (atm `void` returning) functions are overriden.
Further, properties deeper down in the inheritance hierarchy should probably return whatever type they have, instead of `Property&`. This can be realized nicely by overloading with the new type, however that also means that every property in the hierarchy will have to override all of the supported functions, otherwise the following might happen:
```
OptionProperty o;
o.addOption()  // returns OptionProperty
 .setVisible() // returns Property
 .addOption()  // Property does not have this member
```
I can of course do that for all the existing properties, however every property defined outside of core (or rather the master branch) would then have to implement this chaining functionality themselves (unless there are no new functions defined for that property, then it would not matter)

I don't think that I can judge whether that is a big issue or whether it makes sense to integrate this in inviwo, considering the breaking change and the additional overhead for new properties to support it. 
So I guess I need some opinions on that, after all I will not be among the people that would suffer from this change.

### Chaining API Pros and Cons:
**Pro:**
- Cleaner API, more readable
- Strongly encouraging developers to do the whole setup of a property in one place, code is easier semantically grouped (`*` see example on bottom)

**Cons:**
- Breaking change where ever such function is currently overriden
- Additional effort necessary for new properties to support chaining, otherwise the chaining api is inconsistent

#### A quick note on why **I** think its a good idea: 
Since I will probably create some getting started docs for new inviwo users and overall documentation, I want to try to clean up the current api as much as possible. I believe a better api makes it easier to get started and will make inviwo more attractive for new users. 

## Please let me know what you think :)


`*`: Here is what I mean with encourage to semantically group:
Screenshot from 2019-03-19 17-41-05
On the left the individual bits are not really grouped. Look at the left and figure out if there is some magic behavior on `inport_` or `overrideColor_`. Now try the same on the right. It's all in one place. I'm not saying you could not do this on the left, but I think with the right api that's much more likely to happen.

xeTaiz:
> The basecallback pointer is only needed to remove the callback. Do one ever need to remove such callback?

I doubt it. Any time you would need it you are probably using this function wrong

xeTaiz:
Also while making the above example, I think what's definitely a good idea is making a
`Property::visibilityDependsOn(OptionProperty, bool)` and 
`Property::visibilityDependsOn(BoolProperty, bool)`
for shortcuts without a lambda, because it's mostly just unnecessary I think

petersteneteg:
> Also while making the above example, I think what's definitely a good idea is making a
> `Property::visibilityDependsOn(OptionProperty, bool)` and
> `Property::visibilityDependsOn(BoolProperty, bool)`
> for shortcuts without a lambda, because it's mostly just unnecessary I think

That will introduce some strange dependency....

petersteneteg:
I think I like this mostly. Thought I might give a little background, since this is a little different to how apis often are designed in c++ and how many things look in inviwo. Maybe this it more common in python. In c++ we often try and keep the number of public member functions in a class to a minimum. (Like for std vector, is has no find for example.., an often cited exception is std string which has a huge interface, and is often criticized for that...) the advantage of having a minimal interface is that it gets much easier to make sure all class invariants are maintained across all the public functions. Extra convenience functionally can then be written as free functions in terms of that minimal set of exposed functions, and one can the be sure that all the invariants will be upheld since one is only using the public interface, and if that is correct it should not be able to brake any invariants. Think of how algorithms work with containers in c++. That is a very good api design.
For this reason most of the convenience functions in Inviwo lives as free functions. Often in the util namespace.

petersteneteg:
> Since I will probably create some getting started docs for new inviwo users and overall documentation,
Looking forward to this! :)

petersteneteg:
> This can be realized nicely by overloading with the new type

This is not really overloading, one can't overload returntypes :(... rather is is an example of covariant return types...

petersteneteg:
> shadingMode_
>     .addOption("none", "No Shading", ShadingMode::None)
>     .addOption("ambient", "Ambient", ShadingMode::Ambient)
>     .addOption("diffuse", "Diffuse", ShadingMode::Diffuse)
>     .addOption("specular", "Specular", ShadingMode::Specular)
>     .addOption("blinnphong", "Blinn Phong", ShadingMode::BlinnPhong)
>     .addOption("phong", "Phong", ShadingMode::Phong)
>     .setSelectedValue(ShadingMode::Phong)
>     .setCurrentStateAsDefault();

All this should rather be added in the constructor.

Generally as much of the configuration should be done in the constructor as possible.

petersteneteg:
Thing to think about, PropertyOwner::addProperty returns a ref to its argument, where as Property::setVisible returns a reference to it self. 

propertyowner::addProperty could also return a ref to it self allowing composite.addProperty(..).addProperty(...)...

So both return types are kind of valid, what happens in the example above might be surprising to users...

include/inviwo/core/properties/property.h
```diff
@@ -281,6 +282,44 @@ class IVW_CORE_API Property : public PropertyObservable, public Serializable, pu
...
+    /* \brief sets visibility depending another property `prop`, according to `callback`
+     * @param prop is the property on which the visibility depends
+     * @param callback is a function that outputs a visibility boolean value. The function gets `prop` as parameter
+     *
+     * Checks the expression in `callback` every time `prop` is changed and sets the
+     * visibility accordingly. Note that this registers an onChange callback on `prop`, which
+     * might result in poor performance when `prop` is a very frequently changed property.
+     */
+    template<typename P, typename DecisionFunc>
+    const BaseCallBack* visibilityDependsOn(P& prop, DecisionFunc callback) {
```
petersteneteg, line 294: Should this also immediately call lambda to set the visibility so the states are in sync?

ResearchDaniel (commented):
Nice!

xeTaiz:
As discussed in the meeting, the chaining API will be realized in the scope of #461.
This PR will therefore ony include:
- `void PropertyOwner::addProperties()`
- `BaseCallBack* Property::visibilityDependsOn()`
- `BaseCallBack* Property::readonlyDependsOn()`
**without** any chaining.

@r-englund I also looked more into the `addPorts` idea. Since `addPort` is different for `Inport` and `Outport`, I figured this is probably not very useful anymore, since you would have to further split it into `addInports()` and `addOutports()`. Or maybe some template magic might work around this, however there are usually not so many ports that you would want in a processor right? And if you do, probably many of them are optional, which would rather encourage to use chaining over something like `addPorts()` anyways.

petersteneteg (approved):

PR #469, opened 2019-03-25
CMake: truncate module tooltips

martinfalk:
* tooltips shown for modules are truncated since module descriptions provided in some of the readme.md files can be somewhat extensive

### DIFF
--- a/cmake/globalmacros.cmake
+++ b/cmake/globalmacros.cmake
@@ -203,6 +203,12 @@ function(ivw_private_setup_module_data)
     # In that case set to INVIWO<NAME>MODULE_description
     if(EXISTS "${${mod}_path}/readme.md")
         file(READ "${${mod}_path}/readme.md" description)
+        # truncate description since some readme files are quite substantial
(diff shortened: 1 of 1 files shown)


petersteneteg (approved):

cmake/globalmacros.cmake
```diff
@@ -203,6 +203,12 @@ function(ivw_private_setup_module_data)
     # In that case set to INVIWO<NAME>MODULE_description
     if(EXISTS "${${mod}_path}/readme.md")
         file(READ "${${mod}_path}/readme.md" description)
+        # truncate description since some readme files are quite substantial
+        string(LENGTH "${description}" desc_len)
+        if(desc_len GREATER 250)
+            string(SUBSTRING "${description}" 0 250 description)
+            string(JOIN "" description "${description}" "...")
+        endif()
```
r-englund, line 211: If the description is 251 characters, we will remove one and add 3 dots resulting in 253 characters ;)

martinfalk: as if that matters :)

PR #474, opened 2019-03-29
Core: linking between TF and IsoValue properties

martinfalk:
allows direct linking between `IsoTFProperty` and `TransferfunctionProperty` as well as `IsoTFProperty` and `IsoValueProperty`. Only the respective parts, i.e. either the TF or the isovalues, are linked. Everything else is left intact.

### DIFF
--- a/include/inviwo/core/properties/isotfproperty.h
+++ b/include/inviwo/core/properties/isotfproperty.h
@@ -76,7 +76,15 @@ public:
     virtual std::string getClassIdentifierForWidget() const override;
 
     virtual void set(const Property* property) override;
+    /**
+     * \brief sets only the isovalue property to \p p. The transfer function property remains
+     * unchanged.
+     */
     void set(const IsoValueProperty& p);
+    /**
+     * \brief sets only the transfer function property to \p p. The isovalue property remains
+     * unchanged.
+     */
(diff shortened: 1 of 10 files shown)


include/inviwo/core/properties/isotfproperty.h
```diff
@@ -75,6 +75,10 @@ class IVW_CORE_API IsoTFProperty : public CompositeProperty,
 
     virtual std::string getClassIdentifierForWidget() const override;
 
+    virtual void set(const Property* property) override;
+    void set(const IsoValueProperty& p);
```
ResearchDaniel, line 83: Would be good with documentation on what happens to non-iso values. Are they removed?

include/inviwo/core/properties/isotfproperty.h
```diff
@@ -75,6 +75,10 @@ class IVW_CORE_API IsoTFProperty : public CompositeProperty,
...
+    void set(const TransferFunctionProperty& p);
```
ResearchDaniel, line 88: Would be good with documentation on what happens to iso values. Are they removed?

ResearchDaniel (approved):
Would be good with documentation on what happens to the iso/tf when setting it with another one.

PR #475, opened 2019-03-29
BaseGL: custom data range in VolumeCombiner

martinfalk:
* use data/value range from input volumes or custom ranges

The `VolumeCombiner` allows to arbitrarily combine volumes. But so far the output volume reused the data/value range of the first input. Added an option property to use any of the input volume ranges, the combines min/max, or use an entirely custom range.

### DIFF
--- a/modules/basegl/include/modules/basegl/processors/volumeprocessing/volumecombiner.h
+++ b/modules/basegl/include/modules/basegl/processors/volumeprocessing/volumecombiner.h
@@ -39,6 +39,7 @@
 #include <inviwo/core/properties/buttonproperty.h>
 #include <inviwo/core/properties/boolcompositeproperty.h>
 #include <inviwo/core/properties/optionproperty.h>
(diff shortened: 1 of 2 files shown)


r-englund (approved):
Looks good, a convenient new feature.

ResearchDaniel:
Would it not make sense to only be able to change the value range?
The data range should be set according to data type?

martinfalk:
@ResearchDaniel: yes and no. For integer-based types data range should be depending on the data type. But for float/double data, data range is used for normalization with respect to the TF...

ResearchDaniel:
That means that it is still dependent on the data type?
Just that the min/max would need to be computed
Actually, that might be what should be done in all cases.

petersteneteg:
@ResearchDaniel Generally the min max can not be computed from the data. At least not until we have revamped how we handle TFs.

petersteneteg:
Why does this not apply the given function to the given input data range?

PR #476, opened 2019-03-29
Feature/warnfixes

martinfalk:
also fixed the network lock icon (icon was open when network was locked)


src/qt/editor/inviwomainwindow.cpp
```diff
@@ -711,8 +711,8 @@ void InviwoMainWindow::addActions() {
     // Network
     {
         QIcon enableDisableIcon;
-        enableDisableIcon.addFile(":/svgicons/locked.svg", QSize(), QIcon::Normal, QIcon::Off);
-        enableDisableIcon.addFile(":/svgicons/unlocked.svg", QSize(), QIcon::Normal, QIcon::On);
+        enableDisableIcon.addFile(":/svgicons/unlocked.svg", QSize(), QIcon::Normal, QIcon::Off);
+        enableDisableIcon.addFile(":/svgicons/locked.svg", QSize(), QIcon::Normal, QIcon::On);
```
r-englund, line 715: Nice.  This confused me the other day, is my network locked or not, with this commit I might not be confused anymore :) .  

Now they show current state rather then the state it would enter by clicking, right?

r-englund (approved):

PR #477, opened 2019-04-01
Add world up dropdown for fixUp in trackball 

xeTaiz:
The trackball changes in #422 bringt the option to fix horizontal rotations around the world up axis. However I forgot a way to specify the up axis (x, y or z). This adds a dropdown to specify the axis used, defaulting to y.

### DIFF
--- a/src/core/interaction/trackball.cpp
+++ b/src/core/interaction/trackball.cpp
@@ -52,14 +52,24 @@ Trackball::Trackball(TrackballObject* object)
     , isMouseBeingPressedAndHold_(false)
     , lastNDC_(vec3(0.0))
     , gestureStartNDCDepth_(-1)
-    , trackballMethod_("trackballMethod", "Trackball Method")
+    , trackballMethod_("trackballMethod", "Trackball Method",
+                       {{"tb_vt", "Virtual Trackball", 0},
+                        {"tb_tav", "Two Axis Valuator Trackball", 1},
+                        {"tb_fps", "First Person Camera", 2},
+                        {"tb_fodr", "Object follows Cursor", 3}},
+                       0)
     , sensitivity_("sensitivity", "Sensitivity", 3.0f, 0.0f, 10.0f, 0.25f)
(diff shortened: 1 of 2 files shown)


src/core/interaction/trackball.cpp
```diff
@@ -112,6 +114,9 @@ Trackball::Trackball(TrackballObject* object)
     addProperty(sensitivity_);
     addProperty(movementSpeed_);
     addProperty(fixUp_);
+    addProperty(worldUp_);
+    worldUp_.setSelectedIndex(1);
```
r-englund: default selected index can be set as the 4th argument in the constructor 
https://github.com/inviwo/inviwo/blob/23770f502f7a2596eb34f43286f21216fdd57750/include/inviwo/core/properties/optionproperty.h#L121-L125

src/core/interaction/trackball.cpp
```diff
@@ -322,6 +329,16 @@ const vec3 Trackball::getLookRight() const {
     return glm::normalize(glm::cross(getLookTo() - getLookFrom(), getLookUp()));
 }
 
+/* \brief Returns the World Up Vector according to `worldUp_` property. */
+const vec3 Trackball::getWorldUp() const {
+    switch(worldUp_){
+        case 0: return vec3(1,0,0);
+        case 1: return vec3(0,1,0);
+        case 2: return vec3(0,2,0);
```
r-englund: shouldn't case 2 be vec3(0,0,1) ?

r-englund (changes requested):
Do we also want negative x/y/z axis as possible up vectors? And I can also see usecases where a custom up axis is enforced. (For example: My lines in the ventricle has a perceived up-axis that is not aligned with any of the world-axes)

xeTaiz:
Oops, sorry I missed your reply :D
I added the changes you wanted and did some property visibility cleanup.

r-englund (approved):

PR #478, opened 2019-04-01
Feature/various (vectorvis, plotting, core)

r-englund:
Core: 
* Added a onTaskDone callback to the foreachParallelAsync util. This callback will be called by the worker thread once the task is completed. 

VectorVis:
 * Added Curl and Divergence volumes in addition to the velocity volume for Lorentz System processor
 * Minor fixes

Plotting: 
* Added 2 columns (one int and one float) to the synthetic dataset processor that has a constant value. (useful for testing axis where the datarange min == max)


modules/vectorfieldvisualization/src/processors/datageneration/seedpointsfrommask.cpp
```diff
@@ -93,24 +95,32 @@ void SeedPointsFromMask::process() {
             util::IndexMapper3D index(dim);
             vec3 invDim = vec3(1.0f) / vec3(dim);
 
+            std::function<vec3(vec3)> noTransform = [&](vec3 p) -> vec3 { return p; };
+            std::function<vec3(vec3)> withTransform = [&](vec3 p) -> vec3 {
+                auto m = v->getCoordinateTransformer().getDataToWorldMatrix();
```
martinfalk: wouldn't it be more efficient to set `m` already in the capture list?

martinfalk (approved):

r-englund:
@martinfalk Added a fix for the space infront of the headers in the csv-reader

include/inviwo/core/util/foreach.h
```diff
@@ -78,17 +84,21 @@ auto foreach_helper_pool(std::false_type, IT a, IT b, Callback callback,
...
-                                                    size_t jobs = 0) {
+                                                    size_t jobs = 0,
+                                                    OnDoneCallback onTaskDone = []() {}) {
     auto settings = InviwoApplication::getPtr()->getSettingsByType<SystemSettings>();
     auto poolSize = settings->poolSize_.get();
     using IncludeIndexType = typename std::conditional<util::is_callable_with<T, size_t>(callback),
                                                        std::true_type, std::false_type>::type;
     if (poolSize == 0) {
         detail::foreach_helper(IncludeIndexType(), iterable.begin(), iterable.end(), callback);
+        onTaskDone();
```
ResearchDaniel, line 103: why not foreach_helper_pool here?

modules/vectorfieldvisualization/src/processors/datageneration/seedpointsfrommask.cpp
```diff
@@ -93,24 +95,31 @@ void SeedPointsFromMask::process() {
             util::IndexMapper3D index(dim);
             vec3 invDim = vec3(1.0f) / vec3(dim);
 
+            std::function<vec3(vec3)> noTransform = [](vec3 p) -> vec3 { return p; };
+            std::function<vec3(vec3)> withTransform = [m = v->getCoordinateTransformer().getDataToWorldMatrix()](vec3 p) -> vec3 {
+                vec4 WP = m * vec4(p, 1.0f);
+                return vec3(WP) / WP.w;
```
ResearchDaniel: division by w?

modules/vectorfieldvisualizationgl/src/processors/datageneration/lorenzsystem.cpp
```diff
@@ -70,59 +74,68 @@ LorenzSystem::LorenzSystem()
...
-    volume_ = std::make_shared<Volume>(size_.get(), DataVec3Float32::get());
-    volume_->dataMap_.dataRange = vec2(0, 1);
-    volume_->dataMap_.valueRange = vec2(-1, 1);
-    outport_.setData(volume_);
+    auto volume = std::make_shared<Volume>(size_.get(), DataVec3Float32::get());
+    volume->dataMap_.dataRange = vec2(0, 1);
+    volume->dataMap_.valueRange = vec2(-1, 1);
+
+    auto curlvolume = std::make_shared<Volume>(size_.get(), DataVec3Float32::get());
+    curlvolume->dataMap_.dataRange = vec2(0, 1);
```
ResearchDaniel, line 89: Are you sure that the data range should not be equal to value range?

ResearchDaniel (approved):
Some minor comments

include/inviwo/core/util/foreach.h
r-englund: If poolsize is zero we execute it directly since we have no worker threads running

modules/vectorfieldvisualization/src/processors/datageneration/seedpointsfrommask.cpp
r-englund: positions specified as homogeneous coordinates are only considered correct when w == 1

include/inviwo/core/util/foreach.h
```diff
@@ -78,17 +85,21 @@ auto foreach_helper_pool(std::false_type, IT a, IT b, Callback callback,
  * the data structure
  * @param jobs optional parameter specifying how many jobs to create, if jobs==0 (default) it will
  * create pool size * 4 jobs
+ * @param onTaskDone callback that will be called when each job is done
  * @return a vector of futures, one for each job created.
  */
-template <typename Iterable, typename T = typename Iterable::value_type, typename Callback>
+template <typename Iterable, typename T = typename Iterable::value_type, typename Callback,
+          typename OnDoneCallback = std::function<void()>>
```
petersteneteg, line 93: Why do you have a std::function here? Should this argument not always be deduced?

modules/vectorfieldvisualization/src/processors/datageneration/seedpointsfrommask.cpp
```diff
@@ -93,24 +95,31 @@ void SeedPointsFromMask::process() {
             util::IndexMapper3D index(dim);
             vec3 invDim = vec3(1.0f) / vec3(dim);
 
+            std::function<vec3(vec3)> noTransform = [](vec3 p) -> vec3 { return p; };
```
petersteneteg: use 'const auto' here, never assign a lambda to a std::function unless you have to. std::function uses type erasure, and pays a big pref penalty for that compared to using the lambda directly....

petersteneteg (approved):

include/inviwo/core/util/foreach.h
r-englund: It fails to deduce its type when using the default parameter (i.e the user does not pass a callback)

modules/vectorfieldvisualization/src/processors/datageneration/seedpointsfrommask.cpp
r-englund: I redesigned it now. 
I needed it before since I wanted to select between this and the next lambda to avoid an if statement inside the lambda or in the for loop.

PR #479, opened 2019-04-02
BaseGL: Added property for changing indicator size

Alboin:
Added property to change indicator size in VolumeSliceGL since the size was previously determined on the size of the input volume.

Also wanted to create a property for changing the linewidth as it also is depending on ratio between input size and output image size, but it seems that the glLineWidth function call does not change anything and the lines are always 1 pixel wide with respect to the volume resolution.

(Also did some clang formatting)

### DIFF
--- a/modules/basegl/src/processors/volumeslicegl.cpp
+++ b/modules/basegl/src/processors/volumeslicegl.cpp
@@ -92,6 +92,8 @@ VolumeSliceGL::VolumeSliceGL()
     , indicatorColor_("indicatorColor", "Indicator Color", vec4(1.0f, 0.8f, 0.1f, 0.8f), vec4(0.0f),
                       vec4(1.0f), vec4(0.01f), InvalidationLevel::InvalidOutput,
                       PropertySemantics::Color)
+    , indicatorSize_("indicatorSize", "Indicator Size", 4.0f, 0.0f, 100.0f, 0.01f,
+                     InvalidationLevel::InvalidOutput)
     , tfMappingEnabled_("tfMappingEnabled", "Enable Transfer Function", true,
                         InvalidationLevel::InvalidResources)
     , transferFunction_("transferFunction", "Transfer Function", &inport_)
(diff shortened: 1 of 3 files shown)


modules/basegl/src/processors/volumeslicegl.cpp
```diff
@@ -470,7 +478,7 @@ void VolumeSliceGL::renderPositionIndicator() {
 
     utilgl::GlBoolState smooth(GL_LINE_SMOOTH, true);
     utilgl::BlendModeState blend(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA);
-    utilgl::LineWidthState linewidth(2.5f);
+    //utilgl::LineWidthState linewidth(2.5f); // Doesn't seem to make a difference
```
ResearchDaniel: You can remove it. It is not supported in OpenGL anymore.
Use linerenderer instead

ResearchDaniel: Perhaps you could mark LineWidthState as deprecated?

ResearchDaniel (approved):
glLineWidth is not supported in OpenGL core.

r-englund:
> 
> 
> glLineWidth is not supported in OpenGL core.

Only semi-true. It is supported, but only a glLineWidth of 1.0 is guaranteed, anything else is implementation specific, and most (at least nvidia) does only support 1.0.  If we want a wide line, we need to use a quad.

modules/basegl/src/processors/volumeslicegl.cpp
Alboin: How can I mark it deprecated? And should I? As Richard says it might be supported for some people.

ResearchDaniel: Noone should use it IMHO.
Example:

```
[[deprecated("was declared deprecated. Use `onChange(std::function<void()>)` instead")]]
const BaseCallBack* onChange(T* o, void (T::*m)());
```

Alboin: Yeah, but where do I put "[[deprecated]]"? I tried putting it above this line but it just says "attributes are not allowed here":
image
And I don't want to deprecate SimpleState since it's a template struct used for other stuff as well.

ResearchDaniel: Inbetween 
`using PS [[deprecated]] = S*;`
https://en.cppreference.com/w/cpp/language/attributes/deprecated

Alboin: 0ffd3b4 this alright?

PR #482, opened 2019-04-04
Fix/button to fileread conversion

Alboin:
Added conversion from ButtonProperty to FileProperty and DirectoryProperty.

Then added fallback functionality to FileProperty to be able to open the file dialog and read files even if the property Qt widgets hasn't been created yet. Also added this functionality in an override function in DirectoryProperty.

Then made "folder" the default option for VolumeSequenceSource instead of "file".

All this is so that one can open the file dialog of a source processor with a synchronized button, without having to go through the workspace network and click every source processor so that the Qt widgets are created and files can be read.


include/inviwo/core/properties/directoryproperty.h
```diff
@@ -55,6 +55,8 @@ class IVW_CORE_API DirectoryProperty : public FileProperty {
     DirectoryProperty& operator=(const DirectoryProperty& that);
     virtual DirectoryProperty* clone() const override;
 
+    virtual void requestFile() override;
```
ResearchDaniel: Add documentation. Should be requestDirectory?

src/core/properties/directoryproperty.cpp
```diff
@@ -53,4 +54,35 @@ std::string DirectoryProperty::getClassIdentifierForWidget() const {
...
+
+        // Setup Extensions
+        std::vector<FileExtension> filters = this->getNameFilters();
+
+        auto fileDialog = util::dynamic_unique_ptr_cast<FileDialog>(
+            InviwoApplication::getPtr()->getDialogFactory()->create("FileDialog"));
+        if (!fileDialog) {
+            return;
+        }
+        fileDialog->setTitle("Open Folder");
```
ResearchDaniel: These should be set based on the FileProperty

ResearchDaniel (changes requested):

include/inviwo/core/properties/directoryproperty.h
Alboin: Since DirectoryProperty inherit from FileProperty I wanted to override the function that FileProperty runs if the property is a DirectoryProperty. I don't know how a requestDirectory function would get called if such a function was added.

src/core/properties/directoryproperty.cpp
Alboin: Like this fc0bade?

include/inviwo/core/properties/directoryproperty.h
ResearchDaniel: Ah, I see, you should add the corresponding documentation of the FileProperty::requestFile() 
In other words that it will present a dialog to choose directory and its path will be set

src/core/properties/directoryproperty.cpp
ResearchDaniel: No, like in https://github.com/inviwo/inviwo-dev/blob/4fd6dfe2e7df8e05e0fd4b8ce5400a051e33024e/modules/qtwidgets/properties/filepropertywidgetqt.cpp#L133

Line 133 and onwards

src/core/properties/fileproperty.cpp
```diff
@@ -203,6 +205,23 @@ void FileProperty::requestFile() {
             if (filerequestable->requestFile()) return;
         }
     }
+    if (getWidgets().empty()) {
+
+        const std::string filename{this->get()};
```
petersteneteg: unused?

src/core/properties/fileproperty.cpp
```diff
@@ -203,6 +205,23 @@ void FileProperty::requestFile() {
             if (filerequestable->requestFile()) return;
         }
     }
+    if (getWidgets().empty()) {
+
+        const std::string filename{this->get()};
+
+        // Setup Extensions
+        std::vector<FileExtension> filters = this->getNameFilters();
```
petersteneteg: unused

src/core/properties/directoryproperty.cpp
```diff
@@ -53,4 +54,31 @@ std::string DirectoryProperty::getClassIdentifierForWidget() const {
...
 
+void DirectoryProperty::requestFile() {
+    for (auto widget : getWidgets()) {
+        if (auto filerequestable = dynamic_cast<FileRequestable*>(widget)) {
+            if (filerequestable->requestFile()) return;
+        }
+    }
+    if (getWidgets().empty()) {
+
+        const std::string filename{ this->get() };
```
petersteneteg: unused?

petersteneteg (changes requested):

petersteneteg:
I think this approach is very questionable. I don't think I makes any sense to link buttons to file properties.

ResearchDaniel:
@petersteneteg Reopening this again because One cannot call requestFile in python without this code

ResearchDaniel:
But we can remove the conversion from button if it is still in here

petersteneteg:
Why not just expose the file dialog to python, all this requestDialog,really seems to break clear design  boundaries to me. Mixing in GUI stuff where it really does not belong

ResearchDaniel:
Would also be an option but then one would still need to involve the processor to get correct extension etc..
Perhaps the requestFile should not deal with widgets at all?
That would mean that we replace the widget stuff with file dialog. Makes more sense.

I guess both would be good to have anyway

ResearchDaniel:
@petersteneteg Removed button sync and fixed lots of errors. IMHO we should consider removing the widget part and just use the FileDialog in another PR.

petersteneteg (approved):

ResearchDaniel (approved):

PR #488, opened 2019-04-10
Feature/webbrowser

martinfalk:


petersteneteg (approved):

modules/webbrowser/src/processors/webbrowserprocessor.cpp
```diff
@@ -86,27 +89,52 @@ WebBrowserProcessor::WebBrowserProcessor()
...
-                break;
-            case SourceType::WebAddress:
-                fileName_.setVisible(false);
-                url_.setVisible(true);
-                break;
+    addProperty(runJS_);
+    addProperty(js_);
+
+    runJS_.onChange([this]() {
+        browser_->GetMainFrame()->ExecuteJavaScript(js_.get(), "", 1);
```
ResearchDaniel: Script should probably be executed in the process function to make sure that the web page is finished loading when pressing the button.

ResearchDaniel (approved):
Nice! Might want to use a file for the javascript, but I'm fine with a simple string property as well.

modules/webbrowser/src/processors/webbrowserprocessor.cpp
```diff
@@ -266,6 +266,11 @@ void WebBrowserProcessor::OnLoadingStateChange(CefRefPtr<CefBrowser> browser, bo
 }
 
 void WebBrowserProcessor::process() {
+    if (!isBrowserLoading_ && runJSrequested_) {
```
ResearchDaniel: No need to check isBrowserLoading, see isReady in constructor. Also, would not runJS_.isModified() work?

ResearchDaniel (changes requested):

modules/webbrowser/src/processors/webbrowserprocessor.cpp
martinfalk: when relying on the isReady state, isModified should be sufficient.

martinfalk: fixed in latest commit

ResearchDaniel (approved):

PR #498, opened 2019-04-18
Fix/use json in webbrowser

ResearchDaniel:
- Cleanup of property sync handling. Uses json lib for handling json now.
- Support for syncing MinMaxProperty
- Fix bug where checkboxes where not properly synced at startup
- Fix bug with executing empty javascript script

### DIFF
--- a/modules/webbrowser/CMakeLists.txt
+++ b/modules/webbrowser/CMakeLists.txt
@@ -84,7 +84,7 @@ set(HEADER_FILES
     include/modules/webbrowser/properties/boolpropertywidgetcef.h
     include/modules/webbrowser/properties/buttonpropertywidgetcef.h
     include/modules/webbrowser/properties/ordinalpropertywidgetcef.h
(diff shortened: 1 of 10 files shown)


modules/webbrowser/CMakeLists.txt
```diff
@@ -84,6 +84,7 @@ set(HEADER_FILES
     include/modules/webbrowser/properties/boolpropertywidgetcef.h
     include/modules/webbrowser/properties/buttonpropertywidgetcef.h
     include/modules/webbrowser/properties/ordinalpropertywidgetcef.h
+	include/modules/webbrowser/properties/minmaxpropertywidgetcef.h
```
Alboin: Maybe it's just GitHub or should this document be clang-formatted? Seems like the intendation is a bit off on row 87, 111 and perhaps 143.

Alboin (approved):

modules/webbrowser/CMakeLists.txt
ResearchDaniel: Check, fixed

PR #499, opened 2019-04-24
PlottingGL: Added option for switching off ability to select multiple…

Alboin:
… axes. Made it the default option with single axis selection.

### DIFF
--- a/modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
+++ b/modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
@@ -82,6 +82,7 @@ ParallelCoordinates::ParallelCoordinates()
           TransferFunction{
               {{0.0, vec4{1, 0, 0, 1}}, {0.5, vec4{1, 1, 0, 1}}, {1.0, vec4{0, 1, 0, 1}}}}}
 
+    , multipleAxisSelection_("multipleAxisSelection", "Multiple Axis Selection", false)
     , lineSettings_{"lines", "Line Settings"}
     , blendMode_("blendMode", "Blend Mode",
                  {{"additive", "Additive", BlendMode::Additive},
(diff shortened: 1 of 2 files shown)


modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
```diff
@@ -82,6 +82,7 @@ ParallelCoordinates::ParallelCoordinates()
           TransferFunction{
               {{0.0, vec4{1, 0, 0, 1}}, {0.5, vec4{1, 1, 0, 1}}, {1.0, vec4{0, 1, 0, 1}}}}}
 
+    , multipleAxisSelection_("multipleAxisSelection", "Multiple Axis Selection", false)
```
r-englund: Should it be off by default? This changes the old behavior

r-englund (approved):
Nice, have some local changes myself to "fix" this, was gonna create a PR but you were quicker :)

modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
Alboin: That's the question! Would it be a more common case to make multiple selections?

ResearchDaniel:
Nice, but should it be multiple options?
Multiple
Single
Off

ResearchDaniel (approved):

modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
ResearchDaniel: I don't think so. I guess if one wants to highlight the selected axis in a scatter plot matrix or so. But perhaps it is best to have the more general as default anyway?

Alboin:
Made it an OptionProperty with, Single, Multiple and None as alternatives, where Multiple is selected by default.

ResearchDaniel (approved):
Sweet

PR #500, opened 2019-04-25
Feature/pcp features

martinfalk:
* Axes feature a "flipped" flag. If enabled, all labels and ticks appear on the other side of the axis.
* functionality to invert axes in the PCP plot
* PCP handles can be hidden
* some brushing fix (filtering was not always applied in parallel coords plot)

## Flipped _Petal Length_ and _Class_
image

## Regular PCP
image


martinfalk:
An axis can be inverted via a double click with the left mouse button or using the `Invert Range` property of the axis.
image

modules/plottinggl/depends.cmake
```diff
@@ -2,6 +2,7 @@
 # Dependencies for current module
 # List modules on the format "Inviwo<ModuleName>Module"
 set(dependencies
+	InviwoDataFrameModule
```
ResearchDaniel: Indent

modules/plottinggl/include/modules/plottinggl/processors/parallelcoordinates/parallelcoordinates.h
```diff
@@ -194,8 +195,9 @@ class IVW_MODULE_PLOTTINGGL_API ParallelCoordinates : public Processor {
 
         // startFilter, startRegular, startSelected, end
         std::array<size_t, 4> offsets;
-
+        
         std::vector<float> axisPositions;
+        std::vector<float> axisFlipped;
```
ResearchDaniel: float vector? Should be bool? Otherwise comment why

modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
```diff
@@ -668,6 +696,22 @@ void ParallelCoordinates::axisPicked(PickingEvent* p, size_t pickedID, PickType
         p->markAsUsed();
         invalidate(InvalidationLevel::InvalidOutput);
     }
+    if (p->getPressState() == PickingPressState::DoubleClick &&
+        p->getPressItem() == PickingPressItem::Primary) {
+                
+        axes_[pickedID].pcp->invertRange.set(!axes_[pickedID].pcp->invertRange);
```
ResearchDaniel, line 696: In my opinion, axis flipping is not so important that it should be assigned to double-click... 
Context-menu option would be great here I think.

ResearchDaniel (changes requested):
nice, a `vector<float>` that should be `vector<bool>`?
Also, I don't think assigning axis-flipping to double-clicks are the most intuitive, but that might just be me...

modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
```diff
@@ -510,8 +531,13 @@ void ParallelCoordinates::drawHandles(size2_t size) {
             sliderWidgetRenderer_.setUIColor(handleColor_);
         }
 
-        axis.sliderWidget->render(ivec2{ap.first} - ivec2{handleSize_.get(), handleWidth} / 2,
-                                  size);
+        if (axis.pcp->invertRange) {
+            axis.sliderWidget->render(ivec2{ap.first} - ivec2{handleSize_.get(), handleWidth} / 2,
```
petersteneteg: Identical?

petersteneteg (changes requested):

modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
martinfalk: If a context menu event for a specific mouse position would exist, yes.
However, the double click feature was requested by @Alboin  and @petersteneteg

modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
martinfalk: Yes, that was a left-over. If statement was removed.

modules/plottinggl/include/modules/plottinggl/processors/parallelcoordinates/parallelcoordinates.h
martinfalk: Changed to `std::vector<int>` instead. Using int here for performance reasons since bool is not supported as GLSL uniform. A bool vector would internally be converted to an int array prior setting the uniform.

see UniformSetter<std::array<bool, N>>

modules/plottinggl/depends.cmake
martinfalk: fixed

ResearchDaniel (approved):

petersteneteg (approved):

PR #501, opened 2019-04-26
WebBrowser: Added optional dataframe inport

Alboin:
There is some problem with the dataframe to json conversion that I don't get on row 278 when the program throws a stack overflow exception.

I thought maybe I could get some help understanding what I do wrong in this PR, fix it, and then hopefully merge it to master :)

### DIFF
--- a/modules/webbrowser/depends.cmake
+++ b/modules/webbrowser/depends.cmake
@@ -3,4 +3,5 @@
 # List modules on the format "Inviwo<ModuleName>Module"
 set(dependencies
     InviwoOpenGLModule
+	InviwoDataframeModule
 )
--- a/modules/webbrowser/depends.cmake
+++ b/modules/webbrowser/depends.cmake
@@ -3,5 +3,5 @@
 # List modules on the format "Inviwo<ModuleName>Module"
 set(dependencies
     InviwoOpenGLModule
-	InviwoDataframeModule
+    InviwoDataframeModule
 )
--- a/modules/webbrowser/src/processors/webbrowserprocessor.cpp
+++ b/modules/webbrowser/src/processors/webbrowserprocessor.cpp
@@ -277,8 +277,9 @@ void WebBrowserProcessor::process() {
         for (const auto& dataFrame : dataFrames_) {
             nlohmann::json jsonString = *dataFrame;
             std::stringstream data("var data = ", std::ios_base::app | std::ios_base::out);
-            data << jsonString.dump();
-            data << ";onInviwoDataChanged" << dataFrameNr++ << "(data);";
+            data << jsonString.dump() << ";";
+            data << "var dataFrameNr = " << dataFrameNr++ << ";";
+            data << "onInviwoDataChanged(data, dataFrameNr);";
             frame->ExecuteJavaScript(data.str(), frame->GetURL(), 0);
         }
     }
--- a/modules/webbrowser/src/processors/webbrowserprocessor.cpp
+++ b/modules/webbrowser/src/processors/webbrowserprocessor.cpp
@@ -37,6 +37,7 @@
 #include <inviwo/core/util/filesystem.h>
 #include <inviwo/core/util/utilities.h>
 #include <nlohmann/json.hpp>
+#include <inviwo/dataframe/jsondataframeconversion.h>
 
 #include <warn/push>
 #include <warn/ignore/all>
--- a/modules/webbrowser/src/processors/webbrowserprocessor.cpp
+++ b/modules/webbrowser/src/processors/webbrowserprocessor.cpp
@@ -110,7 +110,10 @@ WebBrowserProcessor::WebBrowserProcessor()
     };
     updateVisibility();
 
-    auto reload = [this]() { browser_->GetMainFrame()->LoadURL(getSource()); };
+    auto reload = [this]() {
+        browser_->GetMainFrame()->LoadURL(getSource());
+        reloadAllDataframes_ = true;
+    };
 
     sourceType_.onChange([this, reload, updateVisibility]() {
         updateVisibility();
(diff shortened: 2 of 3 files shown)


ResearchDaniel:
What does the exception say?

modules/webbrowser/src/processors/webbrowserprocessor.cpp
```diff
@@ -267,6 +271,18 @@ void WebBrowserProcessor::process() {
         browser_->GetMainFrame()->ExecuteJavaScript(js_.get(), "", 1);
     }
 
+    if (dataFrames_.isReady()) {
+        auto frame = browser_->GetMainFrame();
+        size_t dataFrameNr = 0;
+        for (const auto& dataFrame : dataFrames_) {
+            nlohmann::json jsonString = *dataFrame;
```
ResearchDaniel: Is dataFrame a port or the actual data?
should it not be dataFrame->getData()

modules/webbrowser/src/processors/webbrowserprocessor.cpp
```diff
@@ -267,6 +271,18 @@ void WebBrowserProcessor::process() {
...
+            std::stringstream data("var data = ", std::ios_base::app | std::ios_base::out);
+            data << jsonString.dump();
+            data << ";onInviwoDataChanged" << dataFrameNr++ << "(data);";
```
ResearchDaniel: Better to have the number as argument: onDataFrameChanged(json, dataFrameNr) ?

modules/webbrowser/src/processors/webbrowserprocessor.cpp
ResearchDaniel: Also, then possible to check dataFrame->isModified() or similar?

Alboin: ```const shared_ptr<const inviwo::DataFrame> &dataFrame```
A pointer to the actual dataframe, so it's not posible to check isModified(). I can run:
```dataFrame->getNumberOfColumns()```
etc and it gives me what seems to be correct results for the dataframe.

File: modules/webbrowser/src/processors/webbrowserprocessor.cpp (hunk 2 above)
Comment (Alboin) [reply modules/webbrowser/src/processors/webbrowserprocessor.cpp]: Sure!

**Comment by Alboin** [comment]
The following exception is thrown on row 10618 in json.hpp, in a template-function so it's a bit hard to track it any further:
```Unhandled exception at 0x00007FF88D6F02E6 (inviwo-module-webbrowser.dll) in inviwo.exe: 0xC00000FD: Stack overflow (parameters: 0x0000000000000001, 0x0000007235403FD0).```

**Comment by ResearchDaniel** [comment]
What if you put a break-point in to_json function and step through?
If you do it with a small dataframe it should be possible to analyse when the exception occur... Or look at the call stack

**Comment by martinfalk** [comment]
What exactly is the reason for having a DataFrame port in the WebBrowser processor? This processor should be as simple as possible and server more as an example. If additional functionality is needed, there should be a new processor instead of (over-)extending WebBrowser.

**Review by martinfalk — CHANGES_REQUESTED** [review]
Need a proper reason for this feature.

**Comment by Alboin** [comment]
> What if you put a break-point in to_json function and step through?
> If you do it with a small dataframe it should be possible to analyse when the exception occur... Or look at the call stack

It crashes before even getting to to_json. Also I notice that it isn't able to build on jenkins, giving the error:

```13:57:12  ../inviwo/modules/webbrowser/src/processors/webbrowserprocessor.cpp:278:28: error: no viable conversion from 'const inviwo::DataFrame' to 'nlohmann::json' (aka 'basic_json<>')```
```13:57:12              nlohmann::json jsonString = *dataFrame;```

**Comment by Alboin** [comment]
> What exactly is the reason for having a DataFrame port in the WebBrowser processor? This processor should be as simple as possible and server more as an example. If additional functionality is needed, there should be a new processor instead of (over-)extending WebBrowser.

@martinfalk I think it's a very small change to the processor that has many use cases.
Do you suggest we make a copy of this processor with the only difference being 12 rows of code that gives it an additional optional inport? We could also split up the PCP processor in one version without brushing and one version with brushing since it's already optional? 😉

**Comment by ResearchDaniel** [comment]
@alboin I think that you have forgotten to include `jsondataframeconversion.h`

**Comment by ResearchDaniel** [comment]
@martinfalk That was my initial thought as well but in the end it does not bloat too much and creating a new processor would require duplicating quite some code unless we use inheritance and that's something we try to avoid for processors...

**Comment by Alboin** [comment]
> @Alboin I think that you have forgotten to include `jsondataframeconversion.h`

🥇 @ResearchDaniel that did the trick!

**Comment by Alboin** [comment]
Although now I get a lot of strange regression test errors...

**Comment by martinfalk** [comment]
> Although now I get a lot of strange regression test errors...

@Alboin this issue was addressed in #502, not necessarily your fault ;-)

File: modules/webbrowser/depends.cmake [hunk 3]
```diff
@@ -3,4 +3,5 @@
 # List modules on the format "Inviwo<ModuleName>Module"
 set(dependencies
     InviwoOpenGLModule
+	InviwoDataframeModule
```
martinfalk: Indentation!

modules/webbrowser/src/processors/webbrowserprocessor.cpp
martinfalk: @ResearchDaniel `dataFrames_` is the port, but there is an iteratable which returns the data directly.

Alternatively, there is `auto changed = inport_.getChangedOutports();` which only returns updated ports.

petersteneteg:
It would make more sense if the JavaScript side could ask for data frame input, and then dynamically add the ports needed

Alboin:
@petersteneteg if that were the case, would it be possible to save workspaces where the port has already been created and connected? So that you don't have to reconnect the port every time you load the workspace.

petersteneteg:
@Alboin Yes that should be possible, but it might require some custom code in the serialization to work properly.

ResearchDaniel:
Closing this PR since. Solution is instead to refactor the WebBrowser processor to make it easier to make custom ones.

PR #503, opened 2019-04-26
Feature/meshinformation

martinfalk:
MeshInformation processor similar to `ImageInformation` and `VolumeInformation`

image

### DIFF
--- a/modules/base/src/processors/meshinformation.cpp
+++ b/modules/base/src/processors/meshinformation.cpp
@@ -1,0 +1,67 @@
+/*********************************************************************************
+ *
+ * Inviwo - Interactive Visualization Workshop
+ *
+ * Copyright (c) 2019 Inviwo Foundation
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ *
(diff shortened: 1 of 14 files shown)


modules/base/src/processors/meshinformation.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+    meshInfo_.setSerializationMode(PropertySerializationMode::None);
+
+    setAllPropertiesCurrentStateAsDefault();
+}
+
+void MeshInformation::process() {
+    auto mesh = mesh_.getData();
+
+    meshInfo_.updateForNewMesh(*(mesh.get()));
```
petersteneteg: *mesh is enough

modules/base/src/properties/bufferinformationproperty.cpp
```diff
@@ -0,0 +1,202 @@
...
+                                                     InvalidationLevel invalidationLevel,
+                                                     PropertySemantics semantics)
+    : CompositeProperty(identifier, displayName, invalidationLevel, semantics)
+    , format_("format", "Format", "")
+    , size_("size", "Size", 0, 0, std::numeric_limits<size_t>::max(), 1,
+            InvalidationLevel::InvalidOutput, PropertySemantics("Text"))
+    , usage_("usage", "Usage")
+    , target_("target", "Target") {
+
+    util::for_each_in_tuple(
```
petersteneteg, line 60: For each argument is a bit simpler

petersteneteg (approved):

modules/base/src/properties/bufferinformationproperty.cpp
martinfalk: sticking with `for_each_in_tuple` and `props()`:
```c++
        util::for_each_in_tuple([](auto& dst, auto& src) { dst = src; }, props(), that.props());
```

PR #506, opened 2019-04-30
PlottingGL: PCP fix

martinfalk:
* fixed a crash caused by refactoring in 547e4a6

### DIFF
--- a/modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
+++ b/modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
@@ -342,6 +342,7 @@ void ParallelCoordinates::createOrUpdateProperties() {
     if (!data->getNumberOfRows()) return;
     if (!data->getNumberOfColumns()) return;
 
+    updating_ = true;
     axisPicking_.resize(data->getNumberOfColumns());
     lines_.axisFlipped.resize(data->getNumberOfColumns());
     for (size_t i = 0; i < data->getNumberOfColumns(); i++) {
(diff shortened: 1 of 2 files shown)


petersteneteg (approved):

modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
```diff
@@ -342,6 +342,7 @@ void ParallelCoordinates::createOrUpdateProperties() {
     if (!data->getNumberOfRows()) return;
     if (!data->getNumberOfColumns()) return;
 
+    updating_ = true;
```
r-englund, line 345: use util::KeepTrueWhileInScope  (see my same fix in https://github.com/inviwo/inviwo/commit/160dcf73ea35d0aa847cc814bc3f3ef3cdffc69f)

r-englund (changes requested):
> fixed a crash caused by refactoring in 547e4a6

There was an issue with this before that commit, but it weren't crashing

modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
martinfalk: there is some code at the end of the function which requires the bool to be reverted to false again. When using KeepTrueWhileInScope one has to do that anyway or introduce another `{ }` block just for the raii thingy.

martinfalk:
It was for me in case the input dataframe was modified.

r-englund (approved):

modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
r-englund: I see

PR #509, opened 2019-05-02
Fix/tf selector

Alboin:
Added a check for preventing the processor from changing selected TF when an unselected TF is modified. Previously it changed the selected TF to the modified even though it wasn't the chosen TF in the OptionProperty.


martinfalk (approved):
good catch!

ResearchDaniel (approved):

PR #510, opened 2019-05-07
serialization fix

petersteneteg:
Property and ProperyOwner need to inherit virtually from Serializable since CompositeProperty inherits from both property and propertyowner, otherwise we get diamond issues.


ResearchDaniel (approved):
Good catch. Would be nice with a comment in the code as well.

martinfalk (approved):

PR #511, opened 2019-05-08
Feature/doxygen fixes

r-englund:
* Some updates to doxygen.cmake 
* Updates to various comments in the code that generated Doxygen Warnings


tools/doxygen/doxygen.cmake
```diff
@@ -41,34 +41,39 @@ function(ivw_get_subdirs_recursive retval start)
...
-    ivw_get_subdirs_recursive(subdirs ${IVW_ROOT_DIR})
-    foreach(subdir IN LISTS subdirs)
-        get_property(targets DIRECTORY ${subdir} PROPERTY BUILDSYSTEM_TARGETS)
-        foreach(target IN LISTS targets)
-            ivw_get_target_property_recursive(dirs ${target} INTERFACE_INCLUDE_DIRECTORIES True)
-            list(APPEND res ${dirs})
-            ivw_get_target_property_recursive(dirs ${target} INCLUDE_DIRECTORIES False)
-            list(APPEND res ${dirs})
-            list(REMOVE_DUPLICATES res)
+    list(LENGTH ivw_doxy_all_dirs len)
```
r-englund, line 46: The first half of this function takes several seconds and is called once per enabled module. This resulted in 1-2 minutes increase to cmake config/generation time. Since it always is the same we now populate it just once per "cmake config/generation".  (I printed the return value before and and after the fix and they are exactly the same)

martinfalk: nice

include/inviwo/core/util/colorconversion.h
```diff
@@ -44,7 +44,7 @@ namespace color {
  * hexcodes with a leading '#' and optional alpha value (single / double digit).
  * In case of 3 respective 6 digits, alpha is implicitly set to 1.0.
  *
- * Supports #RGB, #RGBA, #RRGGBB, #RRGGBBAA
+ * Supports "#RGB", "#RGBA", "#RRGGBB", "#RRGGBBAA"
```
martinfalk, line 47: hehe :)

tools/doxygen/doxygen.cmake
```diff
@@ -501,7 +509,12 @@ function(make_doxygen_target modules_var)
     set_target_properties("DOXY-generate-processor-previews" 
                             PROPERTIES FOLDER "doc" EXCLUDE_FROM_ALL TRUE)
 
-    add_dependencies("DOXY-Inviwo" "DOXY-generate-processor-previews" "DOXY-Clear")
+    option(IVW_DOXYGEN_OPTIONAL_PROCESSOR_PREVIEWS "Turn on this to disable generating processor previews when building doxygen documentation (DOXY-Inviwo)" OFF)
```
martinfalk: Um, turn on to disable but the option is still called `IVW_DOXYGEN_OPTIONAL_PROCESSOR_PREVIEWS`? Maybe have CMake option `ON` by default or rename to `IVW_DOXYGEN_OPTIONAL_PROCESSOR_PREVIEWS_DISABLED` or similar. 
Either way, should remove `"this"` in the string.

martinfalk (changes requested):
The use of `IVW_DOXYGEN_OPTIONAL_PROCESSOR_PREVIEW` is strange.

tools/doxygen/doxygen.cmake
r-englund: I agree that the wording is not the best. 
The reason for the option is to avoid to have to build the whole `inviwo.exe` when you want to build the documentation only. (`DOXY-generate-processor-previews` depends on `inviwo.exe` and if you change the documentation in a file in core, you have to wait 20 min to see the results)

r-englund: The `IVW_DOXYGEN_OPTIONAL_PROCESSOR_PREVIEWS_DISABLED` feels a little like double negation.  The optional is to there to remove the dependency to the preview generator target. The target is still there but has to be built manually when needed. Appending a disabled to the option name feels wrong. Maybe something like, enforce_processor_previews that is on by default?

martinfalk: how about negating the if statement then? Then the previews would be optional.

martinfalk (approved):

Issue #461, opened 2019-03-20
Chaining on Properties


xeTaiz:
So as discussed in the meeting ad described in #458, I will look into building the chaining api as follows:
All methods will return the type of where they are defined. Meaning if e.g. an OptionProperty overrides `setVisible()`, then it shall return an OptionProperty, otherwise it shall return whatever type is closest in the inheritance hierarchy that does override this function.

This means the chaining API **can not** be used in arbitrary order of functions. E.g. the following behavior may occur:
```
OptionProperty o;
o.addOption() // returns OptionProperty&
 .setVisible() // returns Property&, since it is not overriden anywhere
 .addOption() // method does not exist on Property
```

Therefore the api will have to be used from most specific to least specific functions.
However the compiler will tell you if the order is not correct. (It will tell you functions do not exist)

PR #520, opened 2019-05-14
Chaining API

xeTaiz:
Solves Issue #461.
I split my commits into two. The first commit implements the chaining for the properties and for the `addPort()` and the second uses the new apis in a few processors and properties in order to test and demonstrate. Despite the size increase of this PR, I left the second commit in there so reviewers can actually test the first commits functionality quickly.

### DIFF
--- a/include/inviwo/core/properties/optionproperty.h
+++ b/include/inviwo/core/properties/optionproperty.h
@@ -6,7 +6,7 @@
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
- * modification, are permitted provided that the following conditions are met:
+ * modification, are permitted provided that thvoide following conditions are met:
  *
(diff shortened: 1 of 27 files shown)


include/inviwo/core/properties/optionproperty.h
```diff
@@ -6,7 +6,7 @@
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
- * modification, are permitted provided that the following conditions are met:
+ * modification, are permitted provided that thvoide following conditions are met:
```
petersteneteg: ?

petersteneteg (approved):

include/inviwo/core/properties/optionproperty.h
xeTaiz: oh yeah.. I guess that was my cursor flying all over the place :smile: 
fixed it

PR #521, opened 2019-05-15
New API docs css

xeTaiz:
As mentioned in the meeting, there is a new style for the api docs.
I changed the doxygen cmake to just copy over and override doxygens styles.
I may revisit this when looking deeper into doxygen magic, but for now that makes the whole thing a lot prettier:
2019-05-15 11 13 22  2fec5930d2c3


petersteneteg:
Demo http://jenkins.inviwo.org:8080/job/inviwo/job/docs%252Fapi-style/Doxygen/

xeTaiz:
I have seen that there are several images that can't be loaded (some inviwo image and the sponsor images etc.)
I think I will copy all those images into `docs/images` or so and make sure they are copied to the build folder.

tools/doxygen/doxygen.cmake
```diff
@@ -480,6 +480,10 @@ function(make_doxygen_target modules_var)
         WORKING_DIRECTORY ${output_dir}
         COMMENT "Generating Inviwo API documentation with Doxygen"
         VERBATIM
+        COMMAND ${CMAKE_COMMAND} -E copy "${IVW_ROOT_DIR}/docs/css/doxygen.css" "${output_dir}/inviwo/html/doxygen.css"
+        COMMAND ${CMAKE_COMMAND} -E copy "${IVW_ROOT_DIR}/docs/css/navtree.css" "${output_dir}/inviwo/html/navtree.css"
+        COMMAND ${CMAKE_COMMAND} -E copy "${IVW_ROOT_DIR}/docs/css/search.css" "${output_dir}/inviwo/html/search.css"
+        COMMAND ${CMAKE_COMMAND} -E copy "${IVW_ROOT_DIR}/docs/css/tabs.css" "${output_dir}/inviwo/html/tabs.css"
```
r-englund: I think this can be specified in the generated doxy configuration file using HTML_STYLESHEET and HTML_EXTRA_STYLESHEET (http://www.doxygen.nl/manual/config.html#cfg_html_stylesheet) http://www.doxygen.nl/manual/config.html#cfg_html_extra_stylesheet

r-englund (dismissed):
Look much nicer :) 
We need to make sure all images are loaded correctly before merging.

petersteneteg:
images are included by setting the image paths, that is done on
https://github.com/inviwo/inviwo/blob/85f3fae483c9f99a7cc87d6984c743813cd646c5/tools/doxygen/doxygen.cmake#L420
and
https://github.com/inviwo/inviwo/blob/85f3fae483c9f99a7cc87d6984c743813cd646c5/tools/doxygen/doxygen.cmake#L450

No manual copying should be done

r-englund:
> 
> 
> I have seen that there are several images that can't be loaded (some inviwo image and the sponsor images etc.)
> I think I will copy all those images into `docs/images` or so and make sure they are copied to the build folder.

I think it can also be fixed by adding to the IMAGE_PATH in doxy configs, in that way we do not need to manage copies of images

tools/doxygen/doxygen.cmake
petersteneteg: yes

xeTaiz: yes, we found this and prepared an extra stylesheet, however doxygen did not seem to apply the stylesheet properly and there were always things missing, that is why we did the copying in the end

petersteneteg: Make a note explaining that...

xeTaiz:
Okay so I looked a little closer on this `IMAGE_PATH`. Apparently this takes care of images that are referenced in code tags (`@image`), but does not seem to work with markdown files.
However there is a `HTML_EXTRA_FILES` which I am now using for the images to be copied to the html output folder.
I also had to make changes to the README.md, because the images in there are referenced using html tags, which do work with the `HTML_EXTRA_FILES` I believe, however GitHub and Doxygen will expect the images at different paths. Therefore one of the two would be without images.

I changed the images to actual markdown images and wrapped them in a `div` that sets their width (for the sponsor links to be next to each other). However I just found that the width correction does not work with GitHubs markdown (while looking nice in the generated doxygen)

[Take a look at how GitHub butchers the markdown then](https://github.com/xeTaiz/inviwo/tree/docs/api-style). Any ideas to make this work for both GitHub and Doxygen? :/

xeTaiz:
On the `HTML_EXTRA_STYLESHEET` thingy:
I tried again to make it work, however doxygen seems really picky about what it uses from the custom css and what it does not. The breadcrumb-y horizontal bar with the search box seems to ignore basically everything from the custom css and in the actual documentation body there are some more things that just don't get applied. 
I have not deeply investigated why this might be the case, but I also don't feel like copying the 4 files is that bad of an idea, because at least whoever wants to adapt this style can actually be sure to have his changes applied and not get lost somewhere in the doxygen css magic...

martinfalk (dismissed):
Looking way better!
One request: please change the font of the text next to the logo (title bar) to something without serifs. If possible something close to what we use in the official Inviwo Logo.
image

xeTaiz:
> If possible something close to what we use in the official Inviwo Logo.

Do you know which font it uses?

martinfalk:
> Do you know which font it uses?

It appears to be Calibri with a bold face. Alternatively, one could use Segoe or a different sanserif font.

petersteneteg:
Any reason why you put the css files in the docs/ folder, all the old css files are in tools/doxygen/style.

xeTaiz:
> Any reason why you put the css files in the docs/ folder, all the old css files are in tools/doxygen/style.

Because there are more documentation related files in work and I would like to bundle them properly. For that I thought `docs` is a better place than the `tools` folder.
I will move the other files from `/tools/doxygen` accordingly in another PR.

petersteneteg:
Ok, but remove the old ones from the tools/doxygen/style/ in that case

xeTaiz:
> > Do you know which font it uses?
> 
> It appears to be Calibri with a bold face. Alternatively, one could use Segoe or a different sanserif font.

Apparently the html just didn't find the right font ;) It will look like on the website

xeTaiz:
So I moved everything from `tools/doxygen` directly into `docs/`, adapted the cmake files etc.
The correct font is also included now.

One issue that is not resolved yet is the sizing of the images. Since Doxygen and GitHub do different things with the `README.md`, I cannot get nicely sized images on both.
While some markdown flavors (including GitHub) support specifying image width etc. directly, Doxygen's does not. Adding those specifications in the markdown results in no image in Doxygen.
For Doxygen I added `<div class="sponsor-images">` tags around the images and could style the resulting HTML images without affecting GitHubs result negatively. However it results in way too large sponsor images on the GH page ([see here](https://github.com/xeTaiz/inviwo/tree/docs/api-style)).

Any suggestions to fix this?

.gitignore
```diff
@@ -1,7 +1,8 @@
...
 **/.DS_Store
-tools/doxygen/doc
-tools/doxygen/doc-qt
 *.pyc
 pyconfig.ini
-.idea
\ No newline at end of file
+.idea
+*.qch
+build
```
petersteneteg: Don't put build in here, one should nerver build in source, we should not encourage that

.gitignore
```diff
@@ -1,7 +1,8 @@
 inviwo-log.html
 **/.DS_Store
-tools/doxygen/doc
-tools/doxygen/doc-qt
 *.pyc
 pyconfig.ini
-.idea
\ No newline at end of file
+.idea
+*.qch
```
petersteneteg: We usually distribute the qch files right? Why should we stop that?

martinfalk:
@xeTaiz the obvious solution would be to create smaller versions of the logos and save them in the doxygen folder as well. Is the div tag in the readme messing up the scaling  in the markdown or is it just the github page?

docs/doxygen.cmake
```diff
@@ -307,11 +307,11 @@ function(ivw_private_make_help)
         "docpage{2}=\"\\page docpage-\\1 \\2\""
     )
 
-    set(additional_flags_list 
-        "LAYOUT_FILE            = ${HARG_HTML_DIR}/layout.xml"
-        "HTML_STYLESHEET        = ${HARG_HTML_DIR}/stylesheet.css"
-        "HTML_HEADER            = ${HARG_HTML_DIR}/header.html"
-        "HTML_FOOTER            = ${HARG_HTML_DIR}/footer.html"
+    set(additional_flags_list
```
petersteneteg, line 310: Why hard code an not use the function argument?

.gitignore
martinfalk: Yes, they are located in the doc folder of each module.

docs/markdown/GettingStarted_UI.md
```diff
@@ -0,0 +1,89 @@
+# The Inviwo Interface
```
petersteneteg: Great!

.gitignore
xeTaiz: Do we? If we do, how do we make sure they are recent? 
For me they are listed as untracked files and I thought we would not want to store generated files in the repo? 

Maybe we should then make sure that they are either updated on commit or built when compiling inviwo?

petersteneteg: We don't want to require you to run all doxygen stuff to get the help in inviwo, hence we distribute them. When one updates docs for a module one should also make sure to update the qch file for that module.

docs/markdown/GettingStarted_UI.md
xeTaiz: oh that was actually not even supposed to be in there :D
I'll make a separate PR for the markdown files

docs/doxygen.cmake
xeTaiz: To be honest it didn't work and I didn't understand why. But apparently while having this changed I did fix the issue with it, because when I change it now it works :D

xeTaiz:
> @xeTaiz the obvious solution would be to create smaller versions of the logos and save them in the doxygen folder as well. Is the div tag in the readme messing up the scaling in the markdown or is it just the github page?

The `<div>` is what makes it work in Doxygen and it does not affect the GitHub thing at all (probably because it does not find the css as it is in a different place for GH)

The only way to specify the dimensions for GitHub that does not completely break the image for Doxygen is to use an `` tag directly with attributes. However then there is the issue that for GH and Doxygen the images are at different paths. I think I could copy the images to `build/doxy/.../html/docs/images/` to make the paths equal. However that would like kind of odd and would not use Doxygens `HTML_IMAGE_PATH` thingy that actually makes it quite comfortable for all other `.md` files.

.gitignore
xeTaiz: Alright, I added them back

petersteneteg (approved):

xeTaiz:
I had to remove the link to changelogs from the new `Index.md` (since its broken).
I will update the contents of `Index.md` in another PR, together with other markdown things then. So far it is just the `README.md` adapted and with bad links removed.

xeTaiz:
Hmmm... somehow the jenkins does not like those commits. It builds without any issues on my machine.
Unfortunately [all feedback](http://jenkins.inviwo.org:8080/job/inviwo/job/docs%252Fapi-style/7/console) from jenkins is that it fails after the `htmlpublisher` task is started, no more info given. @petersteneteg  any tips on investigating this issue?

r-englund:
> `htmlpublisher`

Think this is true for all jenkins jobs atm? http://jenkins.inviwo.org:8080/job/modules/job/opemmesh%252Fcompilefix/

PR #526, opened 2019-05-24
Feature/image channel split

VisComKreiser:
added processors to split, combine, and replicate image channels.


modules/basegl/glsl/img_channel_combine.frag
```diff
@@ -0,0 +1,53 @@
...
+uniform sampler2D inport1Color;
+uniform sampler2D inport2Color;
+uniform sampler2D inport3Color;
+
+uniform bool use_alpha_texture_ = true;
+uniform float alpha_ = 1.0;
+
+void main() {
+    vec2 texCoords = gl_FragCoord.xy * outportParameters.reciprocalDimensions;
+    float r = texture(inport0Color, texCoords).r;
```
martinfalk, line 44: why only `.r`? Seems rather limiting.

martinfalk: another question is, why do need 4 source textures? How about using swizzling?

modules/basegl/glsl/img_channel_replicate.frag
```diff
@@ -0,0 +1,41 @@
...
+#include "utils/structs.glsl"
+
+uniform ImageParameters outportParameters_;
+uniform sampler2D inport_;
+uniform float alpha_ = 1.0;
+
+void main() {
+    vec2 texCoords = gl_FragCoord.xy * outportParameters_.reciprocalDimensions;
+    vec4 inputColor = texture(inport_, texCoords);
+    vec4 color = vec4(vec3(inputColor.r), alpha_);
```
martinfalk, line 39: Again, why only 'r' ? Use swizzle! 

https://github.com/inviwo/inviwo/blob/18f116ffe038b80b967c70b01bfb3d988fe2b322/include/inviwo/core/datastructures/image/layer.h#L90-L91

modules/basegl/glsl/img_channel_select.frag
```diff
@@ -0,0 +1,41 @@
...
+
+#include "utils/structs.glsl"
+
+uniform ImageParameters outportParameters_;
+
+uniform sampler2D inport_;
+
+void main() {
+    vec2 texCoords = gl_FragCoord.xy * outportParameters_.reciprocalDimensions;
+    float value = CHANNEL;
```
martinfalk, line 38: what is `CHANNEL`? Why not sample the input texture `inport_` and access it using swizzling?
i.e. `texture(inport_, texCoords).CHANNEL` or `texture(inport_, texCoords)[CHANNEL]`

At least predefine `CHANNEL` if it does not exist, otherwise the shader will not compile.

modules/basegl/include/modules/basegl/processors/imageprocessing/imagechannelcombine.h
```diff
@@ -0,0 +1,87 @@
...
+#include <modules/opengl/shader/shader.h>
+
+namespace inviwo {
+/** \docpage{org.inviwo.ImageChannelCombine, Image Channel Combine}
+ * ![](org.inviwo.ImageChannelCombine.png?classIdentifier=org.inviwo.ImageChannelCombine)
+ * Creates a 4 channel image with the connected inputs, from each of them the R channel is used.
+ * If the optional alpha channel is not provided, the alpha property is used to set the alpha value of the whole image
+ *
+ * ### Inports
+ *   * __inport0__ Red input image
```
martinfalk, line 47: So in order to recombine a 4 channel image, one has to split it up into 4 separate images with four processors and then use this one to recombine?

martinfalk: Really weird to have 4 inports, one for each channel.

I'd rather have a processor similar to `VolumeCombiner` with a multi inport.

modules/basegl/include/modules/basegl/processors/imageprocessing/imagechannelreplicate.h
```diff
@@ -0,0 +1,73 @@
...
+
+#include <modules/basegl/baseglmoduledefine.h>
+#include <inviwo/core/common/inviwo.h>
+#include <inviwo/core/properties/ordinalproperty.h>
+#include <modules/basegl/processors/imageprocessing/imageglprocessor.h>
+
+namespace inviwo {
+/** \docpage{org.inviwo.ImageChannelReplicate, Image Channel Replicate}
+ * ![](org.inviwo.ImageChannelReplicate.png?classIdentifier=org.inviwo.ImageChannelReplicate)
+ * Replicates the selected image channel and creates a 4 channel texture with the selected alpha value.
```
martinfalk, line 41: **What** is replicated, does not state it's only the red channel!  Swizzling!

I'd rather have a processor similar to `VolumeCombiner` with a multi inport.

modules/basegl/include/modules/basegl/processors/imageprocessing/imagechannelselect.h
```diff
@@ -0,0 +1,70 @@
...
+ * \brief Creates a single channel texture with the selected channel from the drop down menu.
+ */
+class IVW_MODULE_BASEGL_API ImageChannelSelect : public ImageGLProcessor {
+public:
+    ImageChannelSelect();
+    virtual const ProcessorInfo getProcessorInfo() const override;
+    static const ProcessorInfo processorInfo_;
+
+private:
+    OptionPropertyString channelSelector_; //!< Selection between RED, GREEN, BLUE, and ALPHA
```
martinfalk, line 65: And how would I create a `red, alpha` image?

martinfalk (changes requested):
I'd rather have a processor similar to `VolumeCombiner` which takes a statement like `vec4(img1.rr, img1.aa)` with a multi inport. Not to forget that `Layer` already features a swizzle mask.

modules/basegl/src/processors/imageprocessing/imagechannelcombine.cpp
```diff
@@ -0,0 +1,100 @@
...
+        inport3_.setOptional(true);
+        addPort(inport3_);
+
+        addPort(outport_);
+
+        addProperty(alpha_);
+    }
+
+    void ImageChannelCombine::process() {
+        if (inport0_.getData()->getDimensions() != inport1_.getData()->getDimensions() ||
```
martinfalk, line 70: Why is this check necessary? The textures are sample on the GPU anyway.

modules/basegl/src/processors/imageprocessing/imagechannelcombine.cpp
```diff
@@ -0,0 +1,100 @@
...
+            inport0_.getData()->getDimensions() != inport2_.getData()->getDimensions() ||
+            inport3_.hasData() && inport0_.getData()->getDimensions() != inport3_.getData()->getDimensions()) {
+            LogError("inport dimensions don't match!");
+            return;
+        }
+
+        if (inport0_.isChanged() || inport1_.isChanged() || inport2_.isChanged() || inport3_.isChanged()) {
+            const auto dimensions = inport0_.getData()->getDimensions();
+            auto img = std::make_shared<Image>(dimensions, DataVec4UInt8::get());
```
martinfalk, line 79: This is not good! Why is the image format hard coded here? Either use the common denominator of all input formats or at least use the one of the first inport.

martinfalk (changes requested):
What happens to picking buffer and depth buffer of the split/merged images?
Furthermore, an image can have more than one color layer. How is that handled?

r-englund (changes requested):
I do not understand the reason/usecase for the replicate processor, it just create a copy of the image with the first channel duplicated into the other two channels. This is the purpose of swizzeling. 
I made some suggested changes which uses swizzeling in the split processor and also removed the hardcoded used UInt8 (0dcd403121610bfe3c0ea01c7bd95e3e934bb459)

r-englund:
Another alternative to the split processor is to have 4 image outports instead of just one. One outport for each channel removing the need for 4 processors. Using a custom made FBO with 4 render targets, one for each channel, we would also only need one draw call to split the image instead of 4

r-englund:
rebased into #564

PR #531, opened 2019-05-29
Representation factory

petersteneteg:


include/inviwo/core/common/inviwoapplication.h
```diff
@@ -264,8 +266,11 @@ class IVW_CORE_API InviwoApplication : public Singleton<InviwoApplication>,
     PropertyWidgetFactory* getPropertyWidgetFactory() const;
 
     template <typename BaseRepr>
-    RepresentationConverterFactory<BaseRepr>* getRepresentationConverterFactory() const;
+    RepresentationFactory<BaseRepr>* getRepresentationFactory() const;
```
ResearchDaniel, line 362: Would be good with some documentation here, especially since there are two factories. When does one use one or the other?

include/inviwo/core/common/inviwomodule.h
```diff
@@ -162,6 +166,10 @@ class IVW_CORE_API InviwoModule {
     const std::vector<ProcessorWidgetFactoryObject*> getProcessorWidgets() const;
     const std::vector<PropertyFactoryObject*> getProperties() const;
     const std::vector<PropertyWidgetFactoryObject*> getPropertyWidgets() const;
+
+    const std::vector<BaseRepresentationFactoryObject*> getRepresentationFactoryObjects() const;
```
ResearchDaniel, line 170: Would be good if the purpose of each of the factories in the module interface were explained :)  When to use them, why there is factory and one factory object.

include/inviwo/core/datastructures/representationfactory.h
```diff
@@ -0,0 +1,73 @@
...
+public:
+    using BaseReprId = std::type_index;
+
+    BaseRepresentationFactory() = default;
+    virtual ~BaseRepresentationFactory() = default;
+    virtual BaseReprId getBaseReprId() const = 0;
+};
+
+template <typename Representation>
+class RepresentationFactory
```
ResearchDaniel, line 60: These factories are generally quite hard to understand so it would help with documentation on when and how the should be used :)

include/inviwo/core/datastructures/representationmetafactory.h
```diff
@@ -0,0 +1,64 @@
...
+ *
+ *********************************************************************************/
+#pragma once
+
+#include <inviwo/core/common/inviwocoredefine.h>
+#include <inviwo/core/datastructures/representationfactory.h>
+
+namespace inviwo {
+
+class IVW_CORE_API RepresentationMetaFactory {
```
ResearchDaniel, line 47: Document

src/core/common/inviwoapplication.cpp
```diff
@@ -128,6 +128,7 @@ InviwoApplication::InviwoApplication(int argc, char** argv, std::string displayN
     , propertyConverterManager_{std::make_unique<PropertyConverterManager>()}
     , propertyFactory_{std::make_unique<PropertyFactory>()}
     , propertyWidgetFactory_{std::make_unique<PropertyWidgetFactory>()}
+    , representationMetaFactory_{std::make_unique<RepresentationMetaFactory>()}
```
ResearchDaniel, line 131: missing initialization of representationFactories_ and functors?

ResearchDaniel (changes requested):
What about OpenCL?
Would be good with some documentation since these factories are rather complicated

modules/opengl/src/openglmodule.cpp
```diff
@@ -29,20 +29,62 @@
...
+    : public RepresentationFactoryObjectTemplate<BufferRepresentation, BufferGL> {
+public:
+    virtual std::unique_ptr<BufferRepresentation> create(
+        const typename BufferRepresentation::ReprOwner* buffer) {
+        return std::make_unique<BufferGL>(buffer->getSize(), buffer->getDataFormat(),
+                                          buffer->getBufferUsage(), buffer->getBufferTarget());
+    }
+};
+
+class LayerRAMFactoryObject
```
martinfalk: `LayerRAM`?

martinfalk (approved):
Nice to get rid of the default representations which were always there.
Maybe some lack of documentation

include/inviwo/core/common/inviwoapplication.h
petersteneteg: true

src/core/common/inviwoapplication.cpp
petersteneteg: that done by the modules, Core and OpenGL

petersteneteg:
@ResearchDaniel Have not added any OpenCL representation to the factory yet, should be easy to do... If the representation is not in the factory it will fallback to RAM as before....

petersteneteg:
see #543

PR #532, opened 2019-05-29
Various minor fixes

petersteneteg:

### DIFF
--- a/include/inviwo/core/common/inviwomodule.h
+++ b/include/inviwo/core/common/inviwomodule.h
@@ -253,9 +253,19 @@ public:
     void registerRepresentationConverterFactory(
         std::unique_ptr<BaseRepresentationConverterFactory> converterFactory);
 
+    /**
+     * Register a Settings class and hand over ownership.
+     * @see Settings
+     * @see InviwoApplication::getModuleSettings()
+     */
     void registerSettings(std::unique_ptr<Settings> settings);
+
+    /**
+     * Register a Settings class.
+     * @see Settings
+     * @see InviwoApplication::getModuleSettings()
+     */
     void registerSettings(Settings* settings);
-    void unRegisterSettings(Settings* settings);
 
     InviwoApplication* getInviwoApplication() const;
 
--- a/include/inviwo/core/common/inviwomodule.h
+++ b/include/inviwo/core/common/inviwomodule.h
@@ -165,7 +165,7 @@ public:
     const std::vector<BaseRepresentationConverter*> getRepresentationConverters() const;
     const std::vector<BaseRepresentationConverterFactory*> getRepresentationConverterFactories()
         const;
-    const std::vector<Settings*> getSettings() const;
+    const std::vector<Settings*>& getSettings() const;
 
     void registerCapabilities(std::unique_ptr<Capabilities> info);
 
@@ -181,7 +181,6 @@ public:
     void registerDrawer(std::unique_ptr<MeshDrawer> drawer);
     void registerMetaData(std::unique_ptr<MetaData> meta);
 
-    
     void registerProcessor(std::unique_ptr<ProcessorFactoryObject> pfo);
     template <typename T>
     void registerProcessor();
@@ -255,6 +254,8 @@ public:
         std::unique_ptr<BaseRepresentationConverterFactory> converterFactory);
 
     void registerSettings(std::unique_ptr<Settings> settings);
+    void registerSettings(Settings* settings);
+    void unRegisterSettings(Settings* settings);
 
     InviwoApplication* getInviwoApplication() const;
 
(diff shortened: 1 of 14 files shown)


include/inviwo/core/common/inviwomodule.h
```diff
@@ -255,6 +254,8 @@ class IVW_CORE_API InviwoModule {
         std::unique_ptr<BaseRepresentationConverterFactory> converterFactory);
 
     void registerSettings(std::unique_ptr<Settings> settings);
```
ResearchDaniel, line 261: We should document what happens with ownership for these three

include/inviwo/core/util/sourcecontext.h
```diff
@@ -0,0 +1,142 @@
...
+
+#include <string>
+#include <iostream>
+
+namespace inviwo {
+
+/**
+ * Represents a location in source code
+ */
+class IVW_CORE_API SourceContext {
```
ResearchDaniel, line 43: Would be more clear if this file (and class) would be called SourceCodeContext. Perhaps we should even put this in a subfolder (debug) since it is just for debugging?

include/inviwo/core/util/stdextensions.h
```diff
@@ -33,6 +33,7 @@
 #include <inviwo/core/common/inviwocoredefine.h>
 #include <inviwo/core/util/glm.h>
 #include <inviwo/core/util/foreacharg.h>
+#include <inviwo/core/util/hashcombine.h>
```
ResearchDaniel, line 36: Should update the changelog and state this change

modules/cimg/CMakeLists.txt
```diff
@@ -103,4 +77,44 @@ ivw_register_license_file(NAME tiff TARGET inviwo::tiff MODULE CImg
     FILES ${IVW_EXTENSIONS_DIR}/tiff/COPYRIGHT
 )
 
+
+
+option(IVW_CIMG_USE_OPENEXR  "Add OpenEXR image format support (HDR images) " ON)
```
ResearchDaniel, line 83: Should probably state that there is a problem with exr and that is why it is optional

ResearchDaniel (changes requested):

include/inviwo/core/util/stdextensions.h
petersteneteg: It does not break any code.

include/inviwo/core/common/inviwomodule.h
petersteneteg: the type says it all

martinfalk: agree with @petersteneteg here, in particular when considering lines 313-314

include/inviwo/core/util/sourcecontext.h
martinfalk: That's not just for debugging. Exception handling, stack traces, etc. Regarding the naming scheme, it was previously dubbed `ExceptionContext`. It refers to where its coming from, i.e. the source/reason, not necessarily source code.

modules/cimg/CMakeLists.txt
martinfalk: So you want to add a message explaining why one has to exclude OpenEXR when packaging Inviwo?

martinfalk (approved):

include/inviwo/core/util/stdextensions.h
ResearchDaniel: Ok, not ours at least ;)

modules/cimg/CMakeLists.txt
ResearchDaniel: Something inline with that. At least state what type of a problem there is and that is why we need this workaround.

include/inviwo/core/util/sourcecontext.h
ResearchDaniel: It says "a location in source code" and it has file_ and function_ as member variables :)

include/inviwo/core/common/inviwomodule.h
ResearchDaniel: At least state if the unregisterSettings *must* be called when using registerSettings(Settings*)

include/inviwo/core/util/stdextensions.h
petersteneteg: it was in stdextensions before, and now stdextensions include this one. So it will not break _any_ code as I said.

include/inviwo/core/util/sourcecontext.h
petersteneteg: wanted to be constant with https://en.cppreference.com/w/cpp/experimental/source_location

ResearchDaniel: So we don't want to name it SourceLocation?

petersteneteg: We have a SourceLocation further down... that is new and constepr but has a few restrictions... will add some comments...

martinfalk (approved):

ResearchDaniel (approved):

PR #535, opened 2019-05-29
Feature/minor cleanups

r-englund:
See commit messages

### DIFF
--- a/modules/dataframe/src/datastructures/dataframe.cpp
+++ b/modules/dataframe/src/datastructures/dataframe.cpp
@@ -155,11 +155,13 @@ const std::vector<std::pair<std::string, const DataFormatBase *>> DataFrame::get
     return headers;
 }
 
-std::string DataFrame::getHeader(size_t idx) const { return columns_[idx]->getHeader(); }
+std::string DataFrame::getHeader(size_t idx) const { return columns_.at(idx)->getHeader(); }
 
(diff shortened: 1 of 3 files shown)


martinfalk (approved):

modules/dataframe/src/datastructures/dataframe.cpp
```diff
@@ -155,11 +155,13 @@ const std::vector<std::pair<std::string, const DataFormatBase *>> DataFrame::get
     return headers;
 }
 
-std::string DataFrame::getHeader(size_t idx) const { return columns_[idx]->getHeader(); }
+std::string DataFrame::getHeader(size_t idx) const { return columns_.at(idx)->getHeader(); }
```
petersteneteg, line 158: don't use exception in place of an assertion!

petersteneteg (changes requested):

modules/dataframe/src/datastructures/dataframe.cpp
r-englund: Assertions are only enabled in debug mode....

r-englund: or when we force enable them in cmake

petersteneteg: that is a different issue
You should not depend on exception for detecting programmer errors!

PR #537, opened 2019-06-10
linear interpolation fix for unsigned vectors

r-englund:
When using `glm::mix` with unsigned int vector types (it works for scalars) glm gets the wrong result if the difference is negative: Before the fix `glm::mix(u8vec3{237}, u8vec3{34} , 0.5)` resulted in the value 7 instead of the expected 135.
This issue might have been fixed in latest glm https://github.com/g-truc/glm/issues/866

### DIFF
--- a/include/inviwo/core/util/interpolation.h
+++ b/include/inviwo/core/util/interpolation.h
@@ -85,7 +85,7 @@ inline T linearVectorInterpolation(const T &a, const T &b, P x) {
 
 template <typename T, typename P>
 inline T Interpolation<T, P>::linear(const T &a, const T &b, P x) {
-    using VT = util::value_type<T>::type;
+    using VT = typename util::value_type<T>::type;
     if constexpr (std::is_same_v<T, VT> || std::is_signed_v<VT> ) {
         // GLM fails at correctly linearly interpolate between vectors of unsigned types when the
         // difference is negative
--- a/include/inviwo/core/util/interpolation.h
+++ b/include/inviwo/core/util/interpolation.h
@@ -37,6 +37,7 @@ namespace inviwo {
 
 template <typename T, typename P = double>
 class Interpolation {
+     static_assert(std::is_floating_point_v<P>, "P needs to be a floating point type");
 public:
     Interpolation() = delete;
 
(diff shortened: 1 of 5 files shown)


include/inviwo/core/util/interpolation.h
```diff
@@ -84,7 +85,15 @@ inline T linearVectorInterpolation(const T &a, const T &b, P x) {
...
 inline T Interpolation<T, P>::linear(const T &a, const T &b, P x) {
-    return glm::mix(a, b, x);
+    using VT = typename util::value_type<T>::type;
+    if constexpr (std::is_same_v<T, VT> || std::is_signed_v<VT> ) {
+        // GLM fails at correctly linearly interpolate between vectors of unsigned types when the
+        // difference is negative
+        return glm::mix(a, b, x);
+    } else {
+        using V = typename util::same_extent<T, double>::type;
+        return util::glm_convert<T>(glm::mix(util::glm_convert<V>(a), util::glm_convert<V>(b), x));
```
petersteneteg, line 95: why not just to a > b ? mix(a, b, x) : mix(b, a, 1-x)?

include/inviwo/core/util/interpolation.h
```diff
@@ -84,7 +85,15 @@ inline T linearVectorInterpolation(const T &a, const T &b, P x) {
 
 template <typename T, typename P>
 inline T Interpolation<T, P>::linear(const T &a, const T &b, P x) {
-    return glm::mix(a, b, x);
+    using VT = typename util::value_type<T>::type;
+    if constexpr (std::is_same_v<T, VT> || std::is_signed_v<VT> ) {
```
petersteneteg, line 89: are we requiring 17 now?

petersteneteg (changes requested):

include/inviwo/core/util/interpolation.h
r-englund: will only work for scalars

include/inviwo/core/util/interpolation.h
r-englund: maybe we should

include/inviwo/core/util/interpolation.h
petersteneteg: ahh but this is fixed in GLM, why are you adding a workaround here and not updating glm!?!

petersteneteg: https://godbolt.org/z/HtF2HY

r-englund: yes, it works with the latest glm, but way to cumbersome to update the submodule since it is a fork.

r-englund:
rendered moot since it is fixed in later glm.

PR #542, opened 2019-06-12
Doxygen: GLSL filter fix for include guards

petersteneteg:
Fixes miss matched ifdefs in the doxygen for our shader code


docs/filter/glslfilter.py
```diff
@@ -86,60 +86,83 @@
...
 def writeLine(txt):
     sys.stdout.write(txt)
     
+def writeLines(r):
+    '''dump all lines to stdout'''
+    for s in r:
+        sys.stdout.write(s)
+
 ## parse a shader and generate needed information along on the way
 ## - if comments contain a namespace move it the classname
```
xeTaiz, line 128: Those comments would be nice as a docstring as well

xeTaiz (approved):

docs/filter/glslfilter.py
xeTaiz: for the `parseShader()` function I mean

PR #544, opened 2019-06-12
CImg: TIFF image resolution bug

martinfalk:
* libTIFF stores the image x/y resolution as float and not as rational using two int values


modules/cimg/src/cimgutils.cpp
```diff
@@ -606,20 +606,31 @@ TIFFHeader getTIFFHeader(const std::string& filename) {
...
+    }
+    if (!TIFFGetFieldDefaulted(tif, TIFFTAG_SAMPLEFORMAT, &sampleFormat)) {
+        sampleFormat = 1;
+    }
+
+    float xres = 1.0f;
+    float yres = 1.0f;
+    // X and Y resolution tags are stored as float
+    if (!TIFFGetFieldDefaulted(tif, TIFFTAG_XRESOLUTION, &xres)) {
+        xres = 1.0f;
```
ResearchDaniel, line 623: I guess using 0 would break stuff?

ResearchDaniel (approved):

modules/cimg/src/cimgutils.cpp
martinfalk: Yes, the "physical" extent or print size of an image is obtained by dividing the image width / height by the matching resolution (dpi or dots per cm).

martinfalk:
superseded by #546

PR #555, opened 2019-06-20
VectorVisGL: Added StreamParticle processor using Compute Shaders

r-englund:
canvas2-20190620_085814
canvas2-20190620_085827
canvas2-20190620_085842

### DIFF
--- a/modules/vectorfieldvisualizationgl/glsl/streamparticles.comp
+++ b/modules/vectorfieldvisualizationgl/glsl/streamparticles.comp
@@ -1,0 +1,89 @@
+uniform float stepLength = 0.01;
+uniform int internalSteps = 10;
+uniform sampler3D velocityField;
+
+uniform float minV;
+uniform float maxV;
+
+uniform float minR = 0.025;
+uniform float maxR = 0.035;
+
+uniform sampler2D tf;
+
+layout(std430, binding = 0) buffer posBuffer { vec4 pos[]; };
+layout(std430, binding = 1) buffer lifeBuffer { float life[]; };
+layout(std430, binding = 2) buffer radiBuffer { float radius[]; };
+layout(std430, binding = 3) buffer colBuffer { vec4 col[]; };
+
+vec3 norm(vec3 v) {
+    bool doNormalize = false;
+    if (doNormalize) {
+        float l2 = dot(v, v);
+        if (l2 == 0) return v;
+        return v / sqrt(l2);
+    } else {
+        return v;
+    }
+}
+
+vec4 getColor(float v) {
+    v = (v - minV) / (maxV - minV);
+    return texture(tf, vec2(v, 0.5));
+}
+
+uniform mat4 toWorldMatrix;
+uniform mat4 toTextureMatrix;
+vec3 toWorld(vec3 p) {
+    vec4 P = toWorldMatrix * vec4(p, 1);
+    return P.xyz / P.w;
+}
+
+vec3 toTexture(vec3 p) {
+    vec4 P = toTextureMatrix * vec4(p, 1);
+    return P.xyz / P.w;
+}
+
+bool inside(vec3 p) {
+    if (p.x < 0) return false;
+    if (p.y < 0) return false;
+    if (p.z < 0) return false;
+    if (p.x > 1) return false;
+    if (p.y > 1) return false;
+    if (p.z > 1) return false;
+    return true;
+}
+
+vec3 sample_(vec3 p) {
+    vec3 tex = toTexture(p);
+    if (!inside(tex)) return vec3(0);
+    return texture(velocityField, tex).xyz;
+}
+
+vec3 step(vec3 p) {
+    vec3 k1 = sample_(p);
+    vec3 k2 = sample_(p + norm(k1) * stepLength * 0.5);
+    vec3 k3 = sample_(p + norm(k2) * stepLength * 0.5);
+    vec3 k4 = sample_(p + norm(k3) * stepLength);
+    return p + norm((k1 + k2 + k2 + k3 + k3 + k4) / 6.0) * (stepLength);
+}
+
+layout(local_size_x = 1, local_size_y = 1, local_size_z = 1) in;
+void main() {
+    uint gid = gl_GlobalInvocationID.x;
+    vec3 p = pos[gid].xyz;
+
+    for (int i = 0; i < internalSteps; i++) {
+        p = step(p);
+    }
+
+    float l = length(sample_(p));
+    if (l < 0.001) {
+        life[gid] -= 0.1; // die in 10 frames if stale
+    }else{
+        life[gid] = 1;
+    }
+
+    col[gid] = getColor(l);
+    radius[gid] = mix(0, mix(minR, maxR, l / maxV), life[gid] > 0);
+    pos[gid] = vec4(p, 1);
+}
--- a/modules/vectorfieldvisualizationgl/glsl/streamparticles.comp
+++ b/modules/vectorfieldvisualizationgl/glsl/streamparticles.comp
@@ -1,3 +1,32 @@
+/*********************************************************************************
+ *
+ * Inviwo - Interactive Visualization Workshop
+ *
+ * Copyright (c) 2019 Inviwo Foundation
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ *********************************************************************************/
+
 uniform float stepLength = 0.01;
 uniform int internalSteps = 10;
 uniform sampler3D velocityField;
--- a/modules/vectorfieldvisualizationgl/glsl/streamparticles.comp
+++ b/modules/vectorfieldvisualizationgl/glsl/streamparticles.comp
@@ -29,7 +29,8 @@
  
 layout(local_size_x = 1, local_size_y = 1, local_size_z = 1) in;
 
-#include "utils/classification.glsl"
+#include "utils/classification.glsl" 
+#include "utils/advection.glsl"
 
 #ifndef NUM_ADVECTIONS
 #define NUM_ADVECTIONS 1
@@ -55,40 +56,31 @@ layout(std430, binding = 1) buffer lifeBuffer { float life[]; };
 layout(std430, binding = 2) buffer radiBuffer { float radius[]; };
 layout(std430, binding = 3) buffer colBuffer { vec4 col[]; };
 
-vec3 toTexture(vec3 p) {
-    vec4 P = toTextureMatrix * vec4(p, 1);
-    return P.xyz / P.w;
-}
 
 bool inside(vec3 texCoord) {
     if (any(lessThan(texCoord, vec3(0.0f)))) return false;
     if (any(greaterThan(texCoord, vec3(1.0f)))) return false;
     return true;
+} 
+
+vec3 toTexture(vec3 p) {
+    vec4 P = toTextureMatrix * vec4(p, 1);
+    return P.xyz / P.w;
 }
 
 vec3 sampleVelocity(vec3 worldPos) {
     vec3 tex = toTexture(worldPos);
     if (!inside(tex)) return vec3(0);
     return texture(velocityField, tex).xyz;
-}
+} 
 
-// Advect p one step usig 4th Order Runge-Kutta
-vec3 advect(vec3 p) {
-    const float h = (advectionSpeed * dt);
-    const float h2 = h / 2.0f;
-    vec3 k1 = sampleVelocity(p);
-    vec3 k2 = sampleVelocity(p + k1 * h2);
-    vec3 k3 = sampleVelocity(p + k2 * h2);
-    vec3 k4 = sampleVelocity(p + k3 * h);
-    return p + ((k1 + k2 + k2 + k3 + k3 + k4) / 6.0) * h;
-}
 
 void main() {
     uint gid = gl_GlobalInvocationID.x;
     vec3 p = pos[gid].xyz;
 
     for (int i = 0; i < NUM_ADVECTIONS; i++) { 
-        p = advect(p);
+        p = advectRK4(p, advectionSpeed * dt); 
     }
 
     float v = length(sampleVelocity(p));
--- a/modules/vectorfieldvisualizationgl/glsl/streamparticles.comp
+++ b/modules/vectorfieldvisualizationgl/glsl/streamparticles.comp
@@ -27,44 +27,31 @@
  *
  *********************************************************************************/
 
-uniform float stepLength = 0.01;
-uniform int internalSteps = 10;
-uniform sampler3D velocityField;
+layout(local_size_x = 1, local_size_y = 1, local_size_z = 1) in;
+
+uniform float dt;
+uniform float advectionSpeed; 
+uniform int advectionsPerFrame;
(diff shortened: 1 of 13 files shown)


modules/vectorfieldvisualizationgl/glsl/streamparticles.comp
```diff
@@ -0,0 +1,118 @@
...
+uniform sampler2D tf;
+
+layout(std430, binding = 0) buffer posBuffer { vec4 pos[]; };
+layout(std430, binding = 1) buffer lifeBuffer { float life[]; };
+layout(std430, binding = 2) buffer radiBuffer { float radius[]; };
+layout(std430, binding = 3) buffer colBuffer { vec4 col[]; };
+
+vec3 norm(vec3 v) {
+    bool doNormalize = false;
+    if (doNormalize) {
```
ResearchDaniel: Always false...

modules/vectorfieldvisualizationgl/glsl/streamparticles.comp
```diff
@@ -0,0 +1,118 @@
...
+    vec4 P = toWorldMatrix * vec4(p, 1);
+    return P.xyz / P.w;
+}
+
+vec3 toTexture(vec3 p) {
+    vec4 P = toTextureMatrix * vec4(p, 1);
+    return P.xyz / P.w;
+}
+
+bool inside(vec3 p) {
```
ResearchDaniel: This operation is working in a different coordinate space than the others. I think that we have implementations for bounding box inside elsewhere. Will check

modules/vectorfieldvisualizationgl/glsl/streamparticles.comp
```diff
@@ -0,0 +1,118 @@
...
+    if (p.y < 0) return false;
+    if (p.z < 0) return false;
+    if (p.x > 1) return false;
+    if (p.y > 1) return false;
+    if (p.z > 1) return false;
+    return true;
+}
+
+vec3 sample_(vec3 p) {
+    vec3 tex = toTexture(p);
```
ResearchDaniel: Why underscore?

modules/vectorfieldvisualizationgl/glsl/streamparticles.comp
```diff
@@ -0,0 +1,118 @@
...
+    if (!inside(tex)) return vec3(0);
+    return texture(velocityField, tex).xyz;
+}
+
+vec3 step(vec3 p) {
```
ResearchDaniel: Step? Should probably elaborate on this

modules/vectorfieldvisualizationgl/glsl/streamparticles.comp
```diff
@@ -0,0 +1,118 @@
...
+    uint gid = gl_GlobalInvocationID.x;
+    vec3 p = pos[gid].xyz;
+
+    for (int i = 0; i < internalSteps; i++) {
+        p = step(p);
+    }
+
+    float l = length(sample_(p));
+    if (l < 0.001) {
+        life[gid] -= 0.1; // die in 10 frames if stale
```
ResearchDaniel: This seems a bit arbitrary... 
usually life is based on duration since it was spawned?

modules/vectorfieldvisualizationgl/glsl/streamparticles.comp
```diff
@@ -0,0 +1,118 @@
...
+    }else{
+        life[gid] = 1;
+    }
+
+    col[gid] = getColor(l);
+    radius[gid] = mix(0, mix(minR, maxR, l / maxV), life[gid] > 0);
```
ResearchDaniel: Are there better ways of “killing” a particle? (Discarded through clipping or depth test?)

modules/vectorfieldvisualizationgl/include/modules/vectorfieldvisualizationgl/processors/3d/streamparticles.h
```diff
@@ -0,0 +1,136 @@
...
+#include <inviwo/core/util/clock.h>
+#include <inviwo/core/util/timer.h>
+
+#include <modules/vectorfieldvisualization/ports/seedpointsport.h>
+
+namespace inviwo {
+
+/** \docpage{org.inviwo.StreamParticles, Stream Particles}
+ * ![](org.inviwo.StreamParticles.png?classIdentifier=org.inviwo.StreamParticles)
+ * Explanation of how to use the processor.
```
ResearchDaniel: Add explanation

modules/vectorfieldvisualizationgl/include/modules/vectorfieldvisualizationgl/processors/3d/streamparticles.h
```diff
@@ -0,0 +1,136 @@
...
+ *   * __<Outport1>__ <description>.
+ *
+ * ### Properties
+ *   * __<Prop1>__ <description>.
+ *   * __<Prop2>__ <description>
+ */
+
+/**
+ * \brief VERY_BRIEFLY_DESCRIBE_THE_PROCESSOR
+ * DESCRIBE_THE_PROCESSOR_FROM_A_DEVELOPER_PERSPECTIVE
```
ResearchDaniel: Remove

modules/vectorfieldvisualizationgl/src/processors/3d/streamparticles.cpp
```diff
@@ -0,0 +1,223 @@
...
+
+#include <random>
+
+namespace inviwo {
+
+// The Class Identifier has to be globally unique. Use a reverse DNS naming scheme
+const ProcessorInfo StreamParticles::processorInfo_{
+    "org.inviwo.StreamParticles",  // Class identifier
+    "Stream Particles",            // Display name
+    "Particle",                    // Category
```
ResearchDaniel, line 51: Flow?

modules/vectorfieldvisualizationgl/src/processors/3d/streamparticles.cpp
```diff
@@ -0,0 +1,223 @@
...
+    if (mutex_.try_lock()) {
+        dispatchFront([this]() {
+            invalidate(InvalidationLevel::InvalidOutput);
+            mutex_.unlock();
+        });
+    }
+}
+
+void StreamParticles::initBuffers() {
+    if (!buffersDirty) return;
```
ResearchDaniel: I would put this logic outside of the function

modules/vectorfieldvisualizationgl/src/processors/3d/streamparticles.cpp
```diff
@@ -0,0 +1,223 @@
...
+
+        for (auto z : util::zip(positions, lifes)) {
+            if (get<1>(z) <= 0) {
+                get<0>(z) = seedTransform((*seeds)[dist(rand)]);
+                get<1>(z) = 1.0f;
+                count++;
+            }
+        }
+
+        LogInfo("Reseeded " << count << " particles");
```
ResearchDaniel: Remove

modules/vectorfieldvisualizationgl/include/modules/vectorfieldvisualizationgl/processors/3d/streamparticles.h
```diff
@@ -0,0 +1,136 @@
...
+
+    TemplateOptionProperty<SeedingSpace> seedingSpace_{
+        "seedingSpace",
+        "Seeding Space",
+        {{"data", "Data", SeedingSpace::Data}, {"world", "World", SeedingSpace::World}}};
+
+    FloatProperty stepLength_{"stepLength", "Step Length", 0.01f, 0.0f, 1.0f};
+    IntProperty internalSteps_{"internalSteps", "Internal Steps", 10, 1, 100};
+
+    FloatMinMaxProperty particleSize_{
```
ResearchDaniel: What is the size based on?

modules/vectorfieldvisualizationgl/include/modules/vectorfieldvisualizationgl/processors/3d/streamparticles.h
```diff
@@ -0,0 +1,136 @@
...
+ * ### Properties
+ *   * __<Prop1>__ <description>.
+ *   * __<Prop2>__ <description>
+ */
+
+/**
+ * \brief VERY_BRIEFLY_DESCRIBE_THE_PROCESSOR
+ * DESCRIBE_THE_PROCESSOR_FROM_A_DEVELOPER_PERSPECTIVE
+ */
+class IVW_MODULE_VECTORFIELDVISUALIZATIONGL_API StreamParticles : public Processor {
```
ResearchDaniel, line 84: Should there be one processor for seeding and one for advecting? Opens up for more seed methods and additional particle  attributes

ResearchDaniel (changes requested):

modules/vectorfieldvisualizationgl/glsl/streamparticles.comp
```diff
@@ -0,0 +1,118 @@
...
+    }
+}
+
+vec4 getColor(float v) {
+    v = (v - minV) / (maxV - minV);
+    return texture(tf, vec2(v, 0.5));
+}
+
+uniform mat4 toWorldMatrix;
+uniform mat4 toTextureMatrix;
```
martinfalk: These should be moved to the top, otherwise easy to overlook.

modules/vectorfieldvisualizationgl/glsl/streamparticles.comp
martinfalk: `if (any(p < vec3(0.0f)) || any(p > vec3(1.0f))) ... `

modules/vectorfieldvisualizationgl/glsl/streamparticles.comp
```diff
@@ -0,0 +1,118 @@
...
+
+vec3 step(vec3 p) {
+    vec3 k1 = sample_(p);
+    vec3 k2 = sample_(p + norm(k1) * stepLength * 0.5);
+    vec3 k3 = sample_(p + norm(k2) * stepLength * 0.5);
+    vec3 k4 = sample_(p + norm(k3) * stepLength);
+    return p + norm((k1 + k2 + k2 + k3 + k3 + k4) / 6.0) * (stepLength);
+}
+
+layout(local_size_x = 1, local_size_y = 1, local_size_z = 1) in;
```
martinfalk: move to top?

modules/vectorfieldvisualizationgl/glsl/streamparticles.comp
martinfalk: Life-time is a pretty good start. One problem with discarding particles wrt. depth or clipping is incoherent behavior when moving the camera. In particular since new particles are only inserted at the starting position?

modules/vectorfieldvisualizationgl/include/modules/vectorfieldvisualizationgl/processors/3d/streamparticles.h
```diff
@@ -0,0 +1,136 @@
...
+    IntProperty internalSteps_{"internalSteps", "Internal Steps", 10, 1, 100};
+
+    FloatMinMaxProperty particleSize_{
+        "particleSize", "Paricle Size (visual only)", 0.025f, 0.035f, 0.0f, 1.0f};
+
+    FloatProperty minV_{"minV",
+                        "Min velocity",
+                        0,
+                        0,
+                        std::numeric_limits<float>::max(),
```
martinfalk, line 126: is this clang-formatted?

modules/vectorfieldvisualizationgl/include/modules/vectorfieldvisualizationgl/processors/3d/streamparticles.h
```diff
@@ -0,0 +1,136 @@
...
+    TransferFunctionProperty tf_{"tf", "Velocity mapping"};
+
+    FloatProperty reseedInterval_{"reseedsInterval", "Reseed interval", 1.0f, 0.0f, 10.0f};
+
+    Shader shader_{{{ShaderType::Compute, "streamparticles.comp"}}};
+
+    Timer timer_{Timer::Milliseconds(17), [&]() { update(); }};
+
+    double reseedtime;
+    double prevT = 0;
```
martinfalk: underscores for `reseedtime`, `prevT`, and `c`?

martinfalk (changes requested):
Nice work. Some minor comments.

modules/vectorfieldvisualizationgl/glsl/streamparticles.comp
r-englund: sample is a reserved keyword

modules/vectorfieldvisualizationgl/glsl/streamparticles.comp
ResearchDaniel: I was thinking in terms of performance. I agree that the particles should be killed based on life but is setting size to zero the best way in terms of performance?

r-englund: Well, we can't move them outside of any clipping volumes since there is no info about viewing parameters in this shader. And, it is also possible that the output mesh is used in more than one view and then we have more than one clipping volume to take into consideration.

modules/vectorfieldvisualizationgl/include/modules/vectorfieldvisualizationgl/processors/3d/streamparticles.h
r-englund: No, I wanted to keep this processor as simple as possible. Kind of like an example processor for how compute shaders can be utilized. 
Also, now it has the same seeding strategy as the streamlines processor

modules/vectorfieldvisualizationgl/include/modules/vectorfieldvisualizationgl/processors/3d/streamparticles.h
r-englund: Depends on which processor is used to render them. If it is the Sphere Renderer it will be its radius while while the mesh renderer just ignores it and render it as GL_POINTS

modules/vectorfieldvisualizationgl/include/modules/vectorfieldvisualizationgl/processors/3d/streamparticles.h
r-englund: yes

modules/vectorfieldvisualizationgl/include/modules/vectorfieldvisualizationgl/processors/3d/streamparticles.h
ResearchDaniel: Would be better to have an explicit example processor in that case in my opinion. 
But do as you seem fit

modules/vectorfieldvisualizationgl/include/modules/vectorfieldvisualizationgl/processors/3d/streamparticles.h
ResearchDaniel: The radius will be determined by? Would probably be useful to choose the parameter which the radius is derived from. Same for color.

modules/vectorfieldvisualizationgl/src/processors/3d/streamparticles.cpp
r-englund: Doesn't have to be a flow field, any vector field. Might take a look on what catagories are used for the other processors in the vector vis modules, but that is for another PR

modules/vectorfieldvisualizationgl/glsl/streamparticles.comp
r-englund: using dt now

modules/vectorfieldvisualizationgl/include/modules/vectorfieldvisualizationgl/processors/3d/streamparticles.h
```diff
@@ -96,8 +117,7 @@ class IVW_MODULE_VECTORFIELDVISUALIZATIONGL_API StreamParticles : public Process
                                1,
                                InvalidationLevel::InvalidResources};
 
-    FloatMinMaxProperty particleSize_{
-        "particleSize", "Paricle radius (visual only)", 0.025f, 0.035f, 0.0f, 1.0f};
+    FloatMinMaxProperty particleSize_{"particleSize", "Paricle radius", 0.025f, 0.035f, 0.0f, 1.0f};
```
ResearchDaniel: Paricle -> Particle

modules/vectorfieldvisualizationgl/include/modules/vectorfieldvisualizationgl/processors/3d/streamparticles.h
```diff
@@ -48,23 +48,39 @@ namespace inviwo {
...
+ *   * __Advection Speed__ How fast the particles will move each advection.
+ *   * __Advections per Frame__ Can be used to increase advection speed without loss of
+ * precision.
+ *   * __Paricle radius__ Maps velocity to radius of particle (For rendering that
+ * supports the Radii buffer, e.g. Sphere Renderer).
+ *   * __Min velocity__ See bellow.
+ *   * __Max velocity__ Used together with __Min velocity__ to map velocity magnitude from [Min
+ * velocity, Max velocity] -> [0 1], used for mapping velocity to radius and color.
+ *   * __Velocity mapping__ Transferfunction to map a velocity to color.
+ *   * __Reseed interval__ Seconds between reseeding. When reseeding particles whoes life is zero
```
ResearchDaniel: whoes -> whose

modules/vectorfieldvisualizationgl/glsl/streamparticles.comp
```diff
@@ -89,16 +86,18 @@ void main() {
...
 
-    for (int i = 0; i < NUM_ADVECTIONS; i++) {
+    for (int i = 0; i < NUM_ADVECTIONS; i++) { 
         p = advect(p);
     }
 
-    float l = length(sampleVelocity(p));
-    if (l < 0.001) {
+    float v = length(sampleVelocity(p));
+    v = (v - minV) / (maxV - minV);
```
ResearchDaniel: might want to clamp [0 1] since the bounds are user specified and it might have side-effects for radius and color

ResearchDaniel (approved):

modules/vectorfieldvisualizationgl/glsl/streamparticles.comp
```diff
@@ -0,0 +1,96 @@
...
+uniform float minR;
+uniform float maxR;
+
+uniform sampler2D tf;
+
+uniform mat4 toTextureMatrix;
+
+layout(std430, binding = 0) buffer posBuffer { vec4 pos[]; };
+layout(std430, binding = 1) buffer lifeBuffer { float life[]; };
+layout(std430, binding = 2) buffer radiBuffer { float radius[]; };
```
martinfalk, line 56: typo "radii"

martinfalk (approved):

PR #561, opened 2019-06-25
Core,OpenGL: Volume swizzle mask

martinfalk:
* similar to color Layer swizzle masks

### DIFF
--- a/include/inviwo/core/datastructures/volume/volumeramprecision.h
+++ b/include/inviwo/core/datastructures/volume/volumeramprecision.h
@@ -46,8 +46,10 @@ class VolumeRAMPrecision : public VolumeRAM {
 public:
     using type = T;
 
-    VolumeRAMPrecision(size3_t dimensions = size3_t(128, 128, 128));
-    VolumeRAMPrecision(T* data, size3_t dimensions = size3_t(128, 128, 128));
+    VolumeRAMPrecision(size3_t dimensions = size3_t(128, 128, 128),
+                       const SwizzleMask& swizzleMask = swizzlemasks::rgba);
+    VolumeRAMPrecision(T* data, size3_t dimensions = size3_t(128, 128, 128),
+                       const SwizzleMask& swizzleMask = swizzlemasks::rgba);
     VolumeRAMPrecision(const VolumeRAMPrecision<T>& rhs);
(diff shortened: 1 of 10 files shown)


src/core/datastructures/volume/volume.cpp
```diff
@@ -160,6 +162,21 @@ vec3 Volume::getWorldSpaceGradientSpacing() const {
     return ds;
 }
 
+void Volume::setSwizzleMask(const SwizzleMask& mask) {
+    // update swizzle mask of all representations
+    for (auto rep : representations_) {
+        rep.second->setSwizzleMask(mask);
```
petersteneteg, line 168: Im not sure about this, 
seems strange to update it here and then also in the converter functions...

petersteneteg (changes requested):
I think the state should only be in the representation. And then the converters should take care of updating it... does that makes sense?

include/inviwo/core/datastructures/volume/volumeramprecision.h
```diff
@@ -233,6 +249,19 @@ void VolumeRAMPrecision<T>::setDimensions(size3_t dimensions) {
...
+}
+
+template <typename T>
+SwizzleMask VolumeRAMPrecision<T>::getSwizzleMask() const {
+    return swizzleMask_;
+}
+
 template <typename T>
 double VolumeRAMPrecision<T>::getAsDouble(const size3_t& pos) const {
     return util::glm_convert<double>(data_[posToIndex(pos, dimensions_)]);
```
ResearchDaniel, line 267: Should not the swizzle-mask affect all data-getters?

ResearchDaniel (changes requested):
Would this PR also not mean that the methods using the RAM rep would need to be updated?

include/inviwo/core/datastructures/volume/volumeramprecision.h
r-englund: The LayerRAMPrecision does not use the swizzeling, maybe it should, but I think it would be another PR since that has a quite high impact of existing methods. Also the Volume and Image sampler should consider swizzeling

martinfalk: Agree with @r-englund. That should definitely be implemented. But potentially part of different PR.

martinfalk:
rebased plus additional changes to various representations in PR #568

PR #565, opened 2019-06-26
Docygen: fixes and updates

petersteneteg:


include/inviwo/core/util/clock.h
```diff
@@ -187,7 +187,7 @@ void ScopedClock<Clock>::printAndReset() {
 }
 
 /**
- * \class ScopedClockCPU
+ * \class inviwo::ScopedClockCPU
```
martinfalk, line 190: hm, maybe this should be mentioned in the developer docs that the namespace is required here @xeTaiz

martinfalk (approved):

PR #572, opened 2019-06-27
Base: removed c++17 dep...

petersteneteg:


r-englund:
I can't spot where the change is

r-englund:
> 
> 
> I can't spot where the change is

Found it. its the `std::is_same_v` -> `std::is_same<>::value`

r-englund (approved):
The only change I can see is `std::is_same_v` -> `std::is_same<>::value`. If thats it, then it is approved :)

petersteneteg:
> The only change I can see is `std::is_same_v` -> `std::is_same<>::value`. If thats it, then it is approved :)

Yes, clang format did a big change... sorry about that

PR #578, opened 2019-06-28
Feature/add port unique ptr return ref

r-englund:
Made the addPort(std::unique_ptr<Inport|Outport>) return a reference, now it can be used as addPort(T&)

### DIFF
--- a/include/inviwo/core/processors/processor.h
+++ b/include/inviwo/core/processors/processor.h
@@ -336,11 +336,23 @@ public:
     virtual void serialize(Serializer& s) const override;
     virtual void deserialize(Deserializer& d) override;
 
-    // Assume ownership of port, needed for dynamic ports
+    /**
+     * Add port to processor and pass ownership to processor.
+     * @note Port group is a concept for event propagation. Currently only used for
+     * ResizeEvents, which only propagate from outports to inports in the same port group
+     * @param port to add
+     * @param portGroup name of group to propagate events through (defaults to "default")
+     */
     template <typename T, typename std::enable_if_t<std::is_base_of<Inport, T>::value, int> = 0>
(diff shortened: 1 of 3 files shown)


include/inviwo/core/processors/processor.h
```diff
@@ -397,6 +395,39 @@ class IVW_CORE_API Processor : public PropertyOwner,
...
+}
+
+/**
+ * Add port to processor.
+ * @note Port group is a concept for event propagation. Currently only used for
+ * ResizeEvents, which only propagate from outports to inports in the same port group
+ * @param port to add
+ * @param portGroup name of group to propagate events through (defaults to "default")
+ */
+template <typename T>
```
ResearchDaniel, line 426: Documentation should be moved to definition

include/inviwo/core/processors/processor.h
```diff
@@ -336,6 +336,14 @@ class IVW_CORE_API Processor : public PropertyOwner,
     virtual void serialize(Serializer& s) const override;
     virtual void deserialize(Deserializer& d) override;
 
+    // Assume ownership of port, needed for dynamic ports
+    template <typename T, typename std::enable_if_t<std::is_base_of<Inport, T>::value, int> = 0>
```
ResearchDaniel, line 346: Could add essentially the same documentation as the other addPort functions

ResearchDaniel (approved):
Some documentation could be moved/added. Looks good 👍

include/inviwo/core/processors/processor.h
r-englund: It was duplicated documentation, forgot to remove it after moving the impl out of the class declaration

PR #579, opened 2019-07-01
stop trackball from consuming mouse press events on rotate/zoom. clos…

### LINKED ISSUE #574: Trackball interferes with Canvas context menu
If the network contains a trackball property which handles interaction events, the context menu in a canvas (right mouse button) will no longer show up.

xeTaiz:
Fixes #574


martinfalk (approved):

martinfalk:
Did you test this?

xeTaiz:
I checked it briefly and it worked fine. I also compared how the right click was implemented before I changed this and that was the same way it is now.

Issue #580, opened 2019-07-02
Module name with initial s


petersteneteg:
https://github.com/inviwo/inviwo/commit/e0b12c5443e71fa59577096950fd78d30a729833#r34155242

r-englund:
Adding the comment referenced above for ease of reading: 

> I think I may have encountered a problem which I doesn't seem to have been reported. When making a new module ( through Inviwo ) with a name beginning by 's' or several 's's the cmake doesn't register it under the right name and absorbs the 's's. That causes conflicts as the directory will have the proper name ('spline' for example) but the module will not ('pline'). The CMakeList file also has the proper line: ivw_module(spline) I have tried with a bunch other letters at the beginning and they seem to be working ok. For now I have resorted to naming the module bspline (which is the name of the specific type of spline used).

r-englund:
I\ll bet there is a \s somewhere being interpreted as a whitespace

r-englund:
I think we might have found a bug in cmake, in the function `ivw_private_get_ivw_module_name` in globalutils.cmake we have a regexpt mathcing whitespaces using \\s, it seems that one is also catching the regular s as a whitespace.

PR #582, opened 2019-07-02
CMake: Fix in ivw_private_get_ivw_module_name for cmake incorrectly treating s as whitespace

r-englund:
See #580 

Seems cmake is treating a regular s in regexp mathcing as whitespace when using \\s for searching. changing \\s in to just a regular whitespace or putting the whitespace into brackets [ ] fixes the issue.


r-englund:
Also, there is no need for checking for whitespace before or after the thing we looking for, cmake manage that anyway

r-englund:
the use of `"[ \t\r\n]"` to find whitespaces is taken from https://cmake.org/cmake/help/latest/command/string.html 
> The quoted argument "[ \t\r\n]" specifies a regex that matches any single whitespace character.

petersteneteg (approved):

r-englund:
@petersteneteg I added 1 commit after you approved. Now it checks the whole contents instead of line by line.  Still approved?
Now the regex would correctly parse this example: 
bild

petersteneteg:
The only issue might be if one has two ivw_module and the first has a # in front... but we didn't fix that before either...

petersteneteg:
In principle ivw_module should be at the beginning of the line except for space... but then one could put in inside a if or something so... it's still not fool proof...

r-englund:
The check if it is in a comment and/or if-statement is too tricky to get into a single regex and would require some rework of the function so it out of scope for this PR, this at least fixes #580

PR #592, opened 2019-07-03
Feature/plotting

martinfalk:
* new `AxisStyleProperty` to adjust multiple axis properties at the same time (line/text color, font size, tick length). 
* new processor `ImagePlot` for plotting a 2D image with matching x-y axes. Interaction events are propagated, i.e. it is possible to browse through volume slices (see example)

image


modules/plotting/src/properties/axisstyleproperty.cpp
```diff
@@ -0,0 +1,140 @@
...
+        // use the file name w/o extension as identifier
+        fontFace_.addOption(name, font.first, font.second);
+    }
+    fontFace_.setSelectedIdentifier("Montserrat-Medium");
+    fontFace_.setCurrentStateAsDefault();
+    fontSize_.setSemantics(PropertySemantics("Fontsize"));
+    color_.setSemantics(PropertySemantics::Color);
+
+    fontFace_.onChange([this]() {
+        for (auto a : axes_) {
```
ResearchDaniel, line 69: might want to introduce a network lock for these?

ResearchDaniel (approved):
Sweet!!!
I can imagine that having image(s) in the regular plot should be possible too, basically the merging of ImagePlotProcessor and scatter plot. Would it be possible to reuse this code to do something like that?

petersteneteg (commented):
Some comments

petersteneteg (approved):

modules/plotting/src/properties/axisstyleproperty.cpp
martinfalk: done

martinfalk:
rebased into #599

PR #595, opened 2019-07-04
Feature/property preset manager displayname fix

### LINKED ISSUE #585: PropertyPresetManager overrides the displayname
When setting a property using a preset the display name gets set to the display name to the was set to the property when the preset was created. 
This is especially bad for Workspace and Application level presets. 

**Steps to reproduce**
1. Add a background processor
2. Create a workspace or application preset on `Color 1`
3. Set `Color 2` using the preset from `Color 1`

Edit: Added steps to reproduce

r-englund:
fixes #585

### DIFF
--- a/include/inviwo/core/util/observer.h
+++ b/include/inviwo/core/util/observer.h
@@ -278,7 +278,9 @@ void Observable<T>::startBlockingNotifications() {
 }
 template <typename T>
 void Observable<T>::stopBlockingNotifications() {
-    --notificationsBlocked_;
+    if (notificationsBlocked_ != 0) {
+        --notificationsBlocked_;
+    }
 }
 
 template <typename T>
--- a/src/core/properties/propertypresetmanager.cpp
+++ b/src/core/properties/propertypresetmanager.cpp
@@ -55,6 +55,9 @@ bool PropertyPresetManager::loadPreset(const std::string& name, Property* proper
         // We deserialize into a clone here and link it to the original to only set value not
         // identifiers and such.
         auto temp = std::unique_ptr<Property>(p->clone());
+        // Since Observers are copied when cloning we need to blockNotifications to prevent change
+        // of displayname of widgets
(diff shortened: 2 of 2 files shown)


ResearchDaniel (changes requested):
I think that a better approach would be to remove the observers from the property presets. Then this workaround would not be necessary

include/inviwo/core/util/observer.h
```diff
@@ -278,7 +278,9 @@ void Observable<T>::startBlockingNotifications() {
 }
 template <typename T>
 void Observable<T>::stopBlockingNotifications() {
-    --notificationsBlocked_;
+    if (notificationsBlocked_ != 0) {
```
petersteneteg, line 281: this should never happen if it does that is an error. there should not be an if here. In anything an assert.

r-englund: I called it by mistake instead of `startBlockingNotifications` and it "worked as expected" (i.e. the notifications stopped coming). 
I agree we shouldn't call it and use the `NoficationBlocker` instead, I just did see the `NoficationBlocker` since I just glanced over the interface of `Observer<T>` and saw `stop....()`

petersteneteg: so if you would have had that if there, then you would never have noticed your mistake...

r-englund: > 
> 
> so if you would have had that if there, then you would never have noticed your mistake...

Well, I noticed later when making the other commit that I had used the wrong function. If the if-statement would have been there my mistake would have been spotted earlier since the notifications wouldn't be blocked and my original issue would still have been there

r-englund: Anyway, what about making the startBlocking/StopBlocking methods private and only allow blocking using the `NoficationBlocker `?

petersteneteg: I would use an assert

Issue #238, opened 2018-09-14
Sigar on Arch prevents Build


Sparkier:
**Issue**

Inviwo fails to build on Arch because of sigar.
Errors are the following:
```bash
inviwo/ext/sigar/src/os/linux/linux_sigar.c:1149:22: 
error: called object ‘major’ is not a function or function pointer 
#define ST_MAJOR(sb) major((sb).st_rdev)

inviwo/ext/sigar/src/os/linux/linux_sigar.c:1150:22: 
error: called object ‘minor’ is not a function or function pointer 
#define ST_MINOR(sb) minor((sb).st_rdev)
```
along with other occurences in other lines of the same file.

**Steps to reproduce**

```
git clone https://github.com/inviwo/inviwo.git
git submodule update --init --recursive
mkdir build && cd build
cmake ..
make inviwo
```

**Environment information**
* Operating System: Arch Linux
* Compiler/IDE: gcc
* QT Version: 5.11.1
* Python version: 3.7.0

petersteneteg:
:( there should be a cmake option to disable sigar I think... 
unfortunately sigar is very old and not updated anymore...

ResearchDaniel:
Disable IVW_SIGAR
Could we detect the linux version and disable it by default?

petersteneteg:
Maybe we should just get rid of it...

petersteneteg:
Think this should be fixed by #601

PR #601, opened 2019-07-05
fix sigar not building on linux

xeTaiz:
So, as #238 describes, inviwo does not build sigar on linux (or at least on arch).
After some investigation I found that a few small changes can fix that.
First of all the `sys/sysmacros.h` include is needed to get rid of the initial error.
This then leads to `rpc/rpc.h` includes that are not found, because they are located in `/usr/include/tirpc/rpc/` instead of `/usr/include/rpc/` directly.
I added the include to the sigar's `CMakeLists.txt` and it works perfectly fine now.

The only change that should possibly affect other OS is the `sys/sysmacros.h` include, which I didn't test on windows or mac yet.

Also I'm not sure about the cmake stuff, please take a look at it @petersteneteg.

### DIFF
--- a/cmake/modules/FindTIRPC.cmake
+++ b/cmake/modules/FindTIRPC.cmake
@@ -1,6 +1,8 @@
 # FindTIRPC
 # ---------
 #
+# From https://raw.githubusercontent.com/snort3/snort3/master/cmake/FindTIRPC.cmake
+#
 # Find the native TIRPC includes and library.
 #
 # Result Variables
--- a/cmake/modules/FindTIRPC.cmake
+++ b/cmake/modules/FindTIRPC.cmake
@@ -1,0 +1,44 @@
+# FindTIRPC
+# ---------
+#
+# Find the native TIRPC includes and library.
+#
+# Result Variables
+# ^^^^^^^^^^^^^^^^
+#
+# This module will set the following variables in your project:
+#
+# ``TIRPC_INCLUDE_DIRS``
+#   where to find rpc.h, etc.
+# ``TIRPC_LIBRARIES``
+#   the libraries to link against to use TIRPC.
+# ``TIRPC_VERSION``
+#   the version of TIRPC found.
+# ``TIRPC_FOUND``
+#   true if the TIRPC headers and libraries were found.
+#
+
+find_package(PkgConfig QUIET)
+pkg_check_modules(PC_TIRPC REQUIRED libtirpc)
+
+find_path(TIRPC_INCLUDE_DIRS
+    NAMES netconfig.h
+    PATH_SUFFIXES tirpc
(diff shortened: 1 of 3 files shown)


ResearchDaniel (approved):

r-englund:
> The only change that should possibly affect other OS is the sys/sysmacros.h include, which I didn't test on windows or mac yet.

Since the <sys/sysmacros.h>include was added to linux_sigar.c it should not affect windows or mac builds :)

ext/sigar/CMakeLists.txt
```diff
@@ -9,28 +9,29 @@ project(sigar)
...
+    include/sigar_log.h
+    include/sigar_private.h
+    include/sigar_ptql.h
     include/sigar_util.h
 )
 
 ## linux
 if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
     set(SIGAR_SRC src/os/linux/linux_sigar.c)
+    find_package(TIRPC REQUIRED)
```
r-englund: TIRPC, is that something that all Unix dist comes with or could this potentially introduce new CMake problems on some Unix system?

cmake/modules/FindTIRPC.cmake
```diff
@@ -0,0 +1,46 @@
+# FindTIRPC
```
petersteneteg, line 1: Should this not be in the sigar package?

ext/sigar/CMakeLists.txt
```diff
@@ -9,28 +9,29 @@ project(sigar)
 
 #--------------------------------------------------------------------
 # Define defintions
-if(WIN32) 
-    ## make sure we only use the smallest set of 
-    ## headers on win32. Otherwise we get clashes 
+if(WIN32)
```
petersteneteg, line 12: Since this is an external dependency I would prefer if we kept the changes to a minimum, i.e. don't push any formatting changes...

cmake/modules/FindTIRPC.cmake
xeTaiz: I was unsure. I put it into the folder with the other FindX.cmake files. Shall I move it to `ext/sigar/`?

xeTaiz:
> TIRPC, is that something that all Unix dist comes with or could this potentially introduce new CMake problems on some Unix system?

Apparently RPC was part of glibc until version 2.26, from which on it is removed. My last commit thus checks for `rpc/rpc.h` and tries to find libtircp if unsuccessful

ext/sigar/CMakeLists.txt
xeTaiz: sigar did not see an update for over 4 years and our CMakeLists already kind of replaced every line from the original. I first thought I should fork sigar, apply my changes and switch the submodule to my repo, but I then found it is actually right in our repo, not a submodule,  and already differs significantly.

PR #604, opened 2019-07-08
Webbrowser: Fixed issue with web browser rendering not being updated …

ResearchDaniel:
…according to latest changes. Was necessary to activate the local render context. Made some minor fixes for invalidation of pop-ups (dropdown) following the cef-client offscreen rendering examples.

Puh, this was a hard one to track down!


martinfalk (approved):
nice one :-)

PR #606, opened 2019-07-08
Feature/freetype update

martinfalk:
* freetype is now a submodule
* updated freetype to latest version 2.10.1

## Font rendering in Visual Studio build before (left) and after (right)
image


modules/fontrendering/ext/freetype/CMakeLists.txt
```diff
@@ -1,79 +1,17 @@
-project(freetype)
+set(BUILD_SHARED_LIBS ON)
```
r-englund, line 1: ?? What if it we set it to off in the cmake-gui?

r-englund:
There were no need to do any changes in Inviwo code?

modules/fontrendering/ext/freetype/CMakeLists.txt
martinfalk: Then we have a problem since `freetype.lib` is not created. It is exposed by the freetype CMakeLists.txt anyway... Not the nicest thing, I know.

martinfalk:
Nope, no changes necessary. Only minor adjustments in the CMake stuff.

modules/fontrendering/ext/freetype/CMakeLists.txt
petersteneteg: We also want to build statically so this needs fixing

martinfalk: Tested it again on a clean build. This line can be removed.

martinfalk:
rebased into #607

Issue #608, opened 2019-07-09
PVM reader is broken


martinfalk:
The PVM reader in its current state is severely broken. Internally, it uses C with lots of `malloc` and `realloc` which are not properly matched with `free` in our code.

In case of an error the application exits! For example:
```c++
if (numc<1) ERRORMSG()
```
with `ERRORMSG` being
```c++
#define ERRORMSG() errormsg(__FILE__,__LINE__)

inline void errormsg(const char *file, int line) {
    fprintf(stderr, "fatal error in <%s> at line %d!\n", file, line);
    exit(EXIT_FAILURE);
}
``` 

This needs to be fixed!

related issues #480, #567

PR #610, opened 2019-07-10
PVM: volume reader fixes, #608

martinfalk:
* minor clean up in PVMVolumeReader
* tidds is always linked dynamically due to LGPL


ResearchDaniel (approved):
Looks good as far as I can see

Issue #594, opened 2019-07-04
Copy Constructor of Observer copies the observers of the orignal


r-englund:
The Copy Constructor of Observer copies the observers of the orignal, should it?
https://github.com/inviwo/inviwo/blob/master/src/core/util/observer.cpp#L34-L36
```c++
Observer::Observer(const Observer& rhs) {
    for (auto observable : rhs.observables_) addObservation(observable);
}
```
casuses issue in the `PropertyPresetManager` #585

petersteneteg:
Tread very carefully here, this has lot and lots of implications many of them which is very hard to detect...

r-englund:
![image](https://user-images.githubusercontent.com/534670/60646190-03788580-9e3b-11e9-8931-d81927671163.png)

ResearchDaniel:
Short answer: Yes it should and it is not a bug :)

ResearchDaniel:
I don't exactly remember where it was needed but I guess you can try to remove it and see if anything breaks.

I'm not sure if it is logical that a copy does not observe the same thing though.

PR #613, opened 2019-07-11
Feature/camera fitter

r-englund:


modules/base/include/modules/base/algorithm/camerautils.h
```diff
@@ -0,0 +1,195 @@
...
+ * Setup the camera parameters such that the whole bounding box spaned by basisAndOffset will be
+ * inside the view frustum.
+ *
+ * @param cam the Camera to update
+ * @param basisAndOffset the basis and offset of the bounding box that will fit inside the new view
+ * frustum
+ * @param side Which side of the bounding box that will face the camera
+ * @param fitRatio determines how much spacing will be between edge of view and edge of volume. A
+ * fit ratio of 1 means a perfect fit, not edge between view frustum and volume border (takes aspect
+ * ratio into acount)
```
martinfalk: "determines the spacing between volume and the boundaries of the view frustum. A fit ratio of 1 means a perfect fit, no space between frustum and volume. The aspect ratio is taken into account."

modules/base/include/modules/base/algorithm/camerautils.h
```diff
@@ -0,0 +1,195 @@
...
+ * @param setNearFar set to true to also updated the cameras new/far clip ranges @see
+ * setCameraNearFar
+ * @param setLookRanges set to true to also updated the min/max ranges of the cameras
+ * lookTo/lookFrom ranges @see setCameraLookRanges
```
martinfalk: if true, the camera's look-to/look-from ranges are updated @see ...

modules/base/include/modules/base/algorithm/camerautils.h
```diff
@@ -0,0 +1,195 @@
...
+ */
+IVW_MODULE_BASE_API void setCameraView(CameraProperty &cam, const mat4 &basisAndOffset, Side side,
+                                       float fitRatio = 1.05f, bool setNearFar = false,
+                                       bool setLookRanges = false);
+
+/**
+ * Set the ranges of the look to and look from properties of the camera. Will center around the mid
+ * point of dataToWorld. The lookTo will be set to ranges to stay within the volume. The lookFrom
+ * will be set to the mid point +- the basis vector of dataToWorld times the zoom factor. i.e a zoom
+ * factor of 25 will alow to zoom out to a distance of "25 volumes".
```
martinfalk: "The lookFrom
 * will be set to the mid point +- the basis vector of dataToWorld times the zoom factor. That is a zoom
 * factor of 25 will allow to zoom out to a distance of "25 volumes".

modules/base/src/processors/fitcameratodata.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+#include <modules/base/processors/fitcameratodata.h>
+
+namespace inviwo {
+
+// The Class Identifier has to be globally unique. Use a reverse DNS naming scheme
+const ProcessorInfo FitCameraToData::processorInfo_{
+    "org.inviwo.FitCameraToData",  // Class identifier
+    "Fit Camera To Data",          // Display name
+    "Utility",                     // Category
```
martinfalk: yet another group? ;-)

modules/base/src/processors/fitcameratodata.cpp
```diff
@@ -0,0 +1,107 @@
...
+    CodeState::Stable,             // Code state
+    Tags::CPU,                     // Tags
```
martinfalk, line 40: Maybe "CPU, Camera" ?

martinfalk (approved):

modules/base/include/modules/base/algorithm/camerautils.h
```diff
@@ -0,0 +1,190 @@
...
+ * @param basisAndOffset the basis and offset of the bounding box that will fit inside the new view
+ * frustum
+ * @param side       this side of the bounding box will be facing the camera afterward
+ * @param fitRatio   determines the spacing between volume and the boundaries of the view frustum. A 
+ *                   fit ratio of 1 means a perfect fit, no space between frustum and volume. The aspect 
+ *                   ratio is taken into account.
+ * @param setNearFar the camera's new/far clip ranges are updated if true @see setCameraNearFar
+ * @param setLookRanges   the camera's look-to/look-from ranges are updated if true @see setCameraLookRanges
+ */
+IVW_MODULE_BASE_API void setCameraView(CameraProperty &cam, const mat4 &basisAndOffset, Side side,
```
ResearchDaniel: Suggested API changes:
- Replace side enum with a view vector in world space. The enum could encode view vectors, e.g. Side::XNegative = (-1,0,0).
- Take the bounding geometry + transformation matrix as input, then we can use it for general meshes as well

modules/base/include/modules/base/algorithm/camerautils.h
```diff
@@ -0,0 +1,190 @@
...
+IVW_MODULE_BASE_API std::pair<float, float> computeCameraNearFar(
+    const mat4 &dataToWorld, float maxZoomFactor = 25.f, float nearFarRatio = 1.f / 10000.f);
+
+/**
+ * Sets the near and far clip distances of the camera based on the given bounding volume and zoom
+ * factor. Ensures that the far plane is distant enough to avoid clipping given to current zoomfactor.
+ * @see computeCameraNearFar
+ * @see setCameraLookRanges
+ */
+IVW_MODULE_BASE_API void setCameraNearFar(CameraProperty &cam, const mat4 &dataToWorld,
```
ResearchDaniel: Roughly equivalent to meshutil::computeNearFarPlane? 
Perhaps we should move that one to this file.

modules/base/include/modules/base/algorithm/camerautils.h
```diff
@@ -0,0 +1,190 @@
...
+
+private:
+    CompositeProperty composite_;
+    CompositeProperty lookAt_;
+    CompositeProperty lookAtSettings_;
+    ButtonProperty flipUp_;
+    BoolProperty updateNearFar_;
+    BoolProperty updateLookRanges_;
+    FloatProperty fittingRatio_;
+    ButtonProperty xNegative_;
```
ResearchDaniel, line 144: Suggest to replace the buttons with an option property with the enum options along with a synced vec3 property for camera direction

modules/base/src/algorithm/camerautils.cpp
```diff
@@ -0,0 +1,210 @@
...
+
+    vec3 lookTo(dataToWorld * vec4(vec3(.5f), 1.f));
+
+    vec3 dx(dataToWorld[0]);
+    vec3 dy(dataToWorld[1]);
+    vec3 dz(dataToWorld[2]);
+
+    auto p0 = lookTo + (dx + dy + dz) * zoomRange;
+    auto p1 = lookTo - (dx + dy + dz) * zoomRange;
+    cam.lookFrom_.setMinValue(glm::min(p0, p1));
```
ResearchDaniel, line 152: Use NetworkLock
Also, why not cam.lookFrom_.set(....)?

modules/base/src/algorithm/camerautils.cpp
```diff
@@ -0,0 +1,210 @@
...
+    cam.lookFrom_.setMaxValue(glm::max(p0, p1));
+    p0 = lookTo + (dx + dy + dz);
+    p1 = lookTo - (dx + dy + dz);
+    cam.lookTo_.setMinValue(glm::min(p0, p1));
+    cam.lookTo_.setMaxValue(glm::max(p0, p1));
+}
+
+std::pair<float, float> computeCameraNearFar(const mat4 &dataToWorld, float zoomRange,
```
ResearchDaniel: How does this compare to meshutil:computeNearFarPlanes?
We should probably only have one of them

modules/base/src/processors/fitcameratodata.cpp
ResearchDaniel: Camera would be better?

src/core/properties/propertypresetmanager.cpp
```diff
@@ -55,6 +55,9 @@ bool PropertyPresetManager::loadPreset(const std::string& name, Property* proper
         // We deserialize into a clone here and link it to the original to only set value not
         // identifiers and such.
         auto temp = std::unique_ptr<Property>(p->clone());
+        // Since Observers are copied when cloning we need to blockNotifications to prevent change
```
ResearchDaniel, line 58: Not part of this PR?
We should remove the observers instead?

ResearchDaniel (changes requested):
Really nice features! I think that the API can be generalized:
Suggested API changes:

- Replace side enum with a view vector in world space. The enum could encode view vectors, e.g. Side::XNegative = (-1,0,0). This seem to be used later anyway :)
- Take the bounding geometry + transformation matrix as input. Now it implicitly rely on bounding box [0 1]^3 in texture space?

src/core/properties/propertypresetmanager.cpp
r-englund: I wanted to get this fix in, was too lazy to create a separate PR for it (it was already in my local branch). I do not want to touch the observer interface.

modules/base/src/algorithm/camerautils.cpp
r-englund: I found that one after implementing this one. That one now calls this one.

modules/base/include/modules/base/algorithm/camerautils.h
r-englund: No. I want the buttons to click on. Would just be very annoying having the change the option property all the time when trying to find a nice view.

modules/base/include/modules/base/algorithm/camerautils.h
r-englund: I will not move anything old methods, that might break stuff for other ppl.

modules/base/include/modules/base/algorithm/camerautils.h
r-englund: > Replace side enum with a view vector in world space. The enum could encode view vectors, e.g. 
> Side::XNegative = (-1,0,0).

The fitting will probably not work for arbitrary vector since it works "with the side of the box face the camera". So not going to do that

> Take the bounding geometry + transformation matrix as input, then we can use it for general meshes as well 

That is encoded in the matrix

ResearchDaniel: I don’t think that it can be encoded in a matrix. The matrix assumes that the coordinates lie in texture space between [0 1] and that 0 correspond to the min point and 1 to the max from what I can understand?

r-englund: > I don’t think that it can be encoded in a matrix. The matrix assumes that the coordinates lie in texture space between [0 1] and that 0 correspond to the min point and 1 to the max from what I can understand? 

This function works for meshes aswell, see the helper class bellow.  We could consider adding an overload to this function taking a mesh/list of meshes, feel free to do that in a future PR

modules/base/src/algorithm/camerautils.cpp
ResearchDaniel: The new code is not equivalent, it does not consider offset. Have you tested this with object that are not centered around the world origin?

modules/base/include/modules/base/algorithm/camerautils.h
martinfalk: I would argue against generalization and for the use of the enum with the six sides.

modules/base/src/algorithm/camerautils.cpp
r-englund: near and far plane does not need to consider offset, they are in view space

modules/base/include/modules/base/algorithm/camerautils.h
ResearchDaniel: I think that combining your aspect ratio code with the previous near-far computation would make it work for arbitrary geometry.

modules/base/src/algorithm/camerautils.cpp
r-englund: yes, I have tested this with many volumes, both centered and not centered, with regular basis and rotational basis.

modules/base/include/modules/base/algorithm/camerautils.h
martinfalk: Buttons are more intuitive. Similar, assume the option property is set to "X+", then the camera position is changed. What do you do to reset the camera again to "X+"? With the button it's a single click, using the option property you need to switch to a different view direction first(?).

martinfalk (approved):

modules/base/include/modules/base/algorithm/camerautils.h
ResearchDaniel: I guess it would in its current state need an apply button as well. Less UI space consumed but more interaction. I guess it is more a matter of taste...

modules/base/src/algorithm/camerautils.cpp
ResearchDaniel: Hmm, that is weird. If an object is far away from the camera it is necessary to have a large far distance so it is a combination of camera and object placement. I have some nifty volumes with large offsets that we might be able to try it on and where it failed before I made the other implementation

src/core/properties/propertypresetmanager.cpp
ResearchDaniel: But blocking notifications won’t solve the core issue?

modules/base/include/modules/base/algorithm/camerautils.h
martinfalk: Maybe I'm spoiled by other software ;-)

include/inviwo/core/ports/datainport.h
```diff
@@ -233,17 +233,25 @@ std::shared_ptr<const T> DataInport<T, N, Flat>::getData() const {
 
 template <typename T, size_t N, bool Flat>
 std::vector<std::shared_ptr<const T>> DataInport<T, N, Flat>::getVectorData() const {
-    std::vector<std::shared_ptr<const T>> res(N);
+    std::vector<std::shared_ptr<const T>> res;
+    if constexpr (N > 0) {
```
petersteneteg, line 237: Why?

modules/base/include/modules/base/algorithm/camerautils.h
ResearchDaniel: A different property semantic with condensed button layout for option property had probably been optimal. Or a better packed property group

petersteneteg: A row of icons it whats usually use for this kind of think... guess we could add a RowOfIconsProperty. 
On the other hand neither that or our Button Property are actually Properties, since a property by definition has some state which the button lacks. Could we do that in some better way?

modules/base/include/modules/base/algorithm/camerautils.h
```diff
@@ -0,0 +1,190 @@
...
+void FitCameraPropertiesHelper::init(CameraProperty &camera, Callback getToWorld, Port &port) {
+
+    auto set = [getToWorld, p = &port, cam = &camera, this](auto side) {
+        if (p->hasData()) {
+            setCameraView(*cam, getToWorld(), side, fittingRatio_.get(), updateNearFar_.get(),
+                          updateLookRanges_.get());
+        }
+    };
+
+    xNegative_.onChange([set] { set(Side::XNegative); });
```
petersteneteg, line 184: We should add a constructor argument for this

modules/base/src/algorithm/camerautils.cpp
```diff
@@ -0,0 +1,210 @@
...
+}
+
+void setCameraLookRanges(CameraProperty &cam, const mat4 &dataToWorld, float zoomRange) {
+    NetworkLock lock(&cam);
+
+    vec3 lookTo(dataToWorld * vec4(vec3(.5f), 1.f));
+
+    vec3 dx(dataToWorld[0]);
+    vec3 dy(dataToWorld[1]);
+    vec3 dz(dataToWorld[2]);
```
petersteneteg: Think diagonal = vec3(dataToworld * vec4(1,1,1,0)) might be clearer then dx, dy, dz

modules/base/include/modules/base/algorithm/camerautils.h
r-englund: I've added overloads for meshes now

include/inviwo/core/ports/datainport.h
r-englund: `std::vector<std::shared_ptr<const T>> res(N); `  constructs and vector of size N with empty shared_ptr, in the lines bellow we use push_back to add the data, hence, the first data will be att poistion N+1 with N empty shared_ptr infront of it

petersteneteg: Ahh nice catch

src/core/properties/propertypresetmanager.cpp
r-englund: Well, according to discussions in https://github.com/inviwo/inviwo/issues/594 it is not an issue and works as expected ;)

ResearchDaniel:
Experimented a bit and it is a really nice feature that would be nice to have as toolbar buttons in inviwo!

I think that there are some polishing left to make it a bit more intuitive, both with respect to the functions and with some GUI-namings. This is what I could come up with on the spot:

- Pressing Set Near/Far distance for object more than 25 times the zoom will not make the object visible.
- "Fit Camera To Volume"-> "Center View" 
- "Flip Upvector"-> "Flip Horizontal" or "Flip Up" or "Flip Up Vector" 
-  It is not entirely intuitive what "Set near/far plane" and "Set look to/from Ranges" does. Perhaps adding their corresponding parameters (max zoom out) would clarify? Alternatively, naming them in the lines of "Set Max Zoom-Out 25 (Near/Far Plane)" "Set Max Zoom-Out 25 (Look To/From Bounds)"

Up for discussion :) 
Unfortunately, I don't have much time to work on this, and I do think that it is awesome!

modules/base/include/modules/base/algorithm/camerautils.h
```diff
@@ -0,0 +1,209 @@
...
+/**
+ * Sets the near and far clip distances of the camera based on the given bounding volume and zoom
+ * factor. Ensures that the far plane is distant enough to avoid clipping given to current
+ * zoomfactor.
+ * @see computeCameraNearFar
+ * @see setCameraLookRanges
+ */
+IVW_MODULE_BASE_API void setCameraNearFar(CameraProperty &cam, const mat4 &boundingBox,
+                                          float maxZoomFactor = 25.f,
+                                          float nearFarRatio = 1.f / 10000.f);
```
ResearchDaniel, line 114: Might want to mention that setting a near far ratio below 10^-5/6 might have precision problems.

Rename maxZoomFactor to maxZoomOutFactor

modules/base/include/modules/base/algorithm/camerautils.h
```diff
@@ -0,0 +1,209 @@
...
+ * @param fitRatio   determines the spacing between volume and the boundaries of the view frustum. A
+ *                   fit ratio of 1 means a perfect fit, no space between frustum and volume. The
+ * aspect ratio is taken into account.
+ * @param setNearFar the camera's new/far clip ranges are updated if true @see setCameraNearFar
+ * @param setLookRanges   the camera's look-to/look-from ranges are updated if true @see
+ * setCameraLookRanges
+ */
+IVW_MODULE_BASE_API void setCameraView(CameraProperty &cam, const mat4 &boundingBox, Side side,
+                                       float fitRatio = 1.05f, bool setNearFar = false,
+                                       bool setLookRanges = false);
```
ResearchDaniel, line 65: How about instead the of booleans:
Function overloads with the corresponding parameters given, e.g., 
fitObjectIntoView(cam, boundingBox, side, fitRatio) (equivalent to setNear, setLook = false)
fitObjectIntoView(cam, boundingBox, side, fitRatio, maxZoomOutFactor, nearFarRatio) (equivalent to setNear, setLook = true)

ResearchDaniel: Took the liberty of suggesting a function name change that I think is more expressive of what the function does

r-englund:
Closing this PR since I am leaving for vacation now.  I wanted to move this feature from reserach-dev into core before leaving since people have been asking for it. Might open it again after vacation

PR #615, opened 2019-07-11
Qt: Fix for broken console widget in qt 15.3.0

r-englund:
Just updated to qt 15.3.0 and the Console Widget had the wrong font and looked Terrible 
Before fix:
broken-logconsole
After fix: 
image

### DIFF
--- a/src/qt/editor/consolewidget.cpp
+++ b/src/qt/editor/consolewidget.cpp
@@ -160,7 +160,8 @@ ConsoleWidget::ConsoleWidget(InviwoMainWindow* parent)
     tableView_->verticalHeader()->setVisible(false);
     tableView_->verticalHeader()->setResizeContentsPrecision(0);
     tableView_->verticalHeader()->setSectionResizeMode(QHeaderView::Fixed);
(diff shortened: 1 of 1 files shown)


martinfalk (approved):
I am very much in favor of the new (fixed) look.

petersteneteg (changes requested):
The console should have a fixed format font!

martinfalk:
According to [this](https://support.microsoft.com/en-us/help/90057/microsoft-supplied-monospaced-truetype-fonts), Windows 10 ships Courier New and Lucida sans typewriter.
For some reason, Qt seems to pick Courier as a system default (I presume), which looks horrible. And courier doesn't know about the ellipsis "…" hence the square boxes. Maybe we can tweak it toward Lucida sans typewriter instead (also supports way more unicode characters).

r-englund:
This is caused by a bug introduced in qt 5.12.3 https://bugreports.qt.io/browse/QTBUG-75648 
I just tried with qt 5.13.1-preview and it seems to be fixed.
I guess it does not make sense for us to try to fix it if it is only broken in one Qt version?

r-englund:
image

petersteneteg:
Agree

PR #622, opened 2019-08-05
c++17

petersteneteg:


ResearchDaniel (approved):
Looks good to me

cmake/globalconfig.cmake
```diff
@@ -39,31 +39,39 @@ else() # if IVW_BUILD_VERSION is not set or set to zero
 endif()
 
 #--------------------------------------------------------------------
+# Disallow in-source builds
+if ("${CMAKE_BINARY_DIR}" STREQUAL "${CMAKE_SOURCE_DIR}")
```
r-englund: This would allow a build directory like ${CMAKE_SOURCE_DIR}/build, is that ok/supported?

r-englund (approved):

cmake/globalconfig.cmake
petersteneteg: Humm I wanted that to fail too.. might have to tweak that a bit...

Issue #620, opened 2019-08-05
ToolsMetaMenu.cpp fails to build in MSVC 2019


r-englund:
Compiling ToolsMetaMenu.cpp with MSVC 2019 results in an `An internal error has occurred in the compiler.`
I did some "investigation" by commenting away code and the error is comming from within the includes, after some more commenting of code in the includes I ended up with a situation where it did compile when the following lines where commented and failed when I unommented them: 
https://github.com/inviwo/inviwo/blob/master/tools/meta/include/inviwo/meta/cmake/cmakefile.hpp#L126-L133
I am not sure wether that is the location of the error or not. When I uncomented all the code I had commented out, except the lines mentioned above, I still got the internal error. 

```
1>------ Build started: Project: inviwo-qteditor, Configuration: RelWithDebInfo x64 ------
1>toolsmetamenu.cpp
1>E:\worktrees\inviwo-blodflowexhib\src\qt\editor\toolsmetamenu.cpp : fatal error C1001: An internal error has occurred in the compiler.
1>(compiler file 'd:\agent\_work\3\s\src\vctools\Compiler\Utc\src\p2\main.c', line 160)
1> To work around this problem, try simplifying or changing the program near the locations listed above.
1>Please choose the Technical Support command on the Visual C++
1> Help menu, or open the Technical Support help file for more information
1>INTERNAL COMPILER ERROR in 'C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.22.27905\bin\HostX64\x64\CL.exe'
1>    Please choose the Technical Support command on the Visual C++
1>    Help menu, or open the Technical Support help file for more information
1>C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\MSBuild\Microsoft\VC\v160\Microsoft.CppCommon.targets(429,5): error MSB6006: "CL.exe" exited with code 2.
1>Done building project "inviwo-qteditor.vcxproj" -- FAILED.
========== Build: 0 succeeded, 1 failed, 0 up-to-date, 0 skipped ==========
```

r-englund:
Seems to be a vs 2019 bug realted to static inline in classes that are DLL imported in other projects. Building Inviwo with static linking works. 
https://developercommunity.visualstudio.com/content/problem/670280/vs2019-v1620-ice-when-using-static-inline-const-in.html

PR #625, opened 2019-08-06
Meta: Possible compile fix for MSVC 2019

r-englund:
When updating from MSVC 2017 to 2019 meta failed to build (see #620). 
There is a bug in MSVC 2019 with DLL export/import and static inline const. 
Changing the `static inline const` to  `static inline constexpr` removes the ICE


petersteneteg:
does this compile in 2017?

r-englund:
> 
> 
> does this compile in 2017?

trying now

r-englund:
> 
> 
> > does this compile in 2017?
> 
> trying now

Compiles and runs without errors

petersteneteg (approved):

PR #626, opened 2019-08-07
Better tf image import

petersteneteg:
New version 
bild

### DIFF
--- a/include/inviwo/core/datastructures/transferfunction.h
+++ b/include/inviwo/core/datastructures/transferfunction.h
@@ -134,8 +134,13 @@ public:
 
     virtual std::string getTitle() const override;
 
-
-    static std::vector<TFPrimitiveData> simplify(const std::vector<TFPrimitiveData>& points, double delta);
+    /**
+     * Simplify a vector of TF points by removing points that can be replaced by interpolating
+     * between the previous and the next point, as long as the resulting error is less then delta.
+     * the error is the absolute maximal component wise error at the point to be removed.
+     */
+    static std::vector<TFPrimitiveData> simplify(const std::vector<TFPrimitiveData>& points,
+                                                 double delta = 0.01);
 
 protected:
     void calcTransferValues() const;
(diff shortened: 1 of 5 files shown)


include/inviwo/core/datastructures/transferfunction.h
```diff
@@ -134,6 +134,9 @@ class IVW_CORE_API TransferFunction : public TFPrimitiveSet {
 
     virtual std::string getTitle() const override;
 
+
+    static std::vector<TFPrimitiveData> simplify(const std::vector<TFPrimitiveData>& points, double delta);
```
ResearchDaniel: Add documentation, what is delta? Percieved color difference?, intensity?, or for each rgba-component? How is opacity treated?
What would a good default value be?

src/core/datastructures/transferfunction.cpp
```diff
@@ -217,48 +217,70 @@ void TransferFunction::load(const std::string& filename, const FileExtension& ex
...
 
+std::vector<TFPrimitiveData> TransferFunction::simplify(const std::vector<TFPrimitiveData>& points,
+                                                        double delta) {
+
+    if (points.size() < 3) return points;
+    std::vector<TFPrimitiveData> simple{points};
+
+    // Calculate the error resulting from using a linear interpolation between the prev and next
+    // point instead of including the current one
+    const auto error = [&](size_t i) {
```
ResearchDaniel, line 257: Color and opacity kind of live in different spaces, so perhaps we should separate the threshold into one for color difference and one for opacity?

I think that percieved color difference would make more sense (e.g. convert the color to Luv before using the threshold)

ResearchDaniel (changes requested):
Nice! I miss some documentation about how the threshold in the simplify function and what it does.

Optional improvements:
- Use Luv color space instead of RGB when calculating the error. 
- Simplify could take in a distance computation lambda with the current being the default

r-englund (approved):

src/core/datastructures/transferfunction.cpp
petersteneteg: might be nice, but that's an other feature

ResearchDaniel (approved):

PR #628, opened 2019-08-07
Python: Added a helper for adding module paths

### LINKED ISSUE #551: PythonProcessor: import of local python script fails
When importing a local python script (here `vtkutility.py`), Inviwo reports an error that there is no such module and there is no processor `vtkutiliy`.

### The python script of the processor:
```python
# Name: VtkImageData 

import inviwopy as ivw

import numpy

import vtkutility

class VtkImageData(ivw.Processor):
    def __init__(self, id, name):
    ...
```

### Inviwo error message
```
**PythonProcessorFolderObserver**	Failed to load processor VtkImageData from script: C:/dev/inviwo/inviwo/modules/python3/processors/vtkimagedata.py.
ModuleNotFoundError: No module named 'vtkutility'
At:
  <string>(8): <module>

**PythonProcessorFolderObserver**	Failed to find python processor "vtkutility" in pythons object register
```

petersteneteg:
Closes #551

### DIFF
--- a/modules/python3/CMakeLists.txt
+++ b/modules/python3/CMakeLists.txt
@@ -103,7 +103,7 @@ endif()
 
 #--------------------------------------------------------------------
 # Add Python to package
-option(IVW_PACKAGE_EMBED_PYTHON "Embed a python dist in the installer.\
+option(IVW_PACKAGE_EMBED_PYTHON "Embed a python dist in the installer.\ 
  Otherwise a existing python installation is needed to run inviwo" ON)
 if(IVW_PACKAGE_PROJECT AND BUILD_SHARED_LIBS AND IVW_PACKAGE_EMBED_PYTHON)
     if(WIN32)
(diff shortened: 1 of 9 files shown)


modules/python3/CMakeLists.txt
```diff
@@ -1,168 +1,170 @@
...
+
+add_subdirectory(bindings)
+
+if(IVW_UNITTESTS)
+    add_dependencies(inviwo-unittests-python3 inviwopy)
+endif()
+
+#--------------------------------------------------------------------
+# Add Python to package
+option(IVW_PACKAGE_EMBED_PYTHON "Embed a python dist in the installer.\
```
r-englund: What kind of end of line character is that (`^M`)

petersteneteg: windows?

r-englund (approved):

PR #632, opened 2019-08-07
Core: Make PropertyOwner remove all its properties before destruction…

ResearchDaniel:
Make PropertyOwner remove all its properties before destruction to ensure that PropertyOwnerObserver::onWillRemoveProperty is called.

Otherwise there can be calls to onWillAddProperty but not to onWillRemoveProperty

### DIFF
--- a/include/inviwo/core/processors/canvasprocessor.h
+++ b/include/inviwo/core/processors/canvasprocessor.h
@@ -92,6 +92,7 @@ protected:
 public:
     ImageInport inport_;
 
+    CompositeProperty inputSize_;
     IntVec2Property dimensions_;
     BoolProperty enableCustomInputDimensions_;
     IntVec2Property customInputDimensions_;
@@ -104,7 +105,6 @@ public:
     DirectoryProperty saveLayerDirectory_;
     ButtonProperty saveLayerButton_;
     ButtonProperty saveLayerToFileButton_;
-    CompositeProperty inputSize_;
     BoolProperty fullScreen_;
     EventProperty fullScreenEvent_;
     EventProperty saveLayerEvent_;
--- a/src/core/properties/propertyowner.cpp
+++ b/src/core/properties/propertyowner.cpp
@@ -49,9 +49,9 @@ PropertyOwner::PropertyOwner(const PropertyOwner& rhs)
 
     for (const auto& p : rhs.ownedProperties_) addProperty(p->clone());
 }
-
+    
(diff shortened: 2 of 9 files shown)


petersteneteg:
Where did you run into this? 
You might want to fix operator= also? that only does properties.clear()

ResearchDaniel:
Property-synchronization with webbrowser. I need to remove widgets when a processor is removed. 
=operator does not use add either from what I could tell, not sure if it counts as add/remove. I’ll look into it a bit more

petersteneteg:
https://github.com/inviwo/inviwo/blob/99ffbae57747d92f7a4471b62e9f555d9bf981e9/src/core/properties/propertyowner.cpp#L64
this must result in an add notification?

petersteneteg:
https://github.com/inviwo/inviwo/blob/99ffbae57747d92f7a4471b62e9f555d9bf981e9/src/core/properties/propertyowner.cpp#L62 Humm I wounder if this is even correct? usually we call 
prop1 = rhs.prop1;
prop2 = rhs.prop2;
in the assignment operator of the derived class, but if we have done properties.clear() in this first those properties will not be listed anymore?

ResearchDaniel:
Fixed

src/core/properties/propertyowner.cpp
```diff
@@ -50,11 +50,20 @@ PropertyOwner::PropertyOwner(const PropertyOwner& rhs)
     for (const auto& p : rhs.ownedProperties_) addProperty(p->clone());
 }
 
+PropertyOwner::~PropertyOwner() {
+    while (begin() != end()) {
```
petersteneteg: This crashes on osx...

ResearchDaniel: Should be fixed now.

include/inviwo/core/processors/canvasprocessor.h
```diff
@@ -92,6 +92,7 @@ class IVW_CORE_API CanvasProcessor : public Processor, public ProcessorWidgetMet
 public:
     ImageInport inport_;
 
+    CompositeProperty inputSize_;
```
petersteneteg, line 95: Will this not create a order warning in the constructor?

petersteneteg (approved):

include/inviwo/core/processors/canvasprocessor.h
ResearchDaniel: Yes, fixed it

PR #634, opened 2019-08-08
Webrowser: CEF_TARGET_OUT_DIR need to be set to the executable path.

r-englund:
When building statically `$<TARGET_FILE_DIR:inviwo-module-webbrowser>`  will evaluate to `build/lib/{Config}` and hence the cef dlls will be copied to the lib directory instead of the bin directory. 

An option is to set it to `CMAKE_RUNTIME_OUTPUT_DIRECTORY`, which is the same. (They are both set in https://github.com/inviwo/inviwo/blob/master/cmake/globalconfig.cmake#L115-L120)

### DIFF
--- a/modules/webbrowser/CMakeLists.txt
+++ b/modules/webbrowser/CMakeLists.txt
@@ -68,8 +68,7 @@ set(BUILD_SHARED_LIBS ${BUILD_SHARED_LIBS_TMP})
 
 # Determine the target output directory.
 # Inviwo: use module target directory as output dir$<TARGET_FILE_DIR:inviwo-module-webbrowser>
-#SET_CEF_TARGET_OUT_DIR()
-set(CEF_TARGET_OUT_DIR $<TARGET_FILE_DIR:inviwo-module-webbrowser>)
+set(CEF_TARGET_OUT_DIR ${EXECUTABLE_OUTPUT_PATH}/$<CONFIG>)
 
 # Display configuration settings.
 #PRINT_CEF_CONFIG()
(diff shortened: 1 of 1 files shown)


modules/webbrowser/CMakeLists.txt
```diff
@@ -68,8 +68,7 @@ set(BUILD_SHARED_LIBS ${BUILD_SHARED_LIBS_TMP})
 
 # Determine the target output directory.
 # Inviwo: use module target directory as output dir$<TARGET_FILE_DIR:inviwo-module-webbrowser>
-#SET_CEF_TARGET_OUT_DIR()
-set(CEF_TARGET_OUT_DIR $<TARGET_FILE_DIR:inviwo-module-webbrowser>)
+set(CEF_TARGET_OUT_DIR ${EXECUTABLE_OUTPUT_PATH}/$<CONFIG>)
```
petersteneteg: on a non multi config build like make or ninja there will not be a CONFIG dir

r-englund: :+1: didn't think about that, fixed now

ResearchDaniel (approved):
This should probably be tested on Mac since it might differ a bit. Is it enabled on Jenkins (mac)?

r-englund:
> 
> 
> This should probably be tested on Mac since it might differ a bit. Is it enabled on Jenkins (mac)?

The webbroser module is enabled on jenkins, for example, your PR #632 fails because of it ;)

petersteneteg (approved):

PR #635, opened 2019-08-09
Selected File Extensions

### LINKED ISSUE #590: FileProperty does not store extension/type of selected reader
When a `FileProperty` is serialized, it stores a list of all registered file extensions (`<nameFilter>`). However, neither the currently used reader nor the extension is not stored. This can cause problems when multiple readers are registered for the same file extension.

Example: 
* TIFF-Reader registers `tif` and `tiff
* DICOM-Reader registers `tif` and `dcm`
* TIFF volume data is loaded in `VolumeSource` with `TIFF reader` selected, then the workspace is saved
* Loading the workspace results in an error that the DICOM reader failed to read the tiff file (it might be the first reader in the factory handling a tiff extension)

How should we deal with this?

petersteneteg:
Updates to the FileProperty and Source processors. Better handling of the selected file extensions.

closes #590

bild

### DIFF
--- a/include/inviwo/core/properties/fileproperty.h
+++ b/include/inviwo/core/properties/fileproperty.h
@@ -82,7 +82,14 @@ public:
     virtual FileProperty* clone() const override;
     virtual ~FileProperty() = default;
 
-    virtual void set(const std::string& value) override;
+    /**
+     * Set the file name and also update the selected extension to the first one matching file.
+     */
+    virtual void set(const std::string& file) override;
+    /**
+     * Set the file name and the selected extension.
+     */
+    virtual void set(const std::string& file, const FileExtension& selectedExtension);
     virtual void set(const Property* property) override;
 
     virtual void serialize(Serializer& s) const override;
@@ -92,7 +99,7 @@ public:
     virtual void addNameFilter(FileExtension);
     virtual void addNameFilters(const std::vector<FileExtension>& filters);
     virtual void clearNameFilters();
-    virtual std::vector<FileExtension> getNameFilters() const;
+    virtual const std::vector<FileExtension>& getNameFilters() const;
 
     virtual void setAcceptMode(AcceptMode mode);
     AcceptMode getAcceptMode() const;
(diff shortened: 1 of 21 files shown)


petersteneteg:
review anyone?

modules/base/src/processors/volumesequencesource.cpp
```diff
@@ -167,15 +190,11 @@ void VolumeSequenceSource::loadFolder(bool deserialize) {
...
 }
 
-void VolumeSequenceSource::addFileNameFilters() {
-    file_.clearNameFilters();
-    file_.addNameFilter(FileExtension::all());
-    file_.addNameFilters(rf_->getExtensionsForType<VolumeSequence>());
-}
-
 void VolumeSequenceSource::process() {
     if (file_.isModified() || reload_.isModified() || folder_.isModified() ||
```
r-englund, line 199: do we need to add reader_.isModified() here?

r-englund (approved):

modules/base/src/processors/volumesequencesource.cpp
petersteneteg: yes

include/inviwo/core/properties/fileproperty.h
```diff
@@ -82,7 +82,14 @@ class IVW_CORE_API FileProperty : public TemplateProperty<std::string> {
     virtual FileProperty* clone() const override;
     virtual ~FileProperty() = default;
 
-    virtual void set(const std::string& value) override;
+    /**
+     * Set the file name and also update the selected extension to the first one matching file.
```
ResearchDaniel, line 86: first one matching file? 
You mean?
Set file name and update the selected extension to that of the file.

include/inviwo/core/properties/fileproperty.h
```diff
@@ -101,7 +108,7 @@ class IVW_CORE_API FileProperty : public TemplateProperty<std::string> {
     FileMode getFileMode() const;
 
     void setContentType(const std::string& contentType);
-    std::string getContentType() const;
+    const std::string& getContentType() const;
```
ResearchDaniel, line 111: not part of this PR but a hint on what content type is would be good...

ResearchDaniel (approved):
Looks good. I guess we could hide the reader_ if there is just a single reader?

include/inviwo/core/properties/fileproperty.h
petersteneteg: the file is just a string, and there can be may FileExtensions that match for example *.jpg (one per registered reader)

Issue #513, opened 2019-05-08
automatically set project folders for modules


r-englund:
Many of us have a lot of enabled modules when building our projects in Inviwo, this makes it tedious to browse the projects (in the solution explorer in MSVC). 

I think we, at some point, discussed this at some Inviwo meeting with the suggestion to add each module to a project folder in some smart way. I did some experimenting and got something working (https://github.com/inviwo/inviwo/commit/d19878c64e492b0033ccb49cf0a2c4694caa3571). This places each module in a nice tree structure based on where the modules is located on disc: 
![image](https://user-images.githubusercontent.com/534670/57376284-ece9e100-719f-11e9-896f-addb3ea6e258.png)
IMO that is much nicer then having a long list like this: 
![image](https://user-images.githubusercontent.com/534670/57376491-65e93880-71a0-11e9-9deb-4d779d5e6ec9.png)

martinfalk:
I second that idea. Much nicer and better organization.

petersteneteg:
This needs a more stable way of naming the groups

r-englund:
> 
> 
> This needs a more stable way of naming the groups

Any suggestions?

petersteneteg:
If I had a easy way I would have done it long ago... So far the only way I have thought of is to add some file next to all the modules with metadata.

petersteneteg:
Such meta data could also be useful in a few other places

martinfalk:
I agree with the idea of metadata. But I am afraid that there needs to be some kind of fall-back mechanism. Maybe not for newly created modules where could automatically place a file but for existing ones (potentially external).

PR #637, opened 2019-08-12
CMake: Grouping modules in folders #513

petersteneteg:
bild

Group names defaults to the parent folder name of the module. But one can specify a group name in the ivw_create_module command
i.e.
```
ivw_create_module(GROUP inviwo-core-modules NO_PCH ${SOURCE_FILES} ${HEADER_FILES} ${SHADER_FILES})
```

### DIFF
--- a/cmake/globalmacros.cmake
+++ b/cmake/globalmacros.cmake
@@ -236,7 +236,14 @@ function(ivw_register_modules retval)
     ivw_add_module_option_to_cache(${mod} ON)
 
     foreach(module_path ${IVW_MODULE_DIR} ${IVW_EXTERNAL_MODULES})
-        get_filename_component(group_name ${module_path} NAME)    
+        get_filename_component(group_name ${module_path} NAME)
+        set(group_name "inviwo-${group_name}-modules")
+
+        # Check of there is a meta.cmake
+        # Optionally defines: group_name
+        if(EXISTS "${module_path}/meta.cmake")
+            include("${module_path}/meta.cmake")  
+        endif()
 
         string(STRIP ${module_path} module_path)
         if(NOT EXISTS ${module_path})
@@ -264,7 +271,7 @@ function(ivw_register_modules retval)
                 ivw_private_get_ivw_module_version(${module_path}/${dir}/CMakeLists.txt version)
                 ivw_private_setup_module_data(
                     NAME ${name} 
-                    GROUP "inviwo-${group_name}-modules" 
+                    GROUP ${group_name} 
                     VERSION ${version} 
                     DIR ${dir} 
                     BASE ${module_path}
--- a/cmake/globalmacros.cmake
+++ b/cmake/globalmacros.cmake
@@ -237,7 +237,7 @@ function(ivw_register_modules retval)
 
     foreach(module_path ${IVW_MODULE_DIR} ${IVW_EXTERNAL_MODULES})
         get_filename_component(group_name ${module_path} NAME)
-        set(group_name "inviwo-${group_name}-modules")
+        set(group_name "modules-${group_name}")
 
         # Check of there is a meta.cmake
         # Optionally defines: group_name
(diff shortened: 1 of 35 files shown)


r-englund (changes requested):
We need a way of handling when no group name is specified and the folder is named modules. In the image below a group named `inviwo-modules-modules` is created and it contains modules form 4 different repos. 
image

petersteneteg:
Just rename the directory to a more descriptive name, or add the group names in the cmake files.

petersteneteg:
I don't want any hard coded special cases in the code.

r-englund:
> 
> 
> I don't want any hard coded special cases in the code.

If the `get_filename_component` returns modules, use the parent folder instead, not really a hard-coded special case IMO.

r-englund:
just adding 
```cmake
if(${group_name} STREQUAL modules)
    get_filename_component(moddir ${module_path} DIRECTORY )
    get_filename_component(group_name ${moddir} NAME )      
endif()
```
directly after `get_filename_component(group_name ${module_path} NAME)` would fix it

petersteneteg:
that would be the very definition of a hard coded special case

petersteneteg:
@ResearchDaniel: you get to decide

cmake/globalmacros.cmake
```diff
@@ -259,7 +262,13 @@ function(ivw_register_modules retval)
                 list(APPEND modules ${mod})
                 ivw_private_get_ivw_module_name(${module_path}/${dir}/CMakeLists.txt name)
                 ivw_private_get_ivw_module_version(${module_path}/${dir}/CMakeLists.txt version)
-                ivw_private_setup_module_data(NAME ${name} VERSION ${version} DIR ${dir} BASE ${module_path})
+                ivw_private_setup_module_data(
+                    NAME ${name} 
+                    GROUP "inviwo-${group_name}-modules"
```
ResearchDaniel: Can't the group be just the name, do we need to prepend and append inviwo- -modules? 
I think that it would look much cleaner without it

ResearchDaniel (changes requested):
What if we just loose the "inviwo-" and "-modules" added to the group name. 
Then it would just be less of an issue?

cmake/globalmacros.cmake
petersteneteg: the prefix makes them sort together which I like. the postfix was mostly because I found just having "inviwo-core" a strange since we already have that as a different target

petersteneteg:
At one stage I also had a meta.cmake file in the module folders where you could set the group name. would bringing that back be a solution? @r-englund

cmake/globalmacros.cmake
r-englund: an option ( which I know peter doesn't favour) is to have them in a folder, i.e modules/inviwo-core modules/inviwo-misc etc. Then the default group name could be other, or nothing (so they appear in modules/others or directly in modules). 
Then we could also drop the inviwo prefix

petersteneteg: my issue with that is that it will make it even harder to find "your" code. I also use the collapse all in vs often, and then having more levels means more clicks to get to where i want to go

r-englund: I have had the folder solution locally for quite some time, and use the collapse all feature quite a lot and haven't found it annoying :) And it is only one more click ;)

martinfalk:
> At one stage I also had a meta.cmake file in the module folders where you could set the group name. would bringing that back be a solution? @r-englund

I like the idea of having the `meta.cmake` file as an additional option. Also simplifies moving modules around without having to update the group.

r-englund (approved):
I still think the naming is a bit verbose, I mean, the name is [verbose]-name-[verbose]. The name that you would look for is "hidden" between two [verbose] that are the same for everything. 
What if we drop the '-modules' from the default name? It would still be there for inviwo-core-modules since we have it in the meta.

r-englund:
I like it :) 
image

petersteneteg:
ping @ResearchDaniel

ResearchDaniel:
@petersteneteg Looks good

ResearchDaniel (approved):

PR #645, opened 2019-08-14
OpenGLQT: Issue a resize event after making canvas visible to ensure correct size, closes #536

### LINKED ISSUE #536: Incorrect canvas size when reloading network
# Step to reproduce
1. Load the boron network
2. change the canvas size (to for example 300x300)
3. Reload the network, the canvas is now 16x39 pixels smaller (284x261)

r-englund:
OpenGLQT: Issue a resize event after making canvas visible to ensure correct size, closes #536

### DIFF
--- a/modules/openglqt/src/canvasprocessorwidgetqt.cpp
+++ b/modules/openglqt/src/canvasprocessorwidgetqt.cpp
@@ -93,7 +93,6 @@ CanvasProcessorWidgetQt::CanvasProcessorWidgetQt(Processor* p)
     }
 
     {
-#if QT_VERSION > QT_VERSION_CHECK(5, 6, 0)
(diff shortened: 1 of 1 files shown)


r-englund:
Why this change: 
On some machines it the size of the canvas changed when calling QWidget::show, when inspecting the widget size it was correct before the call and incorrect after the call (see #536 for numbers), adding the QWidget::resize directly after ::show will ensure that the canvas get the correct size after deserialization. 
On machines that did not have the issue, nothing will happen since we issue a resize event with the current size

petersteneteg (approved):
Would be good if you added your explanation to the comment here also....

r-englund:
> 
> 
> Would be good if you added your explanation to the comment here also....

I thought about it, but the comment above already states that

PR #646, opened 2019-08-16
Property assignment

petersteneteg:
Removes operator= from Property and PropertyOwner. The sematics is strange, it is quite hard to get right, and we never use it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 Here we document changes that affect the public API or changes that needs to be communicated to other developers. 
 
+## 2019-08-16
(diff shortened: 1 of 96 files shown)


ResearchDaniel (approved):
Should we perhaps say something about the owner not being transfered when cloning/copying and that the identifier must be changed after these two operations?

petersteneteg:
@ResearchDaniel suggestions about where to say that? 
Property::clone?

ResearchDaniel:
Yes, clone and copy constructor

ResearchDaniel:
By the way, we should update the change log and say that people can just remove the = operator in their properties

martinfalk (approved):

PR #648, opened 2019-08-16
Minmax property fix

petersteneteg:


include/inviwo/core/properties/minmaxproperty.h
```diff
@@ -310,8 +308,10 @@ template <typename T>
 MinMaxProperty<T>& MinMaxProperty<T>::setRangeNormalized(const range_type& newRange) {
     dvec2 val = this->get();
 
-    val = (val - static_cast<double>(range_.value.x)) /
-          (static_cast<double>(range_.value.y) - static_cast<double>(range_.value.x));
+    const auto nomalizedValue =
```
r-englund, line 309: unused?

petersteneteg: ahh forgot to finish that one...

r-englund (approved):

include/inviwo/core/properties/minmaxproperty.h
```diff
@@ -272,29 +272,27 @@ MinMaxProperty<T>& MinMaxProperty<T>::setMinSeparation(const T& newMinSeparation
 }
 
 template <typename T>
-MinMaxProperty<T>& MinMaxProperty<T>::setRange(const range_type& value) {
-    const auto newRange = clamp({glm::min(value.x, value.y), glm::max(value.x, value.y)});
+MinMaxProperty<T>& MinMaxProperty<T>::setRange(const range_type& newRange) {
+    if (range_.update({glm::min(newRange.x, newRange.y), glm::max(newRange.x, newRange.y)})) {
```
ResearchDaniel, line 276: Perhaps add a comment why this is necessary or should be done. Is it not an error if the new range is flipped?

include/inviwo/core/properties/minmaxproperty.h
```diff
@@ -272,29 +272,27 @@ MinMaxProperty<T>& MinMaxProperty<T>::setMinSeparation(const T& newMinSeparation
...
-void MinMaxProperty<T>::set(const range_type& value, const range_type& range, const T& increment,
-                            const T& minSep) {
-
-    const auto newRange = clamp({glm::min(range.x, range.y), glm::max(range.x, range.y)});
+void MinMaxProperty<T>::set(const range_type& newValue, const range_type& newRange,
+                            const T& newIncrement, const T& newMinSep) {
 
     bool modified = false;
-    modified |= range_.update(newRange);
+    modified |= range_.update({glm::min(newRange.x, newRange.y), glm::max(newRange.x, newRange.y)});
```
ResearchDaniel, line 290: Again, is this not an input error?

ResearchDaniel (dismissed):
I’m not sure if incorrect range input should be corrected. Throw invalid argument exception?

include/inviwo/core/properties/minmaxproperty.h
petersteneteg: This is just to be defensive that the new min is less then the new max we have always done this...

include/inviwo/core/properties/minmaxproperty.h
petersteneteg: this is how we have used to do it, if you want to throw go ahead and add that...

include/inviwo/core/properties/minmaxproperty.h
martinfalk: could this be moved into the if statement?

martinfalk (approved):

include/inviwo/core/properties/minmaxproperty.h
petersteneteg: No since it depends on the old range...

PR #649, opened 2019-08-16
PlottingGL: Possible to specify margins including labels for Parallel…

ResearchDaniel:
…Coordinates


modules/plottinggl/include/modules/plottinggl/processors/parallelcoordinates/parallelcoordinates.h
```diff
@@ -211,6 +211,7 @@ class IVW_MODULE_PLOTTINGGL_API ParallelCoordinates : public Processor {
     };
     Lines lines_;
 
+    MarginProperty marginsInternal_; // Margins with/without considering labels
```
petersteneteg: does this have to be a property? why not just use a vec4?

ResearchDaniel: Only because it has the getRect function. I guess one could duplicate that code and put it in the processor or make it a utility function as well. Not sure if it is worth the effort?

petersteneteg: Bringing in a whole composite just to store that vec4 is a bit much I think.
Then you also need to explain why you have a property that is not added to the owner etc.

petersteneteg (approved):

PR #651, opened 2019-08-16
Camera fitting

petersteneteg:
see #613


include/inviwo/core/properties/buttongroupproperty.h
```diff
@@ -0,0 +1,105 @@
...
+ * them.
+ * @see ButtonGroupPropertyWidgetQt
+ */
+class IVW_CORE_API ButtonGroupProperty : public Property {
+public:
+
+    struct IVW_CORE_API Button {
+        std::optional<std::string> name;
+        std::optional<std::string> icon;
+        std::function<void()> action;
```
r-englund, line 53: Our "regular" ButtonProperty can have more than one callback, wouldn't it make sense to be consistent to allow more than one callback here as well?

modules/python3/bindings/src/pyproperties.cpp
```diff
@@ -388,7 +387,40 @@ void exposeProperties(py::module &m) {
...
+             py::arg("identifier"), py::arg("displayName"), py::arg("action"),
+             py::arg("invalidationLevel") = InvalidationLevel::InvalidResources,
+             py::arg("semantics") = PropertySemantics::Default)
         .def("press", &ButtonProperty::pressButton);
+
+    py::class_<ButtonGroupProperty::Button>(m, "ButtonGroupPropertyButton")
+        .def(py::init<std::optional<std::string>, std::optional<std::string>,
+                      std::function<void()>>());
+
+    PyPropertyClass<ButtonGroupProperty, Property>(m, "ButtonGroupProperty")
```
r-englund, line 405: :+1: For remembering the pybindings

r-englund (approved):

include/inviwo/core/properties/buttongroupproperty.h
```diff
@@ -0,0 +1,105 @@
...
+#include <inviwo/core/properties/property.h>
+#include <functional>
+#include <optional>
+
+namespace inviwo {
+
+/**
+ * \ingroup properties
+ * \brief The Button Group property class provides a group of button that you can bind functions to.
+ * The button group property has a widget witch creates a row of buttons and register functions to
```
ResearchDaniel, line 43: Group of buttons
Witch -> which

include/inviwo/core/properties/buttongroupproperty.h
```diff
@@ -0,0 +1,105 @@
...
+ * them.
+ * @see ButtonGroupPropertyWidgetQt
+ */
+class IVW_CORE_API ButtonGroupProperty : public Property {
```
ResearchDaniel, line 47: Why not a composite property?

include/inviwo/core/properties/buttongroupproperty.h
```diff
@@ -0,0 +1,105 @@
...
+public:
```
ResearchDaniel, line 48: What if we instead had a GridLayout property? Any property could be added and one would specify how many columns it should span.

modules/base/include/modules/base/events/viewevent.h
```diff
@@ -0,0 +1,68 @@
...
+#include <inviwo/core/util/constexprhash.h>
+#include <modules/base/algorithm/camerautils.h>
+
+#include <variant>
+
+namespace inviwo {
+
+class IVW_MODULE_BASE_API ViewEvent : public Event {
+public:
+    struct FlipUp {};
```
ResearchDaniel: Empty structs?

modules/base/include/modules/base/processors/fitcameratodata.h
```diff
@@ -0,0 +1,79 @@
...
+
+#include <modules/base/properties/camerafittingproperty.h>
+
+namespace inviwo {
+
+/** \docpage{org.inviwo.FitCameraToData, Fit Camera To Data}
+ * ![](org.inviwo.FitCameraToData.png?classIdentifier=org.inviwo.FitCameraToData)
+ *
+ * Processor demonstrating how camerautil::FitCameraPropertiesHelper can be used to update a camera
+ * based on data from volume and mesh inports.
```
ResearchDaniel, line 46: I don’t think that we should have any processors just for demonstration in base. Either change the description  or move it

modules/base/include/modules/base/properties/camerafittingproperty.h
```diff
@@ -0,0 +1,110 @@
...
+#include <inviwo/core/ports/inport.h>
+#include <inviwo/core/ports/volumeport.h>
+#include <inviwo/core/ports/meshport.h>
+
+#include <functional>
+#include <optional>
+
+namespace inviwo {
+
+class IVW_MODULE_BASE_API CameraFittingProperty : public CompositeProperty {
```
ResearchDaniel: We could explain what this one is for

modules/base/include/modules/base/properties/camerafittingproperty.h
```diff
@@ -0,0 +1,110 @@
...
+public:
+    virtual std::string getClassIdentifier() const override;
+    static const std::string classIdentifier;
+
+    CameraFittingProperty(std::string identifier, std::string displayName,
+                          CameraProperty* camera = nullptr,
```
ResearchDaniel: Why pointer here and & in the other constructors?

modules/base/src/algorithm/camerautils.cpp
```diff
@@ -0,0 +1,207 @@
...
+void setCameraLookRanges(CameraProperty &cam, const mat4 &boundingBox, float zoomRange) {
+    NetworkLock lock(&cam);
+
+    vec3 lookTo(boundingBox * vec4(vec3(.5f), 1.f));
+
+    vec3 dx(boundingBox[0]);
+    vec3 dy(boundingBox[1]);
+    vec3 dz(boundingBox[2]);
+
+    auto p0 = lookTo + (dx + dy + dz) * zoomRange;
```
ResearchDaniel: Should be 0.5* (dx...)?

modules/base/src/processors/fitcameratodata.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+namespace inviwo {
+
+// The Class Identifier has to be globally unique. Use a reverse DNS naming scheme
+const ProcessorInfo FitCameraToData::processorInfo_{
+    "org.inviwo.FitCameraToData",  // Class identifier
+    "Fit Camera To Data",          // Display name
+    "Camera",                      // Category
+    CodeState::Stable,             // Code state
+    "CPU, Camera"                  // Tags
```
ResearchDaniel, line 40: Mesh, Volume

modules/base/src/processors/fitcameratodata.cpp
```diff
@@ -0,0 +1,101 @@
...
+ *********************************************************************************/
+
+#include <modules/base/processors/fitcameratodata.h>
+
+namespace inviwo {
+
+// The Class Identifier has to be globally unique. Use a reverse DNS naming scheme
+const ProcessorInfo FitCameraToData::processorInfo_{
+    "org.inviwo.FitCameraToData",  // Class identifier
+    "Fit Camera To Data",          // Display name
```
ResearchDaniel, line 37: Data->Mesh/Volume would make it easier to search for in the processor list widget

modules/base/src/properties/camerafittingproperty.cpp
```diff
@@ -0,0 +1,165 @@
...
+    settings_.setCollapsed(true);
+}
+
+CameraFittingProperty* CameraFittingProperty::clone() const {
+    return new CameraFittingProperty(*this);
+}
+
+void CameraFittingProperty::invokeEvent(Event* event) {
+    if (auto ve = event->getAs<ViewEvent>()) {
+        std::visit(util::overloaded{[&](camerautil::Side side) { setView(side); },
```
ResearchDaniel: This feels like a very complicated switch case?

modules/base/src/properties/camerafittingproperty.cpp
```diff
@@ -0,0 +1,165 @@
...
+            setCameraView(*camera_, *bb, side, fittingRatio_,
+                          updateNearFar_ ? UpdateNearFar::Yes : UpdateNearFar::No,
+                          updateLookRanges_ ? UpdateLookRanges::Yes : UpdateLookRanges::No);
+        }
+    }
+}
+
+void CameraFittingProperty::fitData() {
+    if (camera_ && getBoundingBox_) {
+        if (auto bb = getBoundingBox_(); bb) {
```
ResearchDaniel: Why ; bb?

include/inviwo/core/properties/buttongroupproperty.h
petersteneteg: I don't see why you should be able to do that. I prefer for keep is a simple as possible. Think it is actually kind of strange. Would rather go the other way and remove that from the button property. Only reason that one has support for more than one is that we reuse the onChange call for the button action. Which is a bit strange in it self...

ResearchDaniel (changes requested):

include/inviwo/core/properties/buttongroupproperty.h
petersteneteg: Well it is not a property owner, just a set of buttons.

include/inviwo/core/properties/buttongroupproperty.h
petersteneteg: No

modules/base/include/modules/base/events/viewevent.h
petersteneteg: Yes, tag types

modules/base/src/properties/camerafittingproperty.cpp
petersteneteg: you should have come to our lecture on std::variant. You can't switch on different types.

modules/base/src/properties/camerafittingproperty.cpp
petersteneteg: that is the actual if check, bb is an optional. Although in this case that check is redundant, a leftover from a previous version

modules/base/include/modules/base/properties/camerafittingproperty.h
petersteneteg: in the other constructors it should never be null

include/inviwo/core/properties/buttongroupproperty.h
petersteneteg: That is way out of scope. And we really don't want general layouting of our properties.

modules/base/src/algorithm/camerautils.cpp
petersteneteg: zoomRange is defined in terms of the size of the bounding box not half its size...

ResearchDaniel: Right, but for example specifying zoomRange=1 will make p0 half of the bounding box away since lookTo is in the center. Might be intentional though?

r-englund: A zoomrange = 1 doesn't really make much sense, that would almost enforce the user to always stay within the volume. Hence we have a default value of 25.

r-englund (changes requested):
Does not work for "rotated basises". 
This is how it looks after pressking Z+ 
image
And continiously presseing fit to view makes it rotate a few degrees each press

Edit: 
The volume us 112x112x46 and its basis is 
image

r-englund:
When pressing Z+ on the same data when compiled with the code in my original PR we get 
image

r-englund:
I've done some more testing, it seems that the buttons are projecting on to the world axis instead of the bounding box axis. The buttons are ment to mean, "Look at the X-side of the boundingBox/volume" etc. 
By changing the following in setCameraView I get the correct behaviour for the side buttons behaviour:
```c++
const auto viewDir = glm::normalize(inViewDir);
const auto lookUp = glm::normalize(inLookUp);
```
```c++
const auto viewDir = glm::normalize(mat3(boundingBox) * inViewDir);
const auto lookUp = glm::normalize(mat3(boundingBox) * inLookUp);
```

r-englund:
See commit https://github.com/inviwo/inviwo/commit/e87f1ba98794dbd6722e5991de68304bd4252a3f for my fixes

petersteneteg:
@r-englund great! thanks for the fix, will cherry pick that...

petersteneteg:
@r-englund @ResearchDaniel @martinfalk I have a question:
At the moment our property list looks like this
bild
the Camera fitting is it's own property. Should it be? or should we just merge that into the cameraproperty? That would be more compact, and you always want them to go together anyway? Or what do you think?

petersteneteg:
Possible new look:
bild

r-englund:
The Settings-composite at the bottom is only for the buttons, hence it should be closer to the buttons, IMO. 
Maybe put the buttons and the settings in a composite? 

BTW, how complicated would it be to have a unique tool tip per button, the tool tip now was a bit odd. I hovered the flip button and the tool tip said somthing with lookAt  (haven't compiled it on current machine)

include/inviwo/core/properties/buttongroupproperty.h
ResearchDaniel: I see several potential problems with using a special property for multiple buttons instead of basing it on a CompositeProperty and using the existing ButtonProperty:
- Duplication of behaviour (the same functionality exist in ButtonProperty)
- Special behaviour required for linking (converters)/python/javascript

I therefore think that incorporating layout capabilities would be a better approach as it would be consistent with other code. The grouping is more of a visual interface thing, it does not affect the behaviour of the button(s)

modules/base/src/algorithm/camerautils.cpp
ResearchDaniel: note that a zoom factor of 25 will let you zoom out 24.5/25.5 volumes.

include/inviwo/core/properties/buttongroupproperty.h
martinfalk: I don't think that is a good idea since it mixes UI layouting stuff with  functionality.

include/inviwo/core/properties/buttongroupproperty.h
martinfalk: Agree with @petersteneteg there. Layouting is an entirely different topic.

modules/base/include/modules/base/processors/fitcameratodata.h
martinfalk: We have that for other modules as well, e.g. the `AxisRenderer`. The `SimpleRaycaster` is in the example module. However, moving would mean that the example module potentially gets a bunch of additional dependencies...

modules/base/src/processors/fitcameratodata.cpp
martinfalk: I find `to data` quite fitting. And searching for `fit` or `camera` will do the trick anyway. In addition, if `mesh, volume` is added to the tags, the processor will be that way.

modules/base/include/modules/base/processors/fitcameratodata.h
petersteneteg: I deleted it

include/inviwo/core/properties/buttongroupproperty.h
r-englund: > Agree with @petersteneteg there. Layouting is an entirely different topic.

+1

martinfalk (commented):
@r-englund 
> I've done some more testing, it seems that the buttons are projecting on to the world axis instead of the bounding box axis. The buttons are ment to mean, "Look at the X-side of the boundingBox/volume" etc.
> By changing the following in setCameraView I get the correct behaviour for the side buttons behaviour...

That's depending on the scenario I'd say :) But I agree, fitting to data should use the data bounding box.

martinfalk:
Pro nesting the buttons under camera:
+ always there if a camera property is used
+ no need to add buttons yourself
+ clear property structure

Cons:
+ always there, but can be turned invisible

Personally, I'd prefer the buttons to be located as close to the top of the camera property as possible (if nested). In particular, since this functionality will most likely be used quite often. Settings can appear all the way at the bottom (since hardly used/adjusted). There could potentially be more settings for the camera in the future as well.

> The Settings-composite at the bottom is only for the buttons, hence it should be closer to the buttons, IMO.
> Maybe put the buttons and the settings in a composite?
> 
> BTW, how complicated would it be to have a unique tool tip per button, the tool tip now was a bit odd. I hovered the flip button and the tool tip said somthing with lookAt (haven't compiled it on current machine)

Should be fairly easy to add individual tool tips.

r-englund:
> always there if a camera property is used

But it need to know about the data (bounding box of the data) to work properly. If all cameras will have the buttons it will only work as expected when the camera is "linked" to the data.

petersteneteg:
But the camera property takes a inport in the constructor...

r-englund:
> 
> 
> But the camera property takes a inport in the constructor...

It is not enforced, the buttons should, in that case, be hidden if a port is not supplied

include/inviwo/core/properties/buttongroupproperty.h
ResearchDaniel: @martinfalk Is that not exactly what ButtonGroup does though? The ButtonGroup is just for layouting?

include/inviwo/core/properties/buttongroupproperty.h
ResearchDaniel: I think we can close this dicussion here as it is done in another thread anyway.

modules/base/src/processors/fitcameratodata.cpp
ResearchDaniel: As long as one can find it by searching for camera/mesh/volume I think we are good

petersteneteg: Following your suggestion I deleted it

include/inviwo/core/properties/buttongroupproperty.h
r-englund: If someone wants to have a unique layout for one of their properties they should make their own widget for it. For example, if someone wants to align buttons on a row, they should make their own widget for that. (As Peter has done here)

r-englund (approved):

include/inviwo/core/properties/buttongroupproperty.h
ResearchDaniel: This is what we have the property semantic for, i.e., different views of the same property. So we could create a horizontal layout semantic for the CompositeProperty. No new ButtonGroup property needed. Just as you say a widget to display buttons in a row

r-englund: But how would we layout Properties that are not ButtonProperties in a horizontal layout? IMO it only makes sense for button properties, hence it should not be a semantic for CompositeProperty

ResearchDaniel: Our FileProperty could have used such a feature for example, e.g., input field and a button. 

I don’t see a major reason to re-implement button behaviour because there are multiple ones. It disrupts other existing functionality such as linking.

r-englund: I guess ButtonGroupProperty could derive from CompositeProperty instead of property and adds it buttons and button properties. But instead of using CompositeProperty's widget it still uses its own widget.

ResearchDaniel: Agree

petersteneteg: I will never implement that. I you want to have a try at layouting our properties. which btw are all designed to all be layouted vertically, in a horiontal layout go ahead. But that is going to be a monumental effort and one that will only very special cases. And which probably will not look half as good as what we have now. And for what use? To redo the fileproperty? I have better things to do.

ResearchDaniel: It should be quite a small effort to change such that ButtonGroupProperty inherit from CompositeProperty and use existing button properties instead of new custom ones? The widget would remain unchanged...

This would make the ButtonGroupProperty work with other code and it would have a custom widget

petersteneteg: no.

r-englund: Thought about it some more. Making ButtonGroupProperty derive from CompositeProperty would make it possible for anyone to add additional properties to it which, IMO makes it a no go. 
Maybe possibly it could work with private inheritance of either CompositeProperty or PropertyOwner and only make public what is needed for the bindings. But I dont have a good feeling about such a solution, it does not feel stable and future proof.

martinfalk: I would say all of this is beyond this PR. I opt for leaving the buttons as they are.

src/core/algorithm/boundingbox.cpp
```diff
@@ -0,0 +1,194 @@
...
+        auto bb = boundingBox(*volume);
+        worldMin = glm::min(worldMin, vec3{bb * vec4{0.0f, 0.0f, 0.0f, 1.0f}});
+        worldMax = glm::max(worldMax, vec3{bb * vec4{1.0f, 1.0f, 1.0f, 1.0f}});
+    }
+    auto m = glm::scale(worldMax - worldMin);
+    m[3] = vec4(worldMin, 1.0f);
+    return m;
+}
+
+std::function<std::optional<mat4>()> boundingBox(const DataInport<Mesh> &mesh) {
```
r-englund, line 128: I suppose this could be templated to support all MeshPorts as: 
```c++
template<unsigned N, bool Flat>
std::function<std::optional<mat4>()> boundingBox(const DataInport<Mesh,N,Flat> &mesh) {
    return [port = &mesh]() -> std::optional<mat4> {
        if (port->hasData()) {
            if constexpr (N == 1){
                return boundingBox(*port->getData());
            } else {
                return boundingBox(port->getVectorData());
            }
        } else {
            return std::nullopt;
        }
    };
}
```

include/inviwo/core/properties/buttongroupproperty.h
r-englund: :+1:

src/core/algorithm/boundingbox.cpp
petersteneteg: There are overloads for the different versions we use, regular multi and flat. Wanted to avoid some includes... but forward decls was difficult

modules/base/src/algorithm/camerautils.cpp
```diff
@@ -0,0 +1,207 @@
...
+        case Side::YNegative:
+            return vec3(1, 0, 0);
+        case Side::YPositive:
+            return vec3(1, 0, 0);
+        case Side::ZNegative:
+            return vec3(0, 1, 0);
+        case Side::ZPositive:
+            return vec3(0, 1, 0);
+        default:
+            return vec3(0);
```
martinfalk: that's a very bad idea. Why not the default up vector, i.e. `(0, 1, 0)` ?

modules/base/src/algorithm/camerautils.cpp
```diff
@@ -0,0 +1,207 @@
...
+        case Side::YNegative:
+            return vec3(0, 1, 0);
+        case Side::YPositive:
+            return vec3(0, -1, 0);
+        case Side::ZNegative:
+            return vec3(0, 0, 1);
+        case Side::ZPositive:
+            return vec3(0, 0, -1);
+        default:
+            return vec3(0);
```
martinfalk: good idea to return 0?

modules/qtwidgets/src/properties/buttongrouppropertywidgetqt.cpp
```diff
@@ -0,0 +1,82 @@
...
+        }
+        if (button.name) {
+            buttonWidget->setText(utilqt::toQString(*button.name));
+        }
+
+        connect(buttonWidget, &QPushButton::released, this, [this, i]() {
+            if (!property_->getReadOnly()) property_->pressButton(i);
+        });
+        setFocusPolicy(buttonWidget->focusPolicy());
+        setFocusProxy(buttonWidget);
```
martinfalk, line 73: Why are `setFocusPolicy()` and `setFocusProxy()` called for each button? Once is sufficient.

src/core/properties/cameraproperty.cpp
```diff
@@ -46,44 +42,33 @@ const std::string CameraProperty::classIdentifier = "org.inviwo.CameraProperty";
 std::string CameraProperty::getClassIdentifier() const { return classIdentifier; }
 
 CameraProperty::CameraProperty(std::string identifier, std::string displayName, vec3 eye,
-                               vec3 center, vec3 lookUp, Inport* inport,
+                               vec3 center, vec3 lookUp, Inport* /*inport*/,
                                InvalidationLevel invalidationLevel, PropertySemantics semantics)
```
martinfalk: `inport` never used or was it just removed?

include/inviwo/core/properties/propertyowner.h
```diff
@@ -81,7 +81,7 @@ class IVW_CORE_API PropertyOwner : public PropertyOwnerObservable,
      */
     virtual void insertProperty(size_t index, Property& property);
 
-    virtual Property* removeProperty(const std::string& identifier);
+    virtual Property* removeProperty(const std::string& identifier);
```
martinfalk, line 84: trailing spaces?

martinfalk (approved):

petersteneteg:
replaced by #656

PR #653, opened 2019-08-19
Feature/line tracing cleanup

r-englund:
Removed some template SFINAE helper functions and use constexpr if instead 
Also added some more regression test to ensure not breaking old stuff.


petersteneteg (approved):
seems reasonable

PR #657, opened 2019-08-23
Feature/pcp colormaps

ResearchDaniel:
Added DataFrameColormapProperty and ColormapProperty to select predefined colormaps for a column.

* Categorical columns automatically use appropriate colormap
* Old workspaces will use previously defined transfer function. Disable "Override" in the new Colormap property

Roughly how it looked before:
image

New 
image

### DIFF
--- a/include/inviwo/core/util/colorbrewer.h
+++ b/include/inviwo/core/util/colorbrewer.h
@@ -32,8 +32,7 @@ This complete file is auto-generated with python script
 tools/codegen/colorbrewer/colorbrewer.py
 */
 
-#ifndef IWW_COLORBREWER_H
-#define IWW_COLORBREWER_H
+#pragma once
 
 #include <inviwo/core/common/inviwocoredefine.h>
 #include <inviwo/core/common/inviwo.h>
@@ -68,7 +67,6 @@ public:
         : Exception(message, context) {}
     virtual ~ColorBrewerTooManyException() throw() {}
 };
-
 // clang-format off
 enum class Colormap {
     Accent_3, Accent_4, Accent_5, Accent_6, Accent_7, Accent_8, 
@@ -108,7 +106,6 @@ enum class Colormap {
     YlOrRd_3, YlOrRd_4, YlOrRd_5, YlOrRd_6, YlOrRd_7, YlOrRd_8, 
     FirstMap=Accent_3, LastMap=YlOrRd_8
 };
-// clang-format on
 
 enum class Category { Diverging, Qualitative, Sequential, NumberOfColormapCategories, Undefined };
 
@@ -151,7 +148,7 @@ enum class Family {
     NumberOfColormapFamilies,
     Undefined
 };
-
+// clang-format on
 template <class Elem, class Traits>
 std::basic_ostream<Elem, Traits> &operator<<(std::basic_ostream<Elem, Traits> &os,
                                              Colormap colormap) {
@@ -421,6 +418,7 @@ std::basic_ostream<Elem, Traits> &operator<<(std::basic_ostream<Elem, Traits> &o
     case Colormap::YlOrRd_6: os << "YlOrRd_6"; break;
     case Colormap::YlOrRd_7: os << "YlOrRd_7"; break;
     case Colormap::YlOrRd_8: os << "YlOrRd_8"; break;
+
     }
     // clang-format on
     return os;
@@ -446,7 +444,6 @@ std::basic_ostream<Elem, Traits> &operator<<(std::basic_ostream<Elem, Traits> &o
             os << "Undefined";
             break;
     }
-
     return os;
 }
 
@@ -491,6 +488,7 @@ std::basic_ostream<Elem, Traits> &operator<<(std::basic_ostream<Elem, Traits> &o
     case Family::YlOrRd: os << "YlOrRd"; break;
     case Family::NumberOfColormapFamilies: os << "NumberOfColormapFamilies"; break;
     case Family::Undefined: os << "Undefined"; break;
+
     }
     // clang-format on
     return os;
@@ -545,16 +543,16 @@ IVW_CORE_API glm::uint8 getMaxNumberOfColorsForFamily(const Family &family);
 IVW_CORE_API std::vector<Family> getFamiliesForCategory(const Category &category);
 
 /**
- * Returns a transfer function for the given parameters. 
- * 
+ * Returns a transfer function for the given parameters.
+ *
  * @param category according to ColorBrewer2
  * @param family color scheme name
  * @param discrete will make each color constant instead of linearly varying inbetween colors.
- * @param divergenceMidpoint in [0 1]. Only used when category is Diverging 
+ * @param divergenceMidpoint in [0 1]. Only used when category is Diverging
  **/
-IVW_CORE_API TransferFunction getTransferfunction(const Category &category, const Family &family, glm::uint8 nColors, bool discrete, double divergenceMidPoint);
+IVW_CORE_API TransferFunction getTransferFunction(const Category &category, const Family &family,
+                                                  glm::uint8 nColors, bool discrete,
+                                                  double divergenceMidPoint);
 
 }  // namespace colorbrewer
 }  // namespace inviwo
-
-#endif  // COLORBREWER_H
--- a/include/inviwo/core/util/colorbrewer.h
+++ b/include/inviwo/core/util/colorbrewer.h
@@ -37,6 +37,7 @@ tools/codegen/colorbrewer/colorbrewer.py
 
 #include <inviwo/core/common/inviwocoredefine.h>
 #include <inviwo/core/common/inviwo.h>
+#include <inviwo/core/datastructures/transferfunction.h>
 #include <vector>
 #include <ostream>
 
@@ -542,6 +543,16 @@ IVW_CORE_API glm::uint8 getMaxNumberOfColorsForFamily(const Family &family);
  * Returns all families contained in the specified category.
  **/
 IVW_CORE_API std::vector<Family> getFamiliesForCategory(const Category &category);
+
+/**
+ * Returns a transfer function for the given parameters. 
+ * 
+ * @param category according to ColorBrewer2
+ * @param family color scheme name
+ * @param discrete will make each color constant instead of linearly varying inbetween colors.
+ * @param divergenceMidpoint in [0 1]. Only used when category is Diverging 
+ **/
+IVW_CORE_API TransferFunction getTransferfunction(const Category &category, const Family &family, glm::uint8 nColors, bool discrete, double divergenceMidPoint);
 
 }  // namespace colorbrewer
 }  // namespace inviwo
--- a/modules/dataframe/include/inviwo/dataframe/properties/colormapproperty.h
+++ b/modules/dataframe/include/inviwo/dataframe/properties/colormapproperty.h
@@ -38,41 +38,85 @@
 #include <inviwo/core/util/colorbrewer.h>
 
 namespace inviwo {
+/**
+ * Continuous is suitable for ordered/sequential data.
+ * Categorical is suitable for data where there should be no magnitude difference between classes,
+ * such as volvo and audi.
+ */
+enum class ColormapType { Continous, Categorical };
 
-class IVW_MODULE_DATAFRAME_API ColorMapProperty : public CompositeProperty {
+template <class Elem, class Traits>
+std::basic_ostream<Elem, Traits>& operator<<(std::basic_ostream<Elem, Traits>& os,
+                                             ColormapType colormap) {
+    // clang-format off
+    switch (colormap) {
+        case ColormapType::Continous: os << "Continous"; break;
+        case ColormapType::Categorical: os << "Categorical"; break;
+    }
+    // clang-format on
+    return os;
+}
+
+/**
+ * \brief Selection of pre-defined color maps based on data type.
+ *
+ * The following data types are supported:
+ * Continous: Ordered/Sequential data progressing from low to high
+ * Continous discrete: Constant inbetween colors to emphasize differences in the scale.
+ * Continuous diverging: Emphasis on mid-point or critical values. Break in the middle.
+ *
+ * Categorical: Suitable for categorical/nominal data where there should be no magnitude difference
+ * between classes. Is discrete and cannot be diverging so those options will be hidden.
+ */
+class IVW_MODULE_DATAFRAME_API ColormapProperty : public CompositeProperty {
 public:
     virtual std::string getClassIdentifier() const override;
     static const std::string classIdentifier;
 
-    ColorMapProperty(std::string identifier,
-                     std::string displayName,
-                     colorbrewer::Category category = colorbrewer::Category::Sequential,
+    ColormapProperty(std::string identifier, std::string displayName,
+                     ColormapType type = ColormapType::Continous,
                      colorbrewer::Family family = colorbrewer::Family::Blues,
-                     size_t numColors = getMinNumberOfColorsForFamily(colorbrewer::Family::Blues), 
+                     size_t numColors = getMinNumberOfColorsForFamily(colorbrewer::Family::Blues),
                      InvalidationLevel invalidationLevel = InvalidationLevel::InvalidOutput,
                      PropertySemantics semantics = PropertySemantics::Default);
 
-    ColorMapProperty(const ColorMapProperty& rhs);
-    virtual ColorMapProperty* clone() const override;
+    ColormapProperty(const ColormapProperty& rhs);
+    virtual ColormapProperty* clone() const override;
 
-    virtual ~ColorMapProperty() = default;
+    virtual ~ColormapProperty() = default;
 
     virtual std::string getClassIdentifierForWidget() const override {
         return CompositeProperty::classIdentifier;
     }
 
-    //virtual void set(const Property* p) override;
-    TransferFunction get() const;
+    // Get settings according to Colorbrewer style
+    virtual colorbrewer::Category getCategory() const;
+    virtual colorbrewer::Family getFamily() const;
 
-    TemplateOptionProperty<colorbrewer::Category> category;
+    /**
+     * Update settings based on Column.
+     * Uses ColormapType::Categorical for CategoricalColumn and ColormapType::Continous otherwise.
+     * Divergence mid point and rawill be computed 
+     */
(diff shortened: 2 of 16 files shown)


include/inviwo/core/util/colorbrewer.h
```diff
@@ -543,6 +544,16 @@ IVW_CORE_API glm::uint8 getMaxNumberOfColorsForFamily(const Family &family);
...
 
+/**
+ * Returns a transfer function for the given parameters. 
+ * 
+ * @param category according to ColorBrewer2
+ * @param family color scheme name
+ * @param discrete will make each color constant instead of linearly varying inbetween colors.
+ * @param divergenceMidpoint in [0 1]. Only used when category is Diverging 
+ **/
+IVW_CORE_API TransferFunction getTransferfunction(const Category &category, const Family &family, glm::uint8 nColors, bool discrete, double divergenceMidPoint);
```
r-englund: This file is generated using a python script, see line 31

modules/dataframe/src/properties/colormapproperty.cpp
```diff
@@ -0,0 +1,166 @@
...
+                                   InvalidationLevel invalidationLevel, PropertySemantics semantics)
+    : CompositeProperty(identifier, displayName, invalidationLevel, semantics)
+    , type("type", "Type")
+    , colormap("colormap", "Colormap")
+    , diverging("diverging", "Diverging", false)
+    , divergenceMidPoint("midPoint", "Mid point", 0.5, 0., 1.)
+    , discrete("discrete", "Discrete")
+    , nColors("nColors", "Classes") {
+    using namespace colorbrewer;
+    auto categories = {ColormapType::Continous, ColormapType::Categorical};
```
petersteneteg: Its better to set this up in the constructor

modules/dataframe/src/properties/colormapproperty.cpp
```diff
@@ -0,0 +1,166 @@
...
+    , diverging(rhs.diverging)
+    , divergenceMidPoint(rhs.divergenceMidPoint)
+    , discrete(rhs.discrete) {
+    addProperty(type);
+    addProperty(colormap);
+    addProperty(diverging);
+    addProperty(divergenceMidPoint);
+
+    addProperty(discrete);
+    addProperty(nColors);
```
petersteneteg, line 115: Need to add the onchange stuff here to I believe

modules/dataframe/src/properties/colormapproperty.cpp
```diff
@@ -0,0 +1,166 @@
...
+}
+
+ColormapProperty* ColormapProperty::clone() const { return new ColormapProperty(*this); }
+
+colorbrewer::Category ColormapProperty::getCategory() const {
+    colorbrewer::Category cat;
+    // clang-format off
+    switch (type) {
+        case ColormapType::Continous:
```
petersteneteg: Why not use the default format?

petersteneteg: Might need a default to

petersteneteg (approved):

modules/dataframe/src/properties/colormapproperty.cpp
ResearchDaniel: > Why not use the default format?

Copied it from a place with many cases. Removed custom format.

> Might need a default to

Better to have an error if more type are added?

petersteneteg: think you will get a compiler warning in that case... you need to handle that one

include/inviwo/core/util/colorbrewer.h
ResearchDaniel: Ah, updated script

modules/dataframe/src/properties/colormapproperty.cpp
ResearchDaniel: done

modules/dataframe/src/properties/colormapproperty.cpp
ResearchDaniel: Not that I could see. I handle all cases

modules/dataframe/include/inviwo/dataframe/properties/colormapproperty.h
```diff
@@ -0,0 +1,123 @@
...
+#include <inviwo/core/properties/optionproperty.h>
+#include <inviwo/core/util/colorbrewer.h>
+
+namespace inviwo {
+/**
+ * Continuous is suitable for ordered/sequential data.
+ * Categorical is suitable for data where there should be no magnitude difference between classes,
+ * such as volvo and audi.
+ */
+enum class ColormapType { Continous, Categorical };
```
martinfalk: Typo: should be `Continuous` 
https://www.merriam-webster.com/dictionary/continuous

modules/dataframe/include/inviwo/dataframe/properties/colormapproperty.h
```diff
@@ -0,0 +1,123 @@
...
+
+template <class Elem, class Traits>
+std::basic_ostream<Elem, Traits>& operator<<(std::basic_ostream<Elem, Traits>& os,
+                                             ColormapType colormap) {
+    // clang-format off
+    switch (colormap) {
+        case ColormapType::Continous: os << "Continous"; break;
```
martinfalk: "Continuous"

modules/dataframe/include/inviwo/dataframe/properties/colormapproperty.h
```diff
@@ -0,0 +1,123 @@
...
+    // clang-format on
+    return os;
+}
+
+/**
+ * \brief Selection of pre-defined color maps based on data type.
+ *
+ * The following data types are supported:
+ * Continous: Ordered/Sequential data progressing from low to high
+ * Continous discrete: Constant inbetween colors to emphasize differences in the scale.
```
martinfalk: typo "continuous"

modules/dataframe/include/inviwo/dataframe/properties/colormapproperty.h
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    // Get settings according to Colorbrewer style
+    virtual colorbrewer::Category getCategory() const;
+    virtual colorbrewer::Family getFamily() const;
+
+    /**
+     * Update settings based on Column.
+     * Uses ColormapType::Categorical for CategoricalColumn and ColormapType::Continous otherwise.
+     * Divergence mid point and rawill be computed
```
martinfalk: `" and rawill be computed"` ?

modules/dataframe/include/inviwo/dataframe/properties/colormapproperty.h
```diff
@@ -0,0 +1,123 @@
...
+     */
+    TransferFunction getTransferFunction() const;
+
+    TemplateOptionProperty<ColormapType> type;
+    TemplateOptionProperty<colorbrewer::Family> colormap;
+    BoolProperty diverging;
+    DoubleProperty divergenceMidPoint;
+    BoolProperty discrete;
+    IntSizeTProperty nColors;
+
```
martinfalk, line 121: remove empty lines

modules/dataframe/include/inviwo/dataframe/properties/dataframecolormapproperty.h
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * It is possible to override the ColormapProperty
+ *
+ */
+class IVW_MODULE_DATAFRAME_API DataFrameColormapProperty : public CompositeProperty {
+public:
+    virtual std::string getClassIdentifier() const override;
+    static const std::string classIdentifier;
+
+    DataFrameColormapProperty(std::string identifier, std::string displayName, DataInport<DataFrame>& port);
```
martinfalk: What about `InvalidationLevel invalidationLevel = InvalidationLevel::InvalidResources, PropertySemantics semantics = PropertySemantics::Default);`

modules/dataframe/include/inviwo/dataframe/properties/dataframecolormapproperty.h
```diff
@@ -0,0 +1,71 @@
...
+
+    DataFrameColumnProperty selectedColorAxis;
+    BoolProperty overrideColormap;
+    TransferFunctionProperty tf;
+
+protected:
+
+    std::vector<ColormapProperty*> colormaps_;
+    std::shared_ptr<const DataFrame> dataframe_;
+    std::shared_ptr<std::function<void()>> colormapChanged_;
```
martinfalk, line 70: Why a shared pointer to a function pointer?

modules/dataframe/include/inviwo/dataframe/properties/colormapproperty.h
```diff
@@ -0,0 +1,123 @@
...
+    void setupForColumn(const Column& col);
+    void setupForColumn(const Column& col, double minVal, double maxVal);
+
+    /**
+     * Get TransferFunction given current settings
+     * @see colorbrewer::getTransferFunction
+     */
+    TransferFunction getTransferFunction() const;
+
+    TemplateOptionProperty<ColormapType> type;
```
martinfalk, line 115: no underscores for member variables?

modules/dataframe/src/properties/colormapproperty.cpp
```diff
@@ -0,0 +1,182 @@
...
+        }
+        nColors = numCategories;
+        discrete = true;
+    } else {
+        type.set(ColormapType::Continous);
+        // Better contrast when using many classes
+        nColors = getMaxNumberOfColorsForFamily(colormap);
+        discrete = false;
+    }
+    divergenceMidPoint.set(0.5 * (minVal + maxVal), minVal, maxVal, 0.1 * (maxVal - minVal));
```
martinfalk: is 10% really a good step value? Shouldn't it be rather 0.01?

modules/dataframe/src/properties/dataframecolormapproperty.cpp
```diff
@@ -0,0 +1,123 @@
...
+namespace inviwo {
+
+const std::string DataFrameColormapProperty::classIdentifier = "org.inviwo.DataFrameColormapProperty";
+std::string DataFrameColormapProperty::getClassIdentifier() const { return classIdentifier; }
+
+DataFrameColormapProperty::DataFrameColormapProperty(std::string identifier,
+                                                     std::string displayName,
+                                                     DataInport<DataFrame>& port)
+    : CompositeProperty(identifier, displayName)
+    , selectedColorAxis{"selectedColorAxis", "Column", port, false, 1}
```
martinfalk, line 44: members without underscore?

modules/dataframe/src/properties/dataframecolormapproperty.cpp
```diff
@@ -0,0 +1,123 @@
...
+    dataframe_ = dataframe;
+    auto oldColormapProperties = colormaps_;
+    colormaps_.clear();
+    for (size_t i = 0; i < dataframe_->getNumberOfColumns(); i++) {
+        auto c = dataframe_->getColumn(i);
+        std::string displayName = "Settings";
+        std::string identifier = util::stripIdentifier(c->getHeader());
+
+        auto colormapProp = [&]() -> ColormapProperty* {
+            if (auto p = getPropertyByIdentifier(identifier)) {
```
martinfalk, line 100: why not remove all properties stored in `colormaps_` beforehand?

martinfalk: you are not replacing them anyway.

modules/dataframe/src/properties/dataframecolormapproperty.cpp
```diff
@@ -0,0 +1,123 @@
...
+            newProp->setVisible(false);
+            newProp->setSerializationMode(PropertySerializationMode::All);
+            insertProperty(1, newProp.release());
+            return ptr;
+        }();
+        colormaps_.push_back(colormapProp);
+        colormapProp->setupForColumn(*c);
+    }
+    // Remove properties that were not reused
+    for (auto p : oldColormapProperties) {
```
martinfalk, line 117: see comment above

modules/plottinggl/include/modules/plottinggl/processors/parallelcoordinates/pcpaxissettings.h
```diff
@@ -35,6 +35,7 @@
 #include <inviwo/core/properties/boolproperty.h>
 #include <modules/opengl/texture/texture2d.h>
 
+#include <inviwo/dataframe/properties/colormapproperty.h>
```
martinfalk: used in here?

modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
```diff
@@ -388,13 +382,14 @@ void ParallelCoordinates::createOrUpdateProperties() {
         // initialize corresponding flipped flag for the line shader
         lines_.axisFlipped[i] = static_cast<int>(prop->invertRange);
 
-        axes_.push_back({prop, std::move(renderer), std::move(slider)});
+        axes_.push_back({prop, std::move(renderer), std::move(slider)});
```
martinfalk: clang format

martinfalk (changes requested):
Nice work.

modules/dataframe/src/properties/colormapproperty.cpp
r-englund: Would it make sense to set it to divergenceMidPoint zero instead of `(0.5 * (minVal + maxVal)` when `minVal < 0` and  `maxVal > 1`?

martinfalk: That might be a good initial guess in this case!

ResearchDaniel: > is 10% really a good step value? Shouldn't it be rather 0.01?

As good as any ;) Changed it to 0.01

> Would it make sense to set it to divergenceMidPoint zero instead of (0.5 * (minVal + maxVal) when minVal < 0 and maxVal > 1?

Good point. Let's try it and see how it works

modules/dataframe/src/properties/dataframecolormapproperty.cpp
ResearchDaniel: Done it because they are public

modules/dataframe/src/properties/dataframecolormapproperty.cpp
ResearchDaniel: They are removed on line 89 and reused if the same column in the new DataFrame is found

modules/plottinggl/include/modules/plottinggl/processors/parallelcoordinates/pcpaxissettings.h
ResearchDaniel: Good catch, not since refactoring. Removed it

ResearchDaniel:
@petersteneteg Some unrelated PR fail on jenkins:

> dlopen /Users/jenkins/jenkins/workspace/inviwo_PR-657/build/bin/./../Frameworks/Chromium Embedded Framework.framework/Chromium Embedded Framework: dlopen(/Users/jenkins/jenkins/workspace/inviwo_PR-657/build/bin/./../Frameworks/Chromium Embedded Framework.framework/Chromium Embedded Framework, 261): image not found

> 11:09:08  Failed to load the CEF framework.

petersteneteg:
@ResearchDaniel 
this is the output on "master"
```
00:00:41.825  + ./inviwo-integrationtests
00:00:41.825  2019-08-21 17:45:48.532 inviwo-integrationtests[43532:15397079] ApplePersistence=NO
00:00:42.378  dlopen /Users/jenkins/jenkins/workspace/inviwo_master/build/bin/./../Frameworks/Chromium Embedded Framework.framework/Chromium Embedded Framework: dlopen(/Users/jenkins/jenkins/workspace/inviwo_master/build/bin/./../Frameworks/Chromium Embedded Framework.framework/Chromium Embedded Framework, 261): image not found
00:00:42.378  Failed to load the CEF framework.
00:00:42.378  [==========] Running 443 tests from 7 test suites.
00:00:43.228  [==========] 443 tests from 7 test suites ran. (1028 ms total)
00:00:43.228  [  PASSED  ] 443 tests.
``` 

so the problem is the segfault not the CEF stuff...

ResearchDaniel:
@martinfalk I think all things are addressed now. Not sure about our naming scheme for public properties, but having underscore is a bit ugly IMHO

petersteneteg:
About the underscore... I guess it is a bit in a gray zone... for a private of protected member in a class we always use trailing _. For a POD struct with all public members we don't add any trailing _.
In most cases a class (at least one with some invariant) never have any public members, and in that case there is no problem. 
The gray zone is for classes with both private and public members. In that case we don't have a consistent style right now. I can argue in both directions.

The reason for have the trailing _ is to easily see in a member function of a class if a variable is a member or a local variable. In a POD struct there is usually no member functions, hence this is not needed.
In a class with member functions on the other hand the need for the _ is equally important irregardless of whether the member is public of not, with regards to being able to differentiate it from a local variable in a member function.

On the other hand when using a class member from the "outside" of the class it reads much nicer without the _.

But since mixing both _ and no _ in the same class makes the primary use case of having them there to begin with void. So if we do that we could just as well remove all _. Hence I think my own conclusion is that in a class or struct one can only omit the _ if all members a public and non of them have a _. i.e. in the case of a POD struct. For all other cases we should use _. But this is just my current opinion, and I could likely be convinced in an other direction if arguments are presented :)

ResearchDaniel:
@petersteneteg good argumentation. I do think that it will look a bit inconsistent from an API perspektivet if some public  membrers have underscore while some don’t. So, i would advocate for the approach taken here :)

petersteneteg:
If we go that way we need to consider if we should drop all _ then...

ResearchDaniel:
@petersteneteg I think that it makes sense to keep the _ in protected/private. Also since we don't want to change that much code.

petersteneteg:
We only need the change a few lines of your code ;)

martinfalk (approved):

r-englund:
Why do we allow public properties in the first place? Why aren't they encapsulated in get functions?

petersteneteg:
@r-englund we have get functions in the property owner which makes all properties kind of public...

ResearchDaniel:
@r-Englund actually, I think that we might have mistake in the beginning when making them private since most properties are parameters that are supposed to be changed by a user/developer

PR #659, opened 2019-08-23
Add options for linking external libraries for the image libs and hdf5

rartino:
I'm sure the bundled libraries are nice in some contexts. However, since Qt and Python are external dependencies, these are likely to have been built against other versions of the same libraries. 

Different OS:es handle the case of dependencies against more than one version of the same dynamic library different, but there is no way it can be handled perfectly safely. In particular on Linux, these libraries are only loaded once, and hence, some parts of the code interact with different versions of the libraries than they were complied against.

This has generated issues that I + students we've supervised in the electronic visualization project have wrestled with for years... Examples:
- In many of our builds, loading the h5py module inside python have crashed Inviwo, because h5py is built against a different version than the bundled one coming with the Inviwo hdf5 processor. HDF5 carefully checks for this mismatch and deliberately crashes the application. 
- In many builds, opening a file dialog crashes Inviwo, because Qt opens the GTK file dialog, which uses the system libpng, which apparently differs enough from the Inviwo-bundled one that the application crashes sometimes.
- I *think* this issue may also be behind all button graphics missing in some builds.

So, it would be nice to at least have the option to turn off bundled libraries and instead use cmake's find_package to build with what the system provides.

Hence, this PR adds two new cmake options:
- IVW_HDF5_USE_EXTERNAL: Link with external HDF5 library instead of building it.
- IVW_IMG_USE_EXTERNAL: Link with external image libraries (zlib, png, jpeg, tiff) instead of building them.

This is tested on Ubuntu Linux 19.04. The PR should have no effect unless the provided options are turned ON. Turning them ON seems to do wonders for my Inviwo stability...


petersteneteg:
Sweet!
This has been on my todo for a long time. But since almost no one here is using linux, (or complaining high enough...) It has been at the bottom of the pile...

ResearchDaniel (approved):
Looks good!
@petersteneteg Packaging (building an installer) will not work when using external libraries? That might be something that we want to tell the user in that case

rartino:
In my own tests I've noticed that the cmake feature I use for making the image libraries globally available under the name the Inviwo build expects, e.g.,:
```
add_library(inviwo::libpng ALIAS PNG::PNG)
```
actually seems to be a new feature in cmake 3.15 when the right hand side of the alias is an external library.

Should I insert a `cmake_minimum_required(3,15)` specifically just in this branch of the build, or does anyone know another way of achieving the same thing without renaming `inviwo::libpng` to `PNG::PNG`, etc., troughout the Inviwo build system?

petersteneteg:
Hi
Is it ok if I cherry pick this add go with the concept a bit further. 
I would like to have all the usage of the libs go through find_packages when ever you need them.
When we build the libs our self we automatically create a set of Find<lib>.cmake files that override the system ones. I have already done this for zlib for example. And would like to do the same for the other libs here.

Also I would prefer to name the cmake options IVW_USE_EXTERNAL_IMG, IVW_USE_EXTERNAL_HDF5 instead. 

Also should we use IVW_USE_EXTERNAL_IMG or one for each of the dependencies? 

we should not use the inviwo::libjpeg alias and use JPEG::JPEG instead. The will remove the requirement of using cmake 3.15... and is also needed to get the find_package stuff to work properly. 
I don't think there is any reason we need to use our own aliases anymore. 

Hence I would prefer not to merge this quite yet...

rartino:
Of course do feel free to edit/rework as you'd like. We'll anyway keep the original PR patch in our set of Inviwo patches for the electron visualization project until you get something equivalent merged.

But, if it is acceptable to merge a patch with `s/inviwo::libjpeg/JPEG::JPEG/` throughout the build system, perhaps that could be a first step, and then the more advanced handling with find_package everywhere could follow later. 

It would make sense to have separate IVW_USE_EXTERNAL_* for all dependencies that usually are available as system dynamic libraries. But perhaps one can also consider a more broad IVW_USE_EXTERNAL_ALL to avoid having to track all of the individual flags down for those who want to link as many libraries as possible.

PR #661, opened 2019-08-26
Feature/volumereader

martinfalk:
Some fixes in the dat-raw and ivf volume readers/writers
* exposed byte offset for data (`DataOffset`) in dat and ivf file (previously only in Raw import dialog)
* ivf writer fix to handle unicode filenames
* ticpp fix for unicode filenames
* minor clean-up

### DIFF
--- a/modules/base/include/modules/base/io/datvolumesequencereader.h
+++ b/modules/base/include/modules/base/io/datvolumesequencereader.h
@@ -45,12 +45,14 @@ namespace inviwo {
  *   - __Rawfile__ The name of the raw data file, should be in the same directory (Mandatory).
  *   - __ByteOrder__ the byte order in the raw data file. (Optional, LittleEndian|BigEndian,
  *     default: LittleEndian).
+ *   - __DataOffset__  offset in byte to where the data starts
  *   - __Resolution | Dimension__ The size of the data grid: nx,ny,nz (Mandatory).
  *   - __Format__ The type of values in the raw file. (Mandatory)
  *   - __Spacing | SliceThickness__ The size of the voxels in the data. (Optional)
  *   - __BasisVector(1|2|3)__ Defines a coordinate system for the data. (Optional, overides spacing,
  *     default: 2*IdentityMatrix);
- *   - __Offset__ Offsets the basisvecors in space. (Optional, defaults to center the data on origo)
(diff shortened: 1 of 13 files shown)


r-englund (approved):
Is there a reason why the byte offset is called `DataOffset` and not `ByteOffset`? Calling it DataOffset makes it ambiguous if it is a byte offset or number of voxel offset.

modules/base/include/modules/base/io/datvolumesequencereader.h
```diff
@@ -45,12 +45,14 @@ namespace inviwo {
...
  *   - __ByteOrder__ the byte order in the raw data file. (Optional, LittleEndian|BigEndian,
  *     default: LittleEndian).
+ *   - __DataOffset__  offset in byte to where the data starts
  *   - __Resolution | Dimension__ The size of the data grid: nx,ny,nz (Mandatory).
  *   - __Format__ The type of values in the raw file. (Mandatory)
  *   - __Spacing | SliceThickness__ The size of the voxels in the data. (Optional)
  *   - __BasisVector(1|2|3)__ Defines a coordinate system for the data. (Optional, overides spacing,
  *     default: 2*IdentityMatrix);
- *   - __Offset__ Offsets the basisvecors in space. (Optional, defaults to center the data on origo)
+ *   - __Offset__ Offsets the basisvecors in space. (Optional, defaults to center the data at the
```
petersteneteg: vector

petersteneteg (approved):

ResearchDaniel:
@r-Englund I had the same thought. Perhaps it should be changed to byte offset since two people thought about it

martinfalk:
@ResearchDaniel, @r-englund `dataOffset` was chosen to be consistent with the existing interface of the `RawVolumeReader`
https://github.com/inviwo/inviwo/blob/5d5f850c197d212fbdd4ffa5bc64af4c4f76e1a4/include/inviwo/core/io/rawvolumereader.h#L56-L57

petersteneteg:
Think we should rename the parameter in the rawreader to byteoffset as well

martinfalk:
ByteOffset could also imply that there is an offset for each data value...

petersteneteg:
Per voxel offset would be called stride something...

martinfalk:
Renamed `dataOffset` to `byteOffset`

petersteneteg (approved):
Nice!

PR #664, opened 2019-08-27
Feature/various

r-englund:
Committed various of my local changes. 
See commit messages.

### DIFF
--- a/include/inviwo/core/util/foreach.h
+++ b/include/inviwo/core/util/foreach.h
@@ -31,11 +31,9 @@
 #define IVW_FOREACH_H
 
 #include <inviwo/core/common/inviwocoredefine.h>
-#include <inviwo/core/util/exception.h>
-#include <inviwo/core/util/stdextensions.h>
 #include <inviwo/core/common/inviwoapplication.h>
 #include <inviwo/core/util/settings/systemsettings.h>
-#include <string>
+
 #include <utility>
 
 namespace inviwo {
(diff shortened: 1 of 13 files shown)


include/inviwo/core/util/foreach.h
```diff
@@ -157,6 +157,14 @@ void forEachParallel(const Iterable& iterable, Callback&& callback, size_t jobs
     }
 }
 
+template <typename Iterable, typename Callback>
+void forEach(const Iterable& iterable, Callback&& callback) {
+    using value_type = decltype(*begin(iterable));
+    using IncludeIndexType = typename util::is_invocable<Callback, value_type, size_t>::type;
```
petersteneteg: use std::is_invocable, 

could just use
``` 
if constexpr ( std::is_invocable_v<Callback, value_type, size_t>) {
    dispatch with index
} else {
    dispatch without index
}
```

include/inviwo/core/util/stdextensions.h
```diff
@@ -123,6 +123,19 @@ T* defaultConstructType() {
 template <class...>
 using void_t = void;
 
+/**
+* Helper struct to allow passing multiple lambda expressions to std::visit.
```
petersteneteg, line 127: Nice

modules/base/src/processors/surfaceextractionprocessor.cpp
```diff
@@ -135,6 +137,7 @@ void SurfaceExtraction::process() {
             result_[i].set(method, iso, color, invert, enclose, 0.0f,
                            dispatchPool([this, vol, method, iso, color, invert, enclose,
                                          i]() -> std::shared_ptr<Mesh> {
+                            RenderContext::getPtr()->activateLocalRenderContext();
```
petersteneteg: lets do this for now, but I think you might be right that it would be better to add this to VolumeGL2RAMConverter or maybe texture... Not sure if this is cheep as long as there is no  actually change of context... or of even that could be expensive?

petersteneteg (changes requested):

r-englund:
Ready for re-review

petersteneteg (approved):

PR #667, opened 2019-08-27
Webbrowser: Fix for "Blocked a frame with origin "null" from accessin…

ResearchDaniel:
…g a frame with origin "null". Protocols, domains, and ports must match.", see https://magpcss.org/ceforum/viewtopic.php?f=14&t=10892


r-englund (approved):

ResearchDaniel:
@petersteneteg The warning check does not seem to work well. Fails due to FileTreeWidget, which has not been modified here. Is there an easy fix for this? Also, we need to make the reason for failure more obvious, just unstable build now

petersteneteg:
@ResearchDaniel if you can figure out how to do that without always doing clean builds I will be happy to accept that PR...

PR #670, opened 2019-08-28
External libs

petersteneteg:
replaces #659


ResearchDaniel (approved):

rartino:
I'm not sure if this is because I am doing something I shouldn't, but when trying the same cmake setup that worked for me for #659, it fails with this PR with:
```
CMake Error at cmake/licenses.cmake:78 (get_target_property):
  get_target_property() called with non-existent target "ZLIB::ZLIB".
Call Stack (most recent call first):
  src/core/CMakeLists.txt:814 (ivw_register_license_file)
```
Where I note that the first `Found ZLIB` comes later. 

Full cmake output of #670: https://pastebin.com/kVt8AZAe
Full cmake output of #659: https://pastebin.com/TtMucqKx

rartino:
I'm guessing the fix for the license file is to move adding a zlib license file from cmake/licenses.cmake to ext/zlib-1.2.11/CMakeLists.txt.

However, if I try that I get other complaints of ZLIB::ZLIB missing: https://pastebin.com/QEHDzU4U

**Edit**: I've updated #659 to the modified version that moves the zlib ivw_register_license_file, so feel free to cherry-pick 1f3a5061f8daf64094453da6d9f4e94cb6966792 into this PR if it seems the right thing to do.

rartino:
I took a crack at adding missing find_package calls to the appropriate places under modules/nifti, so it is fixed in 5c922da0a7074c6b1822ec706fef2366b1211c44 (pushed to #659.) 

Now the cmake step works. I've started a compile and will let you know if the actual build seems ok.

rartino:
The build went through, the system libraries are linked and everything seems stable. Hence, feel free to merge  #659 or, if you want to do further adjustments, cherry-pick into this PR 1f3a5061f8daf64094453da6d9f4e94cb6966792 and 5c922da0a7074c6b1822ec706fef2366b1211c44.

rartino (approved):
This version looks good, builds, and runs as expected for me.

petersteneteg:
Tried this on osx too and is seems to work

PR #673, opened 2019-08-29
Update required Qt version to 5.12

### LINKED ISSUE #671: Update Qt version dependency
**Issue**
Get a compile error from qt-usage even though everything goes through without errors in CMake.

Cmake should produce an error if an old Qt version is used. Get a compile the same compile error as in #398
We need to update cmake scripts as well as the Wiki. Currently we support version 5.12

**Environment information**
* Operating System: Mac
* Compiler/IDE: XCode
* QT Version: 5.8

ResearchDaniel:
Closes issue #671

### DIFF
--- a/src/qt/editor/CMakeLists.txt
+++ b/src/qt/editor/CMakeLists.txt
@@ -112,7 +112,7 @@ ivw_group("Source Files" ${SOURCE_FILES})
 
 #--------------------------------------------------------------------
 # Create library
-find_package(Qt5Core QUIET REQUIRED)
+find_package(Qt5Core 5.12 QUIET REQUIRED)
 find_package(Qt5Gui QUIET REQUIRED)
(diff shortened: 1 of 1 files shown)


src/qt/editor/CMakeLists.txt
```diff
@@ -112,7 +112,7 @@ ivw_group("Source Files" ${SOURCE_FILES})
 
 #--------------------------------------------------------------------
 # Create library
-find_package(Qt5Core QUIET REQUIRED)
+find_package(Qt5Core 5.12 QUIET REQUIRED)
```
petersteneteg, line 115: Should this be quiet here?

petersteneteg (approved):

src/qt/editor/CMakeLists.txt
petersteneteg: seems it does not matter much

PR #681, opened 2019-09-02
QtWidgets: Button group bug fix

ResearchDaniel:


petersteneteg:
How did you end up with an empty button group property? Don't think we support adding buttons after the widgets be created anyway...

ResearchDaniel:
TestProcessor creates an instance of the property without buttons

petersteneteg (approved):

PR #684, opened 2019-09-03
line adjacency 

petersteneteg:
Extended the mesh shader cache to support more kinds of state. Made the line renderer handle adjacency property by inspecting the input data.

### DIFF
--- a/modules/basegl/src/datastructures/meshshadercache.cpp
+++ b/modules/basegl/src/datastructures/meshshadercache.cpp
@@ -37,53 +37,48 @@ namespace inviwo {
 
 MeshShaderCache::Requirement::Requirement(BufferType bufferType, RequireBuffer required,
                                           const std::string& glslType, const std::string& name)
-    : bufferType{bufferType}
-    , required{required}
-    , glslType{glslType}
-    , name{name.empty() ? toString(bufferType) : name} {}
+    : getState{[bufferType, required](const Mesh& mesh, Mesh::MeshInfo) -> int {
+        const auto res = mesh.findBuffer(bufferType);
+        if (res.first) {
+            return res.second;
+        } else if (required == Mandatory) {
+            throw Exception("Unsupported mesh type, a " + toString(bufferType) + " is needed",
+                            IVW_CONTEXT_CUSTOM("MeshShaderCache"));
+        } else {
+            return -1;
+        }
+    }}
+    , updateShader{[glslType, name = name.empty() ? toString(bufferType) : name](
+                       int location, Shader& shader) -> void {
+        if (location >= 0) {
+            const auto& buffername = name;
+            shader[ShaderType::Vertex]->addInDeclaration("in_" + buffername, location, glslType);
+            for (auto& obj : shader.getShaderObjects()) {
+                obj.addShaderDefine("HAS_" + toUpper(buffername));
+            }
+        }
+    }} {}
+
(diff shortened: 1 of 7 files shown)


modules/basegl/src/datastructures/meshshadercache.cpp
```diff
@@ -37,53 +37,48 @@ namespace inviwo {
...
+            return -1;
+        }
+    }}
+    , updateShader{[glslType, name = name.empty() ? toString(bufferType) : name](
+                       int location, Shader& shader) -> void {
+        if (location >= 0) {
+            const auto& buffername = name;
+            shader[ShaderType::Vertex]->addInDeclaration("in_" + buffername, location, glslType);
+            for (auto& obj : shader.getShaderObjects()) {
+                obj.addShaderDefine("HAS_" + toUpper(buffername));
```
ResearchDaniel, line 57: Defines never removed?

modules/basegl/src/processors/linerenderer.cpp
```diff
@@ -80,8 +68,16 @@ LineRenderer::LineRenderer()
...
-
+           {BufferType::PickingAttrib, MeshShaderCache::Optional, "uint"},
+           {[](const Mesh&, Mesh::MeshInfo mi) -> int {
+                return mi.ct == ConnectivityType::Adjacency ||
+                               mi.ct == ConnectivityType::StripAdjacency
+                           ? 1
+                           : 0;
+            },
+            [](int mode, Shader& shader) {
+                shader[ShaderType::Geometry]->addShaderDefine("ENABLE_ADJACENCY", toString(mode));
```
ResearchDaniel, line 79: Never removed?

ResearchDaniel (approved):
Looks good. Might be an issue with defines added but never removed

modules/basegl/src/datastructures/meshshadercache.cpp
petersteneteg: It only does run that when the shader is created, if a different value is needed a different shader is created. Hence it should never remove anything

modules/basegl/src/processors/linerenderer.cpp
petersteneteg: see above

modules/basegl/src/processors/linerenderer.cpp
```diff
@@ -146,53 +130,36 @@ void LineRenderer::process() {
...
-        utilgl::setShaderUniforms(shader, *elem, "geometry");
-        if (autoDrawMode) {
-            drawer.draw();
+    for (const auto& mesh : inport_) {
+        if (mesh->getNumberOfBuffers() == 0) continue;
+        MeshDrawerGL::DrawObject drawer(mesh->getRepresentation<MeshGL>(),
+                                        mesh->getDefaultMeshInfo());
+        if (mesh->getNumberOfIndicies() > 0) {
+            for (std::size_t i = 0; i < mesh->getNumberOfIndicies(); ++i) {
+                if (mesh->getIndexMeshInfo(i).dt != DrawType::Lines) continue;
```
martinfalk, line 139: so one has to explicitly convert a tri mesh into lines before using this processor?

martinfalk (approved):

modules/basegl/src/processors/linerenderer.cpp
petersteneteg: Yes. I think we should optimise for the standard use case.

PR #687, opened 2019-09-04
Feature/shader fixes

r-englund:
* Added onReloadScoped to Shader
* Renamed isReady to checkLinkStatus and introduced a new isReady. When Querying the GL_LINK_STATUS you get the result of the previous call to glLinkProgram, which will not have been called if a shader failed to compile.

### DIFF
--- a/modules/opengl/include/modules/opengl/shader/shader.h
+++ b/modules/opengl/include/modules/opengl/shader/shader.h
@@ -150,6 +150,7 @@ public:
 
     // Callback when the shader is reloaded. A reload can for example be triggered by a file change.
     const BaseCallBack *onReload(std::function<void()> callback);
+    std::shared_ptr<std::function<void()>> onReloadScoped(std::function<void()> callback);
     void removeOnReload(const BaseCallBack *callback);
 
 private:
--- a/modules/opengl/include/modules/opengl/shader/shader.h
+++ b/modules/opengl/include/modules/opengl/shader/shader.h
@@ -120,6 +120,7 @@ public:
     void link();
     void build();
     bool isReady() const;  // returns whether the shader has been built and linked successfully
+    bool checkLinkStatus() const;
 
     GLuint getID() const { return program_.id; }
 
--- a/modules/opengl/include/modules/opengl/shader/shader.h
+++ b/modules/opengl/include/modules/opengl/shader/shader.h
@@ -120,7 +120,6 @@ public:
     void link();
     void build();
     bool isReady() const;  // returns whether the shader has been built and linked successfully
-    bool checkLinkStatus() const;
 
     GLuint getID() const { return program_.id; }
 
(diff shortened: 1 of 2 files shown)


modules/opengl/include/modules/opengl/shader/shader.h
```diff
@@ -120,6 +120,7 @@ class IVW_MODULE_OPENGL_API Shader {
     void link();
     void build();
     bool isReady() const;  // returns whether the shader has been built and linked successfully
+    bool checkLinkStatus() const;
```
petersteneteg: Do we have to introduce an other "ready" function in the public interface.

modules/opengl/src/shader/shader.cpp
```diff
@@ -370,7 +370,9 @@ std::string Shader::processLog(std::string log) const {
     return result.str();
 }
 
-bool Shader::isReady() const {
+bool Shader::isReady() const { return ready_ && checkLinkStatus(); }
```
petersteneteg: do we have to call checkLinkStatus here? would it not be enough to return ready_? we only make that true if we have built and linked right?

modules/opengl/src/shader/shader.cpp
```diff
@@ -395,6 +397,10 @@ const BaseCallBack *Shader::onReload(std::function<void()> callback) {
     return onReloadCallback_.addLambdaCallback(callback);
 }
 
+std::shared_ptr<std::function<void()>> Shader::onReloadScoped(std::function<void()> callback) {
```
petersteneteg, line 400: 👍

petersteneteg (changes requested):
Nice catch with the link error...

modules/opengl/src/shader/shader.cpp
martinfalk: is that state reset when a shader is reloaded?

modules/opengl/include/modules/opengl/shader/shader.h
r-englund: I will make this one private

modules/opengl/src/shader/shader.cpp
petersteneteg: think so

petersteneteg: yes, a file change will trigger a rebuild which starts by setting ready=false

r-englund: By looking at the code it seems we can remove the `&& checkLinkStatus()`, I am testing it now.

martinfalk: 👍

r-englund:
Comments has been addressed, ready for re-review and merge

petersteneteg (approved):

martinfalk (approved):

PR #689, opened 2019-09-04
Core: Removed deprecated methods from TransferFunction

r-englund:
Various method in the TransferFunction class has been marked as deprecated for way over a year now. I keep getting annoyed when IntelliSense suggests them in the completion menu for TF objects.


martinfalk:
good point. There also have been a number of releases in the meantime.

martinfalk (approved):

petersteneteg:
Nice :)

petersteneteg:
Intellisense should obviously not show deprecated methods

PR #695, opened 2019-09-06
QtEditor: Added support for more floating property list widgets

petersteneteg:


martinfalk (approved):
Nice. Might be of interest to add only specific properties, e.g. via drag&drop.

PR #697, opened 2019-09-08
Core//UserInterfaceGL: Treat Stationary TouchState as click on TouchP…

ResearchDaniel:
…ad (Tested on Mac). Use PickingEvent states instead of converting to Mouse/TouchEvent in glUI Element and CameraWidget

- Needs to be tested on Windows
- Need to include touch event states similar to mouse states so that the pressStates preserves the original "button" starting the event sequence. 

These points need to be done before merging.

### DIFF
--- a/include/inviwo/core/interaction/events/touchevent.h
+++ b/include/inviwo/core/interaction/events/touchevent.h
@@ -178,7 +178,7 @@ private:
 class IVW_CORE_API TouchEvent : public InteractionEvent {
 public:
     TouchEvent();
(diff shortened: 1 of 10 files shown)


petersteneteg (approved):
Looks good from what I can see

r-englund:
I've have this branched compiled on my windows machine with touch monitor if you wanna test

ResearchDaniel:
@r-englund can you try parallel coordinates on the touch monitor?

ResearchDaniel:
New in #720

PR #703, opened 2019-09-10
BaseGL: Added min/max ranges to layout components of ImageLayoutGL

ResearchDaniel:
Now possible to restrict sizes of the layout components in ImageLayoutGL.

### DIFF
--- a/modules/basegl/include/modules/basegl/processors/imageprocessing/imagelayoutgl.h
+++ b/modules/basegl/include/modules/basegl/processors/imageprocessing/imagelayoutgl.h
@@ -107,7 +107,6 @@ public:
 protected:
     virtual void process() override;
 
-    void updateMaxSizes(ivec2 outputDim);
     void updateViewports(ivec2 size, bool force = false);
     void onStatusChange(bool propagate = true);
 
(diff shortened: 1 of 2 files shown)


ResearchDaniel:
Discovered one issue when making the window smaller than the specified min/max. Making the canvas larger again does not preserve the previously set min/max. Perhaps we could use the default values for this? E.g. setting the default if a resize event has smaller size and then restoring when/if the canvas gets bigger again.

Also, it would be nice if we could somehow prevent the canvas from getting smaller than the min size

petersteneteg:
whats the reason for not just setting the min and max of the splitter?

ResearchDaniel:
Those are given as percentages while the new min/max are in pixels. Setting splitter min/max will not give a fixed size

petersteneteg:
This seems to me like you are adding a gui option in inviwo to configure a different gui. I think that configuration belong somewhere else.

ResearchDaniel:
I think that it is rather standard layout options, e.g., https://www.w3schools.com/cssref/pr_dim_max-height.asp

ResearchDaniel:
MinMaxSliders, however, might not be the best way to specify them

ResearchDaniel:
Changed to text semantics and removed setting of max value. 
image

ResearchDaniel:
@petersteneteg ping

petersteneteg (approved):

ResearchDaniel:
Wait, I have some improvements coming up

ResearchDaniel:
Now use our new isConnectionActive to deactivate zero-sized layout elements. Also added a grouping for min/max
image

PR #707, opened 2019-09-13
Feature/tfeditor

martinfalk:
Some usability tweaks in the TF editor.
* using spinboxes instead of line edits
* explicit invalid/ambiguous state (which doesn't allow for dragging/mouse wheel changes)
* regex fix when entering floating point numbers, e.g. `0,3 0,2 1 1` (which was previously not accepted)

**non-ambiguous**
image

**ambiguous values**
image


martinfalk:
**before**
image

petersteneteg (approved):

PR #709, opened 2019-09-16
FontRendering/PlottingGL: Bug fixes to prevent errors due to string w…

ResearchDaniel:
…ith space in font rendering and empty axis in parallel coordinates

### DIFF
--- a/modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
+++ b/modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
@@ -395,6 +395,10 @@ void ParallelCoordinates::createOrUpdateProperties() {
 }
 
 void ParallelCoordinates::buildLineMesh() {
+    if (axes_.empty()) {
+        // Nothing to build from
+        return;
+    }
     auto& mesh = lines_.mesh;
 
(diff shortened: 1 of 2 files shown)


modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
```diff
@@ -395,6 +395,10 @@ void ParallelCoordinates::createOrUpdateProperties() {
 }
 
 void ParallelCoordinates::buildLineMesh() {
+    if (axes_.empty()) {
```
petersteneteg: seems strange to handle this here, should be handled in process

petersteneteg (changes requested):

modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
ResearchDaniel: Fixed

petersteneteg: Still no change though?

ResearchDaniel:
@petersteneteg Forgot to push, only commited :D

petersteneteg (approved):

PR #710, opened 2019-09-16
Added auto save and restore support

petersteneteg:


ResearchDaniel (approved):
Looks good from what I can see.
It would be nice if we could move the UndoManager from the network editor to allow any app to use it, future PR.

src/qt/editor/undomanager.cpp
```diff
@@ -43,23 +43,99 @@
...
+};
+
 UndoManager::UndoManager(InviwoMainWindow *mainWindow)
     : mainWindow_(mainWindow)
     , manager_{mainWindow_->getInviwoApplication()->getWorkspaceManager()}
-    , refPath_{filesystem::findBasePath()} {
+    , refPath_{filesystem::findBasePath()}
+    , autoSaver_{std::make_unique<AutoSaver>()} {
 
     mainWindow_->getInviwoApplicationQt()->setUndoTrigger([this]() { pushStateIfDirty(); });
```
ResearchDaniel, line 129: Would be better if we could use the app directly to avoid dependency on IniwoMainWindow.

petersteneteg: The window is used in a few more places.... guess I can disconnect them, but until I have a concrete use case, that would be premature generalization

ResearchDaniel: I now have an app that could use it :) 
but sure, it is another PR

PR #711, opened 2019-09-17
Mesh Plane Clipping

mskr:
This is an improvement to the existing `MeshClipping` processor by @ResearchDaniel, as discussed with him. I moved the clipping routine from the processor to a [shared file](https://github.com/inviwo/inviwo/blob/feature/mesh-plane-clipping/modules/base/include/modules/base/algorithm/mesh/meshclipping.h) in the algorithm folder, and it now supports clipping a mesh successively against different planes. This means, we can now have a processor, that takes a mesh and multiple planes (instead of just one), which I created as [`MeshPlaneClipping`](https://github.com/inviwo/inviwo/blob/feature/mesh-plane-clipping/modules/base/include/modules/base/processors/meshplaneclipping.h) processor. Also, I added a processor, that puts the boundary planes of a given volume into an outport ([`VolumeBoundaryPlanes`](https://github.com/inviwo/inviwo/blob/feature/mesh-plane-clipping/modules/base/include/modules/base/processors/volumeboundaryplanes.h) processor). 

A regression test is included for the new `MeshPlaneClipping` processor, whereas there was none for the old `MeshClipping` processor, which should maybe also be added (@ResearchDaniel). **EDIT**: Just noticed, that there are tests, which all failed on the Jenkins server. Running them on my own machine, the tests `basegl/raycasterclipping` and `basegl/meshclipping` fail, but with no visible errors.

A little bit of background info: This was a byproduct of the MPR (Multi-Planar-Reconstruction) feature we are building. For MPR, I needed to raycast a volume in an arbitrarily oriented "slab", i.e. a plane with thickness, in world space. I chose to construct meshes for the slabs, as entry/exit point geometry. Because the meshes naturally can lie partly outside the volume, I wanted to clip them (otherwise I see clamping artifacts in the raycasted image) at the 6 faces. Thus, I needed to clip against 6 planes, which the old clipping functionality could not do.

_____

**Always**

**New features**
- [x] Code / Processors follows the [Inviwo documentation guidelines](https://github.com/inviwo/inviwo/wiki/Documentation-guide)
- [x] Code / processors are tested in [unittest and/or regresion tests](https://github.com/inviwo/inviwo/wiki/Manual#testing)

**Updated feature / Fixes** 
- [x] Relevant documentation has been updated

**What evironment has the code been compiled and tested on?** 
- Windows 10 64 bit
- Visual Studio 15 2017 [RelWithDebInfo]
- Qt Version 5.13.0
- CMake 3.15.0-rc3

_____

_As this is my first real pull request, please remind me of things that are missing and I will update._


modules/base/include/modules/base/processors/volumeboundaryplanes.h
```diff
@@ -0,0 +1,78 @@
...
+
+#include <inviwo/core/datastructures/geometry/plane.h>
+
+#include <vector>
+
+namespace inviwo {
+
+/** \docpage{org.inviwo.VolumeBoundaryPlanes, Volume Boundary Planes}
+ * ![](org.inviwo.VolumeBoundaryPlanes.png?classIdentifier=org.inviwo.VolumeBoundaryPlanes)
+ * Outputs the six planes that enclose the input volume in world space.
```
ResearchDaniel, line 49: Could mention in which direction the normal will point by default and also the order of the output planes, I.e. which plane that correspond to which side

modules/base/include/modules/base/processors/meshplaneclipping.h
```diff
@@ -0,0 +1,78 @@
...
+    virtual ~MeshPlaneClipping() = default;
+
+    virtual void process() override;
+
+    virtual const ProcessorInfo getProcessorInfo() const override;
+    static const ProcessorInfo processorInfo_;
+
+private:
+    MeshInport inputMesh_;
+    DataInport<std::vector<Plane>> planes_;
```
ResearchDaniel: You can use FlatMultiDataInport<Plane> to support multiple connections, including std:vector<Plane>

modules/base/src/algorithm/mesh/meshclipping.cpp
```diff
@@ -0,0 +1,609 @@
...
+
+namespace inviwo {
+
+namespace meshutil {
+
+namespace detail {
+
+// Check point equality with threshold
+inline bool equal(vec3 v1, vec3 v2, float eps) {
+    return (std::fabs(v1.x - v2.x) < eps && std::fabs(v1.y - v2.y) < eps &&
```
ResearchDaniel: Use glm::equalEpsilon

modules/base/src/algorithm/mesh/meshclipping.cpp
```diff
@@ -0,0 +1,609 @@
...
+    double wsum = 0.f;
+
+    for (size_t i = 0; i < numV; ++i) {
+        ip = (numV - 1 + i) % numV;
+        wi = 2.f * (tanA[i] + tanA[ip]) / ri[i];
+        wsum += wi;
+        baryW[i] = static_cast<float>(wi);
+    }
+
+    if (std::fabs(wsum) > 0.f) {
```
ResearchDaniel: std::abs(wsum) > 0

mskr:
Test basegl/raycasterclipping seems to fail, because the camera has slightly different rotation from ref image, at least on my machine.

Test basegl/meshclipping

modules/base/src/processors/meshplaneclipping.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+    addPort(inputMesh_);
+    addPort(planes_);
+    addPort(outputMesh_);
+    addProperties(clippingEnabled_, capClippedHoles_);
+}
+
+void MeshPlaneClipping::process() {
+    if (clippingEnabled_) {
+        std::shared_ptr<const Mesh> currentMesh = inputMesh_.getData();
```
ResearchDaniel, line 61: You need to make a copy of the input and clip it.

modules/base/src/processors/volumeboundaryplanes.cpp
```diff
@@ -0,0 +1,79 @@
...
+    addProperties(flipPlanes_);
+}
+
+void VolumeBoundaryPlanes::process() {
+    const auto vol = volume_.getData();
+    const auto world = vol->getWorldMatrix();
+    const auto basis = vol->getBasis();
+    const auto offset = vol->getOffset();
+    const auto p0 = world * vec4(offset, 1);
+    const auto p1 = world * vec4(offset + basis[0] + basis[1] + basis[2], 1);
```
ResearchDaniel: Perhaps easier 
```
p0 = dataToWorld*vec4(0, 0, 0, 1)
p1 = dataToWorld*vec4(1)
```
If you are after the min/max points in world space

modules/base/src/processors/volumeboundaryplanes.cpp
```diff
@@ -0,0 +1,79 @@
...
+#include <modules/base/processors/volumeboundaryplanes.h>
+
+namespace inviwo {
+
+// The Class Identifier has to be globally unique. Use a reverse DNS naming scheme
+const ProcessorInfo VolumeBoundaryPlanes::processorInfo_{
+    "org.inviwo.VolumeBoundaryPlanes",  // Class identifier
+    "Volume Boundary Planes",           // Display name
+    "Data Creation",                    // Category
+    CodeState::Experimental,            // Code state
```
ResearchDaniel: Stable?

ResearchDaniel (changes requested):
Nice, only have some minor change requests

modules/base/src/algorithm/mesh/meshclipping.cpp
petersteneteg: Or even better use util::almostEqual the you dont need a stupid epsilon

modules/base/src/algorithm/mesh/meshclipping.cpp
```diff
@@ -0,0 +1,609 @@
...
+// point p (which is inside the polygon) with respect to polygons of vertices (v)
+// Based on Mean Value Coordinates by Hormann/Floater
+inline void barycentricInsidePolygon2D(vec2 p, const std::vector<vec2>& v,
+                                       std::vector<float>& baryW) {
+    size_t numV = v.size();
+    std::vector<dvec2> s(numV);
+    std::vector<double> ri(numV);
+
+    for (size_t i = 0; i < numV; ++i) {
+        // When this difference gets very small, Ai will later be zero, leading to NaNs in baryW,
```
petersteneteg: Seems the paper  https://www.mn.uio.no/math/english/people/aca/michaelf/papers/barycentric.pdf has more robust code for managing these cases

modules/base/src/algorithm/mesh/meshclipping.cpp
```diff
@@ -0,0 +1,609 @@
...
+    float invDenom = 1.f / (d00 * d11 - d01 * d01);
+    vec3 bary;
+    bary.y = (d11 * d20 - d01 * d21) * invDenom;
+    bary.z = (d00 * d21 - d01 * d20) * invDenom;
+    bary.x = 1.f - bary.y - bary.z;
+    return bary;
+}
+
+std::vector<Edge3D> findUniqueEdges(const std::vector<Edge3D>& edges, const float EPSILON) {
+    std::vector<Edge3D> uniqueEdges;
```
petersteneteg: Should just use a std::sort and std::unique

modules/base/src/algorithm/mesh/meshclipping.cpp
```diff
@@ -0,0 +1,609 @@
...
+            if (!found) uniqueEdges.push_back(edge);
+        }
+    }
+    return uniqueEdges;
+}
+
+std::vector<Polygon<Edge3D>> findLoops(const std::vector<Edge3D>& edges, const float EPSILON) {
+    std::vector<Polygon<Edge3D>> polygons;
+    std::vector<Edge3D> connectedEdges;
+    std::vector<Edge3D> unconnectEdges(edges);
```
petersteneteg, line 89: Might want to use a unordered_set here for example, to speed up the lookups

modules/base/src/algorithm/mesh/meshclipping.cpp
```diff
@@ -0,0 +1,609 @@
...
+        Edge3D currentEdge = unconnectEdges.front();
+        connectedEdges.push_back(currentEdge);
+        unconnectEdges.erase(unconnectEdges.begin());
+        bool createdPolygon = false;
+
+        // Search all edges for a connection
+        for (size_t i = 0; i < edges.size(); ++i) {
+            if (equal(edges[i].v1, currentEdge.v2, EPSILON) && edges[i].v2 != currentEdge.v1) {
+                connectedEdges.push_back(Edge3D(currentEdge.v2, edges[i].v2));
+                std::vector<Edge3D>::iterator it =
```
petersteneteg: Use auto for these verbose iterators,,

modules/base/src/algorithm/mesh/meshclipping.cpp
```diff
@@ -0,0 +1,609 @@
...
+                    std::vector<Edge3D>::iterator it =
+                        std::find(unconnectEdges.begin(), unconnectEdges.end(), currentEdge);
+
+                    if (it != unconnectEdges.end()) unconnectEdges.erase(it);
+                }
+
+                Polygon<Edge3D> newPoly(connectedEdges.size());
+
+                for (size_t j = 0; j < connectedEdges.size(); ++j) {
+                    newPoly.at(j) = connectedEdges.at(j);
```
petersteneteg, line 134: Avoid using at for vectors...

modules/base/src/algorithm/mesh/meshclipping.cpp
```diff
@@ -0,0 +1,609 @@
...
+std::vector<Polygon<Edge3D>> simplifyPolygons(const std::vector<Polygon<Edge3D>>& polygons,
+                                              const float EPSILON) {
+    std::vector<Polygon<Edge3D>> simplifiedPolygons;
+
+    for (const auto& poly : polygons) {
+        if (poly.size() == 0) continue;
+
+        std::vector<Edge3D> simplifiedEdges{poly.get(0)};
+        for (size_t i = 1; i < poly.size(); ++i) {
+            const auto pivotEdge = &simplifiedEdges.back();
```
petersteneteg: Why use a pointer? Just use auto&

modules/base/src/algorithm/mesh/meshclipping.cpp
```diff
@@ -0,0 +1,609 @@
...
+            const auto testDir = glm::normalize(testEdge.v2 - testEdge.v1);
+            if (glm::abs(glm::dot(pivotDir, testDir) - 1.0f) < EPSILON) {
+                pivotEdge->v2 = testEdge.v2;
+            } else {
+                simplifiedEdges.push_back(testEdge);
+            }
+        }
+
+        Polygon<Edge3D> newPoly(simplifiedEdges.size());
+        for (size_t i = 0; i < simplifiedEdges.size(); ++i) {
```
petersteneteg, line 179: Use std: copy

modules/base/src/algorithm/mesh/meshclipping.cpp
```diff
@@ -0,0 +1,609 @@
...
+            newPoly.at(i) = simplifiedEdges[i];
+        }
+        simplifiedPolygons.push_back(newPoly);
+    }
+
+    std::vector<Polygon<Edge3D>> notDegenerated;
```
petersteneteg, line 185: Use copy_if

modules/base/src/algorithm/mesh/meshclipping.cpp
```diff
@@ -0,0 +1,609 @@
...
+                pivotEdge->v2 = testEdge.v2;
+            } else {
+                simplifiedEdges.push_back(testEdge);
+            }
+        }
+
+        Polygon<Edge3D> newPoly(simplifiedEdges.size());
+        for (size_t i = 0; i < simplifiedEdges.size(); ++i) {
+            newPoly.at(i) = simplifiedEdges[i];
+        }
```
petersteneteg, line 181: Might check degenerate already here

modules/base/src/algorithm/mesh/meshclipping.cpp
```diff
@@ -0,0 +1,609 @@
...
+        std::vector<vec4> newColors;
+
+        // For polygons with less than 5 unique points, we could do better triangulation without the
+        // centroid, but the following works in any case.
+
+        // Calculate fake centroid
+        // (this will not give us the most optimal triangles,
+        // but it is sufficient to close the hole)
+        vec3 fakeCentroid(0.0f);
+        {
```
petersteneteg: Why not just use the centerpoint?

modules/base/src/algorithm/mesh/meshclipping.cpp
```diff
@@ -0,0 +1,609 @@
...
+// Check point equality with threshold
+inline bool equal(vec3 v1, vec3 v2, float eps) {
+    return (std::fabs(v1.x - v2.x) < eps && std::fabs(v1.y - v2.y) < eps &&
+            std::fabs(v1.z - v2.z) < eps);
+}
+
+// Compute barycentric coordinates/weights for
+// point p (which is inside the polygon) with respect to polygons of vertices (v)
+// Based on Mean Value Coordinates by Hormann/Floater
+inline void barycentricInsidePolygon2D(vec2 p, const std::vector<vec2>& v,
```
petersteneteg: Return the result by value insted

modules/base/src/algorithm/mesh/meshclipping.cpp
```diff
@@ -0,0 +1,609 @@
...
+
+    // Extract vertex data from mesh:
+
+    DrawType drawType = DrawType::Triangles;
+    ConnectivityType connectivityType = ConnectivityType::Strip;
+    const std::vector<vec3>* vertexList = nullptr;
+    const std::vector<vec3>* texcoordlist = nullptr;
+    const std::vector<vec4>* colorList = nullptr;
+    const std::vector<unsigned int>* triangleList = nullptr;
+
```
petersteneteg, line 244: At some point we should do this for all meshes.,. Should not be so hard

modules/base/src/processors/meshplaneclipping.cpp
petersteneteg: Clipmesh returns a new copy... hence this works. But it will create one copy per plane...

modules/base/src/processors/volumeboundaryplanes.cpp
petersteneteg: We should support at least all triangle meshes first...

petersteneteg (changes requested):
Nice work!
Some minor comment, potential improvements and questions

modules/base/src/algorithm/mesh/meshclipping.cpp
```diff
@@ -0,0 +1,609 @@
...
+// Check point equality with threshold
+inline bool equal(vec3 v1, vec3 v2, float eps) {
+    return (std::fabs(v1.x - v2.x) < eps && std::fabs(v1.y - v2.y) < eps &&
+            std::fabs(v1.z - v2.z) < eps);
+}
+
+// Compute barycentric coordinates/weights for
+// point p (which is inside the polygon) with respect to polygons of vertices (v)
+// Based on Mean Value Coordinates by Hormann/Floater
```
petersteneteg: Might want to expose this in the header, might be useful for more people

modules/base/include/modules/base/processors/volumeboundaryplanes.h
martinfalk: Please make sure that is consistent with the existing functionality in Inviwo.

modules/base/src/processors/meshplaneclipping.cpp
mskr: I think not, since clipMeshAgainstPlane creates and returns a new mesh anyway.

modules/base/src/algorithm/mesh/meshclipping.cpp
mskr: It is owed to the Polygon class..

petersteneteg: Yes... sorry, a bit hard to know what is new and what was already there... if it was already there you might disregard those comments...

modules/base/src/algorithm/mesh/meshclipping.cpp
mskr: I tested it and unfortunately the epsilon needs a bit of custom tuning, meaning when we use ULP or machine epsilon, `findLoops` does not find connection points anymore. The current value of `const float EPSILON = 0.00001f` works fine for my tests. I know, this is not optimal and assuming that the intersection points (triangle edges with plane) are exact, machine epsilon **should** work. Please let me know, if any of your tests fail. Otherwise, keep it this way?

petersteneteg: you could use a larger ULP for almostEqual. Otherwise I think the epsilon need to either be derived from the size of the mesh. i.e. norm(mesh bounding box)/10^6 or something like that.
a hardcoded EPSILON like 0.000001 will only work for meshes that has a bounding box of size ~1. But we can often have much larger ~10000 or much smaller ~0.0001 meshes...
I you can derive it from the mesh. then It should at least be possible for the user to supply a epsilon....

modules/base/src/processors/meshplaneclipping.cpp
mskr: ... which is due to Sutherland-Hodgman clipping by nature building a new mesh per plane.

mskr:
As mentioned, the regression tests fail, but with no visible errors. Should I update the ref images then?

modules/base/src/processors/volumeboundaryplanes.cpp
mskr: ... for the MeshPlaneClipping processor, I agree. As this is only the VolumeBoundaryPlanes thing, I suppose I can mark it stable.

martinfalk:
@mskr yes, re-upload the result images from the last regression test run on jenkins.

modules/base/src/algorithm/mesh/meshclipping.cpp
mskr: Do you mean `(sum of points)/(number of points)`? It was my first thought too, but the result does not always lie on the plane.

I should also mention that there was a [previous (somewhat lengthy) code to find the real centroid](https://github.com/inviwo/inviwo/blob/6cadbb3e15476da324c0faa83847c8c8ed2ebcb4/modules/base/src/processors/meshclipping.cpp#L566). At one point during dev, when I saw bugs, I thought it might come from there, so I replaced it by something simpler, and I ended up keeping it, since it does the job.

petersteneteg: maybe Im missing something but why does (sum of points)/(number of points) not always lie on the plane? I use that in some other cases and I never had any problem. In my cases the polygon was always convex? Is that the reason? In the non convex case how is a centroid defined? Might be good with a comment explaining they the simple mean will not work...

modules/base/src/algorithm/mesh/meshclipping.cpp
mskr: Thanks for the suggestion. Tried it, but the compiler failed to create a hash function for Edge3D with the error: 

```
term does not evaluate to a function taking 1 arguments	inviwo-module-base	C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.16.27023\include\xhash	113	
```

modules/base/src/algorithm/mesh/meshclipping.cpp
mskr: Sorry, you are right about the mean. It always lies on the plane, when one sums up only the unique points in the polygon, instead of the points of all edges, like I did. Will replace fakeCentroid by mean then.

I think the mean is also always on the plane for non-convex polygons, although it can lie outside the polygon then. I made a [GeoGebra test where you can play with the scenario](https://www.geogebra.org/classic/cdyuhqjk).

modules/base/src/algorithm/mesh/meshclipping.cpp
mskr: Good point. Though, I would be happy if someone else could do this.

modules/base/src/algorithm/mesh/meshclipping.cpp
petersteneteg: Yes you have to specialize std::hash for Edge3d

modules/base/src/algorithm/mesh/meshclipping.cpp
petersteneteg: Ahh ic, did not see that you used the edge points... cool link, have never seen that one before!

modules/base/include/modules/base/algorithm/mesh/meshclipping.h
```diff
@@ -57,6 +57,13 @@ namespace meshutil {
 IVW_MODULE_BASE_API std::shared_ptr<Mesh> clipMeshAgainstPlane(const Mesh& mesh, const Plane& plane,
                                                                bool capClippedHoles = true);
 
+namespace detail {
+
+IVW_MODULE_BASE_API inline std::vector<float> barycentricInsidePolygon2D(
```
petersteneteg: Should not be inline here...

modules/base/src/algorithm/mesh/meshclipping.cpp
```diff
@@ -41,62 +41,73 @@ namespace detail {
...
+            std::abs(v1.z - v2.z) < eps);
 }
 
 // Compute barycentric coordinates/weights for
 // point p (which is inside the polygon) with respect to polygons of vertices (v)
 // Based on Mean Value Coordinates by Hormann/Floater
-inline void barycentricInsidePolygon2D(vec2 p, const std::vector<vec2>& v,
-                                       std::vector<float>& baryW) {
-    size_t numV = v.size();
+inline std::vector<float> barycentricInsidePolygon2D(vec2 p, const std::vector<vec2>& v) {
```
petersteneteg: Not really much point in using inline in a cpp file

mskr:
[2 tests on branch](http://jenkins.inviwo.org:8080/job/inviwo/job/feature%252Fmesh-plane-clipping/6/execution/node/3/ws/regress/report.html) and [1 test on PR](http://jenkins.inviwo.org:8080/job/inviwo/job/PR-711/7/execution/node/3/ws/regress/report.html) still fail after 6c6f61a4d1f50badaa871a7963806a47281e359c, any ideas what is going on?

ResearchDaniel:
@mskr Looks like there are some numerical differences in the new implementation compared to the previous one? Especially the 3 pixel difference in the meshplaneclipping workspace

mskr:
That is very well possible (whereas it is no visible difference). But it shouldn't even compare against the previous implementation anymore, since I replaced the ref images in the mentioned commit. Am I missing something?

r-englund:
> 
> 
> [2 tests on branch](http://jenkins.inviwo.org:8080/job/inviwo/job/feature%252Fmesh-plane-clipping/6/execution/node/3/ws/regress/report.html) and [1 test on PR](http://jenkins.inviwo.org:8080/job/inviwo/job/PR-711/7/execution/node/3/ws/regress/report.html) still fail after [6c6f61a](https://github.com/inviwo/inviwo/commit/6c6f61a4d1f50badaa871a7963806a47281e359c), any ideas what is going on?

It is very common for images to fail if they were created on one machine (i.e. you machine) and then used on another (i.e Jenkins).  
On a side note, I would remove the bounding box rendering from these regression tests since its is not need to test the clip method. IMO the regression tests should be as minimal as possible and only have the processors that are really needed to produce an image. (I know there are quite some regression tests in Inviwo that does not do that :) )

mskr:
1) Sorry, forgot to say, that I downloaded the images from Jenkins, which is why I am so confused, that the tests fail. Maybe downloading them through the browser changed them slightly..

2) Agreed, will remove bounding boxes and backgrounds.

modules/base/src/algorithm/mesh/meshclipping.cpp
mskr: I implemented the mesh-dependent epsilon, but one case where it still fails are thin "slab" meshes. You can reproduce this in the [Large Epsilon Regression Test](https://github.com/inviwo/inviwo/blob/feature/mesh-plane-clipping/modules/base/tests/regression/meshclipping-large-epsilon/meshclipping-large-epsilon.inv), making a thin "cube" with the Mesh Creator.

It throws [this Exception](https://github.com/inviwo/inviwo/blob/feature/mesh-plane-clipping/modules/base/src/algorithm/mesh/meshclipping.cpp#L144) in findLoops(), saying some edges are fully unconnected. This needs some investigation...

r-englund:
> 
> 
>     1. Sorry, forgot to say, that I downloaded the images from Jenkins, which is why I am so confused, that the tests fail. Maybe downloading them through the browser changed them slightly..
> 
>     2. Agreed, will remove bounding boxes and backgrounds.

Nice. 
I see it still fails though. I added a tolerance (625f9d1) to one of the tests since it is just "random" pixels with very small diff. 
For the other failing test i did not do anything. The reason is, since I have not looked at the code, I can't tell if it is due to "random" or something else. My doubt is based on the fact that the full top plane of the volume is different compared to the ref image. 
image

petersteneteg:
FYI: I'm rebasing this and adding support for all mesh types

PR #712, opened 2019-09-19
Input-selection

petersteneteg:
Make the input selector processsor only evaluare the input that is currently used, by addig a customiation point to the topological sort in the evaluator. A processor can no overload isConnectionActive to specify if a specific connected outport wont be used, and then should not need to be evaluated.


ResearchDaniel (approved):
Tried it and it works like a charm.

PR #713, opened 2019-09-19
Feature/qt gl fixes

martinfalk:


petersteneteg (approved):

modules/opengl/src/canvasprocessorgl.cpp
```diff
@@ -44,4 +47,17 @@ const ProcessorInfo CanvasProcessorGL::getProcessorInfo() const { return process
 
 CanvasProcessorGL::CanvasProcessorGL(InviwoApplication* app) : CanvasProcessor(app) {}
 
+void CanvasProcessorGL::process() {
+    // ensure that the image inport has a GL representation
+    // otherwise the canvas widget will request a new one which in turn creates an fbo within the
+    // context of the canvas, i.e. not the default canvas.
+    //
+    // FIXME: this is not ideal
```
ResearchDaniel, line 55: Is this a Qt only thing? 
Perhaps we should elaborate the explanation:
The process function is guaranteed to be in the default OpenGL context, so we need to ensure that the OpenGL representation (containing an FBO) is created here and not by CanvasWidget. The CanvasWidget may be in a different OpenGL context. 
Perhaps this is how it should be and does not require a FIXME?

modules/opengl/src/canvasprocessorgl.cpp
```diff
@@ -44,4 +47,17 @@ const ProcessorInfo CanvasProcessorGL::getProcessorInfo() const { return process
...
+    if (!inport_.getData()->hasRepresentation<ImageGL>()) {
+        RenderContext::getPtr()->activateDefaultRenderContext();
```
ResearchDaniel, line 57: No need to activate default RenderContext? Should already be active since we are in the process function?

ResearchDaniel (approved):
Some changes may be required

modules/opengl/src/canvasprocessorgl.cpp
martinfalk: Yes, the gl widget in the canvas creates its own context. In some cases, the widget is lazily created just when the process func is called. We tried several approaches. One scenario with repeatedly creating the fbo in the canvas' context: drag&drop of an image, then the new canvas is created at the same time as the network is evaluated (race condition).

modules/opengl/src/canvasprocessorgl.cpp
martinfalk: That is exactly the point. In some borderline cases, the gl widget is created just before this call. Hence activating its context.

PR #714, opened 2019-09-20
PlottingGL: Make PCP axes display the ranges instead of the values. 

ResearchDaniel:
We should display the values next to the slider boxes when hovering, or similar, instead


r-englund (changes requested):
Don't remove displaying the values of the slider, I use them all the time when I use the PCP.

ResearchDaniel:
They are not updated when dragging the sliders anyway (as far as I could tell). Values should be displayed next to the slider buttons.

ResearchDaniel:
@r-englund Added tooltips for values. Not optimal since they are only displayed when hovering, but better than nothing. Future PR could render values as propert text

petersteneteg:
I agree with rickard, seeing the current values on the sliders is much more important the seeing the range. I would not change this. What is the reason for showing the range instead of the values?

petersteneteg (changes requested):
I think the values are much more important to show than the ranges.

petersteneteg:
> They are not updated when dragging the sliders anyway (as far as I could tell). Values should be displayed next to the slider buttons.

Then is is a new bug, they used to be updated.

ResearchDaniel:
@petersteneteg @r-englund I agree. I first thought that the intention was to show the range, since that was what was specified in code but was not shown. Will see if I can change it to the way it was before, i.e. showing the values and updating them all the time.

ResearchDaniel:
Looked into it a bit and the settings says min/max but it displays the values.
image

Options 
1. Removed that option and update the value in its place
2. Fix min/max bug (done here) and add rendering of value (on axis hover or always)

r-englund:
There used to be tooltips in the PCP before, I added it in PR #345, I guess they where removed....
image

petersteneteg:
Guess we might have lost those tooltips when we moved to using the sliders instead...

ResearchDaniel:
Problem with tooltips are that they go away as soon as one clicks (also some delay)

martinfalk:
@ResearchDaniel that is a problem of our implementation. Qt tooltips, by default, are shown at a specific position and stay there until the mouse cursor is moved outside a given rectangle around the trigger position (see https://doc.qt.io/qt-5/qtooltip.html#showText). The disappearing is caused by resetting the tooltip to "", which hides it immediately... call it "feature" by design.

ResearchDaniel:
@martinfalk ok, even though I set the tooltip in the move event?

petersteneteg:
Note: the tool tip is only supposed to be used as a tooltip, i.e. it will only show up after a delay and will generally not follow the mouse around. If you want to show text in a persistent manner while dragging or clicking, some other mechanism should be used.

ResearchDaniel:
Conclusion: This PR fixes the original issue and brings back the old way of inspecting a value (using tooltip). So, I think that we should do the following:
- Merge this PR
- Create a ticket for displaying current value along PCP-axis instead of a tooltip

ResearchDaniel:
@petersteneteg ping

r-englund (approved):

petersteneteg (approved):

PR #718, opened 2019-09-24
BaseGL: ImageLayout deactive zero-sized layout elements to prevent Op…

ResearchDaniel:
…enGL errors. Cap min/max to output dimensions. 

Also performance improvement due to prevension of setVisible state switching

### DIFF
--- a/modules/basegl/src/processors/imageprocessing/imagelayoutgl.cpp
+++ b/modules/basegl/src/processors/imageprocessing/imagelayoutgl.cpp
@@ -261,9 +261,18 @@ void ImageLayoutGL::onStatusChange(bool propagate) {
     }
 }
 
-inline bool ImageLayoutGL::isConnectionActive(Inport*, Outport* from) const {
-    auto p = static_cast<ImageOutport*>(from);
-    return !glm::any(glm::equal(p->getDimensions(), size2_t(0)));
+inline bool ImageLayoutGL::isConnectionActive(Inport* from, Outport* to) const {
+    IVW_ASSERT(from == &multiinport_, "only one inport");
+    const auto ports = multiinport_.getConnectedOutports();
+    auto portIt = std::find(ports.begin(), ports.end(), to);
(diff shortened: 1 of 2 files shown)


petersteneteg (approved):

modules/basegl/src/processors/imageprocessing/imagelayoutgl.cpp
```diff
@@ -166,58 +200,87 @@ void ImageLayoutGL::propagateEvent(Event* event, Outport* source) {
...
+    bottomMinMax_.setVisible(bVisible);
+
     if (propagate) {
         ResizeEvent e(currentDim_);
         propagateEvent(&e, &outport_);
     }
 }
 
+bool ImageLayoutGL::isConnectionActive(Inport* from, Outport* to) const {
+    IVW_ASSERT(from == &multiinport_, "only one inport");
```
r-englund, line 271: I do not get this check, is it not allowed to have only one input? That will happen for sure when setting up a new network.

ResearchDaniel: This processor only has one inport. Perhaps it should say only one inport though

ResearchDaniel: Ah, it does :D

r-englund: Why do we need this test? Is it if someone calls it themself from somewhere? In that case I guess it should say something like "isConnectionActive called with an inport not belling to current processor"

ResearchDaniel: Because the code below assumes that the supplied inport is the multiinport_. Want to make sure that this assumption is valid later even if someone decides to alter the processor.

r-englund: Then use a proper error message

PR #720, opened 2019-09-26
Fix/touchpad2

ResearchDaniel:
New from #697


ResearchDaniel:
@petersteneteg Should be no changes compared to #697

r-englund (approved):
It works on my machine

PR #722, opened 2019-09-27
ImageLayoutGL: Use proper error message in assertion

r-englund:


petersteneteg (approved):

ResearchDaniel (changes requested):
The updated error text will not happen (or should not be necessary to check for at least). The assert will occur if someone decides to add another port and not adjust the code

r-englund:
> The updated error text will not happen (or should not be necessary to check for at least). The assert will occur if someone decides to add another port and not adjust the code

and if that happens the error message should be descriptive

martinfalk (approved):

martinfalk:
Have adjusted the error message.

ResearchDaniel:
A lot of fuss for something that should not happen :) Adjusted the text slightly

ResearchDaniel (approved):

petersteneteg:
Just so you know, the function is automatically added to the assert... https://github.com/inviwo/inviwo/blob/ac876d14cb18ee99e3604e3fce16e9456567f7d4/include/inviwo/core/util/assertion.h#L57

r-englund:
> 
> 
> A lot of fuss for something that should not happen :) Adjusted the text slightly

And all started with me misunderstanding the test-expresion in the first place. Anyway, now we have a descriptive message.

PR #737, opened 2019-10-03
Streamline packaging configuration and make it app-specific.

ResearchDaniel:
- Package your own app on Mac by specifying IVW_APP_INSTALL_NAME
- Moved packaging include from root directory to apps/inviwo
- Moved cmake/packaging to apps/inviwo/packaging to clarify that packaging is application dependent
- Removed unnecessary info_plist_template.txt. Use CMAKE built-in instead
- Introduce IVW_APP_INSTALL_NAME on Mac to enable use of different app name when packaging
- Removed hardcoded Inviwo.app in Webbrowser module, now uses IVW_APP_INSTALL_NAME


apps/inviwo/packaging/packaging.cmake
```diff
@@ -52,10 +52,10 @@ if(IVW_PACKAGE_PROJECT)
     if(WIN32)
         # Need backslash for correct subdirectory paths with NSIS
         set(CPACK_PACKAGE_ICON              "${IVW_ROOT_DIR}\\\\resources\\\\inviwo\\\\inviwo_light.png")
-        set(CPACK_PACKAGE_INSTALL_DIRECTORY "Inviwo\\\\${IVW_VERSION}")
+        set(CPACK_PACKAGE_INSTALL_DIRECTORY "${CPACK_PACKAGE_NAME}\\\\${IVW_VERSION}")
```
petersteneteg: What the reason for moving the out of the cmake folder?

apps/inviwo/CMakeLists.txt
```diff
@@ -173,3 +177,5 @@ endif()
 # Save information for python tools.
 ivw_mod_name_to_class(ivw_module_classes ${enabled_modules})
 ivw_private_create_pyconfig("${IVW_MODULE_DIR};${IVW_EXTERNAL_MODULES}" "${ivw_module_classes}" inviwo)
+
+include(packaging/packaging.cmake)           # Package creation
```
petersteneteg, line 179: What is the semantics in CMake for running CPack from a subdir and not at the top level in the project? 
Is it even possible to run cpack more then ones? how would that even work?

petersteneteg (approved):

apps/inviwo/packaging/packaging.cmake
ResearchDaniel: Because it is app-dependent. This cmake file is *only* used by the inviwo app. Minimals could have their own if we wanted to (or want to give an example).

ResearchDaniel: I think this matters most for MacOS bundles. It is not possible to start different executables in a bundle without explicitly going into the bundle

apps/inviwo/CMakeLists.txt
ResearchDaniel: I don't think that it is possible to run it more than once (although conceptually it should be possible to have multiple package prjoects I guess), but it does not seem to matter that it is not at the top level.
I think that one has to take care not to activate more than one app with an installer.

ResearchDaniel: Perhaps using CPACK_INSTALL_CMAKE_PROJECTS would be an additional way. Slicer seem to use it heavily and includes cpack last
https://github.com/Slicer/Slicer/blob/master/CMake/LastConfigureStep/CMakeLists.txt

ResearchDaniel: They use a "main project", which determines which app to pack

petersteneteg: hehe, they have a CMake setup that is possibly even more complicated then ours :)
Yes, that might be an option to explore...

ResearchDaniel: Still, since this works, I think we can keep it. If we discover issues we can consider changing. I think that this way is a bit cleaner than using a "main project" because one can more easily customize the packaging

petersteneteg (changes requested):
Never merge master into a feature branch. Do a rebase

PR #738, opened 2019-10-03
Feature/sphereselection

martinfalk:
SphereRenderer processor showing selected indices via enlarged and colored sphere glyphs. Selected indices are provided by a Brushing and Linking port (only rendering, no active selection!).

**Note:** at the moment it only supports selection in the first mesh and there is no index mapping, i.e. selection IDs must match with vertex IDs (starting from 0).

### DIFF
--- a/modules/basegl/src/processors/sphererenderer.cpp
+++ b/modules/basegl/src/processors/sphererenderer.cpp
@@ -44,7 +44,7 @@ const ProcessorInfo SphereRenderer::processorInfo_{
     "Sphere Renderer",            // Display name
     "Mesh Rendering",             // Category
     CodeState::Stable,            // Code state
-    Tags::GL,                     // Tags
+    "GL, Brushing, Linking",      // Tags
 };
 const ProcessorInfo SphereRenderer::getProcessorInfo() const { return processorInfo_; }
 
@@ -52,6 +52,7 @@ SphereRenderer::SphereRenderer()
     : Processor()
     , inport_("geometry")
     , imageInport_("imageInport")
+    , brushLinkPort_("brushingAndLinking")
     , outport_("image")
     , renderMode_("renderMode", "Render Mode",
                   {{"entireMesh", "Entire Mesh", RenderMode::EntireMesh},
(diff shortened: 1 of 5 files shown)


modules/basegl/src/processors/sphererenderer.cpp
```diff
@@ -132,13 +142,21 @@ void SphereRenderer::configureShader(Shader& shader) {
...
+    }
+
+    utilgl::BlendModeState blendModeStateGL(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA);
     for (const auto& mesh : inport_) {
         if (mesh->getNumberOfBuffers() == 0) continue;
 
         auto& shader = shaders_.getShader(*mesh);
 
         shader.activate();
-        utilgl::BlendModeState blendModeStateGL(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA);
```
ResearchDaniel, line 141: good catch

modules/basegl/src/processors/sphererenderer.cpp
```diff
@@ -168,19 +186,26 @@ void SphereRenderer::process() {
...
-                                                mesh->getDefaultMeshInfo());
-                utilgl::setShaderUniforms(shader, *mesh, "geometry");
                 drawer.draw(MeshDrawerGL::DrawMode::Points);
-
                 break;
         }
+
+        // render selection only for first mesh input
+        if (!indices.empty() && (mesh == inport_.getData())) {
+            shader.setUniform("selectionMix", 1.0f);
```
ResearchDaniel, line 200: I'm not a big fan of this use of mix unless the idea is to have animation. Better to use `if (selected) ...` otherwise

modules/basegl/src/processors/sphererenderer.cpp
```diff
@@ -168,19 +186,26 @@ void SphereRenderer::process() {
...
+            shader.setUniform("selectionScaleFactor", selectionRadiusFactor_.get());
```
ResearchDaniel, line 201: Would be nice with animation so that the size does not pop

ResearchDaniel (approved):
Looks good to me

PR #743, opened 2019-10-08
Feature/plottting fixes

martinfalk:
Some fixes regarding plotting issues
* `getVisible()` overload in axis settings caused axis properties to disappear/go to hidden
* pcp and axis properties caused recursive re-evaluations of the network since they invalidate the processor
* VolumeAxis: added range mode for basis + offset

### DIFF
--- a/modules/dataframe/include/inviwo/dataframe/datastructures/column.h
+++ b/modules/dataframe/include/inviwo/dataframe/datastructures/column.h
@@ -94,7 +94,7 @@ class TemplateColumn : public Column {
 public:
     using type = T;
 
-    TemplateColumn(const std::string &header);
+    TemplateColumn(const std::string &header, std::shared_ptr<Buffer<T>> buffer = nullptr);
 
     TemplateColumn(const TemplateColumn<T> &rhs);
     TemplateColumn(TemplateColumn<T> &&rhs);
@@ -212,8 +212,8 @@ private:
 };
 
 template <typename T>
-TemplateColumn<T>::TemplateColumn(const std::string &header)
-    : header_(header), buffer_(std::make_shared<Buffer<T>>()) {}
+TemplateColumn<T>::TemplateColumn(const std::string &header, std::shared_ptr<Buffer<T>> buffer)
+    : header_(header), buffer_(buffer ? buffer : std::make_shared<Buffer<T>>()) {}
 
 template <typename T>
 TemplateColumn<T>::TemplateColumn(const TemplateColumn &rhs)
--- a/modules/dataframe/src/properties/dataframecolormapproperty.cpp
+++ b/modules/dataframe/src/properties/dataframecolormapproperty.cpp
@@ -90,6 +90,9 @@ DataFrameColormapProperty::DataFrameColormapProperty(std::string identifier,
 void DataFrameColormapProperty::createOrUpdateProperties(
     std::shared_ptr<const DataFrame> dataframe) {
     if (!dataframe || dataframe->getNumberOfColumns() <= 1) return;
+
+    if (dataframe_ == dataframe) return;
+
     dataframe_ = dataframe;
     auto oldColormapProperties = colormaps_;
     colormaps_.clear();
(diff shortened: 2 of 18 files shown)


martinfalk:
Recursing network evaluation caused by adding the highlighted connection. Note that `Cryo Data Fusion` is above the PCP processor. But there is no change on the outport of `Cryo Data Fusion`.
image

image

modules/dataframe/include/inviwo/dataframe/datastructures/column.h
```diff
@@ -94,7 +94,7 @@ class TemplateColumn : public Column {
 public:
     using type = T;
 
-    TemplateColumn(const std::string &header);
+    TemplateColumn(const std::string &header, std::shared_ptr<Buffer<T>> buffer = nullptr);
```
ResearchDaniel, line 97: Default parameter should be 
std::make_shared<Buffer<T>>()

modules/dataframe/src/properties/dataframecolormapproperty.cpp
```diff
@@ -90,6 +90,9 @@ DataFrameColormapProperty::DataFrameColormapProperty(std::string identifier,
 void DataFrameColormapProperty::createOrUpdateProperties(
     std::shared_ptr<const DataFrame> dataframe) {
     if (!dataframe || dataframe->getNumberOfColumns() <= 1) return;
+
+    if (dataframe_ == dataframe) return;
```
ResearchDaniel, line 94: What if the data frame has changed?

modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
```diff
@@ -219,7 +220,11 @@ ParallelCoordinates::ParallelCoordinates()
         lineShader_.build();
     }
 
-    dataFrame_.onChange([&]() { createOrUpdateProperties(); });
+    dataFrame_.onChange([&]() {
+        if (dataframeCached_ == dataFrame_.getData()) return;
```
ResearchDaniel, line 224: What if the DataFrame has new columns?

ResearchDaniel (changes requested):
Have you tried it with a DataFrame that removes/adds columns? 
We need to compare the actual objects in that case, or at least the columns

modules/dataframe/src/properties/dataframecolormapproperty.cpp
petersteneteg, line 94: Since dataframe_ can be mutated above in the network, this does not mean that it hasn't changed, hence this seems quite error prone...

modules/dataframe/src/properties/dataframeproperty.cpp
```diff
@@ -72,6 +72,9 @@ DataFrameColumnProperty* DataFrameColumnProperty::clone() const {
 
 void DataFrameColumnProperty::setOptions(std::shared_ptr<const DataFrame> dataframe) {
     if (!dataframe || dataframe->getNumberOfColumns() <= 1) return;
+
+    if (dataframe_ == dataframe) return;
```
petersteneteg, line 76: Same as above...

modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
petersteneteg, line 224: Same as above this does not mean that it has the same cols, rows etc...

modules/plottinggl/src/processors/parallelcoordinates/pcpaxissettings.cpp
```diff
@@ -110,6 +110,8 @@ PCPAxisSettings::PCPAxisSettings(const PCPAxisSettings& rhs)
 PCPAxisSettings* PCPAxisSettings::clone() const { return new PCPAxisSettings(*this); }
 
 void PCPAxisSettings::updateFromColumn(std::shared_ptr<const Column> col) {
+    if (col_ == col) return;
```
petersteneteg, line 113: same here, pointer comparisons are dangerous

petersteneteg (changes requested):

modules/dataframe/include/inviwo/dataframe/datastructures/column.h
martinfalk: Why? Then we need to check inside the constructor whether `typename T` matches the column buffer type.

PR #745, opened 2019-10-09
QTWidgets: BoolCompositePropertyWidgetQt to update checkbox widget state when changed from code (python/js)

r-englund:
When setting the BoolProperty of a BoolCompositeProperty through python or javascript the widget was not updated (stayed check/unchecked)


petersteneteg (changes requested):
this should be handled in update from property

r-englund:
Replaced with #747

PR #749, opened 2019-10-11
BaseGL: ViewManager: Only check agains points of used events to determine wether the event was used or not.

r-englund:
Previously, the touch event was always marked as used, even when the propagated event was not marked as used.


ResearchDaniel (approved):
Looks good to me

petersteneteg (approved):

PR #753, opened 2019-10-14
Feature/core python minor additions fixes

petersteneteg:
Python: Fix for the python processor folder observer when a processor
Python: fixes for glm numpy conversions
Core: Added a utility class to redirect a stream to the logcentral
Core: Added system setting to redirect cout and cerr to LogCentral


r-englund (approved):
What happens if we redirect cout/cerr when having attached a console logger? Infinite indirect recursion?

petersteneteg:
That could be interesting :) Have not tested that... Not quite sure how to detect that either

PR #755, opened 2019-10-18
Feature/plottinggl

martinfalk:


modules/plottinggl/src/plotters/persistencediagramplotgl.cpp
```diff
@@ -1,31 +1,31 @@
...
+ *  DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ *  ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ *  (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ *  ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ *  (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ *  SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  *
- *********************************************************************************/
+ * ********************************************************************************/
```
ResearchDaniel: Extra space?

modules/plottinggl/src/plotters/persistencediagramplotgl.cpp
```diff
@@ -70,7 +70,8 @@ PersistenceDiagramPlotGL::Properties::Properties(std::string identifier, std::st
                   vec4(1), vec4(0.1f), InvalidationLevel::InvalidOutput, PropertySemantics::Color)
     , lineColor_("lineColor", "Line Color", vec4(vec3(34, 96, 150) / 255.0f, 1.0f), vec4(0),
                  vec4(1), vec4(0.1f), InvalidationLevel::InvalidOutput, PropertySemantics::Color)
-    , hoverColor_("hoverColor", "Hover color", vec4(1.0f, 0.77f, 0.25f, 1))
+    , hoverColor_("hoverColor", "Hover Color", vec4(1.0f, 0.906f, 0.612f, 1))
```
ResearchDaniel, line 73: Future PR: Define common default colors and/or color schemes for hover and selection so that we don't need to copy them to each processor

modules/plottinggl/src/plotters/persistencediagramplotgl.cpp
```diff
@@ -431,7 +436,21 @@ void PersistenceDiagramPlotGL::setIndexColumn(
     }
 }
 
-void PersistenceDiagramPlotGL::renderAxis(const size2_t &dims) {
+void PersistenceDiagramPlotGL::setSelectedIndices(const std::unordered_set<size_t>& indices) {
```
ResearchDaniel, line 439: Should we copy this or use shared pointers?

modules/plottinggl/src/processors/persistencediagramplotprocessor.cpp
```diff
@@ -40,8 +41,8 @@ const ProcessorInfo PersistenceDiagramPlotProcessor::processorInfo_{
     "org.inviwo.PersistenceDiagramPlotProcessor",  // Class identifier
     "Persistence Diagram Plot Processor",          // Display name
```
ResearchDaniel: Could remove Processor from display name as well

modules/plottinggl/src/processors/persistencediagramplotprocessor.cpp
```diff
@@ -84,25 +100,40 @@ PersistenceDiagramPlotProcessor::PersistenceDiagramPlotProcessor()
...
         auto &indexCol = iCol->getTypedBuffer()->getRAMRepresentation()->getDataContainer();
 
-        auto filteredIndicies = brushing_.getFilteredIndices();
+        auto filteredIndicies = brushingPort_.getFilteredIndices();
         IndexBuffer indicies;
         auto &vec = indicies.getEditableRAMRepresentation()->getDataContainer();
         vec.reserve(dfSize - filteredIndicies.size());
 
         auto seq = util::sequence<uint32_t>(0, static_cast<uint32_t>(dfSize), 1);
         std::copy_if(seq.begin(), seq.end(), std::back_inserter(vec),
```
ResearchDaniel, line 121: Idea for future: add BrushingAndLinking::getVisibleIndices  so that we do not need to do copy_if

ResearchDaniel (approved):
Nice! Have some minor suggestions

modules/plottinggl/src/plotters/persistencediagramplotgl.cpp
martinfalk: Brushing and Linking ports work with explicit sets...

https://github.com/inviwo/inviwo/blob/0e581cfda4ea3d2edc1f4c98a6e8d6256cc298a9/modules/brushingandlinking/include/modules/brushingandlinking/ports/brushingandlinkingports.h#L54-L65

modules/plottinggl/src/plotters/persistencediagramplotgl.cpp
martinfalk: good idea. Maybe even as System settings. However, which module should be responsible for setting those up?

modules/plottinggl/src/plotters/persistencediagramplotgl.cpp
martinfalk: leftover from converting `<space>&` to `&<space>`

modules/plottinggl/src/plotters/persistencediagramplotgl.cpp
ResearchDaniel: Probably not since there might be cases where it is not desirable to have the same. Default linking would probably better

martinfalk: Linking is not the same as default colors.

Issue #727, opened 2019-09-28
Module template file not properly formatted


martinfalk:
Creating a new module from scratch using inviwo-meta results in the new file `<mymodule>moduledhefine.h` not being properly clang formatted.

```
--- /Users/jenkins/jenkins/workspace/inviwo_feature_dataframeqt/inviwo/modules/dataframeqt/include/inviwo/dataframeqt/dataframeqtmoduledefine.h (original)
+++ /Users/jenkins/jenkins/workspace/inviwo_feature_dataframeqt/inviwo/modules/dataframeqt/include/inviwo/dataframeqt/dataframeqtmoduledefine.h (formatted)
@@ -1,21 +1,21 @@
 #pragma once
 
-#ifdef INVIWO_ALL_DYN_LINK  //DYNAMIC
-	// If we are building DLL files we must declare dllexport/dllimport
-	#ifdef IVW_MODULE_DATAFRAMEQT_EXPORTS
-		#ifdef _WIN32
-			#define IVW_MODULE_DATAFRAMEQT_API __declspec(dllexport)
-		#else  //UNIX (GCC)
-			#define IVW_MODULE_DATAFRAMEQT_API __attribute__ ((visibility ("default")))
-		#endif
-	#else
-		#ifdef _WIN32
-			#define IVW_MODULE_DATAFRAMEQT_API __declspec(dllimport)
-		#else
-			#define IVW_MODULE_DATAFRAMEQT_API
-		#endif
-	#endif
-#else  //STATIC
-	#define IVW_MODULE_DATAFRAMEQT_API
+#ifdef INVIWO_ALL_DYN_LINK  // DYNAMIC
+// If we are building DLL files we must declare dllexport/dllimport
+#ifdef IVW_MODULE_DATAFRAMEQT_EXPORTS
+#ifdef _WIN32
+#define IVW_MODULE_DATAFRAMEQT_API __declspec(dllexport)
+#else  // UNIX (GCC)
+#define IVW_MODULE_DATAFRAMEQT_API __attribute__((visibility("default")))
+#endif
+#else
+#ifdef _WIN32
+#define IVW_MODULE_DATAFRAMEQT_API __declspec(dllimport)
+#else
+#define IVW_MODULE_DATAFRAMEQT_API
+#endif
+#endif
+#else  // STATIC
+#define IVW_MODULE_DATAFRAMEQT_API
 #endif
 ```

martinfalk:
[...]

PR #757, opened 2019-10-21
Tools: clang format of templates, #727

martinfalk:


tools/meta/templates/moduledefine.h
```diff
@@ -1,20 +1,20 @@
...
-	#else
-		#ifdef _WIN32
-			#define {{ module/api }} __declspec(dllimport)
-		#else
-			#define {{ module/api }}
-		#endif
-	#endif
+#ifdef INVIWO_ALL_DYN_LINK  // DYNAMIC
+// If we are building DLL files we must declare dllexport/dllimport
+#ifdef IVW_MODULE_{{ upper(module/name) }}_EXPORTS
```
ResearchDaniel, line 5: Hmm, probably should ignore clang format here? The previous formatting made more sense in my opinion

ResearchDaniel (changes requested):

r-englund (approved):

tools/meta/templates/moduledefine.h
petersteneteg: I agree with Daniel, since this file is all macro definitions, and since it is quite hard to follow the logic without the indentation, I would prefer to keep it like that.. guess we should add the format guards though... Seems CMake also can auto generate these now... so maybe we should use that instead at some point...

martinfalk: Then please go ahead and fix this.

martinfalk: so far, no one cared what happened after the initial commit of such files and the subsequent clang format. It's not like these files will be ever looked at again.

PR #766, opened 2019-11-05
OpenEXR/CMake

ResearchDaniel:
- Update OpenEXR to version 2.4 along with (zlib) fixes for making it work
- Removed IVW_PACKAGE_PROJECT since it mostly causes trouble, e.g., no install for targets


cmake/installutils.cmake
```diff
@@ -125,103 +117,121 @@ endfunction()
...
+                endforeach()
             endif()
         endif()
     endforeach()
 endmacro()
 
-function(ivw_register_package name target)
-    get_target_property(incdirs ${target} INTERFACE_INCLUDE_DIRECTORIES)
-
+function(ivw_register_package name)
```
petersteneteg, line 153: Why would you put more then one target in here?

cmake/globalconfig.cmake
```diff
@@ -165,9 +165,6 @@ configure_file(${IVW_CMAKE_TEMPLATES}/inviwocommondefines_template.h
                ${CMAKE_BINARY_DIR}/modules/core/include/inviwo/core/inviwocommondefines.h 
                @ONLY IMMEDIATE)
 
-# Package creation
-option(IVW_PACKAGE_PROJECT "Create Inviwo Package Project" OFF)
```
petersteneteg, line 169: What was the reason for this, I never had any problem with it...

cmake/installutils.cmake
```diff
@@ -125,103 +117,121 @@ endfunction()
...
+    set(incdirs "")
+    foreach(target ${ARGN})
+        get_target_property(dirs ${target} INTERFACE_INCLUDE_DIRECTORIES)
+        set(incdirs "${incdirs} ${dirs}")
```
petersteneteg, line 157: not sure if " " will work well as a separator... might be better to use list(APPEND...

petersteneteg (approved):

cmake/installutils.cmake
ResearchDaniel: Multiple targets associated with one package. See OpenEXR

cmake/globalconfig.cmake
ResearchDaniel: Libraries dependent on our externals (zlib) having install(EXPORT ...) by default produces errors otherwise. Also, I think it is a bit difficult to understand that it is necessary to enable a "packaging" flag to make an installer, but that is not really the resason.

cmake/installutils.cmake
petersteneteg: Ahh ok, guess that makes sense

cmake/globalconfig.cmake
petersteneteg: ok ic... is there any reason not to also remove the package install option? or does that one require additional dependencies, NSIS etc?

ResearchDaniel: I don't think so.

PR #770, opened 2019-11-13
Feature/transform

martinfalk:
* replaced `World Transform` processor with a more generic one supporting arbitrary Transformations using a ListProperty
* some bug fixes

**Before / After**
image
image

### DIFF
--- a/include/inviwo/core/properties/listproperty.h
+++ b/include/inviwo/core/properties/listproperty.h
@@ -141,13 +141,13 @@ public:
     void clear();
 
     /**
-     * \brief add a list entry which is created from the respective prefab object.
+     * \brief construct a list entry which is created from the respective prefab object.
      * This function has no effect if the list size will exceed the maximum number of elements.
      *
      * @param prefabIndex   index of prefab object used for creating the new entry
      * @throw RangeException  in case prefabIndex is invalid
      */
-    void addProperty(size_t prefabIndex);
+    Property* constructProperty(size_t prefabIndex);
 
     /**
      * \brief add \p property as new list entry. The type of the property must match one of the
--- a/include/inviwo/core/properties/listproperty.h
+++ b/include/inviwo/core/properties/listproperty.h
@@ -145,6 +145,7 @@ public:
      * This function has no effect if the list size will exceed the maximum number of elements.
      *
      * @param prefabIndex   index of prefab object used for creating the new entry
+     * @return constructed property, or nullptr if the new property cannot be added to the list
      * @throw RangeException  in case prefabIndex is invalid
      */
     Property* constructProperty(size_t prefabIndex);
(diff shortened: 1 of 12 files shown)


modules/base/src/processors/transform.cpp
```diff
@@ -0,0 +1,163 @@
...
+                               AngleMeasure angleMeasure, InvalidationLevel invalidationLevel,
+                               PropertySemantics semantics)
+    : TransformProperty(identifier, displayName, invalidationLevel, semantics)
+    , mode("mode", "Angle Measure",
+           {{"rad", "Radians", AngleMeasure::Radians}, {"deg", "Degrees", AngleMeasure::Degrees}})
+    , axis("axis", "Axis", glm::normalize(axisValue), vec3{-1.0f}, vec3{1.0f})
+    , angle("angle", "Angle", angleValue, minAngle, maxAngle, increment) {
+    mode.setSelectedValue(angleMeasure);
+    axis.setSemantics(PropertySemantics::SpinBox);
+    addProperties(mode, axis, angle);
```
ResearchDaniel, line 78: We have an angle property semantic as well if you want to use it.

include/inviwo/core/properties/listproperty.h
```diff
@@ -141,13 +141,13 @@ class IVW_CORE_API ListProperty : public CompositeProperty {
...
     /**
-     * \brief add a list entry which is created from the respective prefab object.
+     * \brief construct a list entry which is created from the respective prefab object.
      * This function has no effect if the list size will exceed the maximum number of elements.
      *
      * @param prefabIndex   index of prefab object used for creating the new entry
      * @throw RangeException  in case prefabIndex is invalid
      */
-    void addProperty(size_t prefabIndex);
+    Property* constructProperty(size_t prefabIndex);
```
ResearchDaniel, line 151: Is there a special reason for renaming the function? 
Implementation has a code path for returning null. Would be good to mention if that can occur in the documentation (not sure how that differs from range exception).

ResearchDaniel (approved):
Nice! A bit inconsistent to name the processor just Transform while the property is called TransformProperty. No biggie though.

modules/base/src/processors/transform.cpp
martinfalk: I think in this case a spinbox is better suited since it is more precise.

include/inviwo/core/properties/listproperty.h
martinfalk: Yes, a very specific reason, i.e. function overload. This refactoring was done primarily to prevent future problems e.g. feature creep!
Before, you could not call `ListProperty::addProperty(0);` without explicitly casting the `0` to `size_t` as it is ambiguous.

martinfalk:
Didn't we agree to drop `Processor` for the processors at some point?

include/inviwo/core/properties/listproperty.h
martinfalk: The range exception is tied to the prefab index while not being able to add new items (due to max size) produces a log error and returns `nullptr`. The latter being considered not as severe.

martinfalk: revised the doxygen comment

PR #774, opened 2019-11-21
Feature/imagesubset

ResearchDaniel:
Processor for extracting a subregion of an image. Useful for highlight regions in papers for example.

image

### DIFF
--- a/modules/basegl/src/processors/imageprocessing/imagesubsetgl.cpp
+++ b/modules/basegl/src/processors/imageprocessing/imagesubsetgl.cpp
@@ -30,8 +30,7 @@
 #include <modules/basegl/processors/imageprocessing/imagesubsetgl.h>
 #include <inviwo/core/datastructures/buffer/buffer.h>
 #include <inviwo/core/interaction/events/gestureevent.h>
-#include <inviwo/core/interaction/events/mouseevent.h>
-#include <inviwo/core/interaction/events/touchevent.h>
+#include <inviwo/core/interaction/events/wheelevent.h>
 #include <inviwo/core/network/networklock.h>
 
 #include <modules/opengl/geometry/meshgl.h>
--- a/modules/basegl/src/processors/imageprocessing/imagesubsetgl.cpp
+++ b/modules/basegl/src/processors/imageprocessing/imagesubsetgl.cpp
@@ -1,0 +1,247 @@
+/*********************************************************************************
+ *
+ * Inviwo - Interactive Visualization Workshop
+ *
+ * Copyright (c) 2013-2019 Inviwo Foundation
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ *********************************************************************************/
+
+#include <modules/basegl/processors/imageprocessing/imagesubsetgl.h>
+#include <inviwo/core/datastructures/buffer/buffer.h>
+#include <inviwo/core/interaction/events/gestureevent.h>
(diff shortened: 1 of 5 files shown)


modules/basegl/src/processors/imageprocessing/imagesubsetgl.cpp
```diff
@@ -0,0 +1,239 @@
...
+    , outport_("outputImage")
+    , rangeX_("rangeX", "X Slices", 0, 256, 0, 256, 1, 1)
+    , rangeY_("rangeY", "Y Slices", 0, 256, 0, 256, 1, 1)
+    , shader_("img_texturequad.vert", "img_copy.frag") {
+    addPort(inport_);
+    addPort(outport_);
+    addProperties(rangeX_, rangeY_);
+    dims_ = size2_t(1, 1);
+
+    inport_.setOutportDeterminesSize(true);
```
r-englund: These can go into the constructor of the ports

modules/basegl/src/processors/imageprocessing/imagesubsetgl.cpp
```diff
@@ -0,0 +1,239 @@
...
+    // Create a rectangle mesh for rendering.
+    // Texture coordinates will be modified
+    auto verticesBuffer =
+        util::makeBuffer<vec2>({{-1.0f, -1.0f}, {1.0f, -1.0f}, {-1.0f, 1.0f}, {1.0f, 1.0f}});
+    vec2 texOffset = vec2(0.f) / vec2(dims_);
+    vec2 texEnd = vec2(1.f) / vec2(dims_);
+    texCoordsBuffer_ = util::makeBuffer<vec2>(
+        {texOffset, {texEnd.x, texOffset.y}, {texOffset.x, texEnd.y}, texEnd});
+    auto indices = util::makeIndexBuffer({0, 1, 2, 3});
+    rect_.addBuffer(BufferType::PositionAttrib, verticesBuffer);
```
r-englund: Could use a typed mesh instead: `TypedMesh<buffertraits::PositionsBuffer2D, buffertraits::TexcoordBuffer<2>>`

modules/basegl/src/processors/imageprocessing/imagesubsetgl.cpp
```diff
@@ -0,0 +1,239 @@
...
+        utilgl::DepthFuncState depth(GL_ALWAYS);
+        glDrawArrays(GL_TRIANGLE_STRIP, 0, 4);
+
+        shader_.deactivate();
+        utilgl::deactivateCurrentTarget();
+        TextureUnit::setZeroUnit();
+    }
+}
+
+void ImageSubsetGL::invokeEvent(Event* event) {
```
r-englund, line 137: We have event properties for this. It is much better to use them than to override the invokeEvent method.

modules/basegl/src/processors/imageprocessing/imagesubsetgl.cpp
```diff
@@ -0,0 +1,239 @@
...
+    // Allow panning and prevent interaction events from propagating (does not make
```
r-englund: incomplete comment?

modules/basegl/src/processors/imageprocessing/imagesubsetgl.cpp
```diff
@@ -0,0 +1,239 @@
...
+            event->markAsUsed();
+            break;
+        case TouchEvent::chash():
+            handleTouchEvent(static_cast<TouchEvent*>(event));
+            event->markAsUsed();
+            break;
+        case WheelEvent::chash():
+        case GestureEvent::chash():
+            // GestureEvents are better handles using TouchEvent
+            event->markAsUsed();
```
r-englund, line 156: you mark events as used both here and in the `handleMousePan` and the `handleTouchEvent`, would be enough in those function, and we should only mark them as used when we have used them.

r-englund, line 156: It is good to use `[[fallthrough]];` in switch statements in cases where it is by intention to fall through to the next/default case.  https://en.cppreference.com/w/cpp/language/attributes/fallthrough

r-englund (changes requested):
Nice feature! :) I suggest using `EventProperties` instead of overloading `invokeEvent`. It is generally not a good idea to overload the `invokeEvent`, it seems it's done correctly here but I think it is a bad example in case other devs look at this code to see how to listen to events. 
Also, with `EventProperty`, it is possible to set modifiers (i.e can't pan unless ctrl is pressed) to mouse events (and touch?) incase it is needed in some applicatio

ResearchDaniel:
> Nice feature! :) I suggest using `EventProperties` instead of overloading `invokeEvent`. It is generally not a good idea to overload the `invokeEvent`, it seems it's done correctly here but I think it is a bad example in case other devs look at this code to see how to listen to events.

I agree and that is actually what I did first. Perhaps I did something wrong, but it was tricky to get both mouse and touch to work when using EventProperties. If I recall correctly I had the following issue on on Mac:
- Trackpad issues mouse events without a buttonstate while pressing and moving. Thus, the MouseEventMatcher only matches mouse press, not match press & move. 

To circumvent this I implemented a GeneralEventMatcher. Since I could not change which button that should fire the event it made it useless (it shows up as an empty button that you cannot change in the property list).
In hindsight, I also should have made an issue about it :)

> Also, with `EventProperty`, it is possible to set modifiers (i.e can't pan unless ctrl is pressed) to mouse events (and touch?) incase it is needed in some applicatio

See above, I could not change which mouse button that should trigger the event using the MouseEventProperty

modules/basegl/src/processors/imageprocessing/imagesubsetgl.cpp
ResearchDaniel: In general I agree (also reason for comment a few lines above). I did not think that it made sence to have that kind of interaction in the subset (at least not in the scenario I used it for). 
If someone wants to use it for a different purpose it would be straight forward to introcude a "propagateInteractions" boolean. That is also the reason for keeping the markAsUsed in both functions.

modules/basegl/src/processors/imageprocessing/imagesubsetgl.cpp
ResearchDaniel: See general comment

modules/basegl/src/processors/imageprocessing/imagesubsetgl.cpp
ResearchDaniel: Ok, just copied the code in drawImagePlaneRect :)

modules/basegl/src/processors/imageprocessing/imagesubsetgl.cpp
ResearchDaniel: done

modules/basegl/src/processors/imageprocessing/imagesubsetgl.cpp
ResearchDaniel: done

modules/basegl/src/processors/imageprocessing/imagesubsetgl.cpp
ResearchDaniel: Added more comments

r-englund (approved):

PR #777, opened 2019-11-26
Color conversion: Added rgb2hcl and hcl2rgb

PetiteViking:
I had these two methods in my tensor module but since we already have a util lib for this, I thought I'd put it here.

### DIFF
--- a/include/inviwo/core/util/colorconversion.h
+++ b/include/inviwo/core/util/colorconversion.h
@@ -302,7 +302,8 @@ IVW_CORE_API vec3 Luv2XYZ(const vec3 &Luv, vec3 whitePointXYZ = getD65WhitePoint
 /**
  * \brief convert from hcl to rgb.
  *
- * HCL (hue chroma luminance) is converted to RGB (red green blue) values.
+ * HCL (hue chroma luminance) in range [a, b] is converted to RGB (red green blue) in range [a, b],
+ * where usually a = 0 and b = 1.
  *
  * See https://en.wikipedia.org/wiki/HCL_color_space
  *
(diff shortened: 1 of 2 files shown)


include/inviwo/core/util/colorconversion.h
```diff
@@ -299,6 +299,30 @@ IVW_CORE_API vec3 XYZ2Luv(const vec3 &XYZ, vec3 whitePointXYZ = getD65WhitePoint
...
+/**
+ * \brief convert from hcl to rgb.
+ *
+ * HCL (hue chroma luminance) is converted to RGB (red green blue) values.
+ *
+ * See https://en.wikipedia.org/wiki/HCL_color_space
+ *
+ * @param hcl HCL color
+ * @return RGB color
+ */
```
r-englund, line 312: I would add the possible output ranges of the HCL colors

include/inviwo/core/util/colorconversion.h
```diff
@@ -302,7 +302,8 @@ IVW_CORE_API vec3 Luv2XYZ(const vec3 &Luv, vec3 whitePointXYZ = getD65WhitePoint
 /**
  * \brief convert from hcl to rgb.
  *
- * HCL (hue chroma luminance) is converted to RGB (red green blue) values.
+ * HCL (hue chroma luminance) in range [a, b] is converted to RGB (red green blue) in range [a, b],
+ * where usually a = 0 and b = 1.
```
ResearchDaniel, line 306: Usually :)
I would just write [0 1]. However, is not Hue usually a degree/radian, [-pi pi]?

ResearchDaniel (changes requested):

include/inviwo/core/util/colorconversion.h
martinfalk: Hue, at least for the HSV color space is typically encoded in degree [0,360). We might, however, use the range [0,1] for the sake of simplicity. Not sure regarding HCL

martinfalk: We should specify which range this function expects, not how the color space is "usually" defined ;-)

martinfalk (approved):
Please provide the appropriate data ranges for hcl. RGB should be [0,1] as usual.

PR #786, opened 2019-11-29
StructuredGridEntity: Add getSpacing method

PetiteViking:


include/inviwo/core/datastructures/spatialdata.h
```diff
@@ -316,6 +318,20 @@ StructuredGridEntity<N>::StructuredGridEntity(const Matrix<N + 1, float>& modelM
...
 
+template <unsigned int N>
+Vector<N, float> StructuredGridEntity<N>::getSpacing() const {
+    const auto dimensions = getDimensions();
+    const auto basis = this->getBasis();
+
+    Vector<N, float> spacing;
+
+    for (unsigned int i{0}; i < N; ++i) {
+        spacing[i] = glm::length(basis[i]) / dimensions[i];
```
r-englund, line 329: Shouldn't this be dividing by `dims-1`

JochenJankowai: I was wondering, too. The thing is that in the constructor for `StructuredGridEntity` that takes in spacing as a parameter we do 
```c++
for (unsigned int i = 0; i < N; ++i) {
    basis[i][i] = dimensions[i] * spacing[i];
}
```
so I simply reversed that operation.

I assume it depends on how you define your dimensions, vertex-based or voxel-based.

petersteneteg (changes requested):
I don't like this since it throw away the information about the angles between the axis which may not be orthogonal. This function needs to at least verify that fact. The documentation should also reflect this...

ResearchDaniel:
I think that getSpacing is useful be it needs some documentation since it is not applicable to all cases.
We could consider calling it getVoxelSize and have an additional function that allows you to specify the voxel index to get the size for. This is how OpenVDB does it.
Look at "Voxel Interpretations" 
https://www.openvdb.org/documentation/doxygen/transformsAndMaps.html
and the corresponding functions:
https://www.openvdb.org/documentation/doxygen/classopenvdb_1_1v6__2_1_1math_1_1Transform.html#abe11d7ce015c713f29f88fb376e22d0f

petersteneteg:
Adding an index argument here would be very confusing though since the sizes are all the same by definition. And we will not support varying sizes here.

ResearchDaniel:
@petersteneteg yes, that' true. Was thinking that they could be non-linear, but you are right

JochenJankowai:
@petersteneteg Correct me if I'm wrong (I very well might be) but doesn't the basis matrix essentially consist of the three basis vectors where every column of the matrix defines one vector? If so, then the spacing that I calculate now is correct along the given basis vector. 

Should the basis not be orthogonal, one cannot use the three spacing values to calculate the volume of a voxel but that is another issue that shouldn't be dealt with here. One could possibly conceive a getVoxelSize method, as @ResearchDaniel suggested. Nevertheless, they are two different methods.

I did not quite understand what you meant with "throw away the information about the angles between the axis". The axes and the spacing are two independent things. The basis vectors are not altered and ergo, the angles remain the same.

petersteneteg:
@JochenJankowai True you can do that, and the basis is of course still there. My concern is that a user of this function will assume that the basis is orthogonal and then not consider the angles. I you write your code expressed directly in the basis instead of the spacing making that assumption is way harder. Also it is questionable if what you calculate is actually the "spacing" in the case of a non-orthogonal basis. The spacing of what?

JochenJankowai:
To be perfectly honest, I fail to see how anyone would misinterpret the term *spacing* when talking about volumes. It doesn't even matter whether you define your dimensions vertex-based or volume-based, the spacing will always give you the distance between two vertices along a basis vector. And this holds for non-orthogonal bases. I agree that documentation needs to be added. Notwithstanding, the method name and functionality are in my opinion very clear.

Implementing the method was part of an attempt to integrate the TensorFieldXD classes tighter into Inviwo but in the end, I might just as well implement the methods there. Nobody seems to have missed a getSpacing method so far anyway.

PR #792, opened 2019-12-05
DataFrame: Get column by name

PetiteViking:


modules/dataframe/src/datastructures/dataframe.cpp
```diff
@@ -161,6 +161,24 @@ std::string DataFrame::getHeader(size_t idx) const { return columns_[idx]->getHe
 
 std::shared_ptr<const Column> DataFrame::getColumn(size_t index) const { return columns_[index]; }
 
+std::shared_ptr<Column> DataFrame::getColumn(const std::string &name) {
+    for (const auto col : columns_) {
+        if (col->getHeader() == name) {
+            return col;
+        }
+    }
+    return nullptr;
```
r-englund: Should we throw an exception here? I would prefer an exception here rather than a nullptr exception when I forgot to check if it exists ;)

petersteneteg: if you throw an exception you should return an reference not a pointer... you always have to check if the pointer is null. if it can't be null you should not return a pointer. 
should also use find_if preferably

martinfalk: should be fixed now

martinfalk (approved):

modules/dataframe/include/inviwo/dataframe/datastructures/dataframe.h
```diff
@@ -136,6 +136,8 @@ class IVW_MODULE_DATAFRAME_API DataFrame {
      */
     std::shared_ptr<Column> getColumn(size_t index);
     std::shared_ptr<const Column> getColumn(size_t index) const;
+    std::shared_ptr<Column> getColumn(const std::string &name);
```
ResearchDaniel, line 143: State what happens when
1. No column with that name exist
2. More than one column has the same name

modules/dataframe/src/datastructures/dataframe.cpp
JochenJankowai: @petersteneteg Out of curiosity, why does it matter whether I use find or find_if? Both do exactly what I want.

martinfalk: I guess the point here being that find_if already encapsulates this functionality (except for the comparison, hence find_if). And thus it's just a one liner.

JochenJankowai:
We could also use optionals here...?

petersteneteg:
std::optional does not support references

r-englund:
> 
> 
> std::optional does not support references

We should't not return reference here anyway because we want the lifetime ensuring from the shared_ptr

PR #795, opened 2019-12-05
DataFrame: Drop column method

PetiteViking:


petersteneteg (approved):
We use drop, think we use remove, or erase in most other code?
Would also be nice with a overload for index

r-englund (approved):
Since we have no check for multiple columns with the same header this will drop all columns with that header. I've updated the comment to mention that.

JochenJankowai:
@petersteneteg I was not sure about the naming either. Within the table/frame lingo one would use drop but I agree that most developers would probably rather look for a remove/erase method.

PR #796, opened 2019-12-05
Template column: add iterators

PetiteViking:


ResearchDaniel (approved):

modules/dataframe/include/inviwo/dataframe/datastructures/column.h
```diff
@@ -155,6 +155,19 @@ class TemplateColumn : public Column {
 
     virtual size_t getSize() const override;
 
+    typename std::vector<T>::iterator begin() {
+        return buffer_->getRepresentation<BufferRAM>()->getDataContainer().begin();
```
petersteneteg: Does this even compile, BuffeRAM does not have a getDataContainer only the precision type has that...
buffer_->getRAMRepresentation()->getDataContainer() Should work I think

JochenJankowai: Got those two mixed up. `getRepresentation<BufferRAM>` and `getRAMRepresentation` sound like they would do the same thing :)

PR #799, opened 2019-12-08
Core: Fix datatype in dispatching example code

r-englund:


include/inviwo/core/datastructures/image/layerram.h
```diff
@@ -99,7 +99,7 @@ class IVW_CORE_API LayerRAM : public LayerRepresentation {
      *     using LayerType = util::PrecisionType<decltype(lrprecision)>;
      *     using ValueType = util::PrecisionValueType<decltype(lrprecision)>;
      *
-     *     T* data = lrprecision->getDataTyped();
+     *     auto data = lrprecision->getDataTyped();
```
petersteneteg: Might be nice to use ValueType to explain...

petersteneteg:
hmm seems there are some more issues in that code. Maybe we should use doxygens snippets here... although they are worthless if one is reading the source code and not the doxygen page...

petersteneteg:
Like I don't think that count_if will compile...

petersteneteg (approved):

PR #815, opened 2019-12-17
Feature/miscfixes

martinfalk:
some fixes. More verbose range warnings for ordinal properties (includes now property path as well).

Before, it was hard to locate the property.
```
17:44:49.188	OrdinalProperty<int>	ordinalproperty.h	Invalid range given for "Volume" (IntProperty). Using min range as reference.
```


r-englund (approved):

petersteneteg:
The regression failed, we have a problem with background jobs in the threadpool...

petersteneteg (approved):

PR #816, opened 2019-12-17
Docs/guides

xeTaiz:
Hiho,
here's some more progress on written guides, most notably the guide on [Data Structures & Memory Representations](https://github.com/inviwo/inviwo/blob/docs/guides/docs/markdown/DevGuide_Data_Memory.md), the [Python Processor guide](https://github.com/inviwo/inviwo/blob/docs/guides/docs/markdown/DevGuide_PythonProcessors.md) and the [Inviwopy guide](https://github.com/inviwo/inviwo/blob/docs/guides/docs/markdown/DevGuide_InviwoPy.md) (about using Inviwo from within Python).
The changes to other files only include formatting changes for code in markdown.

Find [the complete up-to-date folder here](https://github.com/inviwo/inviwo/tree/docs/guides/docs/markdown).

**On Data Strucutres and Memory Representations**:
I think most in here should be fine, much is taken from the Manual in the wiki. I also did a more detailed example with `dispatch` compared to the example in the "Build your own processor" guide.
Since there are multiple dispatch functions, like `dispatchFront`, `dispatchPool` and possibly more, which I don't know much detail on, I wonder whether `dispatch` deserves its own little guide? Opinions on this? 

**On the Python Processor guide**:
Am I missing important things here? Also, do you think the example is appropriate, considering that adding handles for the `.npy` extension in existing source processors would probably be the better way to actually bring this functionality? I felt like that's an easy one that actually provides new functionality.

**On the InviwoPy guide**:
What do you think are the best uses of Inviwo inside of Python? I thought that it mostly restricts to embedding a finished workspace and using it from within Python, e.g. passing data from Python to the workspace and retrieving the result. And considering this use case, can you think of a more sensible thing to do than replacing source and sink processors in the network programmatically to build a data interface? And if so, do we need a guide on doing this manually by adding connections etc. or rather some new functionality that can help automate this process? 

Let me know what you think :)


docs/markdown/DevGuide_BuildProcessor.md
```diff
@@ -10,7 +10,7 @@ There are three ways to create a new processor in Inviwo:
     Inviwo-meta does all of those things for you automatically. The following describes the files that need to be created or modified:
     1. You need to create the processors source and header files in `<module-dir>/src/processors/`
```
ResearchDaniel, line 11: Headers should be in `<module-dir>/include/modules/<module-name>/processors/`

docs/markdown/DevGuide_Data_Memory.md
```diff
@@ -0,0 +1,85 @@
+# Data Structures
+In Inviwo the main core data structures (`Volume`, `Image`, `Layer`, `Mesh`, `Buffer`) use a pattern of handles and representations. The volume data structure for example has a handle class called `Volume`. The handle class by itself does not have any data, it only stores metadata, and a list of representations. The representations is where the actual data is stored. The following first explains the basic data structures and how they relate, then shows what different representations exist and how to use them.
+
+## The different structures
+In Inviwo, the `Volume` class wraps volumetric data, more specifically data in a structured grid. This corresponds to what you would expect from a rank 3 tensor (or rank 4 if you have a multi-channel volume).
```
petersteneteg: I would not relate our volumes to tensors. Just say 3 dimensional grid or something like that

docs/markdown/DevGuide_Data_Memory.md
```diff
@@ -0,0 +1,85 @@
...
+
+The `Image` class is used to wrap *Frame buffers*, which are in essence packages for multiple images with potentially different shapes and data types. The actual single images inside those frame buffers are represented by a `Layer` in Inviwo.
```
petersteneteg: > The `Image` class consists of a set of `Layers`, usually a color layer (what we would regularly think off as the "image"), a depth layer, and a picking layer.

I would not talk about frame buffers here that is a OpenGL specific thing, and has in principle nothing to do with our Images.

docs/markdown/DevGuide_Data_Memory.md
```diff
@@ -0,0 +1,85 @@
+# Data Structures
+In Inviwo the main core data structures (`Volume`, `Image`, `Layer`, `Mesh`, `Buffer`) use a pattern of handles and representations. The volume data structure for example has a handle class called `Volume`. The handle class by itself does not have any data, it only stores metadata, and a list of representations. The representations is where the actual data is stored. The following first explains the basic data structures and how they relate, then shows what different representations exist and how to use them.
```
petersteneteg: one could note that the basic Data structures here are the Volume, Layer, and Buffer. Image and Mesh are then containers of Layers, and Buffers, respectively. 

And on a side note, that is a bit strange. Both the Image and the Layer as Representation, and the image representation refers to the layer representations. As an example:
```
//      ┌───────────────┐
//      │     Image     ├────┬────────────────────────┬──────────────────────────┐
//      └───────┬───────┘    │     ┌───────────────┐  │   ┌───────────────┐      │  ┌───────────────┐
//              │            └─────▶  Color Layer  │  └───▶  Depth Layer  │      └──▶  Pick Layer   │
//              │                  └───────┬───────┘      └─────────┬─────┘         └───────┬───────┘
//              │                          │                        │                       │
//              │                          │                        │                       │
//         ┌────┴────┐                 ┌───┴─────────┐          ┌───┴────────┐          ┌───┴─────────┐
//         │         │                 │             │          │            │          │             │
//         │         │                 │     ┌───────▼───────┐  │    ┌───────▼───────┐  │     ┌───────▼───────┐
//         │         │                 │  ┌──▶   Layer GL    │  │ ┌──▶   Layer GL    │  │  ┌──▶   Layer GL    │
//         │ ┌───────▼───────┐         │  │  └───────────────┘  │ │  └───────────────┘  │  │  └───────────────┘
//         │ │   Image GL    ├─────────┼──┴─────────────────────┼─┴─────────────────────┼──┘
//         │ └───────────────┘         │                        │                       │
//         │                           │                        │                       │
//         │                   ┌───────▼───────┐        ┌───────▼───────┐       ┌───────▼───────┐
//         │                ┌──▶   Layer RAM   │     ┌──▶   Layer RAM   │    ┌──▶   Layer RAM   │
// ┌───────▼───────┐        │  └───────────────┘     │  └───────────────┘    │  └───────────────┘
// │   Image RAM   ├────────┴────────────────────────┴───────────────────────┘
// └───────────────┘
```

docs/markdown/DevGuide_Data_Memory.md
```diff
@@ -0,0 +1,85 @@
...
+
+## The different structures
+In Inviwo, the `Volume` class wraps volumetric data, more specifically data in a structured grid. This corresponds to what you would expect from a rank 3 tensor (or rank 4 if you have a multi-channel volume).
+
+The `Image` class is used to wrap *Frame buffers*, which are in essence packages for multiple images with potentially different shapes and data types. The actual single images inside those frame buffers are represented by a `Layer` in Inviwo.
+As an effect, most `Image` ports in Inviwo actually transfer multiple images, or one frame buffer, usually consisting of a color layer, depth layer and optionally a picking layer. In this example the `Image` would contain a `UInt8` color layer, a `Float32` depth layer and a `UInt8` picking layer. The picking layer basically encodes object instance IDs in color, so that a lookup in the picking layer gives the object ID for the pixel of interest. This is used for example to drag'n'drop objects in 3D space.
```
petersteneteg: We should link to a section on picking

docs/markdown/DevGuide_Data_Memory.md
```diff
@@ -0,0 +1,85 @@
...
+Of course you can add more layers to the images in your custom processors as necessary.
+
+The third big data structure in Inviwo is the `Mesh` and its `Buffer`s. A `Buffer` is basically a wrapper for some linearly stored data and the `Mesh` holds a list of index buffers and other buffers, where you can add your data accordingly. That means when you load in a standard mesh, the resulting `Mesh` has the appropriate set of index buffers, a vertex buffer and probably a normal buffer.
```
petersteneteg: I would say Index buffers, and a set of data buffers, vertices, normals, colors, texture coordinates, and so on.

docs/markdown/DevGuide_Data_Memory.md
```diff
@@ -0,0 +1,85 @@
...
+
+The `Mesh` is passed on to the next processor with the `RAMRepresentation`s attached. The `Mesh-BB Renderer` shall be a modified `Mesh Renderer` that also computes the meshes bounding box on CPU and takes care of rendering it. We will only look in detail at the bounding box computation and in particular how the vertex data is accessed.
+In the `Mesh-BB Renderer` processor, the first thing we do is computing the bounding box. As a first step, we will retrieve the appropriate `BufferRAM` representation from the `MeshInport` named `meshInport_`:
+```cpp
+std::shared_ptr<const Mesh> mesh = meshInport_.getData(); // Get Mesh handle
+const auto &buffers = mesh->getBuffers(); // Retrieve vector of attached buffers
+// Iterate through buffers to find the Vertex Position buffer
+auto it = std::find_if(buffers.begin(), buffers.end(), [](const auto &buff) {
```
petersteneteg: could use Mesh::getBuffer(BufferType) here

docs/markdown/DevGuide_Data_Memory.md
```diff
@@ -0,0 +1,85 @@
...
+                });
+                // Convert to pair of dvec4, as we specified in dispatch's type parameter
+                return std::pair<dvec4, dvec4>{util::glm_convert<dvec4>(minmax.first),
+                                               util::glm_convert<dvec4>(minmax.second)};
+        });
+        // minmax now has the lower left and upper right of the meshes bounding box.
+    }
+```
+We start as expected: first the `Mesh` handle is retrieved, then we find the buffer of interest, which is the vertex position buffer in our case. If this buffer exists, we will request its `BufferRAM` representation. Since this representation already exists in an up-to-date state, it can be retrieved without any memory transfer. On this representation the `dispatch<ResultType>(Lambda)` function is called. The actual data access is limited to the Lambda that is passed to `dispatch()` and the result type must match the given type parameter. The actual computations on the data happens inside this Lambda. If any additional variables from the outside scope are required in the Lambda, they can be passed into the Lambdas scope by adding them into the brackets like so:
```
petersteneteg, line 72: you might note that that the lambda has to compile for all our 40 or so kinds of ValueType, int, float, vec2 dvec4 etc. as found in https://github.com/inviwo/inviwo/blob/1b65add43ce3f16f65646234529d235d811176a5/include/inviwo/core/util/formats.h#L55
This can quite often lead to some surprising issues...

docs/markdown/DevGuide_Data_Memory.md
```diff
@@ -0,0 +1,85 @@
...
+```cpp
+rep->dispatch<ResultType>( [ &varFromOutsideScope ] (auto br) { ... })
+```
+The Lambda must also take one parameter, in this case `auto br`, that specifies the actual data type of the elements inside the `Buffer`. By using `auto`, we can let C++ deduce the correct data type and define the operation in a type-agnostic way, so that it works for all different types automatically.
```
petersteneteg: there is actually no deduction going on here. you just provide a function for each possible type, in this case by giving a generic function that will be compiles for all 40 kind of types. We then just call the correct one depending on the type of the ram representation.

docs/markdown/DevGuide_InviwoPy.md
```diff
@@ -0,0 +1,72 @@
+# Using Inviwo from within Python
+While Python Processors (TODO: link) allow you to embed Python code inside Inviwo processors, you can also use Inviwopy to use your Inviwo networks from within a Python application. This guide shows you first how to build Inviwopy, then how to load existing Inviwo workspaces for use in your Python scripts and lastly some ways to modify a loaded Inviwo network from within Python.
+
+## Building InviwoPy
+To use Inviwopy, you have to build the appropriate `.dll`/`.so` yourself, since it is currently not available through Pypi. To do so, enable the `IVW_MODULE_PYTHON3` and `IVW_MODULE_PYTHON3QT` CMake flags. Next you need to specify the Python executable to which the produced library shall be compatible in the `PYTHON_EXECUTABLE` flag.
```
petersteneteg, line 5: there is nothing special here, this is exactly the same as when we build the regular python stuff...

docs/markdown/DevGuide_InviwoPy.md
```diff
@@ -0,0 +1,72 @@
...
+1. Set the `PYTHON_EXECUTABLE` flag to your environment's executable (e.g. `~/.conda/envs/inviwo/bin/python3`)
+
+2. If not set automatically, also adapt the `PYTHON_LIBRARY` flag to `<conda env>/lib/libpython3.6m.so` (according to your Python version).
+
+</p>
+</details>
+
+ Depending on your Python version and operating system this will generate two dynamic libraries with names similar to:
+- `inviwopy.cpython-36m-x76_64-linux-gnu.so`
+- `inviwopyapp.cpython-36m-x76_64-linux-gnu.so`
```
petersteneteg, line 21: just to notet these dll depends on most other dll in the bin folder of the build too...

docs/markdown/DevGuide_PythonProcessors.md
```diff
@@ -0,0 +1,82 @@
...
+        ivw.Processor.__init__(self, id, name)
+        self.outport = VolumeOutport("outport")   # Define Outport
+        self.addOutport(self.outport)            # Add port to processor
+
+        self.file = FileProperty("file", "Numpy Volume (.npy)",
+            invalidationLevel=InvalidationLevel(2)) # invalidate resources on change
+        self.addProperty(self.file)
+        self.array = None    # Init to None
+```
+This `__init__` defines a volume outport to pass the loaded array to the network and a file property to locate the serialized Numpy array. Note that this `FileProperty` has its `invalidationLevel=InvalidationLevel(2)`, which lets the `FileProperty` invalidate the processors resources upon changing the property. This will automatically call the `invalidateResources()` method which will take care of actually loading the Numpy file (see below).
```
petersteneteg: invalidateResources  -> initializeResources

petersteneteg: In principle you can just do the file load in the process. It would not make much difference here...

docs/markdown/DevGuide_PythonProcessors.md
```diff
@@ -0,0 +1,82 @@
...
+
+## NumPy Compatibility
+In order to transfer data between Python and C++, the Inviwo data structures  `Volume` (example below), `Layer` (for `Image`s, [example](https://inviwo.org/assets/media/inviwo-vcbm2019.pdf) slide 34-35) and `Buffer` (for `Mesh`es, [example](https://github.com/inviwo/modules/blob/2f07a0fffe916c413a520644b9fe2e45a3ee60a9/misc/vasp/python/vasputil.py#L109-L123)) can take Numpy arrays (`numpy.ndarray`) for initialization.
+
+Loading a Numpy array from disk, wrapping it in a `Volume` and outputting it to the network can be realized as follows:
+```python
+def initializeResources(self):
+    if Path(self.file.value).exists() and self.file.value.endswith('npy'):
+        self.array = np.load(self.file.value)
+    else: print('Invalid file path.')
```
petersteneteg, line 87: throw an exception instead...

petersteneteg (approved):
Nice work! had some minor comments.

docs/markdown/DevGuide_Data_Memory.md
JochenJankowai: I am not an expert (even though I should be) but rank 4 tensors would not describe multi-channel volumes. They could describe time-dependent scalar fields, though. I have no idea how to describe multi-channel volumes in terms of tensors because afaik, the entries in the tensors are scalars.

Correct me if I'm wrong.

petersteneteg: Lets just not involve tensors in this

docs/markdown/DevGuide_InviwoPy.md
xeTaiz: I agree. I would still leave it here and rather make sure it is the same in the other python guide as well. Or what changes to this part do you suggest?

docs/markdown/DevGuide_Data_Memory.md
JochenJankowai: Always the best solution :)

docs/markdown/DevGuide_PythonProcessors.md
xeTaiz: Isn't this separation why `initializeResources()` exists in the first place? I agree that for this example it does not really matter, but I thought I would rather include this "convention" here to make sure that new users can adopt it, even if the example could be easier the other way.

petersteneteg: Actually I'm not 100% sure on why we have this, you might have to ask Timo :) 
But the main reason I think is if your processor depends on some resources that you don't want to reload on every process call but rather only at the beginning. The typical case in inviwo being a compiled shader. Then you only run initializeResources when you change something that will requite the shader to recompile. 
The main difference here being that for the soruce processor, the only reason it has for ever processing is to load the data. And since there is no reason to call process but for loading the data. There is no need to do that in initializeResources. If on the other hand initializeResources loaded the data, and process then took that data and did some processing on that depending on some property. Then I think that would make more sense. 
In this case you have to store the data as a member variable in the processor just to pass it from initializeResources to process which I think just add unnecessary complexity in this case.

petersteneteg:
@xeTaiz merge?

ResearchDaniel (commented):
Sorry, had one comment that apparently was only pending

PR #819, opened 2019-12-31
Add constexpr_hash overload for std::string_view

PetiteViking:


include/inviwo/core/util/constexprhash.h
```diff
@@ -148,6 +148,10 @@ constexpr uint64_t constexpr_hash(const char (&str)[len]) {
     return detail::crc64(str, len);
 }
 
+constexpr uint64_t constexpr_hash(const std::string_view& str) {
```
petersteneteg: Nice. But since string_view is a reference type, it should always be passed by value...

petersteneteg (changes requested):

petersteneteg (approved):

PR #820, opened 2020-01-02
Add optional BufferTarget to Column

PetiteViking:
... such that the index buffer can be declared as such. This is helpful when looping through the columns while wanting to omit the index column.

This doesn't break anything since it's an optional parameter.

### DIFF
--- a/modules/dataframe/include/inviwo/dataframe/datastructures/column.h
+++ b/modules/dataframe/include/inviwo/dataframe/datastructures/column.h
@@ -94,8 +94,8 @@ class TemplateColumn : public Column {
 public:
     using type = T;
 
-    TemplateColumn(const std::string &header,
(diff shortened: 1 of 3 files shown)


petersteneteg:
Is this the same "kind" of index? The index column would generally not work as an index buffer in a mesh right?

JochenJankowai:
I see your point. The way I am using it it's not the same kind of index.

One could argue though that the data frame doesn't have an intrinsic geometry and we could interpret it as a collection of n-dimensional points. In that case the table index would be the same as the vertex index :)

Setting the buffer target to index would be more for convencience. There are of course other ways to omit the index column when looping throught he columns. So if you feel it's wrongly used we can scrap the PR.

petersteneteg:
Lets see what @r-englund thinks

r-englund:
In my world `BufferTarget::Index` should be interpreted as `Index into something` while the index buffer of the dataframe is `Index OF something`  (something = current row).

PR #823, opened 2020-01-08
ShaderSegements

petersteneteg:
OpenGL: reafactoring of the shader code parsing in ShaderOjects. Adde d support for "ShaderSegements" Small snippets that can be insertered into shader code at specific places

### DIFF
--- a/modules/opengl/include/modules/opengl/shader/linenumberresolver.h
+++ b/modules/opengl/include/modules/opengl/shader/linenumberresolver.h
@@ -1,0 +1,52 @@
+/*********************************************************************************
+ *
+ * Inviwo - Interactive Visualization Workshop
+ *
+ * Copyright (c) 2019 Inviwo Foundation
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ *********************************************************************************/
+#pragma once
+
+#include <modules/opengl/openglmoduledefine.h>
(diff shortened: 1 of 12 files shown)


modules/opengl/include/modules/opengl/shader/linenumberresolver.h
```diff
@@ -0,0 +1,52 @@
...
+
+#include <string>
+#include <utility>
+#include <vector>
+
+namespace inviwo {
+
+struct IVW_MODULE_OPENGL_API LineNumberResolver {
+    LineNumberResolver() = default;
+    void addLine(const std::string& file, size_t line);
```
ResearchDaniel, line 48: Might want to clarify this. File is a path/key?

modules/opengl/include/modules/opengl/shader/linenumberresolver.h
```diff
@@ -0,0 +1,52 @@
...
+
+#include <modules/opengl/openglmoduledefine.h>
+
+#include <string>
+#include <utility>
+#include <vector>
+
+namespace inviwo {
+
+struct IVW_MODULE_OPENGL_API LineNumberResolver {
```
ResearchDaniel, line 43: Is this specific to OpenGL shaders? Perhaps we should put this in a namespace in that case?

A bit tricky to figure out what this class does just by looking at the header. What is the basis for resolving line numbers? Would be good with a  description of what this class does :)

modules/opengl/include/modules/opengl/shader/shadersegment.h
```diff
@@ -0,0 +1,86 @@
...
+        }
+        IVW_MODULE_OPENGL_API friend bool operator>=(const Type& lhs, const Type& rhs) {
+            return !operator<(lhs, rhs);
+        }
+
+    private:
+        std::string value_;
+    };
+
+    Type type;
```
ResearchDaniel: What does type mean?

modules/opengl/include/modules/opengl/shader/shadersegment.h
```diff
@@ -0,0 +1,86 @@
...
+    std::string name;
```
ResearchDaniel, line 80: name == the identifier in the shader code to replace?

modules/opengl/include/modules/opengl/shader/shadersegment.h
```diff
@@ -0,0 +1,86 @@
...
+    std::string snippet;
+    size_t priority = 1000;
```
ResearchDaniel, line 86: Should mention that multiple ShaderSegments with the same identifier will be ordered according to priority. Could also mention this in the class description.

modules/opengl/include/modules/opengl/shader/shadersegment.h
```diff
@@ -0,0 +1,86 @@
...
+ *********************************************************************************/
+#pragma once
+
+#include <modules/opengl/openglmoduledefine.h>
+#include <inviwo/core/common/inviwo.h>
+
+namespace inviwo {
+
+/**
+ * Hold a small text snippet to be inserted into shader code
```
ResearchDaniel, line 37: Inserted at or replaces "name"?

ResearchDaniel (changes requested):
Looks neat but some explanations should be added I think

modules/opengl/include/modules/opengl/shader/linenumberresolver.h
petersteneteg: This as been there for years, I just moved it into its own file

ResearchDaniel: Saw that later

petersteneteg:
@ResearchDaniel poke

modules/opengl/include/modules/opengl/shader/shaderobject.h
```diff
@@ -132,6 +147,10 @@ class IVW_MODULE_OPENGL_API ShaderObject {
     bool hasShaderExtension(const std::string& extName) const;
     void clearShaderExtensions();
 
+    void addSegment(ShaderSegment segment);
+    void removeSegments(const std::string& key);
```
ResearchDaniel: Where does the key come from? 
Might want to return the key when adding a segment?

modules/opengl/include/modules/opengl/shader/shadersegment.h
```diff
@@ -0,0 +1,93 @@
...
+        IVW_MODULE_OPENGL_API friend bool operator>=(const Type& lhs, const Type& rhs) {
+            return !operator<(lhs, rhs);
+        }
+
+    private:
+        std::string value_;
+    };
+
+    Type type;            //!< The placeholder that will be replaced
+    std::string name;     //!< A name of the snippet, used to the LineNumberResolver
```
ResearchDaniel: Typo, used by

ResearchDaniel (approved):
Not sure how the shadersegment key is handled but otherwise looks good.

This will also make it easier to add other things in the future? Like special keywords (thinking of being able to derive properties based on the uniforms)?

petersteneteg:
Not sure about adding properties from uniforms. This is mostly going the other way for adding shader code from properties. Or components, coming in a later PR...

PR #824, opened 2020-01-08
Webbrowser: Fix for non-mac systems due to latest CEF updates

ResearchDaniel:
Webhelper executable had incorrect name.

### DIFF
--- a/modules/webbrowser/src/webbrowsermodule.cpp
+++ b/modules/webbrowser/src/webbrowsermodule.cpp
@@ -124,7 +124,7 @@ WebBrowserModule::WebBrowserModule(InviwoApplication* app)
     auto exeExtension = filesystem::getFileExtension(filesystem::getExecutablePath());
     // Assume that inviwo_web_helper is next to the main executable
     auto exeDirectory = filesystem::getFileDirectory(filesystem::getExecutablePath());
-    
+
     auto locale = app->getUILocale().name();
     if (locale == "C") {
         // Crash when default locale "C" is used. Reproduce with GLFWMinimum application
@@ -166,7 +166,7 @@ WebBrowserModule::WebBrowserModule(InviwoApplication* app)
     CefSettings settings;
     // Non-mac systems uses a single helper executable so here we can specify name
     auto subProcessExecutable = exeDirectory + "/cef_web_helper." + exeExtension;
-        if (!filesystem::fileExists(subProcessExecutable)) {
+    if (!filesystem::fileExists(subProcessExecutable)) {
         throw ModuleInitException("Could not find web helper executable:" + subProcessExecutable);
     }
 
--- a/modules/webbrowser/src/webbrowsermodule.cpp
+++ b/modules/webbrowser/src/webbrowsermodule.cpp
@@ -165,7 +165,8 @@ WebBrowserModule::WebBrowserModule(InviwoApplication* app)
     CefMainArgs args;
     CefSettings settings;
     // Non-mac systems uses a single helper executable so here we can specify name
-    auto subProcessExecutable = exeDirectory + "/cef_web_helper." + exeExtension;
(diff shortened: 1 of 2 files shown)


ResearchDaniel:
Hmm, how did one make jenkins run auto-format without pushing a new commit after adding the auto format label?

JochenJankowai:
I have already started new builds on Jenkins after adding the label but that did not fix the issue. For some reason Jenkins did not push the formatting changes.

petersteneteg:
seems to be some problem with pushing the fix
```gitdb.exc.BadName: Ref 'b265ec06a2' did not resolve to an object```

File: modules/webbrowser/src/webbrowsermodule.cpp [hunk 1]
```diff
@@ -161,16 +161,18 @@ WebBrowserModule::WebBrowserModule(InviwoApplication* app)
...
-    if (!filesystem::fileExists(subProcessExecutable)) {
-        subProcessExecutable =
-            cefParentDir +
-            std::string("/Frameworks/Inviwo Helper.app/Contents/MacOS/Inviwo CEF Helper");
-    }
 #else
     CefMainArgs args;
     CefSettings settings;
+    // Non-mac systems uses a single helper executable so here we can specify name
+    auto subProcessExecutable = exeDirectory + "/cef_web_helper." + exeExtension;
```
martinfalk: this looks shady with the trailing `.` and most likely won't work on Linux (no exeExtension). Doesn't CMake take care of this?

martinfalk (changes requested):

modules/webbrowser/webhelper/CMakeLists.txt
```diff
@@ -23,12 +23,13 @@ set(WEBBROWSER_HELPER_SOURCES
...
 )
 
-# place helper in separate executable. CEF uses "<app name> + Helper" by default
+# place helper in separate executable. 
 set(CEF_HELPER_TARGET "cef_inviwo_helper" CACHE INTERNAL "CEF_HELPER_TARGET")
-set(CEF_HELPER_OUTPUT_NAME "${IVW_APP_INSTALL_NAME} Helper")
+
 
 if (NOT OS_MACOSX)
     add_executable(${CEF_HELPER_TARGET} WIN32 MACOSX_BUNDLE ${WEBBROWSER_HELPER_SOURCES})
```
petersteneteg, line 31: You you need to create a app bundle would it not be easier just do do the plain executable? just skip MACOSX_BUNDLE

modules/webbrowser/src/webbrowsermodule.cpp
ResearchDaniel: Probably right that we need to consider linux here as well. 
The CMake option would be to name the helper according to the main executable (inviwo/qt_minimal and so on) but I would rather not do that (although we are forced to do it on mac) because the module becomes application dependent.

modules/webbrowser/webhelper/CMakeLists.txt
ResearchDaniel: Yes, but I vaguely recall that I tried once and that CEF relies on it being a bundle, but I'm not sure. Probably best to follow their example to keep it consistent with their code.

modules/webbrowser/src/webbrowsermodule.cpp
martinfalk: Any particular reason for doing not the same on other platforms?

ResearchDaniel: Should be fixed now

ResearchDaniel: Yes, they use three executables on Mac so it did not work with specifying a single path

martinfalk (approved):

Issue #771, opened 2019-11-14
Set GL interpolation method for textures


PetiteViking:
I am currently trying to render a volume such that every voxel is shown as a brick, colored
according to the cluster it belongs to (i.e. use `GL_NEAREST` as interpolation method).

There is away to do this with something like

```c++
{
    auto interp = interpolationScheme_.get();
    auto volumeGL = volume_->getRepresentation<VolumeGL>();
    TextureUnit unit;
    volumeGL->bindTexture(unit.getEnum());
    glTexParameteri(GL_TEXTURE_3D, GL_TEXTURE_MIN_FILTER, interp);
    glTexParameteri(GL_TEXTURE_3D, GL_TEXTURE_MAG_FILTER, interp);
    volumeGL->unbindTexture();
}
```

but I think it might be nice to be able to set it as a parameter of the volume (or any texture, for
that matter) like so

```c++
auto interp = interpolationScheme_.get();
auto volumeGL = volume_->getRepresentation<VolumeGL>();
volumeGL->setInterpolationMethod(interp);
```

martinfalk:
Definitely, but this information must be potentially stored on a per-Layer basis. Consider setting the interpolation of a GL representation and the next processor requesting a RAM representation. Then the interpolation information is lost.

See implementation of the swizzle mask!

JochenJankowai:
Fixed in #833

PR #833, opened 2020-01-10
Wrapping

petersteneteg:
Added support for Wrapping and Interpolation in layer and volume


modules/base/src/io/datvolumewriter.cpp
```diff
@@ -66,27 +74,43 @@ void DatVolumeWriter::writeData(const Volume* data, const std::string filePath)
...
+        [&](std::string_view key, InterpolationType val) { fmt::print(ss, "{}: {}", key, val); },
+        [&](std::string_view key, const SwizzleMask& mask) {
+            fmt::print(ss, "{}: {}{}{}{}", key, mask[0], mask[1], mask[2], mask[3]);
+        },
+        [&](std::string_view key, const Wrapping3D& wrapping) {
+            fmt::print(ss, "{}: {} {} {}", key, wrapping[0], wrapping[1], wrapping[2]);
+        },
+        [&](std::string_view key, const auto& vec) {
+            fmt::print(ss, "{}: {}", key, fmt::join(begin(vec), end(vec), " "));
+        }};
```
martinfalk, line 89: that's pretty neat

martinfalk (approved):
looks good

martinfalk:
Takes care of #771

modules/base/src/io/datvolumewriter.cpp
petersteneteg: yes

PR #841, opened 2020-01-23
Feature/brickiterator

martinfalk:
iterator for iterating over a brick (or subvolume) within 3D data, e.g. a volume.

### DIFF
--- a/include/inviwo/core/util/indexmapper.h
+++ b/include/inviwo/core/util/indexmapper.h
@@ -54,7 +54,7 @@ struct IndexMapper<2, IndexType> {
     }
 
 private:
-    const IndexType dimx;
+    IndexType dimx;
 };
 
 template <typename IndexType>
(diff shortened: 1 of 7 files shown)


petersteneteg (approved):

include/inviwo/core/util/indexmapper.h
```diff
@@ -54,7 +54,7 @@ struct IndexMapper<2, IndexType> {
     }
 
 private:
-    const IndexType dimx;
+    IndexType dimx;
```
r-englund, line 57: Why remove const here?

petersteneteg: To allow the assignment operator to work

r-englund: But, there are no assignment operator :P

r-englund (changes requested):
Needs documentation on what it is and how to use it.  (as we discussed IRL)

r-englund (approved):

PR #843, opened 2020-01-23
Core: Serialize SerializationMode, fixes #839

### LINKED ISSUE #839: Property Serialization mode is not remebered for dynamically added properties.
When loading a network with a processor having one or more dynamically added properties with serialization mode set to `All`, the properties recreated by the deserializer does not have the serialization mode set to All. 

For example: The parallel coordinates have a `DataFrameColormapProperty`. The `DataFrameColormapProperty` dynamically adds a bunch of `ColormapProperty` and set them their serialization mode to `All`.   https://github.com/inviwo/inviwo/blob/1f3aba262ce7f3b9a9f5773d08992947c00657b1/modules/dataframe/src/properties/dataframecolormapproperty.cpp#L101-L114

When the network is saved and have been reloaded the serialization mode of the properties are `Default` which means the next time the network is saved information such as display name etc. will be lost. 

**Suggestion:** In `PropertyOwner`, when adding a property to `ownedProperties_` the serialization mode should be set to `All`.   IMO, this makes sense since we need it for properties recreated by the deserializer to get, e.g, displayname correct

r-englund:
Serialize SerializationMode, fixes #839

### DIFF
--- a/include/inviwo/core/processors/compositeprocessor.h
+++ b/include/inviwo/core/processors/compositeprocessor.h
@@ -179,6 +179,8 @@ private:
     virtual void onSetReadOnly(Property* property, bool readonly) override;
     virtual void onSetVisible(Property* property, bool visible) override;
     virtual void onSetUsageMode(Property* property, UsageMode usageMode) override;
+    virtual void onSetSerializationMode(Property* property,
+                                     PropertySerializationMode serializationMode) override;
 
     // ProcessorNetworkObserver overrides
     virtual void onProcessorNetworkDidAddProcessor(Processor*) override;
--- a/include/inviwo/core/processors/compositeprocessor.h
+++ b/include/inviwo/core/processors/compositeprocessor.h
@@ -180,7 +180,7 @@ private:
     virtual void onSetVisible(Property* property, bool visible) override;
     virtual void onSetUsageMode(Property* property, UsageMode usageMode) override;
     virtual void onSetSerializationMode(Property* property,
-                                     PropertySerializationMode serializationMode) override;
+                                        PropertySerializationMode serializationMode) override;
 
     // ProcessorNetworkObserver overrides
     virtual void onProcessorNetworkDidAddProcessor(Processor*) override;
--- a/include/inviwo/core/properties/property.h
+++ b/include/inviwo/core/properties/property.h
@@ -378,7 +378,7 @@ protected:
     void notifyAboutChange();
 
     CallBackList onChangeCallback_;
-    PropertySerializationMode serializationMode_;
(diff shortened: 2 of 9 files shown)


ResearchDaniel:
I think we need to add documentation on what the different modes do.
enum class PropertySerializationMode { Default = 0, All, None };

petersteneteg (changes requested):
Why do we need any on change callback for this?

include/inviwo/core/properties/property.h
```diff
@@ -378,7 +378,7 @@ class IVW_CORE_API Property : public PropertyObservable,
     void notifyAboutChange();
 
     CallBackList onChangeCallback_;
-    PropertySerializationMode serializationMode_;
+    ValueWrapper<PropertySerializationMode> serializationMode_;
```
petersteneteg, line 381: make no sense to put this in a value wrapper

include/inviwo/core/processors/compositeprocessor.h
```diff
@@ -179,6 +179,8 @@ class IVW_CORE_API CompositeProcessor : public Processor,
     virtual void onSetReadOnly(Property* property, bool readonly) override;
     virtual void onSetVisible(Property* property, bool visible) override;
     virtual void onSetUsageMode(Property* property, UsageMode usageMode) override;
+    virtual void onSetSerializationMode(Property* property,
```
petersteneteg, line 182: why?

include/inviwo/core/properties/propertyobserver.h
```diff
@@ -33,6 +33,7 @@
 #include <inviwo/core/common/inviwocoredefine.h>
 #include <inviwo/core/common/inviwo.h>
 #include <inviwo/core/util/observer.h>
+#include <inviwo/core/properties/valuewrapper.h>
```
petersteneteg, line 36: why

src/core/processors/compositeprocessor.cpp
```diff
@@ -319,4 +319,11 @@ void CompositeProcessor::onSetUsageMode(Property* orgProp, UsageMode) {
     }
 }
 
+void CompositeProcessor::onSetSerializationMode(Property* orgProp,
+                                                PropertySerializationMode serializationMode) {
+    if (auto superProperty = getSuperProperty(orgProp)) {
```
petersteneteg, line 324: why? how will this interact with line 274? that is the reasoning here?

petersteneteg (changes requested):

r-englund:
I'll reopen this when I get the time

PR #844, opened 2020-01-24
Base: IvfVolumeReader: Make sure enums are initilized

r-englund:
On my machine (windows, RelWithDebInfo) the enums were initialized to char values (always 64 for some reason) which did not represent any valued enum value and no rendering worked properly. 

Now default initialize them to their default values of Volumes


petersteneteg (changes requested):
This is completly broken, you never set the deserialised values on the volume.

r-englund:
> 
> 
> This is completly broken, you never set the deserialised values on the volume.

I see what you mean, that needs to be fixed. I guess it "worked" since the deserialized values are used when we crate the VolumeDisk representation some lines bellow

petersteneteg (approved):

PR #849, opened 2020-01-30
Mesh and Shader fixes

### LINKED ISSUE #847: parseShaderSource fails to parse with some types of block comments
For example, the following code will not be correctly parsed. While debugging, after the parser has reached the eof  the state mahcine says it is still in `BlockCom1` state. 
```c++
/*
**/
```

### LINKED ISSUE #848: Shader Preivew is not showing parsed code for shaders that has not yet been compiled
**Steps to reproduce**
In an empty network:
1. Add a `Background` processor
2. Open the shader preview for background.frag from the shader menu
3. Press the pound symbol (i.e `#`) to see the preview
4. Result:  
image

petersteneteg:
closes #847
closes #848

### DIFF
--- a/modules/opengl/include/modules/opengl/buffer/bufferobject.h
+++ b/modules/opengl/include/modules/opengl/buffer/bufferobject.h
@@ -81,8 +81,6 @@ public:
     void bindBase(GLuint index) const;
     void bindRange(GLuint index, GLintptr offset, GLsizeiptr size) const;
 
-
-
     /**
      * \brief bind the buffer object and set the vertex attribute pointer
      *
--- a/modules/opengl/include/modules/opengl/buffer/bufferobject.h
+++ b/modules/opengl/include/modules/opengl/buffer/bufferobject.h
@@ -75,10 +75,32 @@ public:
     GLFormat getGLFormat() const;
     const DataFormatBase* getDataFormat() const;
 
+    /**
+     * \brief Calls glBindBuffer.
+     */
     void bind() const;
+
+    /**
+     * \brief Calls glBindBuffer with buffer name 0
+     */
     void unbind() const;
 
+    /**
+     * \brief Calls glBindBufferBase.
+     * Binds the buffer at index of the array of targets
+     * specified by the associated target ( @see getTarget )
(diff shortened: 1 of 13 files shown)


modules/opengl/include/modules/opengl/buffer/bufferobject.h
```diff
@@ -75,6 +78,11 @@ class IVW_MODULE_OPENGL_API BufferObject : public Observable<BufferObjectObserve
     void bind() const;
     void unbind() const;
 
+    void bindBase(GLuint index) const;
```
ResearchDaniel, line 95: Unclear what this means and how it is different from bind()

modules/opengl/include/modules/opengl/buffer/bufferobject.h
```diff
@@ -75,6 +78,11 @@ class IVW_MODULE_OPENGL_API BufferObject : public Observable<BufferObjectObserve
...
+    void bindRange(GLuint index, GLintptr offset, GLsizeiptr size) const;
```
ResearchDaniel: What is index? See comment above

modules/opengl/include/modules/opengl/buffer/bufferobject.h
ResearchDaniel: How about 
```
/*
 * \brief Calls glBindBufferBase. 
 * Binds the buffer at bufferTargetIndex of the array of targets 
 * specified by the associated target ( @see getTarget )
 * Targets must be one of GL_ATOMIC_COUNTER_BUFFER, 
 * GL_TRANSFORM_FEEDBACK_BUFFER, GL_UNIFORM_BUFFER or GL_SHADER_STORAGE_BUFFER
 */ 
void bind(GLuint bufferTargetIndex)
```

ResearchDaniel (changes requested):

r-englund (approved):
Could cherry-pick 2c03b14b6eeea05eb2990dd05cbda4a1940eb191 for updated unittest

ResearchDaniel (approved):

PR #853, opened 2020-02-03
Core: Serialize SerializationMode of properties, fixes #839

### LINKED ISSUE #839: Property Serialization mode is not remebered for dynamically added properties.

r-englund:
Serialize SerializationMode of properties, fixes #839

### DIFF
--- a/src/core/properties/property.cpp
+++ b/src/core/properties/property.cpp
@@ -198,6 +198,7 @@ bool Property::isModified() const { return propertyModified_; }
 void Property::serialize(Serializer& s) const {
     s.serialize("type", getClassIdentifier(), SerializationTarget::Attribute);
     s.serialize("identifier", identifier_, SerializationTarget::Attribute);
+    s.serialize("serializationMode", serializationMode_);
     displayName_.serialize(s, serializationMode_);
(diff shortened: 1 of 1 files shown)


r-englund:
Doesn't work, will need more work

PR #857, opened 2020-02-10
DataFrame: addColumn takes std::vector<T>

PetiteViking:


modules/dataframe/include/inviwo/dataframe/datastructures/dataframe.h
```diff
@@ -211,6 +220,14 @@ std::shared_ptr<TemplateColumn<T>> DataFrame::addColumn(const std::string &heade
     return col;
 }
 
+template <typename T>
+std::shared_ptr<TemplateColumn<T>> DataFrame::addColumn(const std::string &header,
+                                                        const std::vector<T> &data) {
+    auto col = std::make_shared<TemplateColumn<T>>(header);
```
petersteneteg: To avoid the copy it this would be better:
```template <typename T>
std::shared_ptr<TemplateColumn<T>> DataFrame::addColumn(const std::string &header,
                                                        std::vector<T> data) {
    auto ram = std::make_shared<BufferRAMPrecision<T>>(std::move(data));
    auto buf = std::make_shared<Buffer>(ram);
    auto col = std::make_shared<TemplateColumn<T>>(header, buff);
    columns_.push_back(col);
    return col;
}
```

petersteneteg (changes requested):

r-englund (approved):

modules/dataframe/include/inviwo/dataframe/datastructures/dataframe.h
```diff
@@ -221,11 +220,9 @@ std::shared_ptr<TemplateColumn<T>> DataFrame::addColumn(const std::string &heade
 }
 
 template <typename T>
-std::shared_ptr<TemplateColumn<T>> DataFrame::addColumn(const std::string &header,
-                                                        const std::vector<T> &data) {
-    auto col = std::make_shared<TemplateColumn<T>>(header);
-    col->getTypedBuffer()->getEditableRAMRepresentation()->getDataContainer() = data;
-    return col;
+auto DataFrame::addColumn(const std::string &header, const std::vector<T> data) {
```
petersteneteg: don't make it const, then you cant move from it....

petersteneteg (changes requested):

modules/dataframe/include/inviwo/dataframe/datastructures/dataframe.h
JochenJankowai: Interesting! I feel like the compiler could mention that...

petersteneteg: just, unfortunately it just silently falls back to doing a copy :(

petersteneteg (approved):

PR #866, opened 2020-02-13
CImg: OpenMP fix

petersteneteg:


r-englund (changes requested):
Doesn't work on my machine, investigating

r-englund:
> 
> 
> Doesn't work on my machine, investigating

OpenMP_CXX_VERSION is not set on my machine (MSVC 2019 16.3.5) 

The following Vairables starting with OpenMP is set on my machine: 
```
OpenMP_COMPILE_RESULT_CXX_openmp=TRUE
OpenMP_COMPILE_RESULT_C_openmp=TRUE
OpenMP_CXX_FIND_QUIETLY=1
OpenMP_CXX_FLAGS=-openmp
OpenMP_CXX_FOUND=TRUE
OpenMP_CXX_INCLUDE_DIRS=
OpenMP_CXX_LIBRARIES=
OpenMP_CXX_LIB_NAMES=
OpenMP_CXX_SPEC_DATE=200203
OpenMP_C_FIND_QUIETLY=1
OpenMP_C_FLAGS=-openmp
OpenMP_C_FOUND=TRUE
OpenMP_C_INCLUDE_DIRS=
OpenMP_C_LIBRARIES=
OpenMP_C_LIB_NAMES=
OpenMP_C_SPEC_DATE=200203
OpenMP_FINDLIST=C;CXX;Fortran
OpenMP_FOUND=TRUE
OpenMP_ON=ON
OpenMP_SPECTEST_CXX_=TRUE
OpenMP_SPECTEST_C_=TRUE
```

r-englund:
cmake script to print OpenMP varaibles 
```
get_cmake_property(_variableNames VARIABLES)
list (SORT _variableNames)
foreach (_variableName ${_variableNames})
	if(_variableName MATCHES "^OpenMP.*")
		message("${_variableName}=${${_variableName}}")
	endif()
endforeach()
```

petersteneteg:
https://cmake.org/cmake/help/latest/module/FindOpenMP.html
said it should be there... :/

petersteneteg:
can you try with OpenMP_CXX_SPEC_DATE instead?

r-englund:
There seems to be a CMake bug. `FindOpenMP.cmake` fails to set the `OpenMP_CXX_VERSION` when rerunning CMake. [1]
Adding `unset(OpenMP_CXX_SPEC_DATE CACHE)` before the `find_package(OpenMP QUIET)` fixes the issue on my machine 
[1] : https://gitlab.kitware.com/cmake/cmake/issues/19150

r-englund:
Also, it seems that the generator expression used is always true, the ${} around _VERSION is missing and it seems the 3.0 should be without quatation marks. This works on my machine: 
```
$<$<AND:$<VERSION_GREATER_EQUAL:${OpenMP_CXX_VERSION},3.0>,$<BOOL:${IVW_OPENMP_ON}>>:cimg_use_openmp>
```

r-englund:
> 
> 
> can you try with OpenMP_CXX_SPEC_DATE instead?

In theory that would work, but its less future proof than the version comparison. When unsetting the `OpenMP_CXX_SPEC_DATE` the `OpenMP_CXX_VERSION` works as expected. 
There might also be other code, in for example external libraries or external modules that rely on `OpenMP_CXX_VERSION`

r-englund:
the list that cmake uses to convert to the date to version is: 
```cmake
 set(OpenMP_SPEC_DATE_MAP
    # Preview versions
    "201611=5.0" # OpenMP 5.0 preview 1
    # Combined versions, 2.5 onwards
    "201811=5.0"
    "201511=4.5"
    "201307=4.0"
    "201107=3.1"
    "200805=3.0"
    "200505=2.5"
    # C/C++ version 2.0
    "200203=2.0"
    # Fortran version 2.0
    "200011=2.0"
    # Fortran version 1.1
    "199911=1.1"
    # C/C++ version 1.0 (there's no 1.1 for C/C++)
    "199810=1.0"
    # Fortran version 1.0
    "199710=1.0"
  )
  if(MSVC)
    list(APPEND OpenMP_SPEC_DATE_MAP "2019=2.0")
  endif()
```
its seems that MSVC sometimes reports data as 2019 which will make the comparison less trival

petersteneteg:
Maybe just disable it on msvc instead then..

r-englund (approved):

r-englund:
> 
> 
> Maybe just disable it on msvc instead then..

Now when the generator expression has been fixed it will not be enabled on msvc. Since no version is not greater or equal to 3.0

PR #867, opened 2020-02-13
Add min/max to dataframe port inspector info

PetiteViking:


modules/dataframe/include/inviwo/dataframe/datastructures/dataframe.h
```diff
@@ -240,13 +240,40 @@ struct DataTraits<DataFrame> {
...
+
+                        auto [min, max] = std::minmax_element(vec.begin(), vec.end());
+                        minString = std::to_string(*min);
+                        maxString = std::to_string(*max);
+
+                        using ValueType = util::PrecisionValueType<decltype(brprecision)>;
+
+                        if constexpr (std::is_floating_point_v<ValueType>) {
+                            minString = minString.substr(0, minString.rfind('.') + 3);
+                            maxString = maxString.substr(0, maxString.rfind('.') + 3);
```
r-englund: I think this will have problem if thata contains `nan` or `inf`

ResearchDaniel: Could the existing algorithms be used (is it possible to move the info generation to base):
https://github.com/inviwo/inviwo/blob/master/modules/base/include/modules/base/algorithm/dataminmax.h

r-englund: I think the use of the dataminmax in base is smart. But it should still be done here, not moved to base. 
the `util::bufferMinMax` reaturn dvec4 even for scalars and vec2/3, so some care still need to be taken into acount when converting to string.

r-englund: > 
> 
> I think this will have problem if thata contains `nan` or `inf`

You can use `fmt::format("{:.3}", 3.14)` to format all floats to have a maximum of 3 decimals

JochenJankowai:
`NaN` and `Inf` are handled now and the formatting is done via `fmt`.

I do not agree with using the `bufferMinMax` method, as it treats everything as `double` and always returns a `dvec4` which is not only unnecessary but also results in floating point representation in the returned string.

modules/dataframe/include/inviwo/dataframe/datastructures/dataframe.h
martinfalk: and what about data being in the range of -0.00000041 and 0.000000003? -> `min = 0.000, max = 0.000`?

martinfalk: @ResearchDaniel dataminmax has the problem of always returning a `double`.

JochenJankowai: We could use std::scientific for floating point numbers?

martinfalk (approved):

JochenJankowai:
@martinfalk Do you think `std::defaultfloat` gives an appropriate representation? (see image).

dfinfo

petersteneteg (changes requested):
What happend here? did you merge master into this branch? do a rebase please

PR #868, opened 2020-02-14
CMake: Added support for CMake 3.16 PCH

### LINKED ISSUE #827: FYI: Inviwo fails to build on latest version of Visual Studio 2019 (16.4.2 & 16.4.3)
Failed to compile three different cpp files with internal compiler error. This is (most likley) due to a bug in the compiler, not inviwo. 
Probably same issue as https://developercommunity.visualstudio.com/content/problem/856788/compiler-crash-ice-involving-precompiled-headers-a.html 

**Disabling precompiled headers lets you compile without errors**

r-englund:
This commit adds support to use the Precompiled Header support introduced in CMake 3.16. Will fallback to cotire on earlier version and cotire is still an option (opt in) new versions. 
Fixes #827. 

CMake has two ways of using PCH. (1): Each project crates/compiles their own PCH artifacts (2): Projects can reuse PCH artifacts from other projects (e.g. vectorvis can reuse artifacts from core). 

Unfortunately, the second option is not stable and I have opted for defaulting to option one. Which is the same approach that cotire uses.  

Since cotire builds fails on MSVC 2019 i've have not be able to compare build times between CMake PCH and cotire PCH latley. But I did a few month ago and then they were about the same. 

Below are some build time measures on my machine: 

**Inviwo core projects.  A total of 99 projects**
|                       Name | Buildtime                         |
|--------------------------- | --------------------------------- |
|            Reuse-dll-Debug |    4 m 44 sec                     |
|            CMake-lib-Debug |     5 m 9 sec                     |
|            CMake-dll-Debug |    5 m 18 sec                     |
|          Reuse-dll-Release |    6 m 44 sec                     |
|   Reuse-dll-RelWithDebInfo |    6 m 50 sec                     |
|            NoPCH-dll-Debug |     7 m 1 sec                     |
|          CMake-dll-Release |     7 m 3 sec                     |
|            NoPCH-lib-Debug |     7 m 4 sec                     |
|   CMake-dll-RelWithDebInfo |     7 m 9 sec                     |
|          CMake-lib-Release |    7 m 25 sec                     |
|   CMake-lib-RelWithDebInfo |    7 m 27 sec                     |
|   NoPCH-lib-RelWithDebInfo |     9 m 1 sec                     |
|          NoPCH-dll-Release |     9 m 9 sec                     |
|   NoPCH-dll-RelWithDebInfo |    9 m 11 sec                     |
|          NoPCH-lib-Release |    9 m 11 sec                     |
|           Cotire-lib-Debug |    2 m 13 sec (failed to compile) |
|  Cotire-lib-RelWithDebInfo |    2 m 43 sec (failed to compile) |
|         Cotire-lib-Release |    2 m 57 sec (failed to compile) |
|           Cotire-dll-Debug |    4 m 40 sec (failed to compile) |
|         Cotire-dll-Release |    6 m 12 sec (failed to compile) |
|  Cotire-dll-RelWithDebInfo |    6 m 26 sec (failed to compile) |

**Inviwo core projects + most in inviwo-modules + some research-dev + my own modules.  A total of 145 projects**
|                      Name  | Buildtime                         |
|--------------------------- | --------------------------------- |
|            Reuse-dll-Debug |     5 m 4 sec                     |
|            CMake-dll-Debug |     6 m 5 sec                     |
|            CMake-lib-Debug |    6 m 24 sec                     |
|   Reuse-dll-RelWithDebInfo |    7 m 28 sec                     |
|          Reuse-dll-Release |    7 m 37 sec                     |
|   CMake-dll-RelWithDebInfo |    7 m 56 sec                     |
|          CMake-dll-Release |     8 m 9 sec                     |
|            NoPCH-dll-Debug |    8 m 38 sec                     |
|   CMake-lib-RelWithDebInfo |    8 m 55 sec                     |
|            NoPCH-lib-Debug |     9 m 0 sec                     |
|          CMake-lib-Release |    9 m 18 sec                     |
|   NoPCH-dll-RelWithDebInfo |   10 m 51 sec                     |
|   NoPCH-lib-RelWithDebInfo |   10 m 55 sec                     |
|          NoPCH-dll-Release |    11 m 9 sec                     |
|          NoPCH-lib-Release |   11 m 33 sec                     |
|           Cotire-lib-Debug |    2 m 10 sec (failed to compile) |
|  Cotire-lib-RelWithDebInfo |    2 m 42 sec (failed to compile) |
|         Cotire-lib-Release |    2 m 58 sec (failed to compile) |
|           Cotire-dll-Debug |    4 m 50 sec (failed to compile) |
|  Cotire-dll-RelWithDebInfo |    6 m 46 sec (failed to compile) |
|         Cotire-dll-Release |    6 m 48 sec (failed to compile) |


cmake/globalmacros.cmake
```diff
@@ -538,7 +538,15 @@ function(ivw_create_module)
 
     # Optimize compilation with pre-compilied headers
     if(NOT ARG_NO_PCH)
-        ivw_compile_optimize_on_target(${${mod}_target})
+        if(IVW_PRECOMPILED_HEADERS_MODE STREQUAL "Reuse")
```
petersteneteg: this should be inside of ivw_compile_optimize_on_target

modules/base/CMakeLists.txt
```diff
@@ -261,7 +261,9 @@ ivw_add_unittest(${TEST_FILES})
 
 #--------------------------------------------------------------------
 # Create module
-ivw_create_module(${SOURCE_FILES} ${MOC_FILES} ${HEADER_FILES})
+ivw_create_module(NO_PCH ${SOURCE_FILES} ${MOC_FILES} ${HEADER_FILES})
```
petersteneteg, line 264: why?

src/core/CMakeLists.txt
```diff
@@ -907,3 +907,9 @@ else()
     install(FILES ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/$<CONFIG>/inviwo_buildinfo.ini DESTINATION ${IVW_RUNTIME_INSTALL_DIR} COMPONENT modules CONFIGURATIONS Release Debug)
 endif()
 #--------------------------------------------------------------------
+
+
+set(PCH_HEADERS ${HEADER_FILES})
+list(REMOVE_ITEM PCH_HEADERS "${IVW_INCLUDE_DIR}/inviwo/core/common/defaulttohighperformancegpu.h")
```
petersteneteg, line 914: explain why

petersteneteg (changes requested):
some questions.
where does pch get enabled for a target when using the cmake mode?

modules/base/CMakeLists.txt
r-englund: It is turned on by the `ivw_target_pch_headers`-call below.

r-englund: I've done some minor changes so I can probably remove it, investigating.

r-englund:
> e does pch get enabled for a target when using the cmake mode?

By calling `target_precompile_headers()`[1] for a given target with a list of header files. If the `target_precompile_headers` is called with `PUBLIC` or `INTERFACE` any dependent target will also include those headers in their precompiled header. If any of a targets dependencies has called `target_precompile_headers` precompiled headers will be enabled for that target, which can be disabled by setting the target property `DISABLE_PRECOMPILE_HEADERS` 
We call `target_precompile_headers` with PUBLIC in the call to `ivw_target_pch_headers` for base and core.  No other core-module adds more PCH-headers, my initial experimentation did only see a significant compile-time reduction when enabling this for core and base, especially when using `REUSE`

[1] https://cmake.org/cmake/help/latest/command/target_precompile_headers.html

modules/base/CMakeLists.txt
r-englund: Could not remove it, the call to `ivw_target_pch_headers` will then result in cmake error `PRECOMPILE_HEADERS_REUSE_FROM property is already set on target`

src/core/CMakeLists.txt
r-englund: done

cmake/globalmacros.cmake
r-englund: done

r-englund:
New PR #881 due to jenkins regression tests

PR #871, opened 2020-02-18
Feature/stereo

petersteneteg:


src/core/datastructures/camera.cpp
```diff
@@ -271,105 +313,90 @@ bool SkewedPerspectiveCamera::update(const Camera* source) {
...
-            glm::any(glm::notEqual(lhs.frustumSkewOffset_, rhs.frustumSkewOffset_)));
-}
-
-float SkewedPerspectiveCamera::getAspectRatio() const {
-    // Left, right, bottom, top view volume
-    const float width{frustum_.y - frustum_.x};
-    const float height{frustum_.w - frustum_.z};
-    return width / height;
+    return (lhs.equalTo(rhs) || (lhs.fovy_ != rhs.fovy_) ||
+            (lhs.aspectRatio_ != rhs.aspectRatio_) |
```
martinfalk: why not `!operator==()` ?

src/core/datastructures/camera.cpp
```diff
@@ -271,105 +313,90 @@ bool SkewedPerspectiveCamera::update(const Camera* source) {
...
-void SkewedPerspectiveCamera::setAspectRatio(float val) {
-    // Left, right, bottom, top view volume
-    const float width{frustum_.y - frustum_.x};
-    const float height{width / val};
-    frustum_.z = -height / 2.0f;
-    frustum_.w = +height / 2.0f;
-    invalidateProjectionMatrix();
+mat4 SkewedPerspectiveCamera::calculateViewMatrix() const {
+    const vec3 xoffset{offset_.x * glm::normalize(glm::cross(lookTo_ - lookFrom_, lookUp_))};
+    const vec3 yoffset{offset_.y * lookUp_};
```
martinfalk, line 368: is `lookUp_` guaranteed to be orthogonal to view direction at all times?

martinfalk (approved):
nice work!

src/core/datastructures/camera.cpp
petersteneteg: ahh did I not change that one... changed the others...

src/core/datastructures/camera.cpp
petersteneteg: No not if you set it manually....

petersteneteg: but the same is true for the other cameras...

martinfalk: just great... Maybe up should be re-orthogonalized when it is set?

include/inviwo/core/datastructures/camera.h
```diff
@@ -73,13 +73,13 @@ class IVW_CORE_API Camera : public Serializable {
     virtual void configureProperties(CompositeProperty* comp, Config config) = 0;
 
     const vec3& getLookFrom() const;
-    void setLookFrom(vec3 val);
+    virtual void setLookFrom(vec3 val);
```
ResearchDaniel, line 76: Do these need to be virtual?

include/inviwo/core/datastructures/camera.h
```diff
@@ -204,44 +207,43 @@ bool operator!=(const PerspectiveCamera& lhs, const PerspectiveCamera& rhs);
  * @see Camera
  * @see OrthographicCamera
  */
-class IVW_CORE_API OrthographicCamera : public Camera {
+class IVW_CORE_API OrthographicCamera final : public Camera {
 public:
     OrthographicCamera(vec3 lookFrom = vec3(0.0f, 0.0f, 2.0f), vec3 lookTo = vec3(0.0f),
                        vec3 lookUp = vec3(0.0f, 1.0f, 0.0f), float nearPlane = 0.01f,
-                       float farPlane = 10000.0f, vec4 frustum = vec4(-01, 10, -10, 10));
+                       float farPlane = 10000.0f, float width = 10.f, float aspectRatio = 1.f);
```
ResearchDaniel, line 214: Does it not make more sense to specify width, height instead of width, aspectRatio?

ResearchDaniel (commented):
Nice, some minor comments

include/inviwo/core/datastructures/camera.h
martinfalk: nope, since width is the `width` of the orthographic camera frustum in world space and the aspect ratio comes from the canvas/viewport. Two entirely different things!

include/inviwo/core/datastructures/camera.h
martinfalk: Yes! Different cameras do different things. In particular perspective camera vs. orthographic. That was one of the reasons why panning in orthographic was broken.

martinfalk (approved):

include/inviwo/core/datastructures/camera.h
ResearchDaniel: Hm, the projection matrix calculation seem to use width/height derived from the aspect ratio:
    const float halfHeight = halfWidth / aspectRatio_;

petersteneteg: yes, usually we want the aspect of the frustum to be the same as that for the canvas.

ResearchDaniel: I see, I guess it is much easier to deal with the aspect ratio than to change the height. We might want to consider adding some explanation to the parameters.

ResearchDaniel (approved):
Do we need to add converters for this to be backwards compatible?
Should also update the changelog

include/inviwo/core/datastructures/camera.h
martinfalk: agree with that

modules/python3/bindings/src/pypickingmapper.cpp
```diff
@@ -89,28 +108,26 @@ void exposePickingMapper(pybind11::module &m) {
...
-                      [](PickingEvent *e, bool used) {
-                          if (used) {
-                              e->markAsUsed();
-                          } else {
-                              e->markAsUnused();
-                          }
-                      })
-        .def("markkAsUsed", &PickingEvent::markAsUsed)
-        .def("markkAsUnused", &PickingEvent::markAsUnused)
-        .def("setToolTip", &PickingEvent::setToolTip);
```
martinfalk, line 108: is there a reason for removing `setToolTip`?

martinfalk (approved):

modules/python3/bindings/src/pypickingmapper.cpp
petersteneteg: ahh was thinking is was in the base... added it again...

PR #876, opened 2020-02-27
Feature/dataframe info2

petersteneteg:
rebase of #867

### DIFF
--- a/modules/dataframe/include/inviwo/dataframe/datastructures/dataframe.h
+++ b/modules/dataframe/include/inviwo/dataframe/datastructures/dataframe.h
@@ -41,6 +41,7 @@
 #include <inviwo/core/ports/dataoutport.h>
 #include <inviwo/core/util/exception.h>
 #include <unordered_map>
+#include <fmt/format.h>
 
 namespace inviwo {
 class DataPointBase;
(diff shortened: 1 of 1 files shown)


modules/dataframe/include/inviwo/dataframe/datastructures/dataframe.h
```diff
@@ -240,13 +241,59 @@ struct DataTraits<DataFrame> {
...
+                        [](auto brprecision) -> std::pair<std::string, std::string> {
+                            using ValueType = util::PrecisionValueType<decltype(brprecision)>;
+
+                            const auto &vec = brprecision->getDataContainer();
+
+                            if (vec.empty()) {
+                                return {"-", "-"};
+                            }
+
+                            if constexpr (std::is_floating_point_v<ValueType>) {
```
r-englund, line 266: These two tests are unnecessary, the `stringstream` handles them correctly. Especially in the case of `inf`, where one of them can be valid when the other is neg-inf/inf.  
These tests were only needed before when substring based on the dot location was used.

r-englund (approved):
see comment in code about the test for infinity/nan

modules/dataframe/include/inviwo/dataframe/datastructures/dataframe.h
JochenJankowai: True. I also noticed that this will throw a runtime error if the data frame contains vector types. I'll fix that soon.

martinfalk (changes requested):
Nice! Fix runtime errors for non-scalar types.

JochenJankowai:
See #1112

PR #881, opened 2020-03-03
CMake: Added support for CMake 3.16 PCH

r-englund:
Rebase of #868

### DIFF
--- a/cmake/globalconfig.cmake
+++ b/cmake/globalconfig.cmake
@@ -254,13 +254,45 @@ endif()
 
 #--------------------------------------------------------------------
 # Precompile headers
-if(WIN32)
-    option(PRECOMPILED_HEADERS "Create and use precompilied headers" ON)
+set(IVW_DEFAULT_PCH_MODE "No PCH")
+
+set(IVW_TURN_ON OFF) # Default off
+if(DEFINED PRECOMPILED_HEADERS) # Respect the value of our old variable, for backwards compatibility with old cmake-caches
+    set(IVW_TURN_ON ${PRECOMPILED_HEADERS})
+elseif(WIN32)
+    set(IVW_TURN_ON ON) 
+endif()
+
+if(${CMAKE_VERSION} VERSION_LESS "3.16.0") 
+    # CMake added support for precompiled headers in version 3.16, fallback on cotire
+    set(IVW_PCH_OPTIONS "No PCH" "Cotire")
+    set(IVW_PCH_DOCSTRING "Select mode for pch: \nNo PCH: Do not use precompiled headers\nCotire: Use Cotire to generate precompiled headers")
+    if(${IVW_TURN_ON})
+        set(IVW_DEFAULT_PCH_MODE "Cotire")
+    endif()
 else()
-    option(PRECOMPILED_HEADERS "Create and use precompilied headers" OFF)
+    set(IVW_PCH_OPTIONS "No PCH" "CMake" "Reuse" "Cotire")
+    string(CONCAT IVW_PCH_DOCSTRING "Select mode for pch:\n" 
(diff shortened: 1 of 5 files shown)


cmake/globalutils.cmake
```diff
@@ -862,6 +862,26 @@ function(ivw_get_target_property_recursive retval target property alsoInterfaceT
     set(${retval} ${res} PARENT_SCOPE)
 endfunction()
 
+
+#--------------------------------------------------------------------
+# Get the list of targets the given target depends on. 
+function(ivw_get_target_depending_on retval target alsoInterfaceTargets)
```
petersteneteg, line 868: why not just use ivw_get_target_property_recursive with INTERFACE_LINK_LIBRARIES

src/core/CMakeLists.txt
```diff
@@ -911,3 +911,10 @@ else()
     endif()
 endif()
 #--------------------------------------------------------------------
+
+
+set(PCH_HEADERS ${HEADER_FILES})
+# Should not include the following file in PCH or we will get multiple definitions of NvOptimusEnablement and AmdPowerXpressRequestHighPerformance
```
petersteneteg, line 917: hmm those definitions should be in a cpp only the declarations should be in the header!?! @ResearchDaniel Why is it written like that? should be changed.

cmake/globalconfig.cmake
```diff
@@ -254,13 +254,45 @@ endif()
 
 #--------------------------------------------------------------------
 # Precompile headers
-if(WIN32)
-    option(PRECOMPILED_HEADERS "Create and use precompilied headers" ON)
+set(IVW_DEFAULT_PCH_MODE "No PCH")
+
+set(IVW_TURN_ON OFF) # Default off
```
petersteneteg, line 259: pls use a more distinct name..

petersteneteg (changes requested):
Do we want to keep "Resuse" if it does not work and adds quite a bit of complexity to the setup?

cmake/globalconfig.cmake
ResearchDaniel: Suggestion (matches old one):
IVW_PRECOMPILED_HEADERS 
Could prepend _ON or 
IVW_USE_PRECOMPILED_HEADERS

src/core/CMakeLists.txt
ResearchDaniel: No good reason :)

cmake/globalconfig.cmake
```diff
@@ -254,13 +254,45 @@ endif()
...
 else()
-    option(PRECOMPILED_HEADERS "Create and use precompilied headers" OFF)
+    set(IVW_PCH_OPTIONS "No PCH" "CMake" "Reuse" "Cotire")
+    string(CONCAT IVW_PCH_DOCSTRING "Select mode for pch:\n" 
+                                    "• No PCH: Do not use precompiled \n"
+                                    "• CMake: Use CMake functionality for precompiled headers (requires CMake 3.16 or newer) \n"
+                                    "• Reuse: (Unstable) Use CMake functionality for precompiled headers and reuse PCH artifacts from core or base (requires CMake 3.16 or newer) \n"
+                                    "• Cotire: Use Cotire to generate precompiled headers (the old way)")
+    if(${IVW_TURN_ON})
+        set(IVW_DEFAULT_PCH_MODE "CMake")
```
ResearchDaniel, line 281: We should use PRECOMPILED consistently, e.g., IVW_PRECOMPILED_XXX 
This will place them next to each other in the cmake gui.

petersteneteg: think IVW_PRECOMPILED_HEADERS_MODE is the only setting that exposed in the gui. The rest is just regular variables...

PR #885, opened 2020-03-05
BaseGL: RegTet: Simpliedfied volume bounding box regression test

r-englund:
This commits simplifies the volume bounding box regression test. 
I have removed the surface extraction part since this test is supposed to test the bounding box, not surface extraction. 
As a bonus, this test will not randomly break regression tests due to usage of the threadpool


petersteneteg:
I have fixed the random failures I know of for this workspace. The issues are also not just this workspace but for any workspace that uses a poolprocessor with delayed evaluation. If we have more random failures from that we should fix that not remove the uses. Since that would mean that we can't test quite a few processor.

petersteneteg (changes requested):
I prefer to keep this. I serves as a good system test.

r-englund:
Network before: 
image


Network after simplification: 
image

r-englund:
Surface extraction is tested in another regression test, but fine, keep it this way

petersteneteg:
Yes, I know we have different test for that one separately, but the fact that we are having issues this one is what makes it a good test :) Pls let me know if we have more random failures with this one...
I have seen we sometimes get this error now

`Error: (util::saveAllCanvases:79) Canvas is not ready or not valid, no image saved
    Display Name: Canvas
    Identifier: Canvas`

Not sure if that is related exactly...

martinfalk:
I agree with @r-englund and @petersteneteg. The bounding box test should only test the bounding box, while there should be another test for the pool processor, i.e. the old bbox test.

petersteneteg:
Think have the bounding box in a lot of other workspaces too...

PR #886, opened 2020-03-05
BaseGL: Fixe for tube rendering shaders not updated lighting properties

r-englund:


modules/basegl/src/processors/tuberendering.cpp
```diff
@@ -105,6 +105,9 @@ TubeRendering::TubeRendering()
 }
 
 void TubeRendering::initializeResources() {
+    for (auto& item : adjacencyShaders_.getShaders()) {
```
petersteneteg, line 108: good catch

petersteneteg (approved):
I've noticed that the tube render also does not take the basis&offset and or world transform into account ;)

PR #889, opened 2020-03-06
Base: DatVolumeSequenceReader fix for dat-file containing multiple fi…

ResearchDaniel:
…les. Each path in the file should be supplied relative to the containing dat-file, but the reader considered absolute path..

### DIFF
--- a/modules/base/src/io/datvolumesequencereader.cpp
+++ b/modules/base/src/io/datvolumesequencereader.cpp
@@ -250,7 +250,7 @@ std::shared_ptr<DatVolumeSequenceReader::VolumeSequence> DatVolumeSequenceReader
         for (size_t t = 0; t < state.datFiles.size(); ++t) {
             auto datVolReader = std::make_unique<DatVolumeSequenceReader>();
             datVolReader->enableLogOutput_ = false;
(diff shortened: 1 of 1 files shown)


petersteneteg (changes requested):
This is a breaking change is it not? Would it not be better to check if the given path is relative or not?

ResearchDaniel:
The change to make it absolute was breaking, I'm restoring the original behaviour :)

ResearchDaniel:
Also, using absolute paths would prevent one from moving any file

petersteneteg:
Still it van break code

petersteneteg:
In most other places we handle both

ResearchDaniel:
The raw-files read below assumes the same thing. 
Other places handle absolute paths for the specific file, not for ones specified inside another, e.g. DICOM?

ResearchDaniel:
@petersteneteg fixed, not it should handle both cases

petersteneteg (approved):

ResearchDaniel:
@petersteneteg unrelated regression errors are the ones you fixed?

petersteneteg:
@ResearchDaniel Yes so it seems

PR #892, opened 2020-03-10
Feature/various

martinfalk:
Some minor fixes
* fixed content type for the image save dialog, making it remember the last folder
* minor optimization in the raycasting in connection with rendering on top of the background
* fixes for the color map legend (initial layouting and new background mode "solid color")

image

### DIFF
--- a/modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
+++ b/modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
@@ -65,11 +65,14 @@ public:
     bool isColumnSelected(size_t column) const;
 
     void setSelected(const BrushingAndLinkingInport* src, const std::unordered_set<size_t>& idx);
+    void clearSelected();
 
     void setFiltered(const BrushingAndLinkingInport* src, const std::unordered_set<size_t>& idx);
+    void clearFiltered();
 
     void setSelectedColumn(const BrushingAndLinkingInport* src,
                            const std::unordered_set<size_t>& columnIndices);
+    void clearColumns();
 
     const std::unordered_set<size_t>& getSelectedIndices() const;
     const std::unordered_set<size_t>& getFilteredIndices() const;
--- a/modules/brushingandlinking/src/brushingandlinkingmanager.cpp
+++ b/modules/brushingandlinking/src/brushingandlinkingmanager.cpp
@@ -66,14 +66,26 @@ void BrushingAndLinkingManager::setSelected(const BrushingAndLinkingInport*,
     owner_->invalidate(invalidationLevel_);
 }
 
+void BrushingAndLinkingManager::clearSelected() {
+    selected_.clear();
+    owner_->invalidate(invalidationLevel_);
+}
+
 void BrushingAndLinkingManager::setFiltered(const BrushingAndLinkingInport* src,
                                             const std::unordered_set<size_t>& indices) {
     filtered_.set(src, indices);
 }
 
+void BrushingAndLinkingManager::clearFiltered() { filtered_.clear(); }
+
 void BrushingAndLinkingManager::setSelectedColumn(const BrushingAndLinkingInport*,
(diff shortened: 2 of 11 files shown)


modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
```diff
@@ -65,11 +65,14 @@ class IVW_MODULE_BRUSHINGANDLINKING_API BrushingAndLinkingManager {
     bool isColumnSelected(size_t column) const;
 
     void setSelected(const BrushingAndLinkingInport* src, const std::unordered_set<size_t>& idx);
+    void clearSelected();
```
petersteneteg, line 68: what's the thinking here? why not use a empty set together with setSelected? How will this keep track of the sources?

modules/brushingandlinking/src/brushingandlinkingmanager.cpp
```diff
@@ -66,17 +66,29 @@ void BrushingAndLinkingManager::setSelected(const BrushingAndLinkingInport*,
...
+    selected_.clear();
+    owner_->invalidate(invalidationLevel_);
+}
+
 void BrushingAndLinkingManager::setFiltered(const BrushingAndLinkingInport* src,
                                             const std::unordered_set<size_t>& indices) {
     filtered_.set(src, indices);
 }
 
+void BrushingAndLinkingManager::clearFiltered() { filtered_.clear(); }
```
petersteneteg, line 79: invalidation?

modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
petersteneteg: should this be public? a processor using the manager should not call this right?

petersteneteg (approved):

petersteneteg (approved):

modules/brushingandlinking/src/brushingandlinkingmanager.cpp
martinfalk: that is handled internally by `filtered_`

modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
martinfalk: This was based on the fact that none of the set functions in the manager actually consider the source. This is only of importance for the events, not the manager.
https://github.com/inviwo/inviwo/blob/7ba4f7571fc2d0a3a2957cc2f6ad3ba23e6684e1/modules/brushingandlinking/src/brushingandlinkingmanager.cpp#L63-L75

re "public", that was the idea.

petersteneteg: the filter does consider source...

martinfalk: However, the source is only used as key in the unordered map which is going to be cleared.

petersteneteg: But will all the sources of filtering get updated when cleared like this

martinfalk: I assume so. Maybe @r-englund can shed some light on this.

petersteneteg: The filtering comes from the pcp plot for example, from each axis range. How is the pcp plot supposed to know that is should reset all the axis ranges when this happens?!? I think this is quite broken now

r-englund: What are the use of clear? Who/what is calling it?

PR #893, opened 2020-03-11
PlottingGL: enable/disable color legend

martinfalk:
How often has it happened that you wanted to quickly hide a color legend without reconnecting processors in the network?

### DIFF
--- a/modules/basegl/src/processors/tuberendering.cpp
+++ b/modules/basegl/src/processors/tuberendering.cpp
@@ -144,6 +144,8 @@ void TubeRendering::process() {
         return false;
     };
 
+    utilgl::CullFaceState cullstate(GL_FRONT);
+
     const auto draw = [this, hasAnyLine](const Mesh& mesh, Shader& shader, auto test) {
         if (!hasAnyLine(mesh, test)) return;
 
--- a/modules/basegl/src/processors/tuberendering.cpp
+++ b/modules/basegl/src/processors/tuberendering.cpp
@@ -144,8 +144,6 @@ void TubeRendering::process() {
         return false;
     };
 
-    utilgl::CullFaceState cullstate(GL_FRONT);
-
     const auto draw = [this, hasAnyLine](const Mesh& mesh, Shader& shader, auto test) {
         if (!hasAnyLine(mesh, test)) return;
 
--- a/modules/plottinggl/src/processors/colorscalelegend.cpp
+++ b/modules/plottinggl/src/processors/colorscalelegend.cpp
@@ -221,10 +221,16 @@ std::tuple<ivec2, ivec2, ivec2, ivec2> ColorScaleLegend::getPositions(ivec2 dime
 }
 
 void ColorScaleLegend::process() {
-    utilgl::activateTargetAndClearOrCopySource(outport_, inport_);
     if (!enabled_) {
+        if (inport_.isReady()) {
+            outport_.setData(inport_.getData());
+        } else {
+            utilgl::activateAndClearTarget(outport_);
+        }
+
         return;
     }
+    utilgl::activateTargetAndClearOrCopySource(outport_, inport_);
 
     // update the legend range if a volume is connected to inport
     if (volumeInport_.isChanged() && volumeInport_.hasData()) {
--- a/modules/plottinggl/src/processors/colorscalelegend.cpp
+++ b/modules/plottinggl/src/processors/colorscalelegend.cpp
@@ -49,6 +49,7 @@ ColorScaleLegend::ColorScaleLegend()
     , inport_("inport")
     , outport_("outport")
     , volumeInport_("volumeInport")
+    , enabled_("enabled", "Enabled", true)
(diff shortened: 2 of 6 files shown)


modules/plottinggl/src/processors/colorscalelegend.cpp
```diff
@@ -220,11 +221,9 @@ std::tuple<ivec2, ivec2, ivec2, ivec2> ColorScaleLegend::getPositions(ivec2 dime
 }
 
 void ColorScaleLegend::process() {
-    // draw cached overlay on top of the input image
-    if (inport_.isReady()) {
-        utilgl::activateTargetAndCopySource(outport_, inport_);
-    } else {
-        utilgl::activateAndClearTarget(outport_);
+    utilgl::activateTargetAndClearOrCopySource(outport_, inport_);
```
ResearchDaniel: No need to copy data?
outport_.setData(inport_.getData());

ResearchDaniel (changes requested):
Nice, but no need to copy data?

modules/plottinggl/src/processors/colorscalelegend.cpp
martinfalk: nope due to activating and copying/clearing the canvas

modules/plottinggl/src/processors/colorscalelegend.cpp
```diff
@@ -221,10 +221,16 @@ std::tuple<ivec2, ivec2, ivec2, ivec2> ColorScaleLegend::getPositions(ivec2 dime
 }
 
 void ColorScaleLegend::process() {
-    utilgl::activateTargetAndClearOrCopySource(outport_, inport_);
     if (!enabled_) {
+        if (inport_.isReady()) {
```
ResearchDaniel, line 225: isConnected? 
If it is optional it will be ready even if not connected

martinfalk: Then you need to question all places where we use this test. For example here
https://github.com/inviwo/inviwo/blob/1f5e9dfff6b8a62d6079a47ddc47f1816e79d02a/modules/opengl/src/texture/textureutils.cpp#L114-L131

martinfalk: And, if the port is connected but not ready, what should happen then?

ResearchDaniel: You are right, the processor will be ready but not the optional port itself (looked at the documentation).

ResearchDaniel (approved):

modules/basegl/src/processors/tuberendering.cpp
```diff
@@ -144,6 +144,8 @@ void TubeRendering::process() {
         return false;
     };
 
+    utilgl::CullFaceState cullstate(GL_FRONT);
```
r-englund: I guess it doesn't make a difference since the shader doesn't care, but this will remove the front faces, not the back faces. It itches my eyes a bit, I would have used `GL_BACK` instead. :)

petersteneteg: Why is this needed? will this cull the output from the geometry shader? why would that be needed? should have a comment... why do we generate triangels in the geometry shader that we don't want. would it not be better not to create those to begin with...

r-englund: In the geometry shader, we generate a bounding box with all 6 sides. The frag shader does not consider if the current fragment is a front- or backface. The ray-cylinder intersection test will give the same result for both, hence the z-fighting in #897 . 

I guess we could try to figure out which sides are visible given a view direction and only emit those, but I doubt it will be faster (nor slower) than the face culling.

r-englund: But yes, A comment why this is needed might be good.

petersteneteg: go with the comment, that is not obvious...

martinfalk: The alternative would be to use `if (!gl_FrontFacing) discard;` in the fragment shader. See [khronos gl_FrontFacing](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/gl_FrontFacing.xhtml).

petersteneteg:
Could you also make it take the basis into account...

modules/basegl/src/processors/tuberendering.cpp
ResearchDaniel: This is more expensive though
"Face culling allows non-visible triangles of closed surfaces to be removed before expensive Rasterization and Fragment Shader operations."

petersteneteg: would be even better not to create them to begin with...

martinfalk:
I am not going to touch any more of the tube rendering. And for some reason regression tests are failing.
Rebased the legend visibility into #899

Issue #900, opened 2020-03-17
Tests in VectorfieldVisualizationGL module are outdated


martinfalk:
Many of the test workspaces in the `VectorfieldVisualizationGL` module are outdated and use deprecated processors.

For example `vectorfieldvisualizationgl/tests/regression/pathlines/pathlines.inv`
```
Loading old workspace (D:/projects/inviwo/inviwo-dev/modules/vectorfieldvisualizationgl/tests/regression/pathlines/pathlines.inv) 
BaseModule version: 3. Updating to version: 4.
Deserialized property: PathLines.visibleLayer with class identifier: 
org.inviwo.OptionPropertyEnumChar
from a serialized property with a different class identifier: 
org.inviwo.OptionPropertyEnumInt
This Path Lines Processor is Deprecated, use the new Path Lines processor together with the IntegralLineToMesh processor.
```

or `vectorfieldvisualizationgl/tests/regression/streamtubes/streamtubes.inv`
```
Loading old workspace (D:/projects/inviwo/inviwo-dev/modules/vectorfieldvisualizationgl/tests/regression/streamtubes/streamtubes.inv) 
BaseModule version: 3. Updating to version: 4.
Deserialized property: Canvas.visibleLayer with class identifier: 
org.inviwo.OptionPropertyEnumChar 
from a serialized property with a different class identifier: 
org.inviwo.OptionPropertyEnumInt
This Stream Lines Processor is Deprecated, use the new Stream Lines processor together with the IntegralLineToMesh processor.
```

r-englund:
Many? There are 3, out of 18 ;)

Issue #897, opened 2020-03-16
Tube rendering artifacts


martinfalk:
Tube rendering is slightly broken, some tubes are rendered twice.
Input is a mesh with individual lines (`Mesh::MeshInfo(DrawType::Lines, ConnectivityType::None)`). There are no duplicate index pairs. Nonetheless there appear some z-fighting artifacts. In particular noticable with different colors per vertex.

![image](https://user-images.githubusercontent.com/9251300/76738425-e4bce380-676a-11ea-89ae-af608c81fd01.png)

martinfalk:
Depth is set to the same value for both front and back sides.

https://github.com/inviwo/inviwo/blob/bf3ac9ff1e42fb2ac948e7c40009e835c969d060/modules/basegl/glsl/tuberendering.frag#L130

ResearchDaniel:
This whole thing must separate between front/back faces, not only the depth?

martinfalk:
yes. That also indicates that enabling backface culling should at least provide some quick fix.

PR #901, opened 2020-03-17
Feature/tube fix and vectorvis workspaces

r-englund:
* Tube back-face culling, #897 
 * Vector field regression tests updates, #900 

Peter, you mentioned something about basis not being considered? I have done some testing and can't reproduce that.


petersteneteg (approved):

petersteneteg:
Can see if i can reproduce it later

PR #902, opened 2020-03-18
Fixing reference of $(ConfigurationName) in ext/vld/CMakeLists.txt

emmenlau:
This is an almost trivial change. It replaces `$(ConfigurationName)` in CMakeLists.txt with `${CMAKE_CFG_INTDIR}` as discussed [in the cmake documentation](https://cmake.org/cmake/help/latest/variable/CMAKE_CFG_INTDIR.html). This fixes the broken build for [the ninja build system](https://ninja-build.org/).

**Always**

**What envinroment has the code been compiled and tested on?** 
- MSVC 2017 and 2019 with ninja, Ubuntu Linux 18.04 x86_64 with ninja.

### DIFF
--- a/ext/vld/CMakeLists.txt
+++ b/ext/vld/CMakeLists.txt
@@ -13,19 +13,19 @@ set(dll "vld_x64.dll")
 add_custom_target(vld
     ${CMAKE_COMMAND} -E copy_if_different 
         ${CMAKE_CURRENT_SOURCE_DIR}/vld.ini
-        ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/$(ConfigurationName)/vld.ini
+        ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/${CMAKE_CFG_INTDIR}/vld.ini
 
     COMMAND ${CMAKE_COMMAND} -E copy_if_different 
         ${CMAKE_CURRENT_SOURCE_DIR}/bin/${arch}/${dll}
-        ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/$(ConfigurationName)/${dll} 
+        ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/${CMAKE_CFG_INTDIR}/${dll}
 
     COMMAND ${CMAKE_COMMAND} -E copy_if_different 
         ${CMAKE_CURRENT_SOURCE_DIR}/bin/${arch}/dbghelp.dll
-        ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/$(ConfigurationName)/dbghelp.dll
+        ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/${CMAKE_CFG_INTDIR}/dbghelp.dll
 
     COMMAND ${CMAKE_COMMAND} -E copy_if_different 
         ${CMAKE_CURRENT_SOURCE_DIR}/bin/${arch}/Microsoft.DTfW.DHL.manifest
-        ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/$(ConfigurationName)/Microsoft.DTfW.DHL.manifest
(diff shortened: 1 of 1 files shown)


ResearchDaniel:
Hmm, is CMAKE_CFG_INTDIR  not kind of deprecated?

> Note that CMAKE_CFG_INTDIR is no longer necessary for this purpose but has been left for compatibility with existing projects. Instead add_custom_command() recognizes executable target names in its COMMAND option, so ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_CFG_INTDIR}/mytool can be replaced by just mytool.

emmenlau:
> Hmm, is CMAKE_CFG_INTDIR not kind of deprecated?

That's actually correct, but I've previously failed to modify the logic of the multi-configuration-builds in inviwo in a related case, most likely because I run an uncommon setup with only the build tools. This time I wanted to make the least invasive change possible, to preserve the existing logic (just by universally supported means).

So, yes, if you could fire up Visual Studio and test a slightly more modern solution, that would be great!

emmenlau:
...but, just in case nobody has a better suggestion in the close future, then in my humble opinion this PR still improves the code (adds support for ninja) and could be merged as is.

petersteneteg:
Does $<CONFIG> not work here? Think that would be the "proper" solution nowadays...
Also this is only relevant for VLD on windows right?

emmenlau:
> Does $ not work here? Think that would be the "proper" solution nowadays...

Sorry, you have to explain, what is `$`?

> Also this is only relevant for VLD on windows right?

I'm not sure if these instructions are used in other platforms. On Linux I did not observe any problems, so I assume this is not used there(?)

r-englund:
VLD is used only on windows. Since it is using non-standard c++ features only available in the "visual studio c++ compiler" it can't even work on non-visual studio targets. 
When it was added to inviwo visual studio did not have Ninja support and hence it did not need to be considered. 
Compiling this commit on my machine ATM.

ResearchDaniel:
Not sure, but would it be resolved by simply replacing
`${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/$(ConfigurationName) ` -> `${RUNTIME_OUTPUT_DIRECTORY}`

> Multi-configuration generators (VS, Xcode) append a per-configuration subdirectory to the specified directory

https://cmake.org/cmake/help/latest/prop_tgt/RUNTIME_OUTPUT_DIRECTORY.html#prop_tgt:RUNTIME_OUTPUT_DIRECTORY

petersteneteg:
hmm should have looked att the preview... meant 
`$<CONFIG>`

petersteneteg:
RUNTIME_OUTPUT_DIRECTORY will not work

r-englund:
This PR works on my machine, MSVC 2019

ResearchDaniel:
So, it seems like `${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/$<CONFIG>`  should be used

> Note that this variable only has limited support on Ninja Multi-Config. It is recommended that you use the `$<CONFIG>` generator expression instead.

https://cmake.org/cmake/help/latest/variable/CMAKE_CFG_INTDIR.html

emmenlau:
I can only speak for myself, but I'm under the impression that ninja is often used in CI or similar (command line) setups where the build targets a single configuration. So while this change is clearly not perfect, I think its a good step forward.

Off-topic: You may want to consider ninja for your CI too? Its very fast, but more importantly provides a uniform build experience on all major platforms.

petersteneteg:
We do use ninja for our CI and have done so for a long time without issues. But I never think we, at least not me, have use it on windows

Think the whole point if cmake is to give a consistent build regardless of the build system in question...

emmenlau:
> Think the whole point if cmake is to give a consistent build regardless of the build system in question...

And its really good at that, at least for the configure step! Just after configure the mess starts again, and even nowadays I can not build 100% of projects with ninja. A few cases work only with nmake or jom on MSVC, or depending on your taste with msbuild and/or devenv.

Maybe the superbuilds you're using are a better way forward, we build dependencies upfront with their native build system...

ResearchDaniel (approved):
Since it seem to be working I think it is fine. Would could do testing with `$<Config>` later

PR #904, opened 2020-03-23
Core: DataOutport: Added method to move data into the port

r-englund:
Added method to data outport that allows data to be moved into the port (for moveable data). 
Lets you write your process function without the need to create shared_ptr, for example 
```c++ 
void SomePorcessor::process() {
    std::vector<vec3> points;
    /// code to fill the points-vector with data
    myPort_.moveInData(std::move(points));
}
```

* Is the name good? I named it `moveInData` since you move the data into the port. 
* I was also considering making the method taking the data by value (instead of &&) to let the compiler select to use the move constructor, but that would allow for calling the method without using std::move and would then use the copy ctor instead of the move ctor.

### DIFF
--- a/include/inviwo/core/ports/dataoutport.h
+++ b/include/inviwo/core/ports/dataoutport.h
@@ -81,7 +81,7 @@ public:
      * }
      * ```
      */
-    template <typename U = T, typename = std::enable_if_t<std::is_move_constructible_v<T>>>
+    template <typename U = T, typename = std::enable_if_t<std::is_move_constructible_v<U>>>
     void setData(T&& data);
 
     virtual bool hasData() const override;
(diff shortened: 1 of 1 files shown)


martinfalk:
why not simply `setData(T&& data)` ?

r-englund:
> 
> 
> why not simply `setData(T&& data)` ?

I considered that and are not against it, but I wanted to make it clear that you are moving it.

petersteneteg (approved):
Not 100% if it is needed? Maybe it will be more confusing for the user to have overload for both shared_ptr<T> and T&& 
I would use setData(T&&)

r-englund:
> 
> 
> Not 100% if it is needed? 

Depends on how we define needed. One could change the example code above to `myPort_.setData(std::make_shared<std::vector<vec3>>(std::move(points)));`.  And the new method is not needed anymore. 

It is more for convenience, the code in the process methods can after this PR be done a bit cleaner without having the need to create a shared_ptr. This is especially nice when working with `std::vectors` for example, no need for dereferencing, etc.

PR #905, opened 2020-03-26
OpenCL: Fix LayerCL::setDimension bug

ResearchDaniel:
Rather obvious bug fix...

### DIFF
--- a/modules/opencl/src/image/layercl.cpp
+++ b/modules/opencl/src/image/layercl.cpp
@@ -132,11 +132,24 @@ void LayerCL::setDimensions(size2_t dimensions) {
     if (dimensions == dimensions_) {
         return;
     }
+
+    try {
+        auto resizedLayer2D =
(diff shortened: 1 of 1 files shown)


r-englund (approved):

modules/opencl/src/image/layercl.cpp
```diff
@@ -132,7 +132,7 @@ void LayerCL::setDimensions(size2_t dimensions) {
     if (dimensions == dimensions_) {
         return;
     }
-
+    dimensions_ = dimensions;
     cl::Image2D* resizedLayer2D = new cl::Image2D(OpenCL::getPtr()->getContext(), CL_MEM_READ_WRITE,
```
r-englund: This should be a `make_unique`, not `new cl::Image`  :wink:.

ResearchDaniel: fixed it along with an annoying error

PR #906, opened 2020-03-26
Core: ThreadPool: Added name to threadpool threads (windows only)

r-englund:
With this commit, the worker threads will be listed as "Inviwo Worker Thread" in Visual Studio when debugging. 

bild

### DIFF
--- a/src/core/util/threadpool.cpp
+++ b/src/core/util/threadpool.cpp
@@ -31,6 +31,10 @@
 #include <inviwo/core/util/raiiutils.h>
 #include <inviwo/core/util/stdextensions.h>
 
+#ifdef WIN32
+#include <windows.h>
+#endif
+
 namespace inviwo {
 
 // the constructor just launches some amount of workers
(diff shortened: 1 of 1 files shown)


src/core/util/threadpool.cpp
```diff
@@ -31,6 +31,10 @@
 #include <inviwo/core/util/raiiutils.h>
 #include <inviwo/core/util/stdextensions.h>
 
+#ifdef WIN32
+#include <windows.h>
```
petersteneteg, line 35: Should we have those #define nominmax and lean and what ever it is here? or do we set that globally?

petersteneteg (approved):
Cool!
Should we have this for the main thread to? or how is that named now? I recall it being name something with CEF?

src/core/util/threadpool.cpp
r-englund: They are defined at https://github.com/inviwo/inviwo/blob/8503e60c6a176ae50627ce979bcf5fc688bf86af/cmake/compileoptions.cmake#L121-L122

r-englund:
The main thread is named Main Thread, you can see that in the screenshot ;)

Issue #251, opened 2018-09-20
Closing Canvas / ColorDialog breaks window - Qt 5.11


martinfalk:
When closing the color dialog of a color property via the "x" button in the window title bar, the dialog cannot be reopened and only a small empty widget pops up.

![image](https://user-images.githubusercontent.com/9251300/45840619-b5fdb280-bd17-11e8-94ce-52614e954110.png)

Closing the dialog via <Esc> works fine.

martinfalk:
This affects Qt 5.11.1

Also tested with Qt 5.9.1 where the dialog behaves as expected.

martinfalk:
A similar effect occurs in the canvas widget when it is closed using the close button.
After re-showing the canvas widget, the contents remain black and the rendered output is not visible.

Only after hiding the canvas and reshowing it via the canvas processor, the contents are visible.

petersteneteg:
Has anyone tried this in newer qt?

r-englund:
Can reproduce qt 5.13(.0)

r-englund:
Also in qt 5.13.1_preview

r-englund:
Can reproduce with Color Property (tested with background), when double clicken a TF point in the TF editor. 
This also happens with the Ordinal Propety Settings widget and for the min max property settings widget

r-englund:
Seems to be this or related to this qt bug https://bugreports.qt.io/browse/QTBUG-76509?jql=text%20~%20%22QDialog%22

martinfalk:
Running into similar problem with ProcessorWidgetQt, which remains empty after a close.

see also https://bugreports.qt.io/browse/QTBUG-71519

emmwel:
In Qt 5.14, the color dialog reappears incomplete (instead of being empty)

![image](https://user-images.githubusercontent.com/21198447/75983650-2e930780-5ee9-11ea-9e50-f4d01be85722.png)

petersteneteg:
should be fixed now https://github.com/inviwo/inviwo/pull/939/commits/f205caf0d39e6fb83fd15a9b8f24d261d9b928c0

PR #910, opened 2020-03-31
Core: VS 16.5.0 Warning fixes

petersteneteg:
Also a workaround for #251

### DIFF
--- a/modules/qtwidgets/include/modules/qtwidgets/inviwoqtutils.h
+++ b/modules/qtwidgets/include/modules/qtwidgets/inviwoqtutils.h
@@ -214,7 +214,7 @@ IVW_MODULE_QTWIDGETS_API void addImageActions(QMenu& menu, const Image& image,
  * @return title where "[*]" is replaced with '*' if the widget is modified, or '' otherwise
  * \see QWidget::setWindowTitle
  */
-IVW_MODULE_QTWIDGETS_API QString windowTitleHelper(const QString& title, const QWidget* widget);
+ QString windowTitleHelper(const QString& title, const QWidget* widget);
 
 // In a non high dpi system an 'M' measures 11 px. Hence all our old pixels sizes, can be converted
 // to Em sizes by dividing by 11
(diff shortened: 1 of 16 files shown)


src/core/util/stringconversion.cpp
```diff
@@ -43,27 +44,57 @@
...
-}
 #else
-std::wstring toWstring(const std::string& str) {
     auto state = std::mbstate_t();
-    const char* s = str.c_str();
-    size_t len = std::mbsrtowcs(nullptr, &s, 0, &state) + 1;
+    size_t len = std::mbsrtowcs(nullptr, &str.data(), 0, &state) + 1;
     std::wstring result(len, 0);
-    std::mbsrtowcs(&result[0], &s, result.size(), &state);
+    std::mbsrtowcs(result.data(), &str.data(), result.size(), &state);
```
r-englund: `mbsrtowcs` takes `const char **` so you need to pass in `&&str.data()` here and the line above. This is also true for `wcsrtombs`  
https://en.cppreference.com/w/cpp/string/multibyte/mbsrtowcs
https://en.cppreference.com/w/cpp/string/multibyte/wcsrtombs

r-englund (approved):

src/core/util/stringconversion.cpp
petersteneteg: that's not the issue, problem is str.data() returns a const char* r-value which I can not take the adress of.. need to assign it to an l-value first...

modules/qtwidgets/include/modules/qtwidgets/inviwoqtutils.h
```diff
@@ -225,6 +225,19 @@ IVW_MODULE_QTWIDGETS_API QSize emToPx(const QWidget* w, QSizeF);
 IVW_MODULE_QTWIDGETS_API int emToPx(const QWidget* w, double em);
 IVW_MODULE_QTWIDGETS_API int emToPx(const QFontMetrics& m, double em);
 
+
+
+/**
+ * Workaround for closing Widget using the "X" button in the titlebar
+ * This filter intercepts the CloseEvent and ignores it, and then hides the widget
+ */
+struct IVW_MODULE_QTWIDGETS_API WidgetCloseEventFilter : QObject {
```
ResearchDaniel, line 232: Would call this HideOnCloseEventFilter or HideInsteadOfCloseEventFilter or similar

petersteneteg: At least I added a comment ;)

PR #915, opened 2020-04-06
Feature/meshrendering2

AnkeAnke:
**Rasterizers and Renderers**
We now split the fragment list rendering from a monolithic fancy renderer into two stages:
- Rasterizer processors for different kinds of primitives emit rasterizations, functors that generate fragments
- A renderer sets up the fragment lists, calls the rasterizations, and then evaluates the fragments in each pixel

This allows us to mix settings and primitives by using several rasterizers.

**To be done**
- More primitive renderers (lines, points, spheres, raycasts)
  - Example workspaces
  - Testing
- Mixed opacity settings in rasterizers
  - Background image depth is taken into account now
  - Opaque rasterizers are be ignored for fragment list evaluation, need another in-between render target
- Rasterizations should not hold properties, but only the settings of the generating processor

**What environment has the code been compiled and tested on?**
- Pop! OS (Ubuntu) 19.10
- GCC
- Qt 5.14.1 / CMake 3.13.4 / Python 3.7.5
CowsPlosion


modules/meshrenderinggl/CMakeLists.txt
```diff
@@ -1,67 +1,71 @@
...
+    include/modules/meshrenderinggl/rendering/fragmentlistrenderer.h
+    include/modules/meshrenderinggl/meshrenderingglmodule.h
+    include/modules/meshrenderinggl/meshrenderingglmoduledefine.h
+)
+ivw_group("Header Files" ${HEADER_FILES})
+
+#--------------------------------------------------------------------
+# Add source files
+set(SOURCE_FILES
+    src/datastructures/halfedges.cpp
```
petersteneteg, line 24: we want to have 1 .cpp for every .h even of all it does is include it's cpp. Main reason is to make sure each .h can compile in isolation.

modules/meshrenderinggl/glsl/oit/display.frag
```diff
@@ -80,24 +87,29 @@ void main() {
...
 
         // front-to-back shading
         vec4 color = vec4(0);
-        for (int i = 0; i < numFrag; ++i) {
-            vec4 c = uncompressPixelData(fragmentList[i]).color;
+        vec4 nextFragment = selectionSortNext(pixelIdx, 0.0);
+        abufferPixel unpackedFragment = uncompressPixelData(nextFragment);
+        gl_FragDepth = min(backgroundDepth, unpackedFragment.depth);
+
+        while (unpackedFragment.depth >= 0 && unpackedFragment.depth <= backgroundDepth) {
```
petersteneteg, line 104: do while?

modules/meshrenderinggl/glsl/oit/display.frag
```diff
@@ -80,24 +87,29 @@ void main() {
...
+            nextFragment = selectionSortNext(pixelIdx, unpackedFragment.depth);
+            unpackedFragment = uncompressPixelData(nextFragment);
         }
-        float depth = 1.0;
-        if (numFrag > 0) {
-            depth = uncompressPixelData(fragmentList[0]).depth;
-        }
-        gl_FragDepth = depth;
+
         FragData0 = color;
```
petersteneteg, line 113: where is bgColor added?

modules/meshrenderinggl/include/modules/meshrenderinggl/datastructures/rasterization.h
```diff
@@ -0,0 +1,67 @@
...
+namespace inviwo {
+
+/**
+ * \brief A functor class for rendering geometry into a fragment list
+ * Will be applied by a renderer containing an A-buffer.
+ */
+
+class IVW_MODULE_MESHRENDERINGGL_API Rasterization {
+public:
+    virtual ~Rasterization() {}
```
petersteneteg: = default

modules/meshrenderinggl/include/modules/meshrenderinggl/datastructures/rasterization.h
```diff
@@ -0,0 +1,67 @@
...
+    virtual void rasterize(std::function<void(Shader&)> setUniforms) const = 0;
+    virtual void setImageSize(const ivec2& size) const = 0;
+    virtual bool usesFragmentLists() const = 0;
```
petersteneteg, line 63: Should add some comment for these functions

petersteneteg: could have a virtual function for the docs to...

modules/meshrenderinggl/include/modules/meshrenderinggl/ports/rasterizationport.h
```diff
@@ -0,0 +1,44 @@
...
+#pragma once
+
+#include <modules/meshrenderinggl/meshrenderingglmoduledefine.h>
+#include <modules/meshrenderinggl/datastructures/rasterization.h>
+#include <inviwo/core/ports/datainport.h>
+#include <inviwo/core/ports/dataoutport.h>
+
+namespace inviwo {
+
+// using RasterizationInport = DataInport<Rasterization>;
```
petersteneteg: remove

modules/meshrenderinggl/include/modules/meshrenderinggl/processors/meshrasterizer.h
```diff
@@ -1,306 +1,336 @@
...
+
+        void copyFrontToBack();
+
+        void setUniforms(Shader& shader, std::string_view prefix) const;
+
+        bool lastOpaque_;
+    };
+
+    std::array<FaceSettings, 2> faceSettings_;
+
```
petersteneteg, line 280: remove empty lines

modules/meshrenderinggl/include/modules/meshrenderinggl/processors/meshrasterizer.h
```diff
@@ -1,306 +1,336 @@
...
+    /**
+     * \brief Functor object that will render into a fragment list.
+     */
+    class MeshRasterization : public Rasterization {
+    public:
+        const std::vector<std::shared_ptr<const Mesh>> meshes_;
+        std::vector<std::shared_ptr<const Mesh>> enhancedMeshes_;
+
+        const Camera* camera_;
```
petersteneteg: unique ptr

modules/meshrenderinggl/include/modules/meshrenderinggl/processors/meshrasterizer.h
```diff
@@ -1,306 +1,336 @@
...
+         */
+    public:
+        MeshRasterization(const MeshRasterizer& rasterizerProcessor);
+        virtual ~MeshRasterization() override { delete camera_; }
+        void update(const MeshRasterizer& rasterizerProcessor) const;
+        virtual void rasterize(std::function<void(Shader&)> setUniforms) const override;
+        virtual void setImageSize(const ivec2& size) const override { imageSize_ = size; }
+        virtual bool usesFragmentLists() const override {
+            return !forceOpaque_ && FragmentListRenderer::supportsFragmentLists();
+        }
```
petersteneteg: we generally put this at the top

modules/meshrenderinggl/include/modules/meshrenderinggl/processors/rasterizationrenderer.h
```diff
@@ -0,0 +1,171 @@
...
+        IntProperty smoothingSteps_;
+        FloatProperty edgeSmoothing_;
+        FloatProperty haloSmoothing_;
+
+        FragmentListRenderer::IllustrationSettings getSettings() const;
+    };
+    IllustrationSettings illustrationSettings_;
+
+    FragmentListRenderer flr_;
+    typename Dispatcher<void()>::Handle flrReload_;
```
petersteneteg, line 165: use using

modules/meshrenderinggl/include/modules/meshrenderinggl/processors/meshrasterizer.h
```diff
@@ -1,306 +1,336 @@
...
+    /**
+     * \brief This flag is set to true if adjacency information is available in the shader.
+     */
+    bool meshHasAdjacency_;
+    bool supportsFragmentLists_;
+
+    /**
+     * \brief Functor object that will render into a fragment list.
+     */
+    class MeshRasterization : public Rasterization {
```
petersteneteg: I would make this freestanding instead of being a inside meshrasterizer

modules/meshrenderinggl/src/processors/meshrasterizer.cpp
```diff
@@ -1,549 +1,523 @@
...
+    for (auto&& [key, val] : defines) {
+        for (auto&& so : shader_->getShaderObjects()) {
+            so.setShaderDefine(key, val);
+        }
+    }
+
+    shader_->build();
+}
+
+void MeshRasterizer::MeshRasterization::rasterize(
```
petersteneteg: why not pass in the image size here?

modules/meshrenderinggl/src/processors/meshrasterizer.cpp
```diff
@@ -1,549 +1,523 @@
...
+    vec3 lightPos;
+    switch (lightSpace_) {
+        case SimpleLightingProperty::Space::VIEW:
+            lightPos = vec3(camera_->getInverseViewMatrix() * vec4(lightPosition_, 1.0f));
+            break;
+        case SimpleLightingProperty::Space::WORLD:
+        default:
+            lightPos = lightPosition_;
+    }
+    shader_->setUniform("lighting.position", lightPos);
```
petersteneteg: could we not just have the MeshRastersize set all these uniform directly on the shader? why pass them along and then set them, when we could do it directly? we only need to set them when they change or we recompile right?

petersteneteg (changes requested):
Looks good I think, some minor comment.

What did you do to make all the diffs so huge, github seems to think you changes all lines more or less...

modules/meshrenderinggl/include/modules/meshrenderinggl/datastructures/rasterization.h
```diff
@@ -45,10 +45,32 @@ namespace inviwo {
...
+     * \brief Query whether fragments will be emitted.
+     * @return True for order-independent rendering, false for opaque rasterization.
+     */
     virtual bool usesFragmentLists() const = 0;
+
+    /**
+     * \brief Get data description for the network interface.
+     * @return Specific information about this rasterization type/instance.
+     */
+    virtual std::string getDescription() const { return "Rasteriation functor."; }
```
petersteneteg: I would have this return the Document

modules/meshrenderinggl/src/processors/meshrasterizer.cpp
```diff
@@ -523,7 +523,7 @@ void MeshRasterizer::MeshRasterization::update(const MeshRasterizer& rasterizerP
 std::string MeshRasterizer::MeshRasterization::getDescription() const {
     std::stringstream ss;
     ss << "Mesh rasterization functor with " << enhancedMeshes_.size();
-    ss << (enhancedMeshes_.size() == 1) ? "mesh." : " meshes.";
+    ss << ((enhancedMeshes_.size() == 1) ? " mesh." : " meshes.");
```
petersteneteg: fmt::format is quite nice here...

modules/meshrenderinggl/glsl/oit/display.frag
AnkeAnke: We need to fetch one more fragment than we add in to color. So either we do one extra fetch before the loop as now or check the condition inside the loop again. I think this solution is the cleaner one. Also, the first check is with last depth = 0 here explicitly, that would become quite obscure otherways.

modules/meshrenderinggl/glsl/oit/display.frag
AnkeAnke: Blending is set to alpha, so the fragment ops take care of that for us. Since we got the color texture bound as a side effect of binding depth, we could of course blend here ourselves, but I don't see an advantage of that, it's just more code.

modules/meshrenderinggl/src/processors/meshrasterizer.cpp
AnkeAnke: That's a very good point and got rid of a lot of code. Thanks!

modules/meshrenderinggl/include/modules/meshrenderinggl/processors/rasterizationrenderer.h
AnkeAnke: I fell for that btw: It's typename, not typedef...

petersteneteg: ahh... can't read...

petersteneteg (dismissed):

modules/meshrenderinggl/glsl/oit/display.frag
```diff
@@ -107,7 +119,7 @@ void main() {
         // If no fragment, write nothing
         discard;
 #else
-        FragData0 = vec4(0.0f);
+        FragData0 = vec4(1.0f);
```
ResearchDaniel: Why white color? Has this been for debugging? Otherwise I think it should be stated why

modules/meshrenderinggl/include/modules/meshrenderinggl/processors/meshrasterizer.h
```diff
@@ -1,306 +1,315 @@
...
+        BoolProperty enableUniform_;
+        FloatProperty uniformScaling_;
+        // IRIS
+        BoolProperty enableAngleBased_;
+        FloatProperty angleBasedExponent_;
+        BoolProperty enableNormalVariation_;
+        FloatProperty normalVariationExponent_;
+        // Smoke surfaces
+        BoolProperty enableDensity_;
+        FloatProperty baseDensity_;  // k in the paper
```
ResearchDaniel, line 196: Which paper? 
Can state that in the AlphaSettings doc?

modules/meshrenderinggl/include/modules/meshrenderinggl/processors/meshrasterizer.h
```diff
@@ -1,306 +1,315 @@
...
+    enum class ColorSource : int { VertexColor, TransferFunction, ExternalColor };
+    enum class ShadingMode : int {
+        Off,  // no light, no reflection, just diffuse
+        Phong,
+        Pbr
+    };
+    enum class HatchingMode : char { U, V, UV };
+    enum class HatchingBlendingMode : char { Multiplicative, Additive };
+    /**
+     * \brief Hatching settings. These are exactly the parameters from the IRIS-paper
```
ResearchDaniel, line 224: IRIS-paper? Quick google IRIS + hatching did not reveal it...

modules/meshrenderinggl/src/processors/rasterizationrenderer.cpp
```diff
@@ -0,0 +1,180 @@
...
+    callback();
+}
+}  // namespace
+
+// The Class Identifier has to be globally unique. Use a reverse DNS naming scheme
+const ProcessorInfo RasterizationRenderer::processorInfo_{
+    "org.inviwo.RasterizationRenderer",  // Class identifier
+    "Rasterization Renderer",            // Display name
+    "Mesh Rendering",                    // Category
+    CodeState::Experimental,             // Code state
```
ResearchDaniel: Still experimental?

modules/meshrenderinggl/src/processors/rasterizationrenderer.cpp
```diff
@@ -0,0 +1,180 @@
...
+
+    illustrationSettings_.enabled_.readonlyDependsOn(
+        forceOpaque_, [this](const auto& prop) { return !supportesIllustration_ && prop.get(); });
+
+    flrReload_ = flr_.onReload([this]() { invalidate(InvalidationLevel::InvalidResources); });
+}
+
+RasterizationRenderer::IllustrationSettings::IllustrationSettings()
+    : enabled_("illustration", "Use Illustration Effects", false)
+    , edgeColor_("edgeColor", "Edge Color", vec3(0.f, 0.f, 0.f))
```
ResearchDaniel: Could use the new constraints now?
nice to have, not need to...

modules/meshrenderinggl/src/rendering/fragmentlistrenderer.cpp
```diff
@@ -1,562 +1,601 @@
...
+    // get query result
+    GLuint numFrags = 0;
+    glEndQuery(GL_SAMPLES_PASSED);
+    glGetQueryObjectuiv(totalFragmentQuery_, GL_QUERY_RESULT, &numFrags);
+    LGL_ERROR;
+
+    // check if enough space was available
+    if (numFrags > fragmentSize_) {
+        // we have to resize the fragment storage buffer
+        LogInfo("fragment lists resolved, pixels drawn: "
```
ResearchDaniel, line 149: Remove? Just for debugging?

modules/meshrenderinggl/src/rendering/fragmentlistrenderer.cpp
```diff
@@ -1,562 +1,601 @@
...
+    bool hasBackground = bg && bg->hasData();
+
+    // Build shader depending on inport state.
+    if (hasBackground != builtWithBackground_) buildShaders();
+
+    if (!useIllustration) {
+        // render fragment list
+        display_.activate();
+        setUniforms(display_, textureUnits_[0]);
+        if (builtWithBackground_) {
```
ResearchDaniel, line 166: it is just rebuilt above if background exists. Is it possible that the building does not include background anyway? In that case it will be rebuilt every frame?

modules/meshrenderinggl/src/rendering/fragmentlistrenderer.cpp
```diff
@@ -1,562 +1,601 @@
...
+            utilgl::bindAndSetUniforms(display_, textureUnits_, *bg->getData(), "bg",
+                                       ImageType::ColorDepth);
+            display_.setUniform("reciprocalDimensions", vec2(1) / vec2(screenSize_));
```
ResearchDaniel, line 170: Is screenSize_ always the same as the background size?
reciprocalDimensions could possibly be renamed to bgReciprocalDimensions to make its intention more explicit

modules/meshrenderinggl/src/rendering/fragmentlistrenderer.cpp
```diff
@@ -1,562 +1,601 @@
...
+        surfaceInfo.unbind();
+
+        // smoothing: beta + gamma
+        for (int i = 0; i < 2; ++i) {
+            smoothing[i].setSizeInBytes(bufferSize);
+            smoothing[i].unbind();
+        }
+        // reuse fragment lists as neighborhood storage
+
+        LogInfo("Illustration Buffers: additional pixel storage for "
```
ResearchDaniel, line 348: Remove

modules/meshrenderinggl/src/rendering/fragmentlistrenderer.cpp
```diff
@@ -1,562 +1,601 @@
...
+}
+
+void FragmentListRenderer::fillIllustration(TextureUnit& abuffUnit, TextureUnit& idxUnit,
+                                            TextureUnit& countUnit) {
+    // reset counter
+    LGL_ERROR;
+    GLuint v[1] = {0};
+    atomicCounter_.upload(v, sizeof(GLuint));
+    atomicCounter_.unbind();
+    LGL_ERROR;
```
ResearchDaniel, line 361: Is there a graceful way of dealing with these possible errors?

modules/meshrenderinggl/src/rendering/fragmentlistrenderer.cpp
```diff
@@ -1,562 +1,601 @@
...
+
+    // execute sort+fill shader
+    illustration_.fill.activate();
+    setUniforms(illustration_.fill, abuffUnit);
+    illustration_.setUniforms(illustration_.fill, idxUnit, countUnit);
+    if (builtWithBackground_) {
+        utilgl::bindAndSetUniforms(illustration_.fill, textureUnits_,
+                                   *(backgroundPort_.lock()->getData()), "bg",
+                                   ImageType::ColorDepth);
+        illustration_.fill.setUniform("reciprocalDimensions", vec2(1) / vec2(screenSize_));
```
ResearchDaniel, line 370: Same comment as above about background

modules/meshrenderinggl/src/rendering/fragmentlistrenderer.cpp
```diff
@@ -1,562 +1,601 @@
...
+
+    utilgl::GlBoolState depthTest(GL_DEPTH_TEST, false);
+    utilgl::DepthMaskState depthMask(GL_FALSE);
+    utilgl::CullFaceState culling(GL_NONE);
+    utilgl::singleDrawImagePlaneRect();
+    glMemoryBarrier(GL_SHADER_STORAGE_BARRIER_BIT);
+
+    neighbors.deactivate();
+
+    // perform the bluring
```
ResearchDaniel, line 404: blurring

modules/meshrenderinggl/src/processors/meshrasterizer.cpp
```diff
@@ -1,549 +1,500 @@
...
+    for (const auto& [key, val] : uniforms) {
+        std::visit([&, akey = key](
+                       auto& aval) { shader.setUniform(fmt::format("{}{}", prefix, akey), aval); },
+                   val);
+    }
+}
+
+MeshRasterizer::EdgeSettings::EdgeSettings()
+    : CompositeProperty("edges", "Edges")
+    , edgeThickness_("thickness", "Thickness", 2.f, 0.1f, 10.f, 0.1f)
```
ResearchDaniel, line 187: In pixels?

modules/meshrenderinggl/src/processors/meshrasterizer.cpp
```diff
@@ -1,549 +1,500 @@
...
+                    {"tf", "Transfer Function", ColorSource::TransferFunction},
+                    {"external", "Constant Color", ColorSource::ExternalColor}},
+                   2, InvalidationLevel::InvalidResources)
+    , separateUniformAlpha_("separateUniformAlpha", "Separate Uniform Alpha")
+    , uniformAlpha_("uniformAlpha", "Uniform Alpha", 0.5f, 0.f, 1.f, 0.01f)
+    , shadingMode_("shadingMode", "Shading Mode",
+                   {
+                       {"off", "Off", ShadingMode::Off},
+                       {"phong", "Phong", ShadingMode::Phong},
+                       {"pbr", "PBR", ShadingMode::Pbr},
```
ResearchDaniel: Hmm, there are multiple PBR-shading methods. We might want to be explicit if more alternatives are to be added?

modules/meshrenderinggl/src/processors/meshrasterizer.cpp
```diff
@@ -1,549 +1,500 @@
...
+    }
+
+    shader_->build();
+}
+
+void MeshRasterization::rasterize(const ivec2& imageSize,
+                                  std::function<void(Shader&)> setUniformsRenderer) const {
+
+    shader_->activate();
+    if (!shader_->isReady()) return;
```
ResearchDaniel: This should be checked before it is activated?

modules/meshrenderinggl/include/modules/meshrenderinggl/processors/meshrasterizer.h
```diff
@@ -1,306 +1,315 @@
...
+     * \brief Copy all settings and the shader to hand to a renderer.
+     */
+    MeshRasterization(const MeshRasterizer& rasterizerProcessor);
+    void update(const MeshRasterizer& rasterizerProcessor) const;
+    virtual void rasterize(const ivec2& imageSize,
+                           std::function<void(Shader&)> setUniforms) const override;
+    virtual bool usesFragmentLists() const override {
+        return !forceOpaque_ && FragmentListRenderer::supportsFragmentLists();
+    }
+    virtual Document getDescription() const override;
```
ResearchDaniel: Is is possible to move this outside of the class? Printing info is not really its reponsibility

modules/meshrenderinggl/include/modules/meshrenderinggl/datastructures/rasterization.h
```diff
@@ -0,0 +1,79 @@
...
+     * \brief Query whether fragments will be emitted.
+     * @return True for order-independent rendering, false for opaque rasterization.
+     */
+    virtual bool usesFragmentLists() const = 0;
+
+    /**
+     * \brief Get data description for the network interface.
+     * @return Specific information about this rasterization type/instance.
+     */
+    virtual Document getDescription() const;
```
ResearchDaniel: Is is possible to move this outside of the class? Printing info is not really its reponsibility

ResearchDaniel (dismissed):
Nice! Looks very structured for being so complicated. I have some points that might cause bugs though.

modules/meshrenderinggl/src/processors/rasterizationrenderer.cpp
petersteneteg: use 
`edgeColor_("edgeColor", "Edge Color", ordinal::color(vec3(0.0f)))`

ResearchDaniel: Nice, if time admits it could be worthwhile to go through each property and see if constraints should be added

modules/meshrenderinggl/include/modules/meshrenderinggl/processors/meshrasterizer.h
petersteneteg: I is, think of it as data. We do the same thing for Image and Volume. since we put it in a port we want to be able to inspect it. We could even add port inspectors for the resterizsers so we could look at the renderings...

modules/meshrenderinggl/include/modules/meshrenderinggl/datastructures/rasterization.h
petersteneteg: But it is, se above. how else would you describe what's being rendered?

petersteneteg: but we should name it getInfo to be consistent with the other data structures https://github.com/inviwo/inviwo/blob/6068897cfeb1a10a7c5720595f7c6d50b5c6cd02/include/inviwo/core/datastructures/volume/volume.h#L79

petersteneteg: the info is used here
https://github.com/inviwo/inviwo/blob/6068897cfeb1a10a7c5720595f7c6d50b5c6cd02/include/inviwo/core/ports/datainport.h#L267

the default DataTraits implementation will use
https://github.com/inviwo/inviwo/blob/6068897cfeb1a10a7c5720595f7c6d50b5c6cd02/include/inviwo/core/datastructures/datatraits.h#L103

to get the `getInfo` with help from
https://github.com/inviwo/inviwo/blob/6068897cfeb1a10a7c5720595f7c6d50b5c6cd02/include/inviwo/core/util/introspection.h#L199

So we could add a specialisation to the DataTraits class to do this, but then that specialisation would still need to be able to get the information out of the rasteriser some how...

ResearchDaniel: Yes, I think that it is the responsibility of the DataTrait to provide information. The ones in Volume and other data structures are there due to legacy (we did not have DataTraits at the time)...
Having the getInfo in the class breaks the single responsibility principle

petersteneteg: but then you would have to have a way of exposing what's "inside" rasersize any way. And since what is in it depends on which derived class it is, it has to be a virtual function in the base class. Now what should that function return? I can be lots of different thing in a rasterizer, meshes, textures, shaders, etc? hence it probably has to return some kind of generic string description. hence we get what we have now... The reason for having DataTraits it so we can allow for decorating data with info also when we can't add the getInfo method into the data class, like for a std::vector for example.

ResearchDaniel: One trait per derived class would do it. Anyway, I guess convenience wins here

petersteneteg: You can only specialise for the base class not the derived ones, since the port only know knows about the base. Hence that will not work.

ResearchDaniel: True, marked it as resolved

modules/meshrenderinggl/include/modules/meshrenderinggl/processors/meshrasterizer.h
AnkeAnke: Renamed now.

modules/meshrenderinggl/src/datastructures/rasterization.cpp
```diff
@@ -31,7 +31,7 @@
 
 namespace inviwo {
 
-Document Rasterization::getDescription() const {
+Document Rasterization::getInfo() const {
     Document doc;
     doc.append("p", "Rasterization functor.");
     return doc;
```
petersteneteg, line 37: This should describe what kind of rasterisation i.e. fancy/line/sphere etc. could also append the getInfo of the mesh data?

petersteneteg: was meant for the implementation one, think this one can be abstract

modules/meshrenderinggl/include/modules/meshrenderinggl/processors/rasterizationrenderer.h
```diff
@@ -144,8 +144,7 @@ class IVW_MODULE_MESHRENDERINGGL_API RasterizationRenderer : public Processor {
 
     std::shared_ptr<ImageInport> imageInport_;
     ImageOutport outport_;
-
-    BoolProperty forceOpaque_;
+    Image intermediateImage_;
```
petersteneteg, line 147: This could also be an ImageGL I think. but some of the convenience methods might not exits...

modules/meshrenderinggl/include/modules/meshrenderinggl/processors/meshrasterizer.h
AnkeAnke: Now citing both papers in the top, since they are mentioned repeatedly

modules/meshrenderinggl/src/processors/meshrasterizer.cpp
AnkeAnke: Seems so, yes.

modules/meshrenderinggl/src/processors/meshrasterizer.cpp
AnkeAnke: Was actually not implemented in the shader... Removed now.

modules/meshrenderinggl/src/processors/meshrasterizer.cpp
AnkeAnke: Rather not checked, will throw error, which is what we want.

modules/meshrenderinggl/include/modules/meshrenderinggl/processors/rasterizationrenderer.h
AnkeAnke: Exactly. Is more of a hassle and I have to copy code from the glutils. Hope it's not too much of an overhead?

modules/meshrenderinggl/src/datastructures/rasterization.cpp
AnkeAnke: That's the general rasterization info, so only visible if an inherited class does not override the document. Could make it purely virtual of course, but I think the fallback is fine in case of a very simple rasterizer processor.

petersteneteg (dismissed):

modules/meshrenderinggl/src/processors/rasterizationrenderer.cpp
```diff
@@ -134,19 +136,24 @@ RasterizationRenderer::IllustrationSettings::getSettings() const {
...
+    bool containsOpaque = false;
+    for (auto rasterization : rasterizations_) {
+        if (rasterization->usesFragmentLists()) {
+            fragmentLists = true;
+        } else {
+            containsOpaque = true;
         }
     }
+
+    // auto inImage = imageInport_->getData();
```
ResearchDaniel: remove

petersteneteg (approved):

modules/meshrenderinggl/src/rendering/fragmentlistrenderer.cpp
ResearchDaniel: @AnkeAnke ping

modules/meshrenderinggl/src/rendering/fragmentlistrenderer.cpp
ResearchDaniel: @AnkeAnke ping

ResearchDaniel:
@petersteneteg I just pinged @AnkeAnke about the remaining issues. One of them is a possible shader-rebuild every iteration....

AnkeAnke:
The `builtWithBackground_` flag saves whether the shader was build for use with backgrounds. Only if that flag does not agree with the input (i.e., build with background but no background incoming or vice versa), a rebuild is triggered.

ResearchDaniel:
Not if !supportsFragmentLists()
Then it will build every frame?

AnkeAnke:
That's true. I'll fix it with the next PR.

ResearchDaniel:
@petersteneteg please ask before force merging next time :)

PR #917, opened 2020-04-07
GLFW: minor fixes for the fences

petersteneteg:


modules/glfw/src/glfwmodule.cpp
```diff
@@ -75,19 +81,23 @@ void GLFWModule::onProcessorNetworkEvaluationEnd() {
     // its work before we continue. This is needed to make sure that we have textures that are upto
     // data when we render the canvases.
     auto syncObj = glFenceSync(GL_SYNC_GPU_COMMANDS_COMPLETE, 0);
```
ResearchDaniel, line 83: Is this not callled when the exception is thrown? Should check if it exist in that case

modules/glfw/src/glfwmodule.cpp
```diff
@@ -75,19 +81,23 @@ void GLFWModule::onProcessorNetworkEvaluationEnd() {
...
-    auto res = glClientWaitSync(syncObj, GL_SYNC_FLUSH_COMMANDS_BIT, 100000000);
+
+    const GLuint64 timeoutInNanoSec = 50'000'000;  // 50ms
+
+    auto res = glClientWaitSync(syncObj, GL_SYNC_FLUSH_COMMANDS_BIT, timeoutInNanoSec);
+    while (res == GL_TIMEOUT_EXPIRED) {
+        res = glClientWaitSync(syncObj, GL_SYNC_FLUSH_COMMANDS_BIT, timeoutInNanoSec);
```
ResearchDaniel, line 89: Should there be a max timeout here?

ResearchDaniel (dismissed):
There might be an issue with the destructor

modules/glfw/src/glfwmodule.cpp
petersteneteg: maybe, this is more or less what we have in qt, seems to work fine there... usually windows will timeout after 30 sec anyway...

modules/glfw/src/glfwmodule.cpp
petersteneteg: how do you mean? what exception? you mean if a processor fail in the network?

petersteneteg: this will always be call even if a processor fail, but that should not be a problem

ResearchDaniel: There is an exception thrown in the constructor if not glFenceSync exist. Then it is used here

petersteneteg: if the constructor throws then the destructor is never called

petersteneteg:
@ResearchDaniel poke

ResearchDaniel (approved):

petersteneteg:
@ResearchDaniel Thanks!

PR #924, opened 2020-04-24
DataFrame: Make ColorMapProperty default to diverging color map if va…

ResearchDaniel:
Make ColorMapProperty default to diverging color map if values go from negative to positive


martinfalk:
is this a reasonable choice? The color map will also be skewed for data that is not centered around zero.

ResearchDaniel:
I think that it is more reasonable than not having a diverging color map.

martinfalk (approved):

PR #925, opened 2020-04-27
Fix/build issues ubuntu

tirpidz:

**Always**
- [ x] Code is error and warning free
- [ x] Code follows the [Inviwo coding guidelines](https://github.com/inviwo/inviwo/wiki/Coding-Conventions)

- Operating system: ubuntu 18.04
- Compiler/IDE: GCC 8.3.0, vscode
- QT/CMake/Python versions: QT 5.14.1, cmake 3.13.4, Python 3.6.9

The PATH_MAX is undefined in my case, maybe there is something wrong with my CMake config. I would rather fix the CMake config than adding a fallback like it is the case now.


petersteneteg (approved):
Thanks!

PR #930, opened 2020-04-30
Glfw updates

petersteneteg:


CMakeLists.txt
```diff
@@ -27,7 +27,7 @@
  #
  #################################################################################
 
-cmake_minimum_required(VERSION 3.12.0)
+cmake_minimum_required(VERSION 3.12..3.17)
```
martinfalk: 3 dots between min and max version

include/inviwo/core/common/inviwoapplication.h
```diff
@@ -397,14 +399,20 @@ class IVW_CORE_API InviwoApplication : public Singleton<InviwoApplication> {
 
     ///@}
 
+    /**
+     * Sets the a FileSystemObserver for the application
+     * The FileObservers will register them self in the FileSystemObserver from the
```
martinfalk: typo
```
* The FileObservers will register themselves in the FileSystemObserver of the 
* InviwoApplication by default. Usually one would set the FileSystemObserver in the main
```

include/inviwo/core/util/filesystemobserver.h
```diff
@@ -0,0 +1,78 @@
...
+ * @see FileObserver
+ * @see InviwoApplication::setFileSystemObserver
+ */
+class IVW_CORE_API FileSystemObserver {
+public:
+    FileSystemObserver() = default;
+    virtual ~FileSystemObserver() = default;
+
+    /**
+     * Register a FileObserver to get callback when files changes.
```
martinfalk: typo, "change"

modules/glfw/include/modules/glfw/filewatcher.h
```diff
@@ -0,0 +1,72 @@
...
+
+namespace inviwo {
+
+class WatcherThread;
+class FileObserver;
+class InviwoApplication;
+
+/**
+ * An implementation for FileSystemObserver using the windows api.
+ * Currently does nothing on Mac / Linux
```
martinfalk, line 48: is that documented somewhere as a TODO?

modules/glfw/src/canvasglfw.cpp
```diff
@@ -184,136 +191,58 @@ CanvasGLFW* CanvasGLFW::getSharedContext() {
...
-    MouseState state =
-        (thisCanvas->mouseState_ == MouseState::Press ? MouseState::Move : thisCanvas->mouseState_);
-    MouseEvent mouseEvent(thisCanvas->mouseButton_, state, thisCanvas->mouseButton_,
-                          thisCanvas->modifiers_, pos, thisCanvas->getImageDimensions(),
-                          thisCanvas->getDepthValueAtNormalizedCoord(pos));
-
-    thisCanvas->propagateEvent(&mouseEvent);
+std::unique_ptr<Canvas> CanvasGLFW::createHiddenCanvas() {
+    auto res = dispatchFront(
+        [&]() { return std::make_unique<CanvasGLFW>("Background", screenDimensions_); });
```
martinfalk, line 198: might this cause a name clash in case this function is called twice?

modules/glfw/src/filewatcher.cpp
```diff
@@ -0,0 +1,325 @@
...
+                    FindNextChangeNotification(handle);
+                }
+            }
+        }
+        for (auto handle : util::as_range(handles_.begin(), handles_.begin() + active_)) {
+            FindCloseChangeNotification(handle);
+        }
+    }
+
+    std::vector<std::pair<std::string, Action>> getChangedAndUpdateFiles(
```
martinfalk, line 162: `getChangedAndUpdatedFiles`  with "update_d_"?

modules/glfw/src/filewatcher.cpp
```diff
@@ -0,0 +1,325 @@
...
+                    it->second = newTime;
+                }
+            }
+        }
+
+        return changed;
+    }
+
+    static constexpr DWORD filter =
+        FILE_NOTIFY_CHANGE_FILE_NAME | FILE_NOTIFY_CHANGE_FILE_NAME | FILE_NOTIFY_CHANGE_LAST_WRITE;
```
martinfalk: `FILE_NOTIFY_CHANGE_FILE_NAME` is in there twice. Missing another flag?

modules/glfw/src/filewatcher.cpp
```diff
@@ -0,0 +1,325 @@
...
+                  notifyAboutChanges();
+              }
+          })} {}
+
+FileWatcher::~FileWatcher() = default;
+
+void FileWatcher::registerFileObserver(FileObserver* fileObserver) {
+    IVW_ASSERT(std::find(fileObservers_.cbegin(), fileObservers_.cend(), fileObserver) ==
+                   fileObservers_.cend(),
+               "File observer already registered.");
```
martinfalk, line 250: Is it a good idea to use only an assertion here?

martinfalk (dismissed):
nice job.

modules/glfw/include/modules/glfw/filewatcher.h
petersteneteg: In the comment here... No one ever does todos anyway...

modules/glfw/src/canvasglfw.cpp
petersteneteg: It's just a window title not a identifier... also the window is hidden, so you will never see said title... but...

modules/glfw/src/filewatcher.cpp
petersteneteg: no it return the list of new changes, and updates the time stamps of the observed files

modules/glfw/src/filewatcher.cpp
petersteneteg: should be dir not file

modules/glfw/src/filewatcher.cpp
petersteneteg: that is handled internally by the FileObserver hence it should not really happen by user error. So I think the assert is good.

martinfalk (approved):

PR #931, opened 2020-05-05
clone inport data for transforms only when inport has changed.

VisComKreiser:
minimized data copy when properties change.

### DIFF
--- a/modules/base/include/modules/base/processors/transform.h
+++ b/modules/base/include/modules/base/processors/transform.h
@@ -41,6 +41,8 @@
 #include <inviwo/core/ports/datainport.h>
 #include <inviwo/core/ports/dataoutport.h>
 
+#include <memory>
+
 namespace inviwo {
 
 template <typename T>
@@ -56,6 +58,8 @@ public:
 protected:
     DataInport<T> inport_;
     DataOutport<T> outport_;
+
+    std::shared_ptr<T> data_;
 
     TemplateOptionProperty<CoordinateSpace> space_;
     BoolProperty replace_;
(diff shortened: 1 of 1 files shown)


petersteneteg (changes requested):
Hi
I understand why you want to do this, but in principle you are also not allowed to modify data that you have put once put in the inport. Imagine a processor later in the pipe line holding on to that data for background work etc. and then this processor goes and change it... Unfortunately the api we have do not enforce this. You can get around this by verifying that you are the only one holding on to the shared_ptr,... then changing it should be fine.

petersteneteg:
I have also had designs where one might allow to make shallow clones of const volumes and meshes and so. and have copy on write functionality. But I have not had time to implement that.
And Im not sure if much API that would need to break for it to work...

VisComKreiser:
Thanks for the elaborate answer, did not think about that case.
The issue we have is that we can not afford to copy huge volumes and several meshes in our application every frame when we transform things. Another alternative for us would be to incorporate a somewhat global transformation (similar to the transform processors) into the mesh renderer, volume raycaster and volume slice renderer, and attach it to the worldMatrix for rendering to be able to move stuff around all at the same time.
Maybe we can pick up that topic at the next meeting, the fance mesh renderer which is currently developed by the others also might benefit from some solution to this problem.

petersteneteg:
Yes, we can have a discussion in the meeting. Think having a transform in the new rasterisers might be a good idea anyway

AnkeAnke:
I'll add a transform matrix to the rasterizers (fancy mesh rendering etc). Also, since the output of the rasterizers, the rasteriation objects, are tiny, they can be easily copied to add a transform.

VisComKreiser:
> I'll add a transform matrix to the rasterizers (fancy mesh rendering etc). Also, since the output of the rasterizers, the rasteriation objects, are tiny, they can be easily copied to add a transform.

Something similar to the Transform Processors would be nice to have.

petersteneteg:
I'll close this now.

PR #934, opened 2020-05-11
Transform properties for general use and oit

AnkeAnke:
Some useful transform functionality, in general and for order-independent transparency.

* New TransformListProperty in `BaseGL`, allowing to stack affine transformations and a user defined matrix
* Adding further property in `rasterization`, including the switch between model/world manipulation and overwrite/application onto the old transform. Outputs a `SimpleSpatialEntity`, which can be applied on an existing transform and used in place of other `SpatialEntity`s.
We might want to move that into basegl/transform as well.
* Adding that property onto line and mesh rasterizer processor.
* Adding a separate 'transform rasterization' processor, taking in a rasterization object and modifying the transform only. Cheap since rasterizations are light-weight.

Tested on Pop OS! (Ubuntu) 20.04, GCC 9.3.0


modules/base/src/processors/transform.cpp
```diff
@@ -160,4 +160,42 @@ mat4 CustomTransformProperty::getMatrix() const { return matrix; }
...
+              std::vector<std::unique_ptr<Property>> v;
+              v.emplace_back(
+                  std::make_unique<transform::TranslateProperty>("translation", "Translation"));
+              v.emplace_back(std::make_unique<transform::RotateProperty>("rotation", "Rotation"));
+              v.emplace_back(std::make_unique<transform::ScaleProperty>("scaling", "Scaling"));
+              v.emplace_back(
+                  std::make_unique<transform::CustomTransformProperty>("custom", "Custom Matrix"));
+              return v;
+          }())
+    , result_("result", "Result", mat4(1.0f),
```
ResearchDaniel: This property should not be possible to change from the outside (or linked). 
Add constraints

modules/meshrenderinggl/include/modules/meshrenderinggl/datastructures/rasterization.h
```diff
@@ -31,20 +31,49 @@
...
 
 #include <vector>
 #include <unordered_map>
 #include <optional>
 
 namespace inviwo {
 
+namespace rasterization {
+
+struct SimpleSpatialEntity : public SpatialEntity<3> {
```
ResearchDaniel: Not sure what the purpose of this one is?

modules/meshrenderinggl/include/modules/meshrenderinggl/datastructures/rasterization.h
```diff
@@ -31,20 +31,49 @@
...
+
+struct TransformSettings : public CompositeProperty {
+    TransformSettings(const std::string& identifier = "transformSettings",
+                      const std::string& displayName = "Mesh Transforms",
+                      InvalidationLevel invalidate = InvalidationLevel::InvalidOutput);
+    TemplateOptionProperty<CoordinateSpace> space_;
+    BoolProperty replaceTransform_;
+    TransformListProperty transform_;
+
+    SimpleSpatialEntity getTransform() const;
```
ResearchDaniel: This is a bit counter-intuitive

modules/meshrenderinggl/include/modules/meshrenderinggl/datastructures/rasterization.h
```diff
@@ -67,6 +96,18 @@ class IVW_MODULE_MESHRENDERINGGL_API Rasterization {
      * @return Specific information about this rasterization type/instance.
      */
     virtual Document getInfo() const;
+
+    /**
+     * \brief Get a copy of the object. Ownership goes to the caller.
+     * @return A copy with the same data and type as the original.
+     */
+    virtual Rasterization* copy() const = 0;
```
ResearchDaniel: clone?

modules/meshrenderinggl/include/modules/meshrenderinggl/datastructures/rasterization.h
```diff
@@ -67,6 +96,18 @@ class IVW_MODULE_MESHRENDERINGGL_API Rasterization {
...
     virtual Document getInfo() const;
+
+    /**
+     * \brief Get a copy of the object. Ownership goes to the caller.
+     * @return A copy with the same data and type as the original.
+     */
+    virtual Rasterization* copy() const = 0;
+
+    /**
+     * \brief Mot simple spatial 3D data object, hosting a model and world matrix.
```
ResearchDaniel: Mot?

modules/meshrenderinggl/include/modules/meshrenderinggl/datastructures/rasterization.h
```diff
@@ -67,6 +96,18 @@ class IVW_MODULE_MESHRENDERINGGL_API Rasterization {
...
+     * Should be applied to geometry, and then set as a uniform.
+     */
+    rasterization::SimpleSpatialEntity spatialTransformation_;
```
ResearchDaniel: Is it not a bit backwards to have a transform in the rasterizer? 
This is equivalent to a global transformation?

ResearchDaniel (changes requested):
Does it not make more sense to take transformation matrices as optional input (inport) instead of including them in the rasterizer as properties?

petersteneteg:
@ResearchDaniel what would be the reason for setting the transform as a inport? 
We have no precedence for using port for transforms that I know of

petersteneteg:
@ResearchDaniel We can't really make properties unlinkable

modules/base/src/processors/transform.cpp
petersteneteg: We can not make a property unlinkable

modules/meshrenderinggl/include/modules/meshrenderinggl/datastructures/rasterization.h
petersteneteg: its per rasterisation object so it's not global. Many rasterisations are combined in the final rendering. each with individual transforms.

petersteneteg:
We need to consider how this affects picking also.

ResearchDaniel:
> @ResearchDaniel what would be the reason for setting the transform as a inport?
> We have no precedence for using port for transforms that I know of

Properties are for user input, but transforms could also be generated from code, e.g. animations or read from file.  
No need for special handling of the rasterizer, e.g. adding Transform properties for rasterizer, and instead use one way of including transforms.

petersteneteg:
I would not use ports for this.
All our existing processors use properties for transforms.
The whole animation module uses properties for animations. 
I see no reason for using ports for this.

petersteneteg:
@ResearchDaniel We discussed this quite a bit in the last meeting

modules/base/src/processors/transform.cpp
ResearchDaniel: Should be ConstraintBehavior::Ignore?

ResearchDaniel:
> @ResearchDaniel We can't really make properties unlinkable

Gotcha

modules/base/src/processors/transform.cpp
petersteneteg: yes, but that does not affect linking of the value. Only the boundary

ResearchDaniel:
> I would not use ports for this.
> All our existing processors use properties for transforms.

But those are mainly for setting up data. The transform processor that do modify via a property is not suitable for dynamically transforming data.

> The whole animation module uses properties for animations.
> I see no reason for using ports for this.

But this forces a special way of handling transforms for this type of meshrendering, via a property. 
The (global) transformations are then forwarded via the rasterizer as inport data anyway?
If we would use inports we could use the same transformation processor (with a transformation property and transformation outport) for both this module and the one in basegl?

ResearchDaniel:
I guess the difficulty with having an inport is that they are not directly associated, so there could be more/fewer transforms than meshes

petersteneteg:
@ResearchDaniel Not sure I understand your thinking.... maybe we can discuss over zoom later?

modules/meshrenderinggl/include/modules/meshrenderinggl/datastructures/rasterization.h
AnkeAnke: I want to hand a transform around. Currently only in use in the rasterizations, which receive additinal transforms from either LineRasterizer, MeshRasterizer or the TransformRasterization processors. 
Generally, I'd only need a Model and World matrix, but using a light-weight SpatialEntity has the advantage of being able to set it directly. 
[See use here.](https://github.com/inviwo/inviwo/blob/d54984153cdc7b28f85fd3c4746b9228eb8f05f3/modules/meshrenderinggl/src/processors/meshrasterizer.cpp#L479)

I will add documentation there.

modules/meshrenderinggl/include/modules/meshrenderinggl/datastructures/rasterization.h
AnkeAnke: Which part exactly? This property mimics the interface of the transform processor. Since it does not have a SpatialEntity, it can not apply the transform directly, but rather return the transformation itself.
This might get clearer when SimpleSpatialEntity receives some docu.

modules/meshrenderinggl/include/modules/meshrenderinggl/datastructures/rasterization.h
AnkeAnke: This PR is mostly in response to [this pull request](https://github.com/inviwo/inviwo/pull/931), looking for a way of changing transforms without copying the (rather large) data objects. Adding the transform as a property for rendering only solves that.

AnkeAnke (commented):
Will add some documentation to the `SimpleSpatialEntity` and some fixes.

modules/meshrenderinggl/include/modules/meshrenderinggl/datastructures/rasterization.h
petersteneteg: I don't quite understand why you need both a world and model matrix though? when used in this way?

modules/meshrenderinggl/include/modules/meshrenderinggl/datastructures/rasterization.h
ResearchDaniel: Depends on if one view: SpatialEntity = Transform
If so, the naming is intuitive :)

AnkeAnke: Now named CompositeTransform.

modules/meshrenderinggl/include/modules/meshrenderinggl/datastructures/rasterization.h
AnkeAnke: Updated as we discussed at fika.

modules/base/include/modules/base/processors/transform.h
```diff
@@ -43,6 +43,23 @@
 
 namespace inviwo {
 
+/** \docpage{org.inviwo.TransformListProperty, Transform List Property}
+ * ![](org.inviwo.TransformListProperty.png?classIdentifier=org.inviwo.TransformListProperty)
+ * List of transformations being applied on each other.
+ */
+class IVW_MODULE_BASE_API TransformListProperty : public CompositeProperty {
```
petersteneteg: move into its own header under base/properties/transformlistproperty.h

petersteneteg: also add
```c++
```

modules/base/src/processors/transform.cpp
```diff
@@ -160,4 +160,43 @@ mat4 CustomTransformProperty::getMatrix() const { return matrix; }
...
+              v.emplace_back(std::make_unique<transform::ScaleProperty>("scaling", "Scaling"));
+              v.emplace_back(
+                  std::make_unique<transform::CustomTransformProperty>("custom", "Custom Matrix"));
+              return v;
+          }())
+    , result_(
+          "result", "Result", mat4(1.0f),
+          {util::filled<mat4>(std::numeric_limits<float>::lowest()), ConstraintBehavior::Ignore},
+          {util::filled<mat4>(std::numeric_limits<float>::max()), ConstraintBehavior::Ignore},
+          util::filled<mat4>(0.001f), InvalidationLevel::Valid) {
```
petersteneteg: you can set the semantics here to

modules/base/src/processors/transform.cpp
```diff
@@ -160,4 +160,43 @@ mat4 CustomTransformProperty::getMatrix() const { return matrix; }
...
+    result_.setSemantics(PropertySemantics::Text);
+    result_.setReadOnly(true);
```
petersteneteg: add a setCurrentStateAsDefault here

modules/base/include/modules/base/processors/transform.h
petersteneteg: also move the transform properties here

modules/meshrenderinggl/include/modules/meshrenderinggl/datastructures/transformedrasterization.h
```diff
@@ -0,0 +1,74 @@
...
+protected:
+    std::shared_ptr<const Rasterization> innerRasterization_;
+    const mat4 additionalWorldTransform_;
+};
+
+/**
+ * \brief A very simple SpatialEntity<3> to handle a world and model transform
+ * Used by rasterizations for adding transforms before rendering, without copying the mesh data.
+ */
+struct IVW_MODULE_MESHRENDERINGGL_API CompositeTransform : public SpatialEntity<3> {
```
petersteneteg, line 66: Why not just use SpatialEntity directly? this class does nothing now? or did i miss something?

modules/meshrenderinggl/include/modules/meshrenderinggl/datastructures/transformedrasterization.h
```diff
@@ -0,0 +1,74 @@
...
+#pragma once
+
+#include <modules/meshrenderinggl/meshrenderingglmoduledefine.h>
+#include <modules/meshrenderinggl/datastructures/rasterization.h>
+
+namespace inviwo {
+/**
+ * \brief Rasterization wrapper to add an additional transform on the render call.
+ */
+class IVW_MODULE_MESHRENDERINGGL_API TransformedRasterization : public Rasterization {
```
petersteneteg, line 39: Cool I like this!

modules/meshrenderinggl/src/processors/linerasterizer.cpp
```diff
@@ -126,11 +129,18 @@ void LineRasterizer::process() {
         shader.deactivate();
     }
 
-    outport_.setData(new LineRasterization(*this));
+    std::shared_ptr<const Rasterization> rasterization = std::make_shared<LineRasterization>(*this);
+
+    // If transform is applied, wrap rasterization.
+    if (transformSetting_.transforms_.size() > 0) {
+        outport_.setData(
+            new TransformedRasterization(rasterization, transformSetting_.getMatrix()));
```
petersteneteg: use std::make_shared
we should deprecate that overload...

modules/meshrenderinggl/src/processors/linerasterizer.cpp
```diff
@@ -195,13 +205,12 @@ void LineRasterization::rasterize(const ivec2& imageSize,
...
                 auto& shader = lineShaders_->getShader(*mesh, mesh->getIndexMeshInfo(i));
-                if (!shader.isReady()) {
-                    LogWarn("Shader not ready.");
-                    break;
-                }
+                if (!shader.isReady()) break;
 
                 shader.activate();
-                utilgl::setShaderUniforms(shader, *mesh, "geometry");
+                auto transform = CompositeTransform(mesh->getModelMatrix(),
```
petersteneteg, line 212: just use a SpatialEntity directly?

modules/meshrenderinggl/src/processors/meshrasterizer.cpp
```diff
@@ -128,9 +127,8 @@ MeshRasterizer::MeshRasterizer()
     edgeSettings_.visibilityDependsOn(faceSettings_[0].showEdges_, edgevis);
     edgeSettings_.visibilityDependsOn(faceSettings_[1].showEdges_, edgevis);
 
-    camera_.setCollapsed(true);
     lightingProperty_.setCollapsed(true);
-    trackball_.setCollapsed(true);
+    transformSetting_.setCollapsed(true);
```
petersteneteg, line 131: might also add setCurrentStateAsDefault here...

petersteneteg (changes requested):
Cool! I like this design. Just had some minor fixes / suggestions...

modules/meshrenderinggl/include/modules/meshrenderinggl/datastructures/transformedrasterization.h
ResearchDaniel: Or simply just a matrix?

petersteneteg: This is used to add all the uniforms in the shaders in a convenient and standard way...
We also add the inverse etc...

AnkeAnke: SpatialEntity is a purely virtual class, so can't do. This is the most lightweight way of doing that, I suppose.

modules/meshrenderinggl/src/processors/linerasterizer.cpp
AnkeAnke: As said above, that one is purely virtual.

modules/meshrenderinggl/include/modules/meshrenderinggl/datastructures/transformedrasterization.h
AnkeAnke: We could of course make SpatialEntity instantiable by just setting a clone implementation on the base class. I think I prefer this explicit way though.

petersteneteg: ahh ok

modules/meshrenderinggl/src/processors/meshrasterizer.cpp
```diff
@@ -475,7 +480,9 @@ void MeshRasterization::rasterize(const ivec2& imageSize,
         for (auto mesh : enhancedMeshes_) {
```
ResearchDaniel, line 481: outside PR, but why is transferFuncUnit set in rasterize() and not with the other uniforms in the process function?

AnkeAnke: The transfer function needs to bind a texture. We cannot leave that one bound in between, but only set it right before the render call.

ResearchDaniel: Right, thought about that later as well

ResearchDaniel (dismissed):

petersteneteg (dismissed):

modules/meshrenderinggl/src/processors/transformrasterization.cpp
```diff
@@ -0,0 +1,69 @@
...
+}
+
+void TransformRasterization::process() {
+    if (!inport_.hasData()) {
+        outport_.setData(nullptr);
+        return;
+    }
+
+    outport_.setData(
+        new TransformedRasterization(inport_.getData(), transformSetting_.getMatrix()));
```
petersteneteg, line 66: should use make_shared here. We never use new...

modules/base/include/modules/base/properties/transformlistproperty.h
```diff
@@ -0,0 +1,162 @@
...
+
+/** \docpage{org.inviwo.TransformListProperty, Transform List Property}
+ * ![](org.inviwo.TransformListProperty.png?classIdentifier=org.inviwo.TransformListProperty)
+ * List of transformations being applied on each other.
+ */
+class IVW_MODULE_BASE_API TransformListProperty : public CompositeProperty {
+public:
+    TransformListProperty(const std::string& identifier, const std::string& displayName,
+                          InvalidationLevel invalidationLevel = InvalidationLevel::InvalidOutput,
+                          PropertySemantics semantics = PropertySemantics::Default);
```
petersteneteg, line 49: should have a clone and a Copy constructor to... that why the integration tests fail....

VisComKreiser:
Hey, I've also had time to look at it and it feels nice to use. So basically all processors which are currently rendering directly from a volume/mesh, like the axis aligned cut plane or so, would have to be changed to output rasterizers instead of rendering directly on top of the image then?

VisComKreiser (dismissed):

petersteneteg (approved):

petersteneteg (approved):

PR #935, opened 2020-05-12
CMake flag for colorful ninja output

AnkeAnke:
GCC and Clang apparently disable ANSI when their log is redirected, as in ninja calling compiler calls.
Adding simple flag to set the respective compiler flag.

Tested on GCC 9.3.0.

### DIFF
--- a/cmake/compileoptions.cmake
+++ b/cmake/compileoptions.cmake
@@ -33,6 +33,7 @@
 
 option(IVW_TREAT_WARNINGS_AS_ERRORS "Treat compiler warnings as errors" OFF)
 option(IVW_FORCE_ASSERTIONS "Force use of assertions when not in debug mode" OFF)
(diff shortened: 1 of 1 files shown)


petersteneteg (changes requested):
This is probably not the place for such an option, this is for per target settings.
The preferred way for this kind of options I think is to just add them to CMAKE_CXX_FLAGS in the cmake client. Is there a reason why this would not be sufficient? 
Also we always prefix all our options with IVW_

AnkeAnke:
It might be a problem that will come up for others as well, so I added the flag. But I can just fix it locally, that's fine by me. Your call ;)

petersteneteg:
I understand your point, my reluctance comes from the fact that it adds more complexity to an already to complex setup. It also interferes with the "standard" solution, adding the setting to the CMAKE_CXX_FLAGS. Meaning our users now have to find and figure out what our solution does.  And if that is any different from the "standard" one...

PR #940, opened 2020-05-15
Ordinalrefproperty

petersteneteg:
Like ordinal property but with a get and set callback for the value.

### DIFF
--- a/include/inviwo/core/properties/ordinalrefproperty.h
+++ b/include/inviwo/core/properties/ordinalrefproperty.h
@@ -66,7 +66,9 @@ struct OrdinalRefPropertyState {
 
 /**
  * \ingroup properties
- * A property representing a Ordinal value, for example int, floats.
+ * A property representing a reference to an Ordinal value, for example a int, float, or vec3.
+ * The property does not hold or own the value it self. It only contain a set and get callback
+ * function. It does handle min/max/increment for the value.
  */
 template <typename T>
 class OrdinalRefProperty : public Property {
--- a/include/inviwo/core/properties/ordinalrefproperty.h
+++ b/include/inviwo/core/properties/ordinalrefproperty.h
@@ -2,7 +2,7 @@
  *
  * Inviwo - Interactive Visualization Workshop
  *
- * Copyright (c) 2012-2020 Inviwo Foundation
+ * Copyright (c) 2020 Inviwo Foundation
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -35,6 +35,7 @@
 #include <inviwo/core/util/glm.h>
 
 #include <string>
+#include <functional>
 
 #include <fmt/format.h>
 #include <fmt/ostream.h>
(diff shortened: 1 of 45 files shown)


include/inviwo/core/properties/ordinalrefproperty.h
```diff
@@ -0,0 +1,611 @@
...
+    maxValue_.setAsDefault();
+    increment_.setAsDefault();
+    return *this;
+}
+
+template <typename T>
+void OrdinalRefProperty<T>::serialize(Serializer& s) const {
+    Property::serialize(s);
+
+    minConstraint_.serialize(s, this->serializationMode_);
```
ResearchDaniel, line 499: value must be serialized/deserialized?

include/inviwo/core/properties/ordinalrefproperty.h
```diff
@@ -0,0 +1,611 @@
...
+            return PropertySemantics::Text;
+        } else {
+            return PropertySemantics::Default;
+        }
+    }
+};
+
+/**
+ * \ingroup properties
+ * A property representing a Ordinal value, for example int, floats.
```
ResearchDaniel: an Ordinal
Also state its difference to OrdinalProperty

src/core/common/inviwocore.cpp
```diff
@@ -281,61 +305,49 @@ InviwoCore::InviwoCore(InviwoApplication* app)
...
-    registerProperty<DoubleQuaternionProperty>();
+    registerProperty<FilePatternProperty>();
+    registerProperty<DirectoryProperty>();
+
+    // Register ordinal property
+    using OrdinalTypes =
+        std::tuple<float, vec2, vec3, vec4, mat2, mat3, mat4, double, dvec2, dvec3, dvec4, dmat2,
+                   dmat3, dmat4, int, ivec2, ivec3, ivec4, glm::i64, unsigned int, uvec2, uvec3,
+                   uvec4, size_t, size2_t, size3_t, size4_t, glm::fquat, glm::dquat>;
+    util::for_each_type<OrdinalTypes>{}(OrdinalReghelper{}, *this);
```
ResearchDaniel, line 342: Might need to update the json module for the new properties as well

ResearchDaniel (changes requested):
Looks good. Might be a bug in serialization though

include/inviwo/core/properties/ordinalrefproperty.h
petersteneteg: no the value is kept external to the property

ResearchDaniel: But how will the external thingy know about serialization?

petersteneteg: well it is up to the external thing to persist it state. maybe the processor serialise it or maybe the thing is serialisable it self.

ResearchDaniel (dismissed):

include/inviwo/core/properties/ordinalrefproperty.h
```diff
@@ -0,0 +1,613 @@
...
+        set_(val);
+        modified = true;
+    }
+
+    if (modified) this->propertyModified();
+}
+
+template <typename T>
+void OrdinalRefProperty<T>::set(const Property* srcProperty) {
+    if (auto prop = dynamic_cast<const OrdinalRefProperty<T>*>(srcProperty)) {
```
ResearchDaniel, line 410: Is this tio restrictive? Should we make it possible to also combine this one with OrdinalProperty and vice versa?

ResearchDaniel (changes requested):
We need converters to link them with OrdinalProperty?
Also forgot that they need to be registered in the webbrowser module, sorry

include/inviwo/core/properties/ordinalrefproperty.h
petersteneteg: Added the converters

ResearchDaniel (approved):

PR #943, opened 2020-05-19
Core: Better serialization of polymorphic unique_ptr

petersteneteg:

**Always**

**New features**

**Updated feature / Fixes** 

- Operating system
- Compiler/IDE
- QT/CMake/Python versions

### DIFF
--- a/include/inviwo/core/io/serialization/deserializer.h
+++ b/include/inviwo/core/io/serialization/deserializer.h
@@ -35,6 +35,7 @@
 #include <inviwo/core/util/factory.h>
 #include <inviwo/core/util/logfilter.h>
 #include <inviwo/core/util/stdextensions.h>
+#include <inviwo/core/util/detected.h>
 #include <inviwo/core/util/glm.h>
 
 #include <flags/flags.h>
@@ -242,11 +243,11 @@ public:
     template <class Base, class T>
     void deserializeAs(const std::string& key, T*& data);
 
-    template <class T, class D>
-    void deserialize(const std::string& key, std::unique_ptr<T, D>& data);
+    template <class T>
+    void deserialize(const std::string& key, std::unique_ptr<T>& data);
 
-    template <class Base, class T, class D>
-    void deserializeAs(const std::string& key, std::unique_ptr<T, D>& data);
+    template <class Base, class T>
+    void deserializeAs(const std::string& key, std::unique_ptr<T>& data);
 
     void setExceptionHandler(ExceptionHandler handler);
     void handleError(const ExceptionContext& context);
@@ -363,6 +364,12 @@ private:
 };
 
 namespace util {
+
+template <typename T>
+using classIdentifierType = decltype(std::declval<T>().getClassIdentifier());
+
+template <class T>
+using HasGetClassIdentifier = is_detected_exact<std::string, classIdentifierType, T>;
 
 /**
  * A helper class for more advanced deserialization. useful when one has to call observer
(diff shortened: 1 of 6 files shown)


include/inviwo/core/io/serialization/deserializer.h
```diff
@@ -364,6 +365,12 @@ class ContainerWrapper {
 
 namespace util {
 
+template <typename T>
```
ResearchDaniel, line 368: We might want to move these functions to another namespace, detail?

include/inviwo/core/io/serialization/deserializer.h
```diff
@@ -1145,14 +1132,25 @@ void Deserializer::deserialize(const std::string& key, T*& data) {
...
+    auto ptr = data.release();
+
+    if constexpr (util::HasGetClassIdentifier<T>::value) {
+        auto keyNode = retrieveChild(key);
+        if (!keyNode) return;
+
+        const std::string type_attr{
+            detail::getNodeAttribute(keyNode, SerializeConstants::TypeAttribute)};
+
+        if (ptr && !type_attr.empty() && type_attr != ptr->getClassIdentifier()) {
```
ResearchDaniel, line 1146: This handles derived classes?

include/inviwo/core/io/serialization/serializebase.h
```diff
@@ -192,7 +192,7 @@ decltype(auto) toStr(const T& value) {
     } else {
         std::ostringstream stream;
         if constexpr (std::is_same_v<T, double>) {
-            stream.precision(40);
+            stream.precision(17);
```
ResearchDaniel, line 195: Why 17? (Could say the same about 40...) There should be a comment here

src/core/tests/unittests/serializer-polymorphic-test.cpp
```diff
@@ -0,0 +1,245 @@
...
+    s.writeFile(ss);
+
+    std::unique_ptr<Property> prop2 =
+        std::make_unique<DoubleProperty>("double2", "double2", 0.234f);
+    ;
+
+    Deserializer d(ss, "");
+    d.registerFactory(InviwoApplication::getPtr()->getPropertyFactory());
+
+    d.deserialize("prop", prop2);
```
ResearchDaniel, line 120: This makes sense, but what about DoubleProperty -> BoolProperty, which I think also would be valid given the setup here?

ResearchDaniel (approved):
Looks reasonable, but there are some aspects that might need to be considered with respect to converting between the types during serialization

include/inviwo/core/io/serialization/serializebase.h
petersteneteg: thought 16 would be enough for a double, but then one of the unit tests fail to roundtrip a double max. 17 works fine. 40 is just crazy :)

include/inviwo/core/io/serialization/deserializer.h
petersteneteg: yes

include/inviwo/core/io/serialization/deserializer.h
petersteneteg: so if you have a std::unique_ptr<Property> that currently is a BoolProperty but is was serialised as a StringProperty
then it will delete the BoolProperty and create a StringProperty

src/core/tests/unittests/serializer-polymorphic-test.cpp
petersteneteg: Yes that should work fine. What would the problem be?

include/inviwo/core/io/serialization/deserializer.h
petersteneteg: will move in the next pr...

src/core/tests/unittests/serializer-polymorphic-test.cpp
ResearchDaniel: It would be unexpected behaviour that it actually works?
It would mean that one has changed the type of a property, but not the identifier, and not provided a converter for the workspace?

include/inviwo/core/io/serialization/deserializer.h
ResearchDaniel: Then one should have provided a converter instead?
I assume that this can cause some headaches later because one forgot (or was lazy) to do that

src/core/tests/unittests/serializer-polymorphic-test.cpp
petersteneteg: The use case for it is for example the camera, there we have a unique_ptr<Camera> and it can change between various child classes, Perspective, Orthographic etc. And we want to be able to switch between them.

include/inviwo/core/io/serialization/deserializer.h
petersteneteg: this will atm not apply to our properties, since we don't deserialise them using a unique_ptr

PR #944, opened 2020-05-20
Camera updates

petersteneteg:
The camera now uses ref properties instead to unify the state


include/inviwo/core/datastructures/camera.h
```diff
@@ -37,7 +37,18 @@
 
 namespace inviwo {
 
-class CompositeProperty;
+namespace cameradefaults {
+constexpr vec3 lookFrom = vec3(0.0f, 0.0f, 2.0f);
+constexpr vec3 lookTo = vec3(0.0f);
+constexpr vec3 lookUp = vec3(0.0f, 1.0f, 0.0f);
+constexpr float nearPlane = 0.1f;
+constexpr float farPlane = 1000.0f;
```
ResearchDaniel: This was 10000 before.

include/inviwo/core/datastructures/camera.h
```diff
@@ -255,43 +269,41 @@ class IVW_CORE_API OrthographicCamera final : public Camera {
...
 class IVW_CORE_API SkewedPerspectiveCamera final : public Camera {
 public:
-    SkewedPerspectiveCamera(vec3 lookFrom = vec3(0.0f, 0.0f, 2.0f), vec3 lookTo = vec3(0.0f),
-                            vec3 lookUp = vec3(0.0f, 1.0f, 0.0f), float nearPlane = 0.01f,
-                            float farPlane = 10000.0f, float fieldOfView = 60.f,
-                            float aspectRatio = 1.f, vec2 frustumOffset = vec2(0.0f, 0.0f));
+    SkewedPerspectiveCamera(
+        vec3 lookFrom = cameradefaults::lookFrom, vec3 lookTo = cameradefaults::lookTo,
+        vec3 lookUp = cameradefaults::lookUp, float nearPlane = cameradefaults::nearPlane,
+        float farPlane = cameradefaults::farPlane, float fieldOfView = cameradefaults::fieldOfView,
```
ResearchDaniel: Was fov = 60 intentional here since it is used for stereo views? I guess it might not matter...

include/inviwo/core/properties/cameraproperty.h
```diff
@@ -152,21 +151,42 @@ class IVW_CORE_API CameraProperty : public CompositeProperty, public TrackballOb
 
     void invokeEvent(Event* event) override;
 
+    Property* getCameraProperty(const std::string& identifier) const;
+    void addCamerapProperty(Property* camprop);
```
ResearchDaniel: Specify ownership

include/inviwo/core/datastructures/camera.h
```diff
@@ -300,90 +312,99 @@ inline const vec3& Camera::getLookTo() const { return lookTo_; }
 inline const vec3& Camera::getLookUp() const { return lookUp_; }
 
 inline void Camera::setLookFrom(vec3 val) {
-    lookFrom_ = val;
-    invalidateViewMatrix();
+    if (lookFrom_ != val) {
```
ResearchDaniel: Why these checks? The invalidation only sets a flag. I wonder if it makes a difference in terms of performance, do we set it to the same value many times?

include/inviwo/core/properties/cameraproperty.h
```diff
@@ -152,21 +151,42 @@ class IVW_CORE_API CameraProperty : public CompositeProperty, public TrackballOb
...
+
+    virtual void serialize(Serializer& s) const override;
+    virtual void deserialize(Deserializer& d) override;
+
+private:
+    CameraFactory* factory_;
+    OptionPropertyString cameraType_;
```
ResearchDaniel, line 165: Make public?

include/inviwo/core/properties/cameraproperty.h
```diff
@@ -152,21 +151,42 @@ class IVW_CORE_API CameraProperty : public CompositeProperty, public TrackballOb
...
+
+    FloatVec3RefProperty lookFrom_;
+    FloatVec3RefProperty lookTo_;
+    FloatVec3RefProperty lookUp_;
+    FloatRefProperty aspectRatio_;
+    FloatRefProperty nearPlane_;
+    FloatRefProperty farPlane_;
 
 private:
+    bool changeCamera(const std::string& name);
```
ResearchDaniel, line 186: Why private? Should only be done via property?

src/core/datastructures/camera.cpp
```diff
@@ -188,68 +191,51 @@ bool PerspectiveCamera::equal(const Camera& other) const {
...
+const std::string OrthographicCamera::classIdentifier = "OrthographicCamera";
 
-    if (config == Config::Hide) {
-        if (widthProp) widthProp->setVisible(false);
-        return;
+void OrthographicCamera::updateFrom(const Camera* source) {
+    Camera::updateFrom(source);
+    if (auto oc = dynamic_cast<const OrthographicCamera*>(source)) {
+        setWidth(oc->getWidth());
+    } else if (auto pc = dynamic_cast<const PerspectiveCamera*>(source)) {
```
ResearchDaniel: What about skewedperspectiveCamera?

src/core/datastructures/camera.cpp
```diff
@@ -99,87 +103,86 @@ vec3 Camera::getNormalizedDeviceFromNormalizedScreenAtFocusPointDepth(
...
-        return false;
+const std::string PerspectiveCamera::classIdentifier = "PerspectiveCamera";
+
+void PerspectiveCamera::updateFrom(const Camera* source) {
+    Camera::updateFrom(source);
+    if (auto pc = dynamic_cast<const PerspectiveCamera*>(source)) {
+        setFovy(pc->getFovy());
+    } else if (auto oc = dynamic_cast<const OrthographicCamera*>(source)) {
+        setFovy(glm::degrees(
+            2.0f * std::atan(oc->getWidth() / 2.0f / glm::distance(getLookTo(), getLookFrom()))));
```
ResearchDaniel: What about skewedperspectiveCamera?

src/core/properties/cameraproperty.cpp
```diff
@@ -53,19 +54,45 @@ CameraProperty::CameraProperty(const std::string& identifier, const std::string&
...
-              InvalidationLevel::InvalidOutput, PropertySemantics::SpinBox)
-    , aspectRatio_("aspectRatio", "Aspect Ratio", 1.0f, 0.0f, std::numeric_limits<float>::max(),
-                   0.01f, InvalidationLevel::InvalidOutput, PropertySemantics::Text)
-    , nearPlane_("near", "Near Plane", 0.1f, 0.001f, 10.f, 0.001f)
-    , farPlane_("far", "Far Plane", 100.0f, 1.0f, 1000.0f, 1.0f)
+    , lookFrom_(
+          "lookFrom", "Look from", [this]() { return camera_->getLookFrom(); },
+          [this](const vec3& val) { camera_->setLookFrom(val); },
+          {-vec3(100.0f), ConstraintBehavior::Ignore}, {vec3(100.0f), ConstraintBehavior::Ignore},
+          vec3(0.1f), InvalidationLevel::InvalidOutput, PropertySemantics{"SphericalSpinBox"})
```
ResearchDaniel: I hate the spherical spinbox :)
The rotations in that space does not make much sense for camera interaction

src/core/properties/cameraproperty.cpp
```diff
@@ -53,19 +54,45 @@ CameraProperty::CameraProperty(const std::string& identifier, const std::string&
...
+          {std::numeric_limits<float>::max(), ConstraintBehavior::Immutable}, 0.01f,
+          InvalidationLevel::InvalidOutput, PropertySemantics::Text)
+    , nearPlane_(
+          "near", "Near Plane", [this]() { return camera_->getNearPlaneDist(); },
+          [this](const float& val) { camera_->setNearPlaneDist(val); },
+          {0.001f, ConstraintBehavior::Ignore}, {10.0f, ConstraintBehavior::Ignore}, 0.001f)
+    , farPlane_(
+          "far", "Far Plane", [this]() { return camera_->getFarPlaneDist(); },
+          [this](const float& val) { camera_->setFarPlaneDist(val); },
+          {1.0f, ConstraintBehavior::Ignore}, {1000.0f, ConstraintBehavior::Ignore}, 1.0f)
```
ResearchDaniel: Increase max bound. Default of it regular value should be 10000...

src/core/properties/cameraproperty.cpp
```diff
@@ -244,20 +217,55 @@ CameraProperty::operator const Camera&() const { return *camera_; }
...
+
+CameraProperty& CameraProperty::setCamera(const std::string& cameraIdentifier) {
+    cameraType_.setSelectedIdentifier(cameraIdentifier);
+    return *this;
+}
+
+CameraProperty& CameraProperty::setCamera(std::unique_ptr<Camera> newCamera) {
+    if (newCamera) {
+        NetworkLock lock(this);
+        hideConfiguredProperties();
```
ResearchDaniel, line 227: Should the previous visibility state be restored in the end here?

src/core/properties/cameraproperty.cpp
```diff
@@ -355,6 +384,37 @@ void CameraProperty::invokeEvent(Event* event) {
...
+}
+
+void CameraProperty::serialize(Serializer& s) const {
+    CompositeProperty::serialize(s);
+    s.serialize("Camera", camera_);
+
+}
+
+void CameraProperty::deserialize(Deserializer& d) {
+    hideConfiguredProperties();
```
ResearchDaniel: Why this hiding? Might want to clarify it’s purpose.

ResearchDaniel (changes requested):
Wow, a small step for the camera but a giant leap for Inviwo :)

include/inviwo/core/datastructures/camera.h
petersteneteg: It was 0.01 to 10000 before that range might be to big for a float and might give precision issues, that’s why I decreased the default. The automatic fitting also makes this easier to adjust

include/inviwo/core/datastructures/camera.h
petersteneteg: Think 60 was just a bad choice to begin with. 38 is recommended by Martin :)

petersteneteg: It was 60 also for the perspective one, so it should not have anything to do with the stereo aspects

include/inviwo/core/properties/cameraproperty.h
petersteneteg: Will do

include/inviwo/core/datastructures/camera.h
petersteneteg: Mostly to be consistent. We set a flag but then also have to recalculate the matrices...

include/inviwo/core/properties/cameraproperty.h
petersteneteg: Maybe possible now... there are a few invariants that has to be taken care of...

include/inviwo/core/properties/cameraproperty.h
petersteneteg: There is a different one that is public

src/core/datastructures/camera.cpp
petersteneteg: Yes could add that to

src/core/datastructures/camera.cpp
petersteneteg: Yes could add that

src/core/properties/cameraproperty.cpp
petersteneteg: xyz is not much better...

src/core/properties/cameraproperty.cpp
petersteneteg: Why this is only for the slider...

src/core/properties/cameraproperty.cpp
petersteneteg: Configure will show them again

src/core/properties/cameraproperty.cpp
petersteneteg: Mmm will add some comment

include/inviwo/core/datastructures/camera.h
ResearchDaniel: http://gameaccessibilityguidelines.com/if-the-game-uses-field-of-view-3d-engine-only-set-an-appropriate-default-for-the-expected-viewing-environment/

60 is default in Unity as well

ResearchDaniel: Just some other input on the question
https://forum.unity.com/threads/what-is-a-normal-field-of-view.70202/
” For PC games, the screen aspect is important, because players sit roughly at the same distance from the monitor. Let's say now you're using a 21:9 monitor, compared to a 16:9 one, you should see more space horizontally, NOT vertically, right? But in Unity, the aspect remains the same! So if you increase the FOV, you see more space horizontally AND vertically, that's the reason why developers feel weird and uncomfortable when they set the value to 90 or even higher on their camera. In fact, a higher FOV in Unity will make you feel like you're traveling at the speed of light, not only "wider view".

Last question, how to see wider? Change the aspect instead.”

petersteneteg: I think, there is a big difference between a 3D game (like a FPS) and a scientific visualization. I find that 60° create ways too much distortion. @martinfalk might also have opinions here. Also you can always change the camera to fit your preference...

include/inviwo/core/datastructures/camera/cameratools.h
```diff
@@ -0,0 +1,71 @@
...
+
+#include <memory>
+
+namespace inviwo {
+
+class CameraProperty;
+
+namespace util {
+
+IVW_CORE_API FloatRefProperty* getCameraFovProperty(CameraProperty* comp);
```
ResearchDaniel: Should state that null is returned if it does not exist. 
Should be const Camera*

include/inviwo/core/datastructures/camera/cameratools.h
```diff
@@ -0,0 +1,71 @@
...
+
+IVW_CORE_API std::unique_ptr<FloatRefProperty> createCameraFovProperty(
+    std::function<float()> get, std::function<void(const float&)> set);
+
+IVW_CORE_API FloatRefProperty* updateOrCreateCameraFovProperty(
+    CameraProperty* comp, std::function<float()> get, std::function<void(const float&)> set);
+
+
+IVW_CORE_API FloatRefProperty* getCameraWidthProperty(CameraProperty* comp);
```
ResearchDaniel: Same as comment above. Should also state that this property is only valid for orthographic camera?

include/inviwo/core/datastructures/camera/cameratools.h
```diff
@@ -0,0 +1,71 @@
...
+
+IVW_CORE_API std::unique_ptr<FloatRefProperty> createCameraWidthProperty(
+    std::function<float()> get, std::function<void(const float&)> set);
+
+IVW_CORE_API FloatRefProperty* updateOrCreateCameraWidthProperty(
+    CameraProperty* comp, std::function<float()> get, std::function<void(const float&)> set);
+
+
+
+IVW_CORE_API FloatVec2RefProperty* getCameraSeperationProperty(CameraProperty* comp);
```
ResearchDaniel: I did not see (eye)separation in any cameras. Is this property only related the the CameraProperty? I think that it would be good to document where each of these properties stem from, i.e. which Camera types

include/inviwo/core/datastructures/camera/skewedperspectivecamera.h
```diff
@@ -0,0 +1,111 @@
...
+namespace inviwo {
+
+/**
+ * \brief Camera with off axis perspective projection.
+ *
+ * The camera with unsymmetrical frustum for stereo in VR
+ * Kooima, Robert. "Generalized perspective projection." School of Elect. Eng. and Computer Science
+ * (2008): 1-7.
+ * @see Camera
+ * @see SkewedPerspectiveCamera
```
ResearchDaniel: This Reference to itself should be removed

include/inviwo/core/datastructures/camera/cameratools.h
petersteneteg: It is also called offset in the skewedprespective camera

include/inviwo/core/datastructures/camera/cameratools.h
petersteneteg: it tries to find it in the CameraProperty, not in the camera. Guess it could be const syntactically, but semantically I don't think it should be since it returns a non-const property

include/inviwo/core/datastructures/camera/cameratools.h
petersteneteg: It works for any CameraProperty. but it would only be needed by an orthographic camera.

include/inviwo/core/datastructures/camera/cameratools.h
ResearchDaniel: Aha, eye separation is in my opinion more descriptive. Would be good if our naming is consistent in the different places at least.

include/inviwo/core/datastructures/camera/cameratools.h
ResearchDaniel: Ok, but later you have createOrUpdate functions indicating that it might return null (i.e. not work)

src/core/properties/cameraproperty.cpp
petersteneteg: maybe it would be better to use xyz and then add some more buttons (?) for various view changes? (zoom+-, various rotations)

ResearchDaniel: Definitely! Rotate left/right/up/down, translate left/right (perhaps forward/backward) and zoom in/out

src/core/datastructures/camera/skewedperspectivecamera.cpp
```diff
@@ -113,11 +160,11 @@ mat4 SkewedPerspectiveCamera::calculateProjectionMatrix() const {
 
 void SkewedPerspectiveCamera::serialize(Serializer& s) const {
     Camera::serialize(s);
-    s.serialize("fovy", fovy_);
+    s.serialize("fov", fovy_);
```
ResearchDaniel, line 167: No. We have two field of views, one vertical and one horizontal (derived from the aspect ratio if I recall correctly). This is the vertical field of view. Could be better named

petersteneteg: only one is ever stored. the other one is never used, and if it would be, it would be computed with the vertical fov and aspect ratio. Renaming this makes it much easier to convert old workspaces since the vertical fov property has always been serialised as "fov" the camera it self has never been serialised before, so changing this here will not break anything.

petersteneteg:
How would I add my SGCT camera then? should that go in the core?

src/core/datastructures/camera/skewedperspectivecamera.cpp
ResearchDaniel: ok

ResearchDaniel:
> How would I add my SGCT camera then? should that go in the core?

I was not aware that we had more cameras. Probably not, I guess it depends on how specialized it is to SGCT.

petersteneteg:
that's the reason I went down this rabbit hole ;)

src/core/properties/cameraproperty.cpp
martinfalk: careful here! Zoom means you change the fov but not the position. You just ensure that an object at a certain distance, i.e. the look at, stays about the same size on the screen.
Moving the camera closer/further away is a dolly.

include/inviwo/core/datastructures/camera.h
petersteneteg: sketchup uses a default of 35°

martinfalk: A vertical fov of 38° is closer to human vision. Notice the extreme perspective distortions happening at 60 degrees. And I am with Peter here that visualization is _not_ the same as games. In games you want to see as much of your surroundings as possible. Here we mostly have a tiny (square) canvas showing one dataset. Another thing to consider: to be more accurate one should actually use orthographic. But that messes with 3D perception of shapes/objects.

include/inviwo/core/datastructures/camera/camera.h
```diff
@@ -0,0 +1,200 @@
...
+
+namespace inviwo {
+
+namespace cameradefaults {
+constexpr vec3 lookFrom = vec3(0.0f, 0.0f, 2.0f);
+constexpr vec3 lookTo = vec3(0.0f);
+constexpr vec3 lookUp = vec3(0.0f, 1.0f, 0.0f);
+constexpr float nearPlane = 0.1f;
+constexpr float farPlane = 1000.0f;
+constexpr float fieldOfView = 38.f;
```
ResearchDaniel, line 45: Ok, but I think that we should state why we choose this

include/inviwo/core/datastructures/camera.h
petersteneteg: and orthographic "zoom" (changing width) now also works with the mouse

src/core/properties/cameraproperty.cpp
petersteneteg: on the other hand we really don't want to zoom using FOV since that will also be quite confusing in many cases...

include/inviwo/core/datastructures/camera/camera.h
```diff
@@ -0,0 +1,200 @@
...
+    const vec3& getLookTo() const;
+    virtual void setLookTo(vec3 val);
+
+    const vec3& getLookUp() const;
+    virtual void setLookUp(vec3 val);
+
+    virtual float getAspectRatio() const;
+    virtual void setAspectRatio(float val);
+
+    virtual void zoom(float factor, std::optional<mat4> boundingBox) = 0;
```
ResearchDaniel, line 96: Should this functionality be in the camera class? It breaks the single responsibility principle in my opinion. Is it possible to move this the camerautil?

ResearchDaniel (approved):
Some unresolved comments remaining that are “nice to have”

include/inviwo/core/datastructures/camera/camera.h
petersteneteg: problem in that different cameras "zoom" differently, our perspective camera does a "dolly" moving closer to the look to point, where as a orthtographic camera changes it viewport width. Hence it is not so easy having it in a util class...

include/inviwo/core/datastructures/camera/camera.h
petersteneteg: as of now this selected to break as few regressions as possible. Might add do a different PR with update to just these values...

PR #947, opened 2020-06-02
NetworkVisitor

petersteneteg:
Core: Added support of network visitation in the form of a NetworkVisitor that visits processors and properties

### DIFF
--- a/include/inviwo/core/network/networkvisitor.h
+++ b/include/inviwo/core/network/networkvisitor.h
@@ -38,27 +38,66 @@ class Processor;
 class PropertyOwner;
 class Property;
 class CompositeProperty;
+class CanvasProcessor;
 
 /**
- * \brief A vistor base for visiting an inviwo network
+ * @brief Visitor pattern base for visiting an Inviwo ProcessorNetwork
  */
 class IVW_CORE_API NetworkVisitor {
 public:
     virtual ~NetworkVisitor() = default;
 
+    /**
+     * @brief Visit a Processor
+     * Adding and removing processors while visiting are not supported
+     * @return visit all child properties if true else go to next processor
+     */
     virtual bool visit(Processor&) { return true; }
-    virtual bool visit(PropertyOwner&) { return true; }
-    virtual bool visit(Property&) { return true; }
+
+    /**
+     * @brief Visit a CanvasProcessor
+     * Adding and removing processors while visiting are not supported
+     * @return visit all child properties if true else go to next processor
+     */
+    virtual bool visit(CanvasProcessor&) { return true; }
+
+    /**
+     * @brief Visit a CompositeProperty
+     * Adding and removing properties while visiting are not supported
+     * @return visit all child properties if true else go to next CompositeProperty
+     */
     virtual bool visit(CompositeProperty&) { return true; }
+
+    /**
+     * @brief Visit a Property
+     * Adding and removing properties while visiting are not supported
+     * @return not used for Properties
+     */
+    virtual bool visit(Property&) { return true; }
 };
 
+/**
+ * @brief A Helper class to construct a NetworkVisitor from a set of lambda functions
+ *
+ * ```{.cpp}
+ * LambdaNetworkVisitor visitor{[](Property& property) {
+ *                                  // Do something for a properties
+ *                                  return true;
+ *                              },
+ *                              [](Processor&) {
+ *                                  // Do something for a processor
+ *                                  return true;
+ *                              }};
+ * processorNetwork->accept(visitor);
+ * ```
+ */
 template <typename... Funcs>
 struct LambdaNetworkVisitor : NetworkVisitor, Funcs... {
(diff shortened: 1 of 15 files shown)


include/inviwo/core/network/networkvisitor.h
```diff
@@ -0,0 +1,70 @@
...
+    virtual ~NetworkVisitor() = default;
+
+    virtual bool visit(Processor&) { return true; }
+    virtual bool visit(PropertyOwner&) { return true; }
+    virtual bool visit(Property&) { return true; }
+    virtual bool visit(CompositeProperty&) { return true; }
+};
+
+template <typename... Funcs>
+struct LambdaNetworkVisitor : NetworkVisitor, Funcs... {
```
ResearchDaniel, line 95: This sounds really convenient, but I have no idea how to use it :D
A hint of an example or short description would be helpful

include/inviwo/core/network/networkvisitor.h
```diff
@@ -0,0 +1,70 @@
...
+class CompositeProperty;
+
+/**
+ * \brief A vistor base for visiting an inviwo network
+ */
+class IVW_CORE_API NetworkVisitor {
+public:
+    virtual ~NetworkVisitor() = default;
+
+    virtual bool visit(Processor&) { return true; }
```
ResearchDaniel, line 55: Specify what happens when returning true/false

include/inviwo/core/network/processornetwork.h
```diff
@@ -249,6 +250,8 @@ class IVW_CORE_API ProcessorNetwork : public Serializable,
      */
     void clear();
 
+    void accept(NetworkVisitor& visor);
```
ResearchDaniel, line 257: Should add some details here about the order. Both in terms of network topology and also the order of the visit functions ( first processor?) 
What happens with the classes deriving from multiple ones, e.g Processor is a PropertyOwner
A short description or link to the visitor pattern would also help newcomers

include/inviwo/core/network/networkvisitor.h
```diff
@@ -0,0 +1,70 @@
...
+
+namespace inviwo {
+
+class Processor;
+class PropertyOwner;
+class Property;
+class CompositeProperty;
+
+/**
+ * \brief A vistor base for visiting an inviwo network
```
ResearchDaniel: Visitor pattern base for...

ResearchDaniel (changes requested):
Looks good, but I think that each accept method could explain what it is doing. 
Also, the behavior with respect to for example Processor/PropertyOwner is not clear. Will for example PropertyOwner ever be called if I visit a network or is it just Processor/Property? Same goes for CompositeProperty I guess

include/inviwo/core/network/networkvisitor.h
petersteneteg: added an example

include/inviwo/core/network/networkvisitor.h
petersteneteg: done

include/inviwo/core/network/processornetwork.h
petersteneteg: generally for a visitor pattern you will always get the most the overload for the most derived type.

petersteneteg: The PropertyOwner would not be used in most scenarios, but wanted it there for completeness.

ResearchDaniel: Come to think of it we probably also want to state if adding/removing elements (e.g. properties) are allowed during traversal.

include/inviwo/core/properties/propertyowner.h
```diff
@@ -163,18 +161,22 @@ class IVW_CORE_API PropertyOwner : public PropertyOwnerObservable,
...
-                                   subProperties.end());
+
+    for (auto* property : properties_) {
+        if (auto p = dynamic_cast<T*>(property)) {
+            foundProperties.push_back(p);
+        }
+    }
+
+    if (recursiveSearch) {
+        for (auto* copy : compositeProperties_) {
```
ResearchDaniel, line 174: Copy?

src/core/properties/propertyowner.cpp
```diff
@@ -170,35 +171,33 @@ std::vector<Property*> PropertyOwner::getPropertiesRecursive() const {
...
-            if (comp) {
-                property = comp->getPropertyByIdentifier(path[i]);
-                if (!property) return nullptr;
-            } else {
-                return nullptr;
-            }
-            ++i;
+    if (path.empty()) return nullptr;
+
+    auto last = --path.end();
```
ResearchDaniel, line 188: path.back()
Ah, you mean lastIt

ResearchDaniel (approved):
I’m not sure if the CanvasProcessor should have special treatment in the visitor. convenient, yes, but Once you go down that road: what about other types of processors?

One would instead have to check that type in my opinion. Makes the interface a bit cleaner

petersteneteg:
see #952

PR #949, opened 2020-06-03
Feature/webbrowser refactoring

ResearchDaniel:
- Use one CefClient and RenderHandler for multiple browsers as it is originally designed for. The CefClient can now be accessed via the WebBrowserModule.

- Changed redirection of https://inviwo to api://inviwo to avoid confusion with the https scheme

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 Here we document changes that affect the public API or changes that needs to be communicated to other developers. 
 
+## 2020-06-03 WebBrowser Javascript API
+Changed redirection of `https://inviwo` to `api://inviwo` to avoid confusion with the https scheme.
+Your html-files will need to update from 
+'https://inviwo/modules/yourmodule' 
+to 
+'api://inviwo/modules/yourmodule' 
+
 ## 2020-04-03 OrdinalPropertyState
 Added a OrdinalPropertyState helper for constructing ordinal properties.
 And a factory function `util::ordinalColor` for OrdinalProperties representing Colors
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,11 @@
 Here we document changes that affect the public API or changes that needs to be communicated to other developers. 
 
 ## 2020-06-03 WebBrowser Javascript API
-Changed redirection of `https://inviwo` to `api://inviwo` to avoid confusion with the https scheme.
+Changed redirection of `https://inviwo` to `inviwo://` to avoid confusion with the https scheme.
 Your html-files will need to update from 
 'https://inviwo/modules/yourmodule' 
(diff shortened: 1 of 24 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 Here we document changes that affect the public API or changes that needs to be communicated to other developers. 
 
+## 2020-06-03 WebBrowser Javascript API
+Changed redirection of `https://inviwo` to `api://inviwo` to avoid confusion with the https scheme.
+Your html-files will need to update from 
+'https://inviwo/modules/yourmodule' 
+to 
+'api://inviwo/modules/yourmodule'
```
petersteneteg: is api recommended practice? why not use inviwo:// instead? 
also not that new modules don't use the modules prefix in the path but rather inviwo or some other name specific t the authors. Hence for new core modules you would get api://inviwo/inviwo/newmodule
if it is possible I think inviwo://mymodule/... would be best? why do you even need the "module" in there? should not the module id be enough?

ResearchDaniel: good idea!

ResearchDaniel: we have both app/ and modules/, hence the separation

petersteneteg: would it be possible to just use core / app / module_identifier? Guess we could assume there are no module name app...

ResearchDaniel: fixed

modules/animation/src/animationmodule.cpp
```diff
@@ -82,6 +83,10 @@ struct OrdinalReghelper {
         trackRegHelper<PropertyType>(am);
         interpolationRegHelper<PropertyType, LinearInterpolation>(am);
         interpolationRegHelper<PropertyType, ConstantInterpolation>(am);
+        using PropertyRefType = OrdinalRefProperty<T>;
+        trackRegHelper<PropertyRefType>(am);
+        interpolationRegHelper<PropertyRefType, LinearInterpolation>(am);
```
petersteneteg, line 88: Nice

modules/webbrowser/include/modules/webbrowser/webbrowserclient.h
```diff
@@ -53,9 +54,9 @@ namespace inviwo {
 
 /* \class WebBrowserClient
  * CefClient with custom render handler and call redirections.
- * Calls to 'https://inviwo/modules/yourmodule' will be redirected to yourmodule
+ * Calls to 'api://inviwo/modules/yourmodule' will be redirected to yourmodule
```
petersteneteg: inviwo://

petersteneteg (dismissed):

petersteneteg (approved):

PR #951, opened 2020-06-07
OpenGLQt: Use OpenGLWidget

ResearchDaniel:
Replaces deprecated use of QGLWidget.
Tested on windows.
Will remove CanvasQGLWidget if everything works


modules/openglqt/src/canvasqopenglwidget.cpp
```diff
@@ -36,92 +36,94 @@
...
+    if (QOpenGLContext::globalShareContext()) {
         std::string preferProfile = OpenGLCapabilities::getPreferredProfile();
-        if (preferProfile == "core")
-            sharedFormat_.setProfile(QSurfaceFormat::CoreProfile);
-        else if (preferProfile == "compatibility")
-            sharedFormat_.setProfile(QSurfaceFormat::CompatibilityProfile);
+        auto gsc = QOpenGLContext::globalShareContext();
+        if (preferProfile == "core") {
+            gsc->format().setProfile(QSurfaceFormat::CoreProfile);
+        } else if (preferProfile == "compatibility") {
```
petersteneteg: should we remove compatibility?

modules/openglqt/src/openglqtmodule.cpp
```diff
@@ -67,7 +67,8 @@ OpenGLQtModule::OpenGLQtModule(InviwoApplication* app)
 
     // Create GL Context
     CanvasQt::defineDefaultContextFormat();
-    sharedCanvas_ = util::make_unique<CanvasQt>(size2_t(16, 16), "Default");
+
+    sharedCanvas_ = util::make_unique<CanvasQt>(nullptr, size2_t(16, 16), "Default");
```
petersteneteg: would it not be possible to use the auto QOpenGLContext::globalShareContext(); here instead?

modules/opengl/src/canvasgl.cpp
```diff
@@ -95,7 +96,12 @@ void CanvasGL::renderLayer(size_t idx) {
 bool CanvasGL::ready() {
     if (ready_) {
         return true;
-    } else if (glCheckFramebufferStatus(GL_FRAMEBUFFER) == GL_FRAMEBUFFER_COMPLETE) {
+    } else if (glCheckFramebufferStatus(GL_FRAMEBUFFER) == GL_FRAMEBUFFER_COMPLETE &&
+               // Capabilities might not have been registered yet in (OpenGLQt) module constructor
+               ShaderManager::getPtr()
```
petersteneteg: if we use the QOpenGLContext::globalShareContext() the OpenGLQtModule then this should never happen?

petersteneteg (dismissed):

modules/openglqt/src/openglqtmodule.cpp
ResearchDaniel: Perhaps, if we have a QSurface for it. Have not looked into it but this might be a way to go:
https://doc.qt.io/qt-5/qoffscreensurface.html

petersteneteg: we don't need any actual fbo or anything here, just a global context.. the canvas is only there since that's usually the only way to get to that context...

modules/openglqt/include/modules/openglqt/hiddencanvasqt.h
```diff
@@ -31,35 +31,68 @@
...
+/*
+ * Convenience class for creating an QOffscreenSurface with an QOpenGLContext.
+ * 
+ * This class can be used for concurrent OpenGL operations in background threads.
+ * The class must be created in the main thread. 
+ * initializeGL must be called before use, but can be called in a different thread.
+ *
+ * @note Most Canvas overriden functions are non-functional except HiddenCanvasQt::activate()
+ */
+class IVW_MODULE_OPENGLQT_API HiddenCanvasQt : public Canvas {
```
petersteneteg, line 55: Some time ago I had to refactor the HiddenCanvas on the GLFW side to provide for sgct, to there is now a ContextHolder in RenderContext that is used for creating context in the backgrounds threads, By default it will wrapper the old HiddenCanvas. If we now refactor the qt classes also we might not even need the hidden canvas anymore just a ContextHolderQt 

here is the default implementation
https://github.com/inviwo/inviwo/blob/0d24ee698c8f47da4e58ffbb93b6577a6aba2834/src/core/util/rendercontext.cpp#L35-L44

here is how it's done in glfw now
https://github.com/inviwo/inviwo/blob/0d24ee698c8f47da4e58ffbb93b6577a6aba2834/modules/glfw/src/glfwmodule.cpp#L47-L60

petersteneteg: Humm maybe it still would be needed... but one does not really need to return a Canvas from createHiddenCanvas an new ContextHolder would be enough in principle. If that could simplify stuff.

CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
 Here we document changes that affect the public API or changes that needs to be communicated to other developers. 
 
+## 2020-06-09 Remove use of deprecated QGLWidget
+Your main file need be updated to enable shared OpenGL context
```
petersteneteg: but this is only if you have your own apps right? might want to point that out.

petersteneteg (dismissed):

modules/openglqt/include/modules/openglqt/hiddencanvasqt.h
ResearchDaniel: Yes, having to use a Canvas for the context holder does not make much sense because now we must override a bunch of irrelevant functions

petersteneteg: With some minor modifications the the RenderContext class there should the no longer be any need for Canvases there I think

ResearchDaniel: Let's do another PR for that

petersteneteg (changes requested):
What did you do in git now? you have added bunch of commits from master in this PR? Just to a proper  rebase

ResearchDaniel:
I thought I did: git rebase master
I guess something went wrong in some step

PR #952, opened 2020-06-08
Network visitor 

petersteneteg:
rebase of #947

### DIFF
--- a/include/inviwo/core/network/lambdanetworkvisitor.h
+++ b/include/inviwo/core/network/lambdanetworkvisitor.h
@@ -108,5 +108,5 @@ struct LambdaNetworkVisitor : NetworkVisitor, Funcs... {
     }
 };
 template <class... Funcs>
-LambdaNetworkVisitor(Funcs...) -> LambdaNetworkVisitor<Funcs...>;
+LambdaNetworkVisitor(Funcs...)->LambdaNetworkVisitor<Funcs...>;
 }  // namespace inviwo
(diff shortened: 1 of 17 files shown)


petersteneteg:
@ResearchDaniel rebased this one...

include/inviwo/core/network/lambdanetworkvisitor.h
```diff
@@ -0,0 +1,112 @@
...
+        } else {
+            return true;
+        }
+    }
+
+    virtual bool visit([[maybe_unused]] Property& property) override {
+        if constexpr (util::is_detected_exact_v<bool, PropertyOverload, decltype(*this)>) {
+            return operator()(property);
+        } else {
+            return true;
```
ResearchDaniel, line 97: Should be false?

include/inviwo/core/properties/propertyowner.h
```diff
@@ -153,39 +154,31 @@ class IVW_CORE_API PropertyOwner : public PropertyOwnerObservable,
     // allocated on the heap.
     std::vector<std::unique_ptr<Property>> ownedProperties_;
 
+    void forEachProperty(std::function<void(Property&)> callback,
+                         bool recursiveSearch = false) const;
```
ResearchDaniel, line 158: Const?

ResearchDaniel (approved):
Check the comments

include/inviwo/core/properties/propertyowner.h
petersteneteg: all the other property getter are const...

include/inviwo/core/network/lambdanetworkvisitor.h
petersteneteg: we generally want to do a recursive traversal which is what the parameter signals, although for a property that is not used, it's only there to be consistent with the other ones...

PR #956, opened 2020-06-12
OpenGL perf and bug fixes

petersteneteg:


martinfalk (dismissed):

modules/opengl/src/image/imagegl.cpp
```diff
@@ -82,18 +82,18 @@ void ImageGL::reAttachAllLayers(ImageType type) {
 void ImageGL::activateBuffer(ImageType type) {
     frameBufferObject_.activate();
 
-    std::vector<GLenum> drawBuffers{frameBufferObject_.getDrawBuffers()};
-    if (!drawBuffers.empty()) {
-        GLsizei numBuffersToDrawTo = static_cast<GLsizei>(drawBuffers.size());
+    drawBuffers_ = frameBufferObject_.getDrawBuffers();
```
ResearchDaniel, line 85: I don't fully understand why there is a need to cache these? They are in the frame buffer already?

petersteneteg: We modify them so we have to make a copy

modules/basegl/src/datavisualizer/volumeraycastvisualizer.cpp
```diff
@@ -95,17 +96,17 @@ std::vector<Processor*> VolumeRaycastVisualizer::addVisualizerNetwork(Outport* o
     auto vbb = net->addProcessor(util::makeProcessor<VolumeBoundingBox>(GP{8, 3}));
     auto lrp = net->addProcessor(util::makeProcessor<LineRendererProcessor>(GP{8, 6}));
 
-    static_cast<FloatVec4Property*>(bak->getPropertyByIdentifier("bgColor1"))
+    dynamic_cast<FloatVec4Property*>(bak->getPropertyByIdentifier("bgColor1"))
```
ResearchDaniel, line 99: No reason to dynamic cast if not checking for null afterwards?

modules/opengl/src/image/imagegl.cpp
ResearchDaniel: But why cache them when they are copied every time? They do not seem to be used elsewhere in imagegl?

modules/basegl/src/datavisualizer/volumeraycastvisualizer.cpp
```diff
@@ -95,17 +96,17 @@ std::vector<Processor*> VolumeRaycastVisualizer::addVisualizerNetwork(Outport* o
...
 
     // set shading mode in volume raycaster to 'no shading'
-    static_cast<OptionPropertyInt*>(vrc->getPropertyByIdentifier("shadingMode", true))->set(0);
+    dynamic_cast<OptionPropertyInt*>(vrc->getPropertyByIdentifier("shadingMode", true))->set(0);
 
-    static_cast<FloatProperty*>(lrp->getPropertyByPath({{"lineSettings"}, {"lineWidth"}}))
+    dynamic_cast<FloatProperty*>(lrp->getPropertyByPath({{"lineSettings"}, {"lineWidth"}}))
         ->set(1.5f);
-    static_cast<FloatVec3Property*>(vrc->getPropertyByIdentifier("lookFrom", true))
+    dynamic_cast<FloatVec3RefProperty*>(vrc->getPropertyByIdentifier("lookFrom", true))
```
ResearchDaniel, line 109: We should have a common interface for these if we don’t already

ResearchDaniel (approved):
There are some minor comments that might need to be addressed

modules/basegl/src/datavisualizer/volumeraycastvisualizer.cpp
petersteneteg: a dynamic cast will fail with a crash, a static cast will just corrupt data and go on...

petersteneteg: i.e. it is much easier do debug a dynamic cast failing

modules/basegl/src/datavisualizer/volumeraycastvisualizer.cpp
petersteneteg: what do you mean? an interface for what?

modules/opengl/src/image/imagegl.cpp
petersteneteg: I want to avoid having to allocate that array every time... that new actually showed up on some of my profiles...

modules/basegl/src/datavisualizer/volumeraycastvisualizer.cpp
petersteneteg: I don't have access to the camera in the VolumeRaycaster it is a protected member...

ResearchDaniel: For OrdinalProperty/OrdinalRefProperty in this case it does not make sense to need to separate between the two

modules/opengl/src/image/imagegl.cpp
ResearchDaniel: Ok, the workaround should be documented because it is hard to know and could be removed later otherwise

modules/basegl/src/datavisualizer/volumeraycastvisualizer.cpp
petersteneteg: They don't have identical interfaces, sometime I guess one could get away with using a common interface, but then you would still need to cast to that one, and I don't think we do that very often anyway.

PR #957, opened 2020-06-12
Fixing fragment renderer resize

### LINKED ISSUE #955: Crash in Rasterization Renderer when disconnecting optional background inport
Probably caused by `imageInport_->getData()` return nullptr in the code refrenced below: 

https://github.com/inviwo/inviwo/blob/d1749ba220ba089cc78391da222461e38ad40bd4/modules/meshrenderinggl/src/processors/rasterizationrenderer.cpp#L114-L115

AnkeAnke:
Fixing rasterization renderer crash on background removal.

### DIFF
--- a/modules/meshrenderinggl/src/processors/rasterizationrenderer.cpp
+++ b/modules/meshrenderinggl/src/processors/rasterizationrenderer.cpp
@@ -111,8 +111,12 @@ RasterizationRenderer::RasterizationRenderer()
 
     flrReload_ = flr_.onReload([this]() { invalidate(InvalidationLevel::InvalidResources); });
 
-    imageInport_->onChange(
-        [this]() { intermediateImage_.setDimensions(imageInport_->getData()->getDimensions()); });
+    imageInport_->onChange([this]() {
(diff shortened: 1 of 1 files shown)


r-englund (dismissed):

modules/meshrenderinggl/src/processors/rasterizationrenderer.cpp
```diff
@@ -111,8 +111,12 @@ RasterizationRenderer::RasterizationRenderer()
 
     flrReload_ = flr_.onReload([this]() { invalidate(InvalidationLevel::InvalidResources); });
 
-    imageInport_->onChange(
-        [this]() { intermediateImage_.setDimensions(imageInport_->getData()->getDimensions()); });
+    imageInport_->onChange([this]() {
+        if (imageInport_->hasData())
```
petersteneteg: We always use braces with ifs...

petersteneteg: thanks, that makes me happy :)

petersteneteg (approved):

PR #960, opened 2020-06-17
Core: StringProperty: Keep linebreaks in tooltips

r-englund:
Before fix:
image
After fix:
image

### DIFF
--- a/src/core/properties/stringproperty.cpp
+++ b/src/core/properties/stringproperty.cpp
@@ -51,7 +51,7 @@ Document StringProperty::getDescription() const {
     using P = Document::PathComponent;
     Document doc = TemplateProperty<std::string>::getDescription();
     auto b = doc.get({P("html"), P("body")});
-    auto pre = b.append("pre", value_.value, {{"style","font"}});
+    auto pre = b.append("pre", value_.value);
     return doc;
 }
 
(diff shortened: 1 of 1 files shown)


petersteneteg (dismissed):
I would have put the string in a &lt;pre> element instead, but I guess this works to

r-englund:
&lt;pre> changes the font to a monospace font, in the example above where there is code in the string property that might make sense, but in other cases it feels strange, for example
image

r-englund:
Hrm, ´pre´ might still be preferable, since it keeps extra whitespaces etc. (Code indentation for the if-statement is missing in the example above)

r-englund:
image

martinfalk:
That's a long tooltip ;-) Might cause some problems on smaller screen resolutions in general.

src/core/properties/stringproperty.cpp
```diff
@@ -51,7 +51,9 @@ Document StringProperty::getDescription() const {
     using P = Document::PathComponent;
     Document doc = TemplateProperty<std::string>::getDescription();
     auto b = doc.get({P("html"), P("body")});
-    b.append("p", value_.value);
+    auto val = value_.value;
+    replaceInString(val, "\n", "<br \>\n");
```
martinfalk: should read `<br />` !

martinfalk (changes requested):

martinfalk (dismissed):

PR #963, opened 2020-06-17
Fancy mesh rendering examples

AnkeAnke:
Adding minimum alpha setting for Martin.

Adding example workspaces for mesh rasterization:

Different alpha settings (opaque, uniform, angle-based, density-based/smoke surface)
exAlphas
 
Geometry contouring (edges, silhouettes, illustration and hatching)
exContours

Multiple mesh rendering (compositing with other renderers, transforming rasterizations)
exCow

Composite renderings (as for paper renderings)
exComposites


modules/meshrenderinggl/glsl/fancymeshrenderer.frag
```diff
@@ -187,27 +188,31 @@ vec4 performShading() {
...
 #ifdef ALPHA_ANGLE_BASED
         alpha *= pow(acos(angle) * 2 / M_PI, alphaSettings.angleExp);
+        alpha += alphaSettings.minAlpha;
 #endif
 #ifdef ALPHA_NORMAL_VARIATION
         float nv_dzi = dFdxFinest(normal.z);
         float nv_dzj = dFdyFinest(normal.z);
         float nv_curvature = min(1, nv_dzi * nv_dzi + nv_dzj * nv_dzj + 0.0000001);
         alpha *= min(1, pow(nv_curvature, alphaSettings.normalExp * 0.5) * 10);
+        alpha += alphaSettings.minAlpha;
```
petersteneteg, line 202: will this not give a max of 1+minAlpha? there are some off constants in this code :)

modules/meshrenderinggl/src/rendering/fragmentlistrenderer.cpp
```diff
@@ -146,9 +146,6 @@ bool FragmentListRenderer::postPass(bool useIllustration, const Image* backgroun
     // check if enough space was available
     if (numFrags > fragmentSize_) {
         // we have to resize the fragment storage buffer
-        LogInfo("fragment lists resolved, pixels drawn: "
```
petersteneteg, line 149: maybe we could expose the memory usage as a readonly property or something? other PR probably...

petersteneteg (changes requested):

modules/meshrenderinggl/src/rendering/fragmentlistrenderer.cpp
AnkeAnke: Maybe. On the other hand I never experienced any memory overflow on a 4K screen, so hopefully the error will never be a problem.

modules/meshrenderinggl/glsl/fancymeshrenderer.frag
AnkeAnke: Very true.

modules/meshrenderinggl/src/rendering/fragmentlistrenderer.cpp
petersteneteg: When i rewrote that code I did some tests, and as long as you only use one, renderer it not to bad. but If you have a couple of renderers and use illustration buffers, then you easily use several GBs of memory...

petersteneteg (approved):

martinfalk:
Nice!

PR #965, opened 2020-06-18
Fixing gcc/clang warnings

AnkeAnke:
Mostly unused variable and redundant move.
Export of specialized template classes failed on clang (silently) gcc.
Tested on gcc-7/Ubuntu 20 (Pop! OS) and MSVC 14/Windows 10 (and clang/jenkins).


petersteneteg (dismissed):

modules/dataframe/src/datastructures/dataframe.cpp
```diff
@@ -161,6 +161,19 @@ DataFrame::DataFrame(const DataFrame &df) {
     }
 }
 
+DataFrame &DataFrame::operator=(const DataFrame &df) {
+    for (const auto &col : df.columns_) {
```
r-englund: Need to `clear` `columns_` on `this` first. 
Also, need to check such that `this != *df`

modules/dataframe/src/datastructures/dataframe.cpp
```diff
@@ -161,6 +161,19 @@ DataFrame::DataFrame(const DataFrame &df) {
...
+        columns_.emplace_back(col->clone());
+    }
+    return *this;
+}
+
+DataFrame &DataFrame::operator=(const DataFrame &&df) {
```
r-englund: move ctor and assign-op take non-const input or else you cant move out data from the rhs

r-englund: Also, I think the move ctor and assign-op could use `= default`

r-englund (changes requested):

modules/dataframe/src/datastructures/dataframe.cpp
```diff
@@ -162,6 +162,8 @@ DataFrame::DataFrame(const DataFrame &df) {
 }
 
 DataFrame &DataFrame::operator=(const DataFrame &df) {
+    if (this == &df) return;
```
petersteneteg: should always return *this...

modules/dataframe/src/datastructures/dataframe.cpp
AnkeAnke: Fixed the first. Movign the std vector is more efficient than copying, thus the move overrides.

r-englund (dismissed):

petersteneteg (approved):

PR #967, opened 2020-06-18
Cmake refactor deps

petersteneteg:


apps/inviwo/packaging/packaging.cmake
```diff
@@ -103,15 +101,8 @@ else()
...
-if(NOT APPLE)
-    install(DIRECTORY ${IVW_ROOT_DIR}/data/images           DESTINATION data      COMPONENT images)
-    install(DIRECTORY ${IVW_ROOT_DIR}/data/scripts          DESTINATION data      COMPONENT scripts)
-    install(DIRECTORY ${IVW_ROOT_DIR}/data/volumes          DESTINATION data      COMPONENT volumes)
-    install(DIRECTORY ${IVW_ROOT_DIR}/data/workspaces       DESTINATION data      COMPONENT workspaces)
-
-    install(DIRECTORY ${IVW_ROOT_DIR}/tests/images          DESTINATION tests     COMPONENT images)
-    install(DIRECTORY ${IVW_ROOT_DIR}/tests/volumes         DESTINATION tests     COMPONENT volumes)
-endif()
+install(DIRECTORY ${IVW_ROOT_DIR}/data/  DESTINATION ${IVW_RESOURCE_INSTALL_PREFIX}data  COMPONENT Datasets)
```
ResearchDaniel, line 104: This can be other things than datasets such as scripts and transfer functions?

CMakeLists.txt
```diff
@@ -53,18 +47,27 @@ endwhile()
...
 
 #Verify that git submodules has been cloned
 include(cmake/verifysubmodules.cmake)
 verify_submodules("${CMAKE_CURRENT_LIST_DIR}/.gitmodules")
 
 option(BUILD_SHARED_LIBS "Build shared libs, else static libs" ON)
 
-# Needed for find_package(PythonLibsNew 3), which is an improved version of CMake shipped find_pythonLibs
-set(CMAKE_MODULE_PATH "${CMAKE_BINARY_DIR}/pkg/;${CMAKE_MODULE_PATH};${CMAKE_SOURCE_DIR}/ext/pybind11/tools")
+file(REMOVE_RECURSE ${CMAKE_BINARY_DIR}/pkg)
```
ResearchDaniel, line 64: This seems a bit off, why remove?

CMakeLists.txt
```diff
@@ -53,18 +47,27 @@ endwhile()
...
+set(CMAKE_FIND_PACKAGE_PREFER_CONFIG ON)
+
+list(APPEND CMAKE_MODULE_PATH 
+    ${PROJECT_SOURCE_DIR}/support/cmake/modules
```
ResearchDaniel, line 68: support/cmake ? Is it moved?

CMakeLists.txt
```diff
@@ -75,30 +78,37 @@ set(CMAKE_MODULE_PATH "${CMAKE_BINARY_DIR}/pkg/;${CMAKE_MODULE_PATH};${CMAKE_SOU
...
-option(IVW_TINY_GLFW_APPLICATION "Build Inviwo Tiny GLFW Application" OFF)
-option(IVW_TINY_QT_APPLICATION   "Build Inviwo Tiny QT Application" OFF)
-
-if(IVW_QT_APPLICATION AND NOT IVW_QT_APPLICATION_BASE)
-    set(IVW_QT_APPLICATION_BASE ON CACHE BOOL "Build base for qt applications. \
-            Used by Qt network editor and Qt minimal application" FORCE)
-    message(STATUS "IVW_QT_APPLICATION_BASE was set to build, due to dependency towards IVW_QT_APPLICATION")
+option(IVW_APP_QTBASE       "Build base for Qt apps. Used by the Qt network editor and the Tiny Qt app" ON)
+option(IVW_APP_INVIWO       "Build Inviwo Qt network editor application" ON)
+option(IVW_APP_MINIMAL_GLFW "Build Inviwo Tiny GLFW Application" OFF)
```
ResearchDaniel, line 92: Can these be moved to respective app? It would cleanup this file a bit and make it easier for people to add their own

CMakeLists.txt
```diff
@@ -75,30 +78,37 @@ set(CMAKE_MODULE_PATH "${CMAKE_BINARY_DIR}/pkg/;${CMAKE_MODULE_PATH};${CMAKE_SOU
...
+        "Build base for qt applications. Used by Qt network editor and Qt minimal application" 
+        FORCE
+    )
+    message(STATUS "IVW_APP_QTBASE was set to build, due to dependencies")
 endif()
-ivw_enable_modules_if(IVW_QT_APPLICATION QtWidgets)
-ivw_enable_modules_if(IVW_INTEGRATION_TESTS GLFW Base)
-ivw_enable_modules_if(IVW_TINY_GLFW_APPLICATION GLFW)
+
+ivw_enable_modules_if(IVW_APP_INVIWO QtWidgets)
```
ResearchDaniel, line 104: Same here, can these be moved to respective app?

CMakeLists.txt
```diff
@@ -75,30 +78,37 @@ set(CMAKE_MODULE_PATH "${CMAKE_BINARY_DIR}/pkg/;${CMAKE_MODULE_PATH};${CMAKE_SOU
...
+ivw_enable_modules_if(IVW_APP_MINIMAL_QT QtWidgets)
+ivw_enable_modules_if(IVW_APP_MINIMAL_GLFW GLFW)
+ivw_enable_modules_if(IVW_APP_INVIWO_DOME SGCT)
+ivw_enable_modules_if(IVW_APP_PYTHON Python3 Python3Qt QtWidgets)
+
+option(IVW_TEST_INTEGRATION_TESTS "Build inviwo integration test" ON)
```
ResearchDaniel, line 110: I think that IVW_TESTS would be better. TEST can be interpreted as unfinished stuff (e.g, we are testing some new functionality).
This would then be IVW_TESTS_INTEGRATION

CMakeLists.txt
```diff
@@ -112,52 +122,19 @@ endif()
...
-
-option (IVW_USE_EXTERNAL_IMG "Link with external image libraries (zlib, png, jpeg, tiff) instead of building them." OFF)
-if (NOT IVW_USE_EXTERNAL_IMG)
-  add_subdirectory(ext/zlib-1.2.11) # Defines target ZLIB::ZLIB
-  add_subdirectory(ext/libpng)      # Defines target PNG::PNG
-  add_subdirectory(ext/libjpeg)     # Defines target JPEG::JPEG
-  add_subdirectory(ext/tiff)        # Defines target TIFF::TIFF
+add_subdirectory(ext)                        # Add external libraries.
+if(IVW_TEST_INTEGRATION_TESTS OR IVW_TEST_UNIT_TESTS)
+    add_subdirectory(tests/testutil)         # Add test utils
```
ResearchDaniel, line 127: Can this be moved to tests?

CMakeLists.txt
```diff
@@ -112,52 +122,19 @@ endif()
...
+endif()                    
+ivw_register_modules(all_modules)            # Add modules
+add_subdirectory(tools/meta)                 # Add Inviwo meta
+add_subdirectory(src/qt)                     # Add Qt
+add_subdirectory(apps)                       # Add applications
+ivw_add_external_projects()                  # Add external projects
+if(IVW_TEST_INTEGRATION_TESTS)
```
ResearchDaniel, line 134: How about adding the tests directory and doing this logic there?

CMakeLists.txt
```diff
@@ -112,52 +122,19 @@ endif()
...
-  add_subdirectory(ext/zlib-1.2.11) # Defines target ZLIB::ZLIB
-  add_subdirectory(ext/libpng)      # Defines target PNG::PNG
-  add_subdirectory(ext/libjpeg)     # Defines target JPEG::JPEG
-  add_subdirectory(ext/tiff)        # Defines target TIFF::TIFF
+add_subdirectory(ext)                        # Add external libraries.
+if(IVW_TEST_INTEGRATION_TESTS OR IVW_TEST_UNIT_TESTS)
+    add_subdirectory(tests/testutil)         # Add test utils
+endif()                    
+ivw_register_modules(all_modules)            # Add modules
+add_subdirectory(tools/meta)                 # Add Inviwo meta
```
ResearchDaniel, line 130: These comments are not very helpful :) 
Not part of this PR but you could consider changing or removing them

apps/inviwo/CMakeLists.txt
```diff
@@ -84,47 +83,23 @@ elseif(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
         MACOSX_BUNDLE_COPYRIGHT "C) 2012-${YEAR} Inviwo Foundation"
     )
     #Following http://www.cmake.org/Wiki/BundleUtilitiesExample
-    install(
```
ResearchDaniel, line 87: Can this be removed?

cmake/compileoptions.cmake
```diff
@@ -62,7 +62,7 @@ function(ivw_define_standard_properties)
             list(APPEND comp_opts "/wd4251") # needs dll-interface   https://msdn.microsoft.com/en-us/library/esew7y1w.aspx
             list(APPEND comp_opts "/wd4505") # unreferenced funtion  https://msdn.microsoft.com/en-us/library/mt694070.aspx
             list(APPEND comp_opts "/w35038") # class member reorder
-            if(NOT IVW_OPENMP_ON)
+            if(NOT IVW_USE_OPENMP)
```
ResearchDaniel, line 65: IVW_CFG_USE_OPENMP

cmake/installutils.cmake
```diff
@@ -27,9 +27,14 @@
 # 
 #################################################################################
 
+# Components
+# Application   All the executables and dll etc needed to run the att
```
ResearchDaniel, line 31: att->app

cmake/installutils.cmake
```diff
@@ -27,9 +27,14 @@
...
+# Development   Lib files, pdb files, headers, 
+# Datasets      General datasets i.e. boron etc.
```
ResearchDaniel, line 33: Also other things

ResearchDaniel:
Had some comments that probably apply to follow up PR

ResearchDaniel:
@petersteneteg can you see them?

petersteneteg:
Can not find any :(

ResearchDaniel (commented):
Start of review before the PR closed :)

apps/inviwo/packaging/packaging.cmake
petersteneteg: Yes, but why are they in the "data" folder then?

CMakeLists.txt
petersteneteg: I need to remove the old ones, if one is no longer building them

CMakeLists.txt
petersteneteg: tried to figure out where that one came from, should be removed... good find

CMakeLists.txt
petersteneteg: no we need them here to enable the modules for them (ivw_enable_modules_if) The alternative could maybe be that we just asset in the app that the needed modules are enabled and then you have to manually add them.
Which is more or less what you would do for you own app at the moment, but it would make the regular use case a bit less easy...

CMakeLists.txt
petersteneteg: that would not work

CMakeLists.txt
petersteneteg: Would you do IVW_TESTS_UNIT then? that does not seem right to me

CMakeLists.txt
petersteneteg: I prefer having the check on the "outside" of the add_subdirectory. Think that makes it easier to understand.

cmake/compileoptions.cmake
petersteneteg: We have IVW_USE_* for enabling various dependences, like sigar, openmp, openexr, and IVW_USE_EXTERNAL_* for defining where to get it

cmake/installutils.cmake
petersteneteg: true

cmake/installutils.cmake
petersteneteg: fixed

apps/inviwo/CMakeLists.txt
petersteneteg: ivw_default_install_targets(inviwo) does that for all platforms

CMakeLists.txt
petersteneteg: the testutils has to be added before the modules and the integration tests after the modules...

CMakeLists.txt
petersteneteg: true

PR #974, opened 2020-06-25
Feature/base python bindings

r-englund:


petersteneteg (commented):
Cool remember that both c++ and cmakefiles should end with an empty line

r-englund:
> 
> 
> Cool remember that both c++ and cmakefiles should end with an empty line

Saw that, will be in next commit :)

modules/base/bindings/include/modules/base/pythonbindings/io/volumewriting.h
```diff
@@ -0,0 +1,39 @@
...
+#pragma once
+
+#include <warn/push>
+#include <warn/ignore/shadow>
+#include <pybind11/pybind11.h>
+#include <warn/pop>
+
+namespace inviwo {
+
+void exposeVolumeWriteMehtods(pybind11::module& m);
```
ResearchDaniel: Typo

modules/base/bindings/src/basebindings.cpp
```diff
@@ -0,0 +1,60 @@
...
+#ifdef IVW_ENABLE_MSVC_MEM_LEAK_TEST
+    VLDDisable();
+#endif
+
+    py::module::import("inviwopy");
+
+    m.doc() = "Gives python scripts access to inviwo functionality in base";
+
+    auto ioMod = m.def_submodule("io", "Exposes various input/output functions");
+    auto utilMod = m.def_submodule("util", "Exposes various util functions");
```
ResearchDaniel, line 50: Do we have or can we make a convention for these? Should we follow the category in the processor or something?
io is good but if we can avoid util it would be good

modules/base/bindings/src/io/volumewriting.cpp
```diff
@@ -0,0 +1,60 @@
...
+#include <warn/ignore/shadow>
+#include <pybind11/pybind11.h>
+#include <pybind11/stl.h>
+#include <warn/pop>
+
+namespace inviwo {
+
+void exposeVolumeWriteMehtods(pybind11::module& m) {
+
+    m.def("saveDatVolume", &util::writeDatVolume);
```
ResearchDaniel, line 46: This could also use the volume writer factory and then be more general saveVolume(....) and saveVolumeSequence

ResearchDaniel (dismissed):

modules/base/bindings/src/io/volumewriting.cpp
petersteneteg: guess one can expose that factory in core?

r-englund: The factories require an application and modules to be registered to work, these functions work without an application. But true, we could add that in addition, when an application is avaiable.

modules/base/bindings/src/basebindings.cpp
r-englund: Maybe change to something like based on `Volume Operation`

modules/base/bindings/src/io/volumewriting.cpp
petersteneteg: think both are useful

r-englund:
rebased, see #1056

PR #976, opened 2020-06-30
Feature/dataframejoin

martinfalk:
utility functions and processor for joining two DataFrames

### DIFF
--- a/modules/dataframe/include/inviwo/dataframe/datastructures/column.h
+++ b/modules/dataframe/include/inviwo/dataframe/datastructures/column.h
@@ -46,7 +46,7 @@ class BufferBase;
 
 class IVW_MODULE_DATAFRAME_API InvalidConversion : public Exception {
 public:
-    InvalidConversion(const std::string &message = "",
+    InvalidConversion(const std::string& message = "",
                       ExceptionContext context = ExceptionContext())
         : Exception(message, context) {}
     virtual ~InvalidConversion() throw() {}
@@ -60,12 +60,12 @@ class IVW_MODULE_DATAFRAME_API Column {
 public:
     virtual ~Column() = default;
 
-    virtual Column *clone() const = 0;
+    virtual Column* clone() const = 0;
 
-    virtual const std::string &getHeader() const = 0;
-    virtual void setHeader(const std::string &header) = 0;
+    virtual const std::string& getHeader() const = 0;
+    virtual void setHeader(const std::string& header) = 0;
 
-    virtual void add(const std::string &value) = 0;
+    virtual void add(const std::string& value) = 0;
 
     virtual std::shared_ptr<BufferBase> getBuffer() = 0;
     virtual std::shared_ptr<const BufferBase> getBuffer() const = 0;
@@ -94,33 +94,33 @@ class TemplateColumn : public Column {
 public:
     using type = T;
 
-    TemplateColumn(const std::string &header,
+    TemplateColumn(const std::string& header,
                    std::shared_ptr<Buffer<T>> buffer = std::make_shared<Buffer<T>>());
 
-    TemplateColumn(const std::string &header, std::vector<T> data);
+    TemplateColumn(const std::string& header, std::vector<T> data);
 
-    TemplateColumn(const TemplateColumn<T> &rhs);
-    TemplateColumn(TemplateColumn<T> &&rhs);
+    TemplateColumn(const TemplateColumn<T>& rhs);
+    TemplateColumn(TemplateColumn<T>&& rhs);
 
-    TemplateColumn<T> &operator=(const TemplateColumn<T> &rhs);
-    TemplateColumn<T> &operator=(TemplateColumn<T> &&rhs);
+    TemplateColumn<T>& operator=(const TemplateColumn<T>& rhs);
+    TemplateColumn<T>& operator=(TemplateColumn<T>&& rhs);
 
-    virtual TemplateColumn *clone() const override;
+    virtual TemplateColumn* clone() const override;
 
     virtual ~TemplateColumn() = default;
 
-    virtual const std::string &getHeader() const override;
-    void setHeader(const std::string &header) override;
+    virtual const std::string& getHeader() const override;
+    void setHeader(const std::string& header) override;
 
-    virtual void add(const T &value);
+    virtual void add(const T& value);
     /**
      * \brief converts given value to type T, which is added to the column
      *
      * @param value
      * @throws InvalidConversion if the value cannot be converted to T
      */
-    virtual void add(const std::string &value) override;
-    virtual void set(size_t idx, const T &value);
+    virtual void add(const std::string& value) override;
+    virtual void set(size_t idx, const T& value);
 
     T get(size_t idx) const;
     T operator[](size_t idx) const;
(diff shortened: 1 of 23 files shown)


petersteneteg (dismissed):

modules/dataframe/include/inviwo/dataframe/datastructures/column.h
```diff
@@ -94,33 +94,42 @@ class TemplateColumn : public Column {
...
-    virtual void set(size_t idx, const T &value);
+    virtual void add(const std::string& value) override;
+
+    /**
+     * \brief appends all rows from column \p col
+     *
+     * @param col
+     * @throws Exception if data format does not match
+     */
+    virtual void append(const Column& col) override;
```
ResearchDaniel, line 133: Documentation should be in base class and copied here using doxygen command?

modules/dataframe/include/inviwo/dataframe/util/dataframeutil.h
```diff
@@ -0,0 +1,97 @@
...
+#include <inviwo/dataframe/dataframemoduledefine.h>
+
+#include <inviwo/core/common/inviwo.h>
+#include <inviwo/core/datastructures/buffer/buffer.h>
+#include <inviwo/core/io/serialization/serializable.h>
+#include <inviwo/dataframe/datastructures/dataframe.h>
+
+namespace inviwo {
+
+namespace dataframeutil {
```
ResearchDaniel: An option would be to just have dataframe as namespace. Breaking change but moving the file will be a breaking change anyway

modules/dataframe/include/inviwo/dataframe/util/dataframeutil.h
```diff
@@ -0,0 +1,97 @@
...
+void IVW_MODULE_DATAFRAME_API copyBufferRange(std::shared_ptr<const BufferBase> src,
+                                              std::shared_ptr<BufferBase> dst, ivec2 range,
+                                              size_t dstStart = 0);
+
+/**
+ * \brief create a new DataFrame by appending the columns of DataFrame \p right to DataFrame \p left
+ *
+ * @param ignoreDuplicates duplicate columns, i.e. same column header, are ignored if true
+ * @param fillMissingRows  if true, missing rows in either DataFrame are filled with 0 or
+ *                         "undefined" (for categorical columns)
```
ResearchDaniel, line 54: Should we use NaN instead on 0? Or force specification of value to fill in?
Forcing 0 can cause issues later

modules/dataframe/include/inviwo/dataframe/util/dataframeutil.h
```diff
@@ -0,0 +1,97 @@
...
+std::shared_ptr<DataFrame> IVW_MODULE_DATAFRAME_API appendColumns(const DataFrame& left,
+                                                                  const DataFrame& right,
+                                                                  bool ignoreDuplicates = false,
+                                                                  bool fillMissingRows = false);
+
+/**
+ * \brief create a new DataFrame by appending the rows of DataFrame \p bottom to DataFrame \p top
+ *
+ * @param matchByName    if true, column headers are used for matching columns. Otherwise columns
+ *                       are matched by order (default)
```
ResearchDaniel, line 67: Does it not make more sense to match by name by default?

modules/dataframe/src/util/dataframeutil.cpp
```diff
@@ -0,0 +1,325 @@
...
+#include <inviwo/core/util/zip.h>
+#include <inviwo/core/util/assertion.h>
+
+#include <fmt/format.h>
+
+namespace inviwo {
+
+namespace dataframeutil {
+
+std::shared_ptr<BufferBase> cloneBufferRange(std::shared_ptr<const BufferBase> buffer,
```
ResearchDaniel, line 44: Not part of this PR but we have placed this type in buffer/algorithm/ for other type (volume/image)

petersteneteg (dismissed):

modules/dataframe/src/util/dataframeutil.cpp
```diff
@@ -0,0 +1,325 @@
...
+
+    addColumns(left, rows.first, false);
+    addColumns(right, rows.second, true);
+
+    dataframe->updateIndexBuffer();
+    return dataframe;
+}
+
+std::shared_ptr<DataFrame> combineDataFrames(std::vector<std::shared_ptr<DataFrame>> dataFrames,
+                                             bool skipIndexColumn, std::string skipcol) {
```
ResearchDaniel, line 226: Simply combine(...) because it is in the namespace?

ResearchDaniel (dismissed):

modules/dataframe/include/inviwo/dataframe/util/dataframeutil.h
martinfalk: Not sure. At the moment we definitely don't have a proper way to encode "missing" or empty data. In Sql it's `NULL`. `"undefined"` for categorical columns is already stretching it a bit.

modules/dataframe/include/inviwo/dataframe/util/dataframeutil.h
martinfalk: wouldn't say so. Common scenario from my perspective is to merge two dataframes from different sources, i.e. different headers.

modules/dataframe/src/util/dataframeutil.cpp
martinfalk: that was pre-existing prior this PR (similar to `cloneBufferRange`)

modules/dataframe/include/inviwo/dataframe/util/dataframeutil.h
ResearchDaniel: I think that allowing values to fill in would be the way to go. The default could be NaN for numerical and undefined for categorical though. Then it is kind of up to the developer to choose 0 if that is more appropriate

modules/dataframe/include/inviwo/dataframe/util/dataframeutil.h
ResearchDaniel: Ok, we had another scenario where the columns were moved, but I guess either is valid

modules/dataframe/include/inviwo/dataframe/util/dataframeutil.h
martinfalk: NaN does not exist for integral types though.

modules/dataframe/include/inviwo/dataframe/util/dataframeutil.h
martinfalk: done. Also mentioned in changelog.

modules/dataframe/include/inviwo/dataframe/util/dataframeutil.h
petersteneteg: One should definitely not use NaN. It's not supposed to represent missing data, it is the result it a failed computation.
if you want to have a state for missing you have to provide that externally as in using `optional<int>` or `optional<float>`

petersteneteg: But having the data stored in buffers will not work. Hence it you want to have missing data in a data frame it needs to be dramatically redesigned.

petersteneteg (approved):

modules/dataframe/include/inviwo/dataframe/util/dataframeutil.h
ResearchDaniel: One should not use 0 to represent missing data either. Let the developer decide at least

petersteneteg: The thing is the current data frame can NOT represent missing data, there is nothing to decide.

ResearchDaniel: With that reasoning there should be an exception thrown and that is not really an option either.

petersteneteg: there should only be an inner join. we can't support any other one.

PR #985, opened 2020-07-03
External app pre-module registration hook

ResearchDaniel:
Provide a hook for external applications such that they can perform the same setup as the internal apps.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -105,6 +105,9 @@ ivw_enable_modules_if(IVW_APP_PYTHON Python3 Python3Qt QtWidgets)
 option(IVW_TEST_INTEGRATION_TESTS "Build inviwo integration test" ON)
 ivw_enable_modules_if(IVW_TEST_INTEGRATION_TESTS GLFW Base)
 
+# Provide optional hook for external application setup 
+ivw_external_apps_pre_module_registration_setup()
+
 # Try to find qt and add it if it is not already in CMAKE_PREFIX_PATH
 if(NOT "${CMAKE_PREFIX_PATH}" MATCHES "[Qq][Tt]")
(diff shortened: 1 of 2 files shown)


petersteneteg:
Any reason why you can not just rise a cmake error in your external app if the module is missing, and ask the user to enable it? 
I would rather that we make it possible to add inviwo as a subproject, than that we add more complexity for including external projects like this

ResearchDaniel:
Convenience and consistency with the “internal” code was the reason

petersteneteg:
Mmm I can understand that. But Im trying to reduce the complexity of our cmake setup, and this would go in the opposite direction of that...

CMakeLists.txt
```diff
@@ -125,7 +128,7 @@ endif()
 ivw_register_modules(all_modules)            # Add inviwo modules
 add_subdirectory(src/qt)                     # Add the Qt network editor
 add_subdirectory(apps)                       # Add all the applications, uses the modules.
-ivw_add_external_projects()                  # Add external projects
+ivw_add_external_apps()                      # Add external projects
```
petersteneteg, line 131: this is not just for apps, way rename it? it also makes it inconsistent with the cmake flag use and the comments...

petersteneteg (changes requested):

CMakeLists.txt
ResearchDaniel: I’ve only seen it used for apps and thought it would be more consistent with the other names (IVW_APP).

ResearchDaniel:
Ok, let’s skip this PR then and see what should be done once the complexity is reduced.

petersteneteg:
reducing complexity is more of a long term goal... but ideally your app should be able to consume inviwo as a dependency not make it self a part of it....

PR #989, opened 2020-07-06
Webbrowser: setupResources was not called after module registration.

ResearchDaniel:
Also added regression test to make sure this does not go unseen again.


martinfalk (dismissed):

petersteneteg (dismissed):

petersteneteg (dismissed):

petersteneteg (approved):

ResearchDaniel:
yay, finally! 
We might need to tune the test wait-delay

PR #996, opened 2020-07-13
Feature/converter

martinfalk:
* Volume data format converter with optional mapping of data values
* DataFrame converter for `double` columns to `float32`
* improved CSV reader distinguishes `int`, `float`, and `string`/categorical (optional double precision)

### DIFF
--- a/modules/base/include/modules/base/processors/volumeconverter.h
+++ b/modules/base/include/modules/base/processors/volumeconverter.h
@@ -1,0 +1,84 @@
+/*********************************************************************************
+ *
+ * Inviwo - Interactive Visualization Workshop
+ *
+ * Copyright (c) 2020 Inviwo Foundation
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ *********************************************************************************/
+
+#pragma once
+
+#include <modules/base/basemoduledefine.h>
+#include <inviwo/core/processors/processor.h>
+#include <inviwo/core/properties/stringproperty.h>
+#include <inviwo/core/properties/boolproperty.h>
+#include <inviwo/core/properties/optionproperty.h>
+#include <inviwo/core/properties/minmaxproperty.h>
+#include <modules/base/properties/datarangeproperty.h>
+#include <inviwo/core/ports/volumeport.h>
+
+namespace inviwo {
+
+/** \docpage{org.inviwo.VolumeConverter, Volume Converter}
+ * ![](org.inviwo.VolumeConverter.png?classIdentifier=org.inviwo.VolumeConverter)
+ * Converts the data type of a volume to a given output data format. The number of channels remains
+ * unchanged.
+ *
+ * ### Inports
+ *   * __inport__    volume input
+ *
+ * ### Outports
+ *   * __outport__   converted input volume
+ *
+ * ### Properties
+ *   * __Format__    data format of the output volume. The number of channels remains
+ *                   the same. If identical to input data, no conversion will be performed.
+ *   * __Use Data Mapping__   If enabled, the processor will utilize data mapping between different
+ *                   integer formats. That is, each data value is normalized using the data range of
+ *                   the input volume before being adjusted to the target format (using the range of
+ *                   the data type). Float formats are __not__ normalized!
+ */
+class IVW_MODULE_BASE_API VolumeConverter : public Processor {
+public:
+    VolumeConverter();
+    virtual ~VolumeConverter() = default;
+
+    virtual void process() override;
+
+    virtual const ProcessorInfo getProcessorInfo() const override;
+    static const ProcessorInfo processorInfo_;
+
+private:
+    VolumeInport inport_;
+    VolumeOutport outport_;
+
+    StringProperty inputFormat_;
+    TemplateOptionProperty<DataFormatId> format_;
+    BoolProperty enableDataMapping_;
+
+    DataRangeProperty dataRange_;
+    DoubleMinMaxProperty outputDataRange_;
+};
+
+}  // namespace inviwo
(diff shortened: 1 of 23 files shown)


modules/base/src/processors/volumeconverter.cpp
```diff
@@ -0,0 +1,203 @@
...
+        if (dataRange_.getCustomRangeEnabled()) {
+            return {dataRange_.getCustomDataRange(), dataRange_.getCustomValueRange()};
+        } else {
+            if (enableDataMapping_) {
+                if (auto format = DataFormatBase::get(format_);
+                    format->getNumericType() != NumericType::Float) {
+                    DataMapper mapper{format};
+                    return {mapper.dataRange, mapper.valueRange};
+                } else {
+                    return {dvec2{0.0, 1.0}, dvec2{0.0, 1.0}};
```
ResearchDaniel, line 179: We should use the min/max of the data?

modules/base/include/modules/base/processors/volumeconverter.h
```diff
@@ -0,0 +1,84 @@
...
+ * ### Inports
+ *   * __inport__    volume input
+ *
+ * ### Outports
+ *   * __outport__   converted input volume
+ *
+ * ### Properties
+ *   * __Format__    data format of the output volume. The number of channels remains
+ *                   the same. If identical to input data, no conversion will be performed.
+ *   * __Use Data Mapping__   If enabled, the processor will utilize data mapping between different
```
ResearchDaniel: Suggestion: Normalize to target range
Example, 
uint8 [0 255], value 255 -> uint16 [0 65536], value 65536

Should only be visible if the input format is integer type.

modules/base/src/processors/volumeconverter.cpp
```diff
@@ -0,0 +1,203 @@
...
+
+                auto dstVol = std::make_shared<VolumeRAMPrecision<T>>(
+                    dims, src->getSwizzleMask(), src->getInterpolation(), src->getWrapping());
+                T* dstData = dstVol->getDataTyped();
+
+                if (mapData) {
+                    const dvec2 srcRange{
+                        (src->getDataFormat()->getNumericType() != NumericType::Float)
+                            ? src->dataMap_.dataRange
+                            : dvec2{0.0, 1.0}};
```
ResearchDaniel, line 69: Not sure I understand this special treatment. Is it because the range is hard coded in the other parts?

modules/base/src/processors/volumeconverter.cpp
```diff
@@ -0,0 +1,203 @@
...
+                    std::transform(
+                        srcData, srcData + glm::compMul(dims), dstData,
+                        [srcRange, dstRange](auto& v) {
+                            return static_cast<T>((static_cast<Tdouble>(v) - srcRange.x) /
```
ResearchDaniel, line 73: Ideally, we would use the DataMapper here

modules/dataframe/include/inviwo/dataframe/datastructures/column.h
```diff
@@ -321,10 +321,14 @@ namespace detail {
 template <typename T, typename std::enable_if<std::is_integral<T>::value, int>::type = 0>
 void add(Buffer<T>* buffer, const std::string& value) {
     T result;
-    std::istringstream stream(value);
-    stream >> result;
-    if (stream.fail()) {
-        throw InvalidConversion("cannot convert \"" + value + "\" to target type");
+    if (value.empty()) {
+        result = T{0};  // no special value indicating missing data for integral types
```
ResearchDaniel, line 325: Why set 0 if the data is missing?

modules/dataframe/include/inviwo/dataframe/io/csvreader.h
```diff
@@ -71,6 +71,7 @@ class IVW_MODULE_DATAFRAME_API CSVReader : public DataReaderType<DataFrame> {
 
     void setDelimiters(const std::string& delim);
     void setFirstRowHeader(bool hasHeader);
+    void setDoublePrecision(bool doubleprec);
```
ResearchDaniel: This deserves some explanation. 
UseDoublePrecisionForFloatingPointNumbers

modules/dataframe/tests/unittests/csvreader-test.cpp
```diff
@@ -131,7 +131,7 @@ TEST(CSVdata, emptyFields) {
     auto value = dataframe->getColumn(3)->get(1, true)->toString();
     EXPECT_EQ("", value) << "empty field at end of row";
     value = dataframe->getColumn(1)->get(2, false)->toString();
-    EXPECT_EQ("nan", value) << "empty numeric field at begin of row";
+    EXPECT_EQ("0", value) << "empty numeric field at begin of row";
```
ResearchDaniel, line 134: No, how should we deal with missing numbers?
This will break existing code assumptions. We need a way to deal with missing numbers if this is removed

ResearchDaniel (changes requested):

modules/base/src/processors/volumeconverter.cpp
martinfalk: Try it! ;-) Had it there in the first place gives a lot template errors regarding not being able to static_cast<ValueType> to T. Furthermore, one has to use map dataRange to valueRange with valueRange being the range of the destination format. Bit of a misnomer.

modules/base/src/processors/volumeconverter.cpp
martinfalk: No, this is for float types so that their range is not affected by a renormalization to [0,1] as all the integer formats.

modules/base/include/modules/base/processors/volumeconverter.h
martinfalk: this is exactly what happens. Float values are not (re-)normalized. If this bool option is selected, incoming values are normalized (except float) and then adjusted to the integer target ranges.

modules/base/src/processors/volumeconverter.cpp
martinfalk: Why? When a DataMapper is used, there is never a normalization to [0,1] for floating point types.

modules/base/include/modules/base/processors/volumeconverter.h
martinfalk: And I really dislike fields which disappear and re-appear based on other options. Every time this happens the property widget layout changes and most often properties jump up and down. That is, you cannot quickly enable and disable it since the property moved.

modules/dataframe/include/inviwo/dataframe/datastructures/column.h
martinfalk: What else should we use for int? We are tied to integer values due to the use of `Buffer`. Also, the CSV reader now distinguishes between float and int columns. One reason for this is that joining DataFrames columns with different types will be interesting to implement. For now, only matching columns (same title, same format) are handled.

modules/dataframe/tests/unittests/csvreader-test.cpp
martinfalk: "nan" still holds true for float32/float64 cols! Note that this column is an integer column, hence no `nan`.

martinfalk: Missing numbers -> need to re-design the entire DataFrame/Column datastructures to take care of this. One option would be to use `std::optional` for each single value... Might not be what we want. There are also more issues we need to address like the implicit (or rather explicit) index column at index 0. When copying columns that one needs special handling. One might even remove it from a dataframe... Needs to be discussed before changing anything.

modules/dataframe/include/inviwo/dataframe/io/csvreader.h
martinfalk: `setEnableDoublePrecision` + doxygen comment?

modules/base/src/processors/volumeconverter.cpp
martinfalk: DataMapper also has no interface for DataRange to normalized and vice versa.

modules/base/include/modules/base/processors/volumeconverter.h
ResearchDaniel: Good, then the example could be included in the comment :)

Then we can make it read-only at least?

modules/base/src/processors/volumeconverter.cpp
ResearchDaniel: Yeah, that should be changed in the datamapper at some point

modules/dataframe/include/inviwo/dataframe/datastructures/column.h
ResearchDaniel: std::numeric_limits<T>::quiet_NaN
(Which will result in 0 anyway though)

modules/dataframe/tests/unittests/csvreader-test.cpp
ResearchDaniel: I guess one option is also to keep references to missing values. Really not optimal but would go in line with the current implementation

ResearchDaniel (dismissed):

modules/base/include/modules/base/processors/volumeconverter.h
martinfalk: Unfortunately, read-only doesn't solve this either.
Assuming your source format is int8, target format is float. With data mapping int values are renormalized to [0,1] (using data range) and the destination range is set to [0,1].
Without data mapping, int8 values are casted into float without any changes.

martinfalk: Clarified that in the doxygen comment

modules/dataframe/include/inviwo/dataframe/datastructures/column.h
martinfalk: hmm. Needs discussion. For now, there is a special overload for floating point values when setting data values from `std::string`.

ResearchDaniel (approved):

PR #997, opened 2020-07-18
fix compile error on windows 10 with vs2019

maidamai0:
Fix compile error with:
* Windows 10
* Visual studio 2019 build tools
* Ninja

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@ inviwo-log.html
 pyconfig.ini
 .idea
 .vscode
+.clangd
+compile_commands.json
--- a/modules/dataframe/src/util/dataframeutil.cpp
+++ b/modules/dataframe/src/util/dataframeutil.cpp
@@ -27,6 +27,8 @@
  *
  *********************************************************************************/
 
+ #include <optional>
+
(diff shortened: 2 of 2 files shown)


maidamai0:
These two format error seems to be caused by old version `clang-format`. I formatted the code with the `clang-format` installed with `llvm-10.0`.

r-englund (approved):

petersteneteg:
We try and use the same clangformat version that is shipping with visual studio. Think we use the one in vs2017 right now. Should probably update that to the one in vs2019. Unfortunately newer versions of clangformat does not format in the same way as older ones...

maidamai0:
Jenkins uses clang-format 6.0 installed with llvm-6
``` shell
python3 ../inviwo/tools/jenkins/check-format.py --binary /usr/local/Cellar/llvm@6/6.0.1_1/bin/clang-format /Users/jenkins/jenkins/
clang-format version 6.0.1 (tags/RELEASE_601/final)
```

petersteneteg:
Which I think is the same one that ships with msvc 2017

petersteneteg:
see #1002

PR #999, opened 2020-07-23
fix: windeploy not found

maidamai0:
windeploy not found if user set Qt path with CMAKE_PREFIX_PATH instead of QTDIR

- Windows 10
- CMake & Visual studio 2019 build tools & Ninja
- Qt 15.0
- CMake 3.17.0-rc1
- Python 3.8.4


ResearchDaniel (approved):

ResearchDaniel:
Thanks!

PR #1000, opened 2020-07-24
Feature/postprocessing depthoffield

nataxe:


modules/postprocessing/glsl/dof_approx.frag
```diff
@@ -0,0 +1,49 @@
...
+
+uniform sampler3D lightField;
+uniform sampler3D lightFieldDepth;
+uniform vec3 dim;
+
+void main() {
+    ivec2 xy = ivec2(gl_FragCoord.xy);
+
+    vec4 acc = vec4(0);
+    int weight = 0;
```
ResearchDaniel: This could be a float to avoid the conversion later

modules/postprocessing/glsl/dof_exact.frag
```diff
@@ -0,0 +1,39 @@
...
+
+uniform sampler2D oldImgColor;
+uniform sampler2D newImgColor;
+uniform int nOldImages;
+
+void main() {
+    vec4 oldColor = texelFetch(oldImgColor, ivec2(gl_FragCoord.xy), 0);
+    vec4 newColor = texelFetch(newImgColor, ivec2(gl_FragCoord.xy), 0);
+    FragData0 = oldColor * float(nOldImages) / float(nOldImages + 1) +
+                newColor * 1.0 / float(nOldImages + 1);
```
ResearchDaniel: Remove * 1 ?
Can write 
(oldColor * float(nOldImages) + newColor ) / float(nOldImages + 1)

modules/postprocessing/include/modules/postprocessing/processors/depthoffield.h
```diff
@@ -0,0 +1,150 @@
...
+#include <modules/opengl/texture/textureutils.h>
+#include <modules/opengl/volume/volumeutils.h>
+#include <modules/opengl/volume/volumegl.h>
+
+namespace inviwo {
+
+/** \docpage{org.inviwo.DepthOfField, Depth Of Field}
+ * ![](org.inviwo.DepthOfField.png?classIdentifier=org.inviwo.DepthOfField)
+ *
+ * Use any pinhole image as input and connect the camera property to the rendering processor.
```
ResearchDaniel: pinhole camera-generated image

You could move the first sentence last so that we first explain what it does and then how it can be used

modules/postprocessing/include/modules/postprocessing/processors/depthoffield.h
```diff
@@ -0,0 +1,150 @@
...
+ * Applies a depth of field effect where out-of-focus regions appear blurred. The image should be
+ * generated with a perspective camera but may use any rendering technique.
+ *
+ * ### Inports
+ *   * __ImageInport__ Input image.
+ *   * __MeshInport__ World space point to focus on. The first point in the first position buffer
```
ResearchDaniel, line 69: Requiring a mesh is a bit of a hack.
Ideally, you would simply use a vec3

@martinfalk @petersteneteg any thought on this

modules/postprocessing/include/modules/postprocessing/processors/depthoffield.h
```diff
@@ -0,0 +1,150 @@
...
+ * will be used.
+ *
+ * ### Outports
+ *   * __ImageOutport__ Output image.
+ *
+ * ### Properties
+ *   * __Aperture__ The diameter of the simulated lens. Affects blur strength.
+ *   * __FocusDepth__ The depth in the scene that is in focus.
```
ResearchDaniel: Is this not given by the input position?
Ah, you should probably mention that manual focus needs to be enabled :)

modules/postprocessing/include/modules/postprocessing/processors/depthoffield.h
```diff
@@ -0,0 +1,150 @@
...
+    std::vector<float> haltonY_;
+    Shader addSampleShader_;
+
+    // Approximative algorithm
+    size3_t dimLightField_;
+    std::shared_ptr<Volume> lightField_;
+    std::shared_ptr<Volume> lightFieldDepth_;
+    std::shared_ptr<Image> haltonImg_;
+    Shader addToLightFieldShader_;
+    Shader averageLightfieldShader_;
```
ResearchDaniel: Typo, averageLightFieldShade

modules/postprocessing/src/processors/depthoffield.cpp
```diff
@@ -0,0 +1,427 @@
...
+    auto mouseEvent = static_cast<MouseEvent*>(e);
+    size2_t clickPos(mouseEvent->x(), mouseEvent->y());
+    double depthNdc =
+        inport_.getData()->getDepthLayer()->getRepresentation<LayerRAM>()->getAsNormalizedDouble(
+            clickPos);
+
+    double nearClip = camera_.getNearPlaneDist();
+    double farClip = camera_.getFarPlaneDist();
+    double depthEye = farClip * nearClip / (farClip - depthNdc * (farClip - nearClip));
+    focusDepth_.set(depthEye);
```
ResearchDaniel: Use camera::getWorldPosFromNormalizedDeviceCoords ?

modules/postprocessing/src/processors/depthoffield.cpp
```diff
@@ -0,0 +1,427 @@
...
+    , outport_("outport")
+    , aperture_("aperture", "Aperture", 0.5f, 0.01f, 1.f, 0.01f)
+    , focusDepth_("focusDepth", "Focus depth", 8.f, 1.0f, 15.0f, 0.1f)
+    , manualFocus_("manualFocus", "Manual focus", true)
+    , approximate_("approximate", "Approximate", false)
+    , viewCountExact_("viewCountExact", "View count", 40, 10, 200)
+    , viewCountApprox_("viewCountApprox", "Rendered view count", 5, 1, 12)
+    , simViewCountApprox_("simViewCountApprox", "Simulated view count", 40, 10, 200)
+    , clickToFocus_(
+          "clickToFocus", "Click to focus", [this](Event* e) { clickToFocus(e); },
```
ResearchDaniel: You could add the manual focus check in the lambda

modules/postprocessing/src/processors/depthoffield.cpp
```diff
@@ -0,0 +1,427 @@
...
+        clickToFocus_.setReadOnly(!manualFocus_);
+    });
+
+    addSampleShader_.onReload([this]() { invalidate(InvalidationLevel::InvalidOutput); });
+    addToLightFieldShader_.onReload([this]() { invalidate(InvalidationLevel::InvalidOutput); });
+    averageLightfieldShader_.onReload([this]() { invalidate(InvalidationLevel::InvalidOutput); });
+}
+
+void DepthOfField::setApproximate(bool approximate) {
+    viewCountExact_.setVisible(!approximate);
```
ResearchDaniel: You can use Property::visibilityDependsOn and set it in the constructor instead

modules/postprocessing/src/processors/depthoffield.cpp
```diff
@@ -0,0 +1,427 @@
...
+
+void DepthOfField::process() {
+    int maxEvalCount = approximate_ ? viewCountApprox_.get() : viewCountExact_.get();
+    std::shared_ptr<const Image> img = inport_.getData();
+    size2_t dim = img->getDimensions();
+    TextureUnitContainer cont;
+
+    if (evalCount_ >= maxEvalCount) {
+        // Catch final network evaluation caused by camera resetting
+        evalCount_ = 0;
```
ResearchDaniel: Does this mean that the network is processed one time too much?

modules/postprocessing/src/processors/depthoffield.cpp
```diff
@@ -0,0 +1,427 @@
...
+            warpToLightfieldCPU(img, nearClip, farClip, fovy, focusDepth, dim, cameraPos);
+        }
+    } else {
+        addToAccumulationBuffer(img, cont);
+    }
+
+    if (evalCount_ < maxEvalCount - 1) {
+        // Prepare camera for rendering again
+        if (!approximate_) {
+            nextOutImg_->copyRepresentationsTo(prevOutImg_.get());
```
ResearchDaniel: Can’t you swap the pointers instead of copying?

modules/postprocessing/src/processors/depthoffield.cpp
```diff
@@ -0,0 +1,427 @@
...
+        }
+        moveCamera(camera, maxEvalCount, focusDepth);
```
ResearchDaniel: You are not allowed to change properties during network evaluation (e.g. in the process function). Use InviwoApplication::dispatchFront and do the changes in the supplied lambda instead. See SimpleRaycaster for an example.

modules/postprocessing/src/processors/depthoffield.cpp
```diff
@@ -0,0 +1,427 @@
...
+        // Set output
+        if (approximate_) {
+            synthesizeLightfield(cont);
+        } else {
+            outport_.setData(nextOutImg_);
+        }
+
+        // Reset camera
+        evalCount_ = (ogType_ == SkewedPerspectiveCamera::classIdentifier) ? 0 : maxEvalCount;
+        camera_.setCamera(std::unique_ptr<Camera>(ogCamera_->clone()));
```
ResearchDaniel: See comment about changing properties in the process function

modules/postprocessing/src/processors/depthoffield.cpp
```diff
@@ -0,0 +1,427 @@
...
+    }
+}
+
+void DepthOfField::setupRecursion(size2_t dim, int maxEvalCount, std::shared_ptr<const Image> img) {
+    ogCamera_.reset(camera_.get().clone());
+    ogType_ = camera_.get().getClassIdentifier();
```
ResearchDaniel: You probably don’t need to store the class identifier. 
Use dynamic cast each time instead

modules/postprocessing/src/processors/depthoffield.cpp
```diff
@@ -0,0 +1,427 @@
...
+    if (ogType_ != PerspectiveCamera::classIdentifier) {
+        LogWarn("Intended for use with Perspective Camera. Unexpected behavior may follow.");
```
ResearchDaniel, line 171: I think that it is probably better to throw an exception when things are unsupported.

modules/postprocessing/src/processors/depthoffield.cpp
```diff
@@ -0,0 +1,427 @@
...
+    }
+    camera_.setCamera(SkewedPerspectiveCamera::classIdentifier);
```
ResearchDaniel: Any camera linked to this processor will be changed to skewed perspective camera. Is that intended? 
Above you warn about using non-perspective camera and then you set it to be skewed perspective. So other camera types, e.g. orthographic, will never be used anyway?

modules/postprocessing/src/processors/depthoffield.cpp
```diff
@@ -0,0 +1,427 @@
...
+
+    if (approximate_) {
+        // Prepare light field
+        dimLightField_ = size3_t(dim.x, dim.y, viewCountApprox_.get() + simViewCountApprox_.get());
```
ResearchDaniel: You don’t need to store this, just get it from lightField_? It will avoid possible bugs

modules/postprocessing/src/processors/depthoffield.cpp
```diff
@@ -0,0 +1,427 @@
...
+        lightField_ = std::make_shared<Volume>(dimLightField_, DataFormat<vec4>::get());
```
ResearchDaniel: You can create the volume(s) from a VolumeRamPrecision<vec4> instead. Then you don’t need to do the getEditableRepresentation step and also get the correct type directly.

modules/postprocessing/src/processors/depthoffield.cpp
```diff
@@ -0,0 +1,427 @@
...
+        lightFieldDepth_ = std::make_shared<Volume>(dimLightField_, DataFormat<float>::get());
+        float* lightFieldDepthData = static_cast<float*>(
+            lightFieldDepth_->getEditableRepresentation<VolumeRAM>()->getData());
+        std::fill(lightFieldDepthData,
+                  lightFieldDepthData + dimLightField_.x * dimLightField_.y * dimLightField_.z,
+                  -1.f);
+
+        // Prepare Halton sequence
+        haltonX_ = util::haltonSequence<float>(2, simViewCountApprox_.get());
+        haltonImg_ = std::make_shared<Image>(size2_t(simViewCountApprox_.get(), 1),
```
ResearchDaniel: See comment above. Use ImageRAMPrecision instead

modules/postprocessing/src/processors/depthoffield.cpp
```diff
@@ -0,0 +1,427 @@
...
+            outport_.setData(nextOutImg_);
+        }
+
+        // Reset camera
+        evalCount_ = (ogType_ == SkewedPerspectiveCamera::classIdentifier) ? 0 : maxEvalCount;
+        camera_.setCamera(std::unique_ptr<Camera>(ogCamera_->clone()));
+    }
+}
+
+void DepthOfField::setupRecursion(size2_t dim, int maxEvalCount, std::shared_ptr<const Image> img) {
```
ResearchDaniel: This function does a lot of allocations and if I understand it correctly it is called as soon as the camera is changed.
You could resize the volumes/images and initialize them only if they have the incorrect size?

modules/postprocessing/src/processors/depthoffield.cpp
```diff
@@ -0,0 +1,427 @@
...
+        return vec2(0, 0);
+    } else {
+        float currAngle = 2.0 * M_PI * (float(evalCount) - 0.5) / float(maxEvalCount - 1);
+        return aperture_.get() / 2.0 * vec2(glm::cos(currAngle), glm::sin(currAngle));
+    }
+}
+
+void DepthOfField::addToAccumulationBuffer(std::shared_ptr<const Image> img,
+                                           TextureUnitContainer& cont) {
+    utilgl::activateTargetAndCopySource(*nextOutImg_, inport_, ImageType::ColorOnly);
```
ResearchDaniel, line 270: Is it necessary to copy the input?

modules/postprocessing/src/processors/depthoffield.cpp
```diff
@@ -0,0 +1,427 @@
...
+
+    addToLightFieldShader_.activate();
+    utilgl::bindAndSetUniforms(addToLightFieldShader_, cont, inport_, ImageType::ColorDepth);
+    utilgl::bindAndSetUniforms(addToLightFieldShader_, cont, *haltonImg_, "halton",
+                               ImageType::ColorOnly);
+
+    addToLightFieldShader_.setUniform("nearClip", float(nearClip));
+    addToLightFieldShader_.setUniform("farClip", float(farClip));
+    addToLightFieldShader_.setUniform("evalCount", int(evalCount_));
+    addToLightFieldShader_.setUniform("aperture", float(aperture_.get()));
```
ResearchDaniel, line 303: We have convenience functions that use the property identifier as name of uniform. Check the Raycaster for an example. Then you can set the uniforms with a single line of code. You can use it for convenience and Better code readability

modules/postprocessing/glsl/dof_approx.comp
```diff
@@ -0,0 +1,79 @@
...
+    float radius = aperture / 2.0 * sqrt(texelFetch(haltonColor, ivec2(viewIndex, 0), 0).r);
+    float angle = float(viewIndex - 1) / float(nSimViews) * 2.0 * 3.1415926538;
+    vec2 simCameraPos = radius * vec2(cos(angle), sin(angle));
+
+    // Warp to simulated view.
+    vec2 disparity = (1.0 / zWorld - 1.0 / focusDepth) * (cameraPos - simCameraPos);
+    vec2 simScreenPos = screenPos + disparity * dim.y / (2.0 * tan(fovy / 2.0));
+    ivec3 pos = ivec3(round(simScreenPos.x), round(simScreenPos.y), viewIndex + nViews);
+    float currDepth = imageLoad(lightFieldDepth, pos).r;
+    if (currDepth > 0) {
```
ResearchDaniel, line 73: && < zWorld

modules/postprocessing/src/processors/depthoffield.cpp
```diff
@@ -0,0 +1,427 @@
...
+
+        // Prepare Halton sequences
+        haltonX_ = util::haltonSequence<float>(2, maxEvalCount);
+        haltonY_ = util::haltonSequence<float>(3, maxEvalCount);
+    }
+
+    // Update focus depth and aperture ranges based on the depths in the scene, ignoring
+    // background points and including the current value.
+    const float* inputDepthData = static_cast<const float*>(
+        img->getRepresentation<ImageRAM>()->getDepthLayerRAM()->getData());
```
ResearchDaniel, line 215: Much performance would be gained if this computation is done one the gpu with parallel reduce.

Use parallel reduce or minmax_element on the cpu 
https://en.cppreference.com/w/cpp/algorithm/minmax_element

modules/postprocessing/src/processors/depthoffield.cpp
```diff
@@ -0,0 +1,427 @@
...
+    for (const float* f = inputDepthData; f < inputDepthData + dim.x * dim.y; f++) {
+        minDepthNdc = std::min(minDepthNdc, *f);
+        if (*f < 1) {
+            maxDepthNdc = std::max(maxDepthNdc, *f);
+        }
+    }
+
+    double nearClip = camera_.getNearPlaneDist();
+    double farClip = camera_.getFarPlaneDist();
+    float minDepthWorld = nearClip * farClip / (minDepthNdc * (nearClip - farClip) + farClip);
```
ResearchDaniel: I think the camera has these transforms? ClipToWorld

modules/postprocessing/src/processors/depthoffield.cpp
```diff
@@ -0,0 +1,427 @@
...
+    float maxDepthWorld = nearClip * farClip / (maxDepthNdc * (nearClip - farClip) + farClip);
+
+    focusDepth_.setMinValue(std::min(minDepthWorld, focusDepth_.get()));
```
ResearchDaniel: Not allowed to set properties in the process function.
Also, either lock the network or use property.set with all values at the same time because otherwise the network will be evaluated multiple times

modules/postprocessing/src/processors/depthoffield.cpp
```diff
@@ -0,0 +1,427 @@
...
+    focusDepth_.setMaxValue(std::max(maxDepthWorld, focusDepth_.get()));
+    focusDepth_.set(focusDepth_.get(), std::min(minDepthWorld, focusDepth_.get()),
+                    std::max(maxDepthWorld, focusDepth_.get()),
+                    (maxDepthWorld - minDepthWorld) / 100.0);
+
+    int depthScale = floor(log10(minDepthWorld));
+    float minAperture = pow(10, depthScale - 1);
+    float maxAperture = pow(10, depthScale);
+    aperture_.set(aperture_.get(), std::min(minAperture, aperture_.get()),
```
ResearchDaniel: Since you set many properties here you want to lock the network using NetworkLock

modules/postprocessing/src/processors/depthoffield.cpp
```diff
@@ -0,0 +1,427 @@
...
+
+    double currDepth = lightFieldDepth->getAsDouble(pos);
+    if (currDepth < 0 || currDepth > zWorld + 0.01) {
+        lightField->setFromDVec4(pos, color);
+        lightFieldDepth->setFromDouble(pos, zWorld);
+    }
+}
+
+void DepthOfField::synthesizeLightfield(TextureUnitContainer& cont) {
+    utilgl::activateTargetAndCopySource(outport_, inport_, ImageType::ColorOnly);
```
ResearchDaniel, line 398: Do you need to copy input?

ResearchDaniel (changes requested):
Very nice work!
I have a lot of comments but many are on a nice-to-have level. 
Also, it would be nice (not necessary though) if the two algorithms could be separated more so that it is easier to follow. This would also prevent cases were algorithm-specific data is left around (lightField stuff) when switching algorithms.

A feature request: is it possible to somehow animate DoF on and off? For example by changing the aperture? In that case the off-mode could simple forward the inport data. This could create nice effects when transitioning between focus points :)

modules/postprocessing/glsl/dof_approx.frag
nataxe: Done

modules/postprocessing/glsl/dof_exact.frag
nataxe: Done

modules/postprocessing/include/modules/postprocessing/processors/depthoffield.h
nataxe: Done

modules/postprocessing/src/processors/depthoffield.cpp
nataxe: I see how this would be a good way of abstracting out camera-specific details. However, this change would introduce lots of extra work:
- Converting our vector [0, width-1]x[0, height-1]x[0,1] to [-1,1]x[-1,1]x[-1,1].
- camera::getWorldPosFromNormalizedDeviceCoords itself does more work than we need, including multiple matrix multiplications.
- The result is a world space position, from which we would then need to calculate the distance vector to the camera and project it onto the camera's view vector.

modules/postprocessing/src/processors/depthoffield.cpp
nataxe: When we have finished generating the image, we want to reset the camera to its original state which can trigger an additional network evaluation (which we want to ignore).

modules/postprocessing/src/processors/depthoffield.cpp
nataxe: Yes, this is intended. We need a skewed perspective camera since we're rendering multiple times with slightly offset view positions, but with the focus plane aligned each time. Orthographic cameras will never be used. Cameras that are already skewed may produce unexpected the result since the initial offset requires complex modifications to the algorithm.

modules/postprocessing/glsl/dof_approx.comp
nataxe: This does not compile for me. Not sure why but that's why I wrote this awkward nested conditional in the first place.

modules/postprocessing/src/processors/depthoffield.cpp
nataxe: This creates other issues down the line, particularly when using this data in shaders. If it's just a matter of preference I'd rather leave it as it is.

modules/postprocessing/src/processors/depthoffield.cpp
nataxe: Same as above

modules/postprocessing/src/processors/depthoffield.cpp
nataxe: I need it to copy the depth map.

modules/postprocessing/src/processors/depthoffield.cpp
nataxe: I'm a little bit unsure about making this change since I don't currently have access to a machine where I can test the compute shader. Additionally, as far as I can tell, this can only be used on properties/ports which have identifiers so the majority of these lines would be kept the same anyway.

modules/postprocessing/src/processors/depthoffield.cpp
nataxe: See above

modules/postprocessing/src/processors/depthoffield.cpp
nataxe: Which version of parallel reduce do you mean? I tried to google it but could only find experimental versions? I can't use minmax_element since I want to ignore the value 1.

modules/postprocessing/src/processors/depthoffield.cpp
ResearchDaniel: What kind of issue with shaders does it create?

modules/postprocessing/src/processors/depthoffield.cpp
ResearchDaniel: You can pass a lambda for special comparison in std::minmax_element

I think one have to roll your own parallel reduce algorithm for the GPU unfortunately (our of scope here). For OpenCL we have used CLOGS ( https://github.com/bmerry/clogs ). Should be done outside of this module in that case though.

modules/postprocessing/src/processors/depthoffield.cpp
nataxe: Yes I need the depth map

modules/postprocessing/src/processors/depthoffield.cpp
nataxe: Other classes are not necessarily unsupported but certain settings will give strange results.

nataxe:
Thanks for the comments! I've fixed most comments and added replies otherwise. Animated transitions sounds really nice but is out of scope for this PR. :)

modules/postprocessing/include/modules/postprocessing/processors/depthoffield.h
nataxe: This is the current solution suggested by Tino (my supervisor in this project). We didn't want to use vec3 since we couldn't find interesting processors with vec3 outports.

modules/postprocessing/src/processors/depthoffield.cpp
nataxe: Right, but the lambda should only return true or false, indicating whether or not a < b. I'm not sure how I would ignore the value 1 without it affecting either the min or the max value since they are both evaluated using the same lambda function.

modules/postprocessing/src/processors/depthoffield.cpp
nataxe: I couldn't easily convert it to a VolumeGL and since I don't currently have access to a machine which supports compute shaders, I feel uncertain about making such big changes.

modules/postprocessing/glsl/dof_approx.comp
ResearchDaniel: I see, weird. Then write a comment and say which GPU and driver version it happened for.

modules/postprocessing/src/processors/depthoffield.cpp
ResearchDaniel: Ah, then you can use std::accumulate if you want. See here for an example 
https://github.com/inviwo/inviwo/blob/2205ee567b625a65c8aeccb78cd56ea15d5c4e6c/modules/base/include/modules/base/algorithm/dataminmax.h

modules/postprocessing/src/processors/depthoffield.cpp
ResearchDaniel: Ok. 
Setting the camera property would also set the near/far
https://github.com/inviwo/inviwo/blob/2205ee567b625a65c8aeccb78cd56ea15d5c4e6c/modules/opengl/glsl/utils/structs.glsl

ResearchDaniel:
Ok, but that would be the way to go in case an animation is desired?

Forgot this before, could you add a reference to the paper the algorithm is from in case you have one? Either in the processor documentation or somewhere in the implementation

modules/postprocessing/src/processors/depthoffield.cpp
ResearchDaniel: Ok. I think that I saw this code pop up a couple of times. Perhaps it should be extracted to a function and reused? I might be mistaken though since there might be small differences

modules/postprocessing/include/modules/postprocessing/processors/depthoffield.h
ResearchDaniel: Ok, even though it is a bit more work on the network level it might be better to create a processor that extracts a point from a mesh (or buffer or data frame)

nataxe: Okay, once that is done, it will be very straightforward to modify this processor. I'm not sure what kind of data might be interesting so I think it's out of scope for this PR.

modules/postprocessing/src/processors/depthoffield.cpp
nataxe: Done

modules/postprocessing/src/processors/depthoffield.cpp
nataxe: Hmm yeah I see how that would be a better design choice... Is it acceptable the way it is though? As I've mentioned, I don't have access to a machine to test the compute shader and I'm worried I might break something. (I was only supposed to be working on this project until the end of june but there have been various delays).

modules/postprocessing/src/processors/depthoffield.cpp
nataxe: Thanks! Done

modules/postprocessing/glsl/dof_approx.comp
nataxe: I don't know the driver version but I added a comment with the GPU.

nataxe:
Yes I think so, probably interpolating the aperture size for turning the effect on/off and interpolating the focus depth for transitioning the focus. 

I added a comment about the report it is based on (my master thesis from this spring) but it hasn't been uploaded yet so unfortunately I can't add a link. :(

petersteneteg (commented):
postprocessing.qch should not be deleted

modules/postprocessing/glsl/dof_approx.comp
```diff
@@ -0,0 +1,81 @@
...
+layout (binding = 1, r32f) uniform image3D lightFieldDepth;
+
+void main() {
+    vec2 screenPos = gl_GlobalInvocationID.xy;
+    vec4 color = texelFetch(inportColor, ivec2(screenPos), 0);
+    float zNdc = texelFetch(inportDepth, ivec2(screenPos), 0).r;
+    float zWorld = nearClip * farClip / (zNdc * (nearClip - farClip) + farClip);
+
+    imageStore(lightField, ivec3(screenPos, evalCount), color);
+    imageStore(lightFieldDepth, ivec3(screenPos, evalCount), vec4(zWorld,0,0,0));
```
petersteneteg, line 58: can this store not be merged with the one at the bottom? now some codepaths will store twice?

nataxe: I'm not sure what you mean by merge. The different imageStore calls always operate on different indices in the 3D image.

nataxe:
postprocessing.qch has not been deleted as far as I can see. To me, it shows that it 88% has been removed but I'm not sure exactly what has been deleted. I have only modified it by calling `make DOXY-HELP-PostProcessing` .

modules/postprocessing/glsl/dof_approx.comp
petersteneteg: ahh I see, ok.

nataxe:
> postprocessing.qch should not be deleted

It has not been deleted as far as I can see. To me, it shows that it 88% has been removed but I'm not sure exactly what has been deleted. I have only modified it by calling `make DOXY-HELP-PostProcessing`.

ResearchDaniel (dismissed):

petersteneteg (approved):

Issue #875, opened 2020-02-27
Volume offset affects scaling of volume slices


josejonasj:
**Issue**

A change in the offset of a volume source affects the scaling in volume slices.

For instance, when using the _diagnosticlayout_head_ example the default offset is:

<img width="221" alt="01" src="https://user-images.githubusercontent.com/58182481/75399308-e6ad2880-58c9-11ea-92de-f6f75f891efb.PNG">

and the result is:

<img width="256" alt="02" src="https://user-images.githubusercontent.com/58182481/75399317-ead94600-58c9-11ea-8718-e952b18c776c.png">

But if the user changes the Z value from -102.4 to -502.4 like this:

<img width="217" alt="03" src="https://user-images.githubusercontent.com/58182481/75399328-ef056380-58c9-11ea-9196-eaef2ebbee04.PNG">

the result is:

<img width="256" alt="04" src="https://user-images.githubusercontent.com/58182481/75399339-f298ea80-58c9-11ea-8855-f490caca6ae5.png">

So, the Z offset is squeezing the front and top volume slices.

**Steps to reproduce**

- Use the _diagnosticlayout_head_ example (located in Inviwo/0.9.11/data/workspaces).
- Select the _Volume Source_ processor.
- Check _Override_ property on _Basis and offset_ section.
- Uncheck the _Center Automatically_ property in the same section.
- Modify any of the offset values.

**Environment information**
* Operating System: Windows 10 (1903)
* Compiler/IDE: Visual Studio 2019
* QT Version: 5.13.1
* Python version: 3.8.1

**Log**

No errors found

Thanks!

petersteneteg:
Interesting, thanks for reporting, I will try and investigate

josejonasj:
If someone could give me a hint of how to address it I could try to fix it because I need it...

petersteneteg:
probably something gets wrong in this function or related calls https://github.com/inviwo/inviwo/blob/975b66f5d19be24c72510f939793cc160e8f4caa/modules/basegl/src/processors/volumeslicegl.cpp#L320

josejonasj:
OK, thanks a lot.

PR #1003, opened 2020-08-20
BaseGL: VolumeSliceGL fixes. Closes issue #640. Closes issue #875

ResearchDaniel:
- Output size is now determined by the connected canvas instead of the volume diagonal dimensions (issue #640).
- Changing Volume offset does not alter the slicing (issue #875)
- Indicator interaction can now handle aspect ratios != 1 (the indicator was placed at wrong location before)
- Minor indicator mesh optimization - update instead of recreating the mesh all the time.


petersteneteg (dismissed):
Nice!!! you even got a util::table in there, have not seen that in a while :)

ResearchDaniel:
> Nice!!! you even got a util::table in there, have not seen that in a while :)

Thanks, but it was in there already :)
I did have a weird compile error on Jenkins when trying to pass the vertices in the TypedMesh constructor.

petersteneteg:
yes, that is a strange error, from what I can see that should have worked.

petersteneteg (dismissed):

petersteneteg (approved):

PR #1008, opened 2020-08-27
BaseGL: Entryexitpoint fix for left handed systems

petersteneteg:


ResearchDaniel (approved):
Might want to add in the documentation that both left- and right handed coordinate spaces are supported. 
Although it might make sense to ensure that input is given in a specific coordinate system??

petersteneteg:
@ResearchDaniel: Not sure I want to touch to much on this topic. Problem where was that if you use a volume with a transform that flips one of the axis, then it goes from right handed to left handed...

PR #1009, opened 2020-08-28
AxisRenderer3D: Given caption text rotation is now used.

jjeessppeer:
The rotation given from the PlotTextProperty is now used when AxisRenderer3D renders text, much like AxisRenderer does.


martinfalk (approved):
Thanks for the contribution!

PR #1014, opened 2020-09-03
Feature/plottinggl

martinfalk:
various fixes, mostly plotting related.

- made FrameBufferObject, TextRenderer, and AxisRenderer move-assignable 
    * deleted copy constructor and assignment operator
- Default RenderContext activated
    * before adding properties to a ListPropertyWidget
    * before removing properties from a ListPropertyWidget
    * before adding a visualizer
- more fixes (TextRenderer, ScatterPlotGL, ...)

### DIFF
--- a/modules/fontrendering/include/modules/fontrendering/textrenderer.h
+++ b/modules/fontrendering/include/modules/fontrendering/textrenderer.h
@@ -110,7 +110,11 @@ struct TextTextureObject {
 class IVW_MODULE_FONTRENDERING_API TextRenderer {
 public:
     TextRenderer(const std::string& fontPath = util::getDefaultFontPath() + "/arial.ttf");
-    virtual ~TextRenderer();
+    TextRenderer(const TextRenderer& rhs) = delete;
+    TextRenderer(TextRenderer&& rhs);
+    TextRenderer& operator=(const TextRenderer& rhs) = delete;
+    TextRenderer& operator=(TextRenderer&& rhs) noexcept;
+    ~TextRenderer();
 
     /**
      * \brief replace the currently loaded font face with a new one
--- a/modules/fontrendering/include/modules/fontrendering/textrenderer.h
+++ b/modules/fontrendering/include/modules/fontrendering/textrenderer.h
@@ -111,7 +111,7 @@ class IVW_MODULE_FONTRENDERING_API TextRenderer {
 public:
     TextRenderer(const std::string& fontPath = util::getDefaultFontPath() + "/arial.ttf");
     TextRenderer(const TextRenderer& rhs) = delete;
-    TextRenderer(TextRenderer&& rhs);
+    TextRenderer(TextRenderer&& rhs) noexcept;
     TextRenderer& operator=(const TextRenderer& rhs) = delete;
     TextRenderer& operator=(TextRenderer&& rhs) noexcept;
     ~TextRenderer();
--- a/modules/fontrendering/include/modules/fontrendering/textrenderer.h
+++ b/modules/fontrendering/include/modules/fontrendering/textrenderer.h
@@ -109,7 +109,7 @@ struct TextTextureObject {
  */
 class IVW_MODULE_FONTRENDERING_API TextRenderer {
 public:
-    TextRenderer(const std::string &fontPath = util::getDefaultFontPath() + "/arial.ttf");
+    TextRenderer(const std::string& fontPath = util::getDefaultFontPath() + "/arial.ttf");
     virtual ~TextRenderer();
 
     /**
@@ -119,9 +119,9 @@ public:
      * @throws Exception      if the font file could not be opened
      * @throws FileException  if the font format is unsupported
      */
-    void setFont(const std::string &fontPath);
+    void setFont(const std::string& fontPath);
 
-    void render(const std::string &str, const vec2 &posf, const vec2 &scale, const vec4 &color);
+    void render(const std::string& str, const vec2& posf, const vec2& scale, const vec4& color);
 
     /**
      * \brief renders the given string with the specified color at position x, y in normalized
@@ -133,7 +133,7 @@ public:
      * @param scale  scaling factor from screen space (pixel) to normalized device coords
      * @param color  color of rendered text
      */
-    void render(const std::string &str, float x, float y, const vec2 &scale, const vec4 &color);
+    void render(const std::string& str, float x, float y, const vec2& scale, const vec4& color);
 
     /**
      * \brief renders the given string with the specified color into a texture.
@@ -143,11 +143,11 @@ public:
      * @param color  color of rendered text
      * @param clearTexture   if true, the texture is cleared before rendering the text
      */
-    void renderToTexture(std::shared_ptr<Texture2D> texture, const std::string &str,
-                         const vec4 &color, bool clearTexture = true);
+    void renderToTexture(std::shared_ptr<Texture2D> texture, const std::string& str,
+                         const vec4& color, bool clearTexture = true);
 
-    void renderToTexture(const TextTextureObject &texObject, const std::string &str,
-                         const vec4 &color, bool clearTexture = true);
+    void renderToTexture(const TextTextureObject& texObject, const std::string& str,
+                         const vec4& color, bool clearTexture = true);
     /**
      * \brief renders the given string with the specified color into a subregion of the texture.
      *
@@ -158,19 +158,19 @@ public:
      * @param color          color of rendered text
      * @param clearTexture   if true, the texture is cleared before rendering the text
      */
-    void renderToTexture(const TextTextureObject &texObject, const size2_t &origin,
-                         const size2_t &size, const std::string &str, const vec4 &color,
+    void renderToTexture(const TextTextureObject& texObject, const size2_t& origin,
+                         const size2_t& size, const std::string& str, const vec4& color,
                          bool clearTexture = true);
-    void renderToTexture(std::shared_ptr<Texture2D> texture, const size2_t &origin,
-                         const size2_t &size, const std::string &str, const vec4 &color,
+    void renderToTexture(std::shared_ptr<Texture2D> texture, const size2_t& origin,
+                         const size2_t& size, const std::string& str, const vec4& color,
                          bool clearTexture = true);
 
-    void renderToTexture(std::shared_ptr<Texture2D> texture, const std::vector<size2_t> &origin,
-                         const std::vector<size2_t> &size, const std::vector<std::string> &str,
-                         const vec4 &color, bool clearTexture = true);
+    void renderToTexture(std::shared_ptr<Texture2D> texture, const std::vector<size2_t>& origin,
+                         const std::vector<size2_t>& size, const std::vector<std::string>& str,
+                         const vec4& color, bool clearTexture = true);
 
     void renderToTexture(std::shared_ptr<Texture2D> texture,
-                         const std::vector<TexAtlasEntry> &entries, bool clearTexture = true);
+                         const std::vector<TexAtlasEntry>& entries, bool clearTexture = true);
 
     /**
      * \brief computes the glyph bounding box of a given string in normalized device coordinates
@@ -185,7 +185,7 @@ public:
      *
      * \see computeBoundingBox
      */
-    vec2 computeTextSize(const std::string &str, const vec2 &scale);
+    vec2 computeTextSize(const std::string& str, const vec2& scale);
 
     /**
      * \brief computes the glyph bounding box of a given string in pixels (screen space). The
@@ -197,7 +197,7 @@ public:
      *
      * \see computeBoundingBox
      */
-    size2_t computeTextSize(const std::string &str);
+    size2_t computeTextSize(const std::string& str);
 
     /**
      * \brief computes the bounding boxes of both text and all glyphs for a given string in pixels
@@ -211,7 +211,7 @@ public:
      * @param str  input string
      * @return bounding box of the given string
      */
-    TextBoundingBox computeBoundingBox(const std::string &str);
+    TextBoundingBox computeBoundingBox(const std::string& str);
 
     void setFontSize(int val);
     int getFontSize() const { return fontSize_; }
@@ -241,7 +241,7 @@ public:
      */
     int getBaseLineDescender() const;
 
-    void configure(const FontSettings &settings);
+    void configure(const FontSettings& settings);
 
 protected:
     static std::shared_ptr<Shader> getShader();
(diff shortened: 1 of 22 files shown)


modules/plottinggl/src/utils/axisrenderer.cpp
```diff
@@ -190,11 +212,11 @@ void AxisRendererBase::renderAxis(Camera* camera, const vec3& start, const vec3&
         drawer.draw();
     };
 
-    drawMesh(axisMesh, settings_.getWidth(), false);
+    drawMesh(axisMesh, settings_.get().getWidth(), false);
```
ResearchDaniel, line 194: Possible to use settings_->getXX instead of settings.get().getXX

modules/fontrendering/include/modules/fontrendering/textrenderer.h
```diff
@@ -316,7 +316,7 @@ class IVW_MODULE_FONTRENDERING_API TextRenderer {
     int fontSize_;        //<! font size in pixel
     double lineSpacing_;  //!< spacing between two lines in percent (default = 0.2)
 
-    const int glyphMargin_;
+    static constexpr int glyphMargin_ = 2;
```
ResearchDaniel: Is this really the same for all instances? 
That probably deserves a comment in that case.

modules/qtwidgets/src/properties/collapsiblegroupboxwidgetqt.cpp
```diff
@@ -515,6 +516,7 @@ void CollapsibleGroupBoxWidgetQt::addButtonLayout(QGridLayout* layout, int row,
     auto removePropertyBtn = createButton("removeListItemButton", str);
 
     connect(removePropertyBtn, &QToolButton::clicked, this, [layout, prop, buttonWidget]() {
+        RenderContext::getPtr()->activateDefaultRenderContext();
```
ResearchDaniel, line 521: This (workaround) deserves a comment

modules/qtwidgets/src/properties/listpropertywidgetqt.cpp
```diff
@@ -71,6 +72,7 @@ ListPropertyWidgetQt::ListPropertyWidgetQt(ListProperty* property)
 
     connect(addItemButton_, &QToolButton::clicked, this, [&]() {
         if (listProperty_ && listProperty_->getPrefabCount() > 0) {
+            RenderContext::getPtr()->activateDefaultRenderContext();
```
ResearchDaniel, line 77: This (workaround) deserves a comment

ResearchDaniel (dismissed):
Some changes that could be made

modules/plottinggl/src/utils/axisrenderer.cpp
martinfalk: `settings_`  is a `std::reference_wrapper<AxisSettings>` and thus needs `get()` to get the reference.

modules/fontrendering/include/modules/fontrendering/textrenderer.h
martinfalk: it was a `const 2` previously set in the constructor. And the name is exactly what it is. The margin/spacing around a glyph in the atlas to prevent interpolation artifacts.

modules/qtwidgets/src/properties/collapsiblegroupboxwidgetqt.cpp
martinfalk: right. Or rather a better concept of not using (and enforcing) GL-related things within properties :)

modules/fontrendering/include/modules/fontrendering/textrenderer.h
martinfalk: fixed.

modules/qtwidgets/src/properties/collapsiblegroupboxwidgetqt.cpp
martinfalk: fixed

modules/opengl/src/buffer/framebufferobject.cpp
```diff
@@ -78,7 +78,7 @@ void FrameBufferObject::defineDrawBuffers() {
 }
 
 void FrameBufferObject::deactivate() {
-    if (isActive()) {
+    if (isActive() && (static_cast<GLuint>(prevFbo_) != id_)) {
```
petersteneteg: Should the same thing be done in activate?

modules/plotting/src/properties/marginproperty.cpp
```diff
@@ -42,10 +42,10 @@ MarginProperty::MarginProperty(
     float left, InvalidationLevel invalidationLevel /*= InvalidationLevel::InvalidOutput*/,
     PropertySemantics semantics /*= PropertySemantics::Default*/)
     : CompositeProperty(identifier, displayName)
-    , top_("top", "Top", top, 0.0f, 100.0f, 1.0f, invalidationLevel, semantics)
-    , right_("right_", "Right", right, 0.0f, 100.0f, 1.0f, invalidationLevel, semantics)
-    , bottom_("bottom", "Bottom", bottom, 0.0f, 100.0f, 1.0f, invalidationLevel, semantics)
-    , left_("left", "Left", left, 0.0f, 100.0f, 1.0f, invalidationLevel, semantics) {
+    , top_("top", "Top", top, 0.0f, 200.0f, 1.0f, invalidationLevel, semantics)
```
petersteneteg: set the constraint ConstraintBehavior to ignore on max

modules/plottinggl/src/processors/scatterplotmatrixprocessor.cpp
```diff
@@ -179,7 +180,7 @@ ScatterPlotMatrixProcessor::ScatterPlotMatrixProcessor()
 }
 
 template <typename T>
-struct RangeIterator : public std::iterator<std::forward_iterator_tag, T, T, const T *, T> {
+struct RangeIterator : public std::iterator<std::forward_iterator_tag, T, T, const T*, T> {
```
petersteneteg: remove unused

modules/plottinggl/src/utils/axisrenderer.cpp
```diff
@@ -122,6 +122,28 @@ AxisRendererBase::AxisRendererBase(const AxisSettings& settings)
 AxisRendererBase::AxisRendererBase(const AxisRendererBase& rhs)
     : settings_{rhs.settings_}, shaders_{getShaders()} {}
 
+AxisRendererBase::AxisRendererBase(AxisRendererBase&& rhs)
```
petersteneteg: use defualt?

modules/plottinggl/src/utils/axisrenderer.cpp
```diff
@@ -122,6 +122,28 @@ AxisRendererBase::AxisRendererBase(const AxisSettings& settings)
...
+    : settings_{rhs.settings_}
+    , textRenderer_{std::move(rhs.textRenderer_)}
+    , quadRenderer_{std::move(rhs.quadRenderer_)}
+    , axisPickingId_{rhs.axisPickingId_}
+    , meshes_{std::move(rhs.meshes_)}
+    , caption_{std::move(rhs.caption_)}
+    , shaders_{std::move(rhs.shaders_)} {}
+
+AxisRendererBase& AxisRendererBase::operator=(AxisRendererBase&& rhs) noexcept {
```
petersteneteg: use default?

modules/plottinggl/src/utils/axisrenderer.cpp
```diff
@@ -209,9 +231,20 @@ AxisRenderer::AxisRenderer(const AxisSettings& settings)
                        [&](auto&& p) { return p.second; });
     }} {}
 
-void AxisRenderer::render(const size2_t& outputDims, const size2_t& startPos, const size2_t& endPos,
+AxisRenderer::AxisRenderer(AxisRenderer&& rhs)
+    : AxisRendererBase(rhs), labels_{std::move(rhs.labels_)} {}
```
petersteneteg: needs AxisRendererBase(std::move(rhs)) or just use default

modules/plottinggl/src/utils/axisrenderer.cpp
```diff
@@ -209,9 +231,20 @@ AxisRenderer::AxisRenderer(const AxisSettings& settings)
...
+
+AxisRenderer& AxisRenderer::operator=(AxisRenderer&& rhs) noexcept {
```
petersteneteg: default

modules/plottinggl/src/utils/axisrenderer.cpp
```diff
@@ -358,9 +393,20 @@ AxisRenderer3D::AxisRenderer3D(const AxisSettings& settings)
                        [](auto& tick) { return tick.second; });
     }} {}
 
+AxisRenderer3D::AxisRenderer3D(AxisRenderer3D&& rhs)
```
petersteneteg: same as above

petersteneteg (changes requested):

modules/plottinggl/src/utils/axisrenderer.cpp
martinfalk: with default, the following assertions fail:
```c++
    static_assert(std::is_move_assignable_v<AxisRendererBase>);
    static_assert(std::is_move_assignable_v<AxisRenderer>);
```

modules/opengl/src/buffer/framebufferobject.cpp
martinfalk: potentially yes. The [OpenGL specs](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glBindFramebuffer.xhtml) are not referring to binding the same fbo.

modules/plottinggl/src/utils/axisrenderer.cpp
martinfalk: fixed

modules/plottinggl/src/utils/axisrenderer.cpp
martinfalk: fixed

ResearchDaniel (dismissed):

modules/opengl/src/image/imagecompositor.cpp
```diff
@@ -69,4 +85,36 @@ void ImageCompositor::composite(ImageInport& source, ImageOutport& target, Image
...
+        colorTex_ = *layer->getTexture();
+    }
+    {
+        auto layer = glrep->getDepthLayerGL();
+        IVW_ASSERT(layer, "no depth layer in target image");
+        depthTex_ = *layer->getTexture();
+    }
+    {
+        auto layer = glrep->getPickingLayerGL();
+        IVW_ASSERT(layer, "no picking layer in target image");
```
ResearchDaniel, line 99: This will be the case if picking is not enabled in the settings so this should not be an assert.

modules/opengl/src/image/imagecompositor.cpp
```diff
@@ -53,14 +64,19 @@ void ImageCompositor::composite(const Image& source, Image& target, ImageType ty
 
 void ImageCompositor::composite(ImageInport& source, ImageOutport& target, ImageType type) {
     if (source.isReady() && target.hasData()) {
+        copyTextures(*target.getData());
```
ResearchDaniel, line 67: is it not possible to call composite(*source.getData(), *target.getData(), type) to avoid code duplication?

modules/opengl/src/image/imagecompositor.cpp
```diff
@@ -53,14 +64,19 @@ void ImageCompositor::composite(const Image& source, Image& target, ImageType ty
...
+
         utilgl::activateTarget(target, type);
         shader_.activate();
+        utilgl::GlBoolState depthTest(GL_DEPTH_TEST, false);
+
+        GLint currentFbo = 0;
```
ResearchDaniel: Remove?

modules/opengl/include/modules/opengl/image/imagecompositor.h
```diff
@@ -50,7 +52,13 @@ class IVW_MODULE_OPENGL_API ImageCompositor {
     void composite(ImageInport& source, ImageOutport& target, ImageType type);
 
 private:
+    void copyTextures(const Image& target);
+    void bindTextures(TextureUnitContainer& cont, const std::string& id);
+
     Shader shader_;
+    Texture2D colorTex_;
```
ResearchDaniel, line 59: is not the interface for `composite(...)` incorrect?
It is not obvious that it should combine source and target into target.

It should be `composite(const Image& first, const Image& second, Image& target, ImageType type);` in my opinion.
Not part of this PR, but the documentation about what the supplied fragment shader should take as input and produce is missing.

ResearchDaniel (changes requested):
I think that the ImageCompositor has the wrong interface, so the intermediate copying is the wrong fix

modules/opengl/include/modules/opengl/image/imagecompositor.h
martinfalk: This code was >5 years old and it broke when cleaning up the FBO state. That is two regression tests using the Geometry Picking processor (aka. `meshpicking.h`) were failing.
This is nothing more than a hot fix. I don't even know why we would need such a special compositing in the first place.
Even the shader looks like it does nothing special, so ActivateAndCopy might actually do the expected thing.

https://github.com/inviwo/inviwo/blob/master/modules/opengl/glsl/composite.frag

modules/opengl/src/image/imagecompositor.cpp
martinfalk: hmm, left-over from debugging.

modules/opengl/src/image/imagecompositor.cpp
martinfalk: not really. `target.getData()` returns a `std::shared_ptr<const Image>`, note the `const` here. Hence, it will not be possible to render into the image.

modules/opengl/src/image/imagecompositor.cpp
martinfalk: Doesn't an image "always" have a picking buffer, even when disabled? If not, then this must be documented somewhere.

martinfalk: Having assertions enforce and picking turned off, image outports still have a picking buffer.

Try: Random Sphere Generator, Sphere Renderer, Canvas

Canvas shows the "picking buffer"
image

martinfalk: see also info from `Image Information` processor
image

martinfalk:
> I think that the ImageCompositor has the wrong interface, so the intermediate copying is the wrong fix

It has been like this for years ;-) Processors using this ImageCompositor are:
- `DrawLines`
- `DrawPoints`
- `ImageCompositeProcessorGL`
- `MeshPicking`

Potentially all these processors do weird stuff to get the compositing done. For example, why not use regular blending with the depth test?

modules/opengl/src/image/imagecompositor.cpp
ResearchDaniel: Hmm, then I think that the picking setting can be removed

ResearchDaniel (dismissed):
Ok, let's revise the ImageComposite in another PR

modules/plottinggl/src/utils/axisrenderer.cpp
petersteneteg: why? your implementation looks like a default one?

petersteneteg: or have you skipped some members?

modules/opengl/src/image/imagecompositor.cpp
petersteneteg: the settings has nothing to do with the image layer, but with propagation of picking events

modules/opengl/include/modules/opengl/image/imagecompositor.h
petersteneteg: where is it use can we remove it?

petersteneteg:
> * `DrawLines`
> * `DrawPoints`

There two are very old and also quite embarrassing. I would prefer to remove them

> * `ImageCompositeProcessorGL`
> * `MeshPicking`

These two we could maybe update and fix?

modules/plottinggl/src/utils/axisrenderer.cpp
martinfalk: using default now after fixing all members.

petersteneteg (dismissed):

modules/opengl/src/buffer/framebufferobject.cpp
martinfalk: addressed in 23d2fbecb6fb40619e99ea94dc8b49243c356f16

petersteneteg (approved):

PR #1025, opened 2020-09-09
unicode

petersteneteg:
Better error message when registering a processor
Fixes for unicode conversion.in filesystem

### DIFF
--- a/include/inviwo/core/processors/processorutils.h
+++ b/include/inviwo/core/processors/processorutils.h
@@ -119,7 +119,8 @@ std::unique_ptr<T> makeProcessor(ivec2 pos, Args&&... args) {
  * @param app the InviwoApplication needed to get the modules
  * @return the InviwoModule that registered the processor or nullptr if not found
  */
-IVW_CORE_API InviwoModule* getProcessorModule(const Processor* processor, const InviwoApplication& app);
+IVW_CORE_API InviwoModule* getProcessorModule(const Processor* processor,
+                                              const InviwoApplication& app);
 
 /**
  * @brief Find which module that registered a processor
--- a/include/inviwo/core/processors/processorutils.h
+++ b/include/inviwo/core/processors/processorutils.h
@@ -35,11 +35,13 @@
 #include <inviwo/core/util/glmvec.h>
 
 #include <type_traits>
+#include <string_view>
 
 namespace inviwo {
 
 class Processor;
 class ProcessorMetaData;
+class InviwoModule;
 
 namespace util {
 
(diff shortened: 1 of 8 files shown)


martinfalk (dismissed):

include/inviwo/core/processors/processorutils.h
```diff
@@ -111,6 +113,23 @@ std::unique_ptr<T> makeProcessor(ivec2 pos, Args&&... args) {
     return p;
 }
 
+/**
+ * @brief Find which module that registered a processor
+ * @param processor the processor to look for
+ * @param app the InviwoApplication needed to get the modules
+ * @return the InviwoModule that registered the processor or nullptr if not found
+ */
+IVW_CORE_API InviwoModule* getProcessorModule(Processor* processor, const InviwoApplication& app);
```
ResearchDaniel: should be const Processor*
Implementation should check that processor is not null.

ResearchDaniel (changes requested):

include/inviwo/core/processors/processorutils.h
petersteneteg: fixed

src/core/util/stringconversion.cpp
```diff
@@ -60,9 +60,20 @@ std::wstring toWstring(const std::string& str) {
...
     auto state = std::mbstate_t();
     auto sptr = str.data();
+    const char* loc = nullptr;
     size_t len = std::mbsrtowcs(nullptr, &sptr, 0, &state);
+    if (len == static_cast<std::size_t>(-1)) {
+        loc = std::setlocale(LC_CTYPE, nullptr);
+        std::setlocale(LC_CTYPE, "en_US.UTF-8");
+        len = std::mbsrtowcs(nullptr, &sptr, 0, &state);
+        if (len == static_cast<std::size_t>(-1)) {
+            if (loc) std::setlocale(LC_CTYPE, loc);
```
ResearchDaniel, line 71: Should be an else here for throwing the exception?

include/inviwo/core/processors/processorutils.h
ResearchDaniel: No changes pushed yet...

ResearchDaniel (changes requested):

src/core/util/stringconversion.cpp
petersteneteg: no, I just want to restore the local before I leave by throwing

petersteneteg:
@ResearchDaniel done

include/inviwo/core/processors/processorutils.h
ResearchDaniel: Still not changed

petersteneteg: that was very strange, I had fixed it locally and had now changes according to git, but it still was not committed, I restarted VS and the changes was gone. seems VS did something strange...
Should be fixed now

ResearchDaniel (dismissed):

ResearchDaniel (approved):

PR #1028, opened 2020-09-11
WebBrowser: registering of processor callbacks

martinfalk:
Added functionality for registering processor callbacks in the browser and invoking them from Javascript. Useful for communicating with the processor. For example in brushing&linking by sending indices of selected elements.

```c++
    auto browserClient = app->getModuleByType<WebBrowserModule>()->getBrowserClient();
    ...
    browserClient->setBrowserParent(browser_, this);
    selectionCallback_ = browserClient->registerCallback(browser_, "selectionChanged",
                                    [&](const std::string& str) { LogInfo("callback: " + str); });
```

Can be invoked via 
```javascript
var data = [ 1, 2, 3, 4, 5 ];
inviwo.invokeCallback("myCallback", data);
```
Payload is stringified into JSON.

### DIFF
--- a/modules/webbrowser/data/js/inviwoapi.js
+++ b/modules/webbrowser/data/js/inviwoapi.js
@@ -162,6 +162,22 @@ class InviwoAPI {
             }
         });
     }
+    /*
+     * Call a previously registered processor callback.
+     * @param callback  name of the callback
+     * @param data      payload will be stringified into JSON
+     */
+    async invokeCallback(callback, data) {
+        window.cefQuery({
+            request: JSON.stringify({
+                'command': 'callback',
+                'callback': callback,
+                'data': JSON.stringify(data)
+            }),
+            onSuccess: function (response) { },
+            onFailure: function (error_code, error_message) { }
+        });
+    }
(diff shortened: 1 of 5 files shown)


modules/webbrowser/data/js/inviwoapi.js
```diff
@@ -162,6 +162,22 @@ class InviwoAPI {
             }
         });
     }
+    /*
+     * Call a previously registered processor callback.
+     * @param callback  name of the callback
+     * @param data      payload will be stringified into JSON
```
ResearchDaniel, line 168: Why not in JSON format already? 
Would make it more like the other functions.

ResearchDaniel (dismissed):
Nice

modules/webbrowser/data/js/inviwoapi.js
martinfalk: doesn't really matter, does it? If you put in a string, it stays a string. Putting in an object transforms it to JSON.

ResearchDaniel (approved):

PR #1029, opened 2020-09-17
Feature/dataframe filtering

martinfalk:


modules/dataframe/include/inviwo/dataframe/util/dataframeutil.h
```diff
@@ -109,8 +111,66 @@ std::shared_ptr<DataFrame> IVW_MODULE_DATAFRAME_API
...
+ *                              [](const int& arg) { return arg == 5;  }, 
+ *                              [](const auto&) { return true; });
+ * \endcode
+ *
+ * @param col   column containing data for filtering
+ * @param pred  predicate to check values from \p col
+ * @return list of row indices where rows fulfill the predicate
+ */
+template <typename Pred>
+std::vector<size_t> filteredRows(std::shared_ptr<const Column> col, Pred pred);
```
ResearchDaniel, line 142: Can usage be simplified if the column type is somehow specified? E.g. passing the filter along to the dispatching?
And/or might be better to have an example that checks for floating point/integral numbers.

ResearchDaniel (dismissed):

modules/dataframe/include/inviwo/dataframe/util/dataframeutil.h
martinfalk: The main problem here is that you don't know the type of your column. Hence, the predicate needs to handle all the different types, mainly distinguish between strings/categorical and numerical values.
If you know the column type, then one could potentially have a templated function based on the type (which is needed for the predicate). 
Making such a function a member function of Column does not work as it cannot be virtual due to the template.

ResearchDaniel (approved):

petersteneteg (approved):

PR #1030, opened 2020-09-17
Core: PickingMapper fix

martinfalk:
minor fix preventing problems when initializing a `PickingMapper` with size 0.

### DIFF
--- a/src/core/interaction/pickingmapper.cpp
+++ b/src/core/interaction/pickingmapper.cpp
@@ -33,8 +33,6 @@
 
 namespace inviwo {
 
-PickingMapper::PickingMapper(PickingManager* manager) : manager_(manager) {}
-
 PickingMapper::PickingMapper(Processor* processor, size_t size,
                              std::function<void(PickingEvent*)> callback, PickingManager* manager)
     : manager_(manager)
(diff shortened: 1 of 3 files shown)


ResearchDaniel (changes requested):
I think that we need to take care of possible nullptrs in all places, e.g.:
```
size_t PickingMapper::getPickingId(size_t id)
vec3 getColor(size_t id) const
size_t PickingMapper::getSize() const
```

martinfalk:
> I think that we need to take care of possible nullptrs in all places, e.g.:

Good point. Will take care of it.

petersteneteg (dismissed):

src/core/interaction/pickingmapper.cpp
```diff
@@ -89,21 +87,40 @@ void PickingMapper::resize(size_t newSize) {
...
 void PickingMapper::setEnabled(bool enabled) {
     if (pickingAction_) pickingAction_->setEnabled(enabled);
 }
 
-size_t PickingMapper::getPickingId(size_t id) const { return pickingAction_->getPickingId(id); }
+size_t PickingMapper::getPickingId(size_t id) const {
+    if (pickingAction_) {
+        return pickingAction_->getPickingId(id);
+    } else {
+        return 0;
```
ResearchDaniel: This might result in errors later because picking id 0 is valid?
Would it be better to throw or assert?
Documentation should be updated :)

ResearchDaniel (dismissed):
+1

src/core/interaction/pickingmapper.cpp
martinfalk: Hm, valid point. Going for throwing an exception here would be best then.

petersteneteg: In principle the only time pickingAction_ can be nullptr now is in a moved from object right, then you are not really allowed to call those function anyway...

ResearchDaniel: same for the color then

martinfalk: So, how should we handle those cases?

petersteneteg: it's a precondition that's broken. There are a few valid option, do nothing, assert, throw, return something anyway. I don't think we have a standard way of handling this in our guidelines, But I think I would lean towards an assert

ResearchDaniel (approved):

petersteneteg (approved):

PR #1031, opened 2020-09-19
Fixing same-depth fragments

AnkeAnke:
Fixing fragment list rendering for multiple fragments of exact same depth.
MartinZ-Fighting
Lower right: Twice the same mesh at the same time.

There is still z-fighting, in the sense that the fragment first in the linked list is alpha-blended first. You see that when changing the order of rasterizations in the port (disconnect and reconnect) - one mesh will be "before" the other. 
Not sure if we need that fixed, it's a border case and I find it more of a hidden feature right now. Opinions?

### DIFF
--- a/modules/meshrenderinggl/glsl/oit/sort.glsl
+++ b/modules/meshrenderinggl/glsl/oit/sort.glsl
@@ -46,7 +46,7 @@ void fillFragmentArray(uint idx, out int numFrag);
 // Fill local memory array with lowest values
 void fillFragmentArraySortedUntilDepth(uint pixelIdx, float maxDepth, out int numFrag);
 // Selection sort, allowing to get the first elements of a larger series
-vec4 selectionSortNext(uint headPtr, float lastDepth);
+vec4 selectionSortNext(uint headPtr, float lastDepth, inout uint lastPtr);
 
 // Bubble sort used to sort fragments
 void bubbleSort(int array_size) {
@@ -105,9 +105,10 @@ void fillFragmentArraySortedUntilDepth(uint pixelIdx, float maxDepth, out int nu
     // Load fragments into a local memory array for sorting
     numFrag = 0;
     float depth = 0.0;
+    uint lastPtr = 0;
 
     while (numFrag < ABUFFER_SIZE) {
-        vec4 val = selectionSortNext(pixelIdx, depth);
+        vec4 val = selectionSortNext(pixelIdx, depth, lastPtr);
         depth = val.y;
         if (depth < 0.0 || depth > maxDepth) return;
         fragmentList[numFrag] = val;
@@ -115,17 +116,22 @@ void fillFragmentArraySortedUntilDepth(uint pixelIdx, float maxDepth, out int nu
     }
 }
 
-vec4 selectionSortNext(uint headPtr, float lastDepth) {
+vec4 selectionSortNext(uint headPtr, float lastDepth, inout uint lastPtr) {
     vec4 closestVal = vec4(-1.0);
     float minDepth = 1.0;
+    bool passedLastPtr = false;
+    uint minPtr = 0;
     while (headPtr != 0) {
         vec4 val = readPixelStorage(headPtr - 1);
-        if (val.y > lastDepth && val.y < minDepth) {
+        if (headPtr == lastPtr) passedLastPtr = true;
+        else if (val.y < minDepth && (val.y > lastDepth || (passedLastPtr && val.y == lastDepth) )) {
             minDepth = val.y;
             closestVal = val;
+            minPtr = headPtr;
         }
         headPtr = floatBitsToUint(val.x);
     }
+    lastPtr = minPtr;
     return closestVal;
 }
 
--- a/modules/meshrenderinggl/glsl/oit/sort.glsl
+++ b/modules/meshrenderinggl/glsl/oit/sort.glsl
@@ -45,7 +45,8 @@ void bitonicSort(int n);
 void fillFragmentArray(uint idx, out int numFrag);
 // Fill local memory array with lowest values
 void fillFragmentArraySortedUntilDepth(uint pixelIdx, float maxDepth, out int numFrag);
-// Selection sort, allowing to get the first elements of a larger series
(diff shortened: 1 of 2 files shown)


modules/meshrenderinggl/glsl/oit/sort.glsl
```diff
@@ -105,27 +105,33 @@ void fillFragmentArraySortedUntilDepth(uint pixelIdx, float maxDepth, out int nu
...
         numFrag++;
     }
 }
 
-vec4 selectionSortNext(uint headPtr, float lastDepth) {
+vec4 selectionSortNext(uint headPtr, float lastDepth, inout uint lastPtr) {
     vec4 closestVal = vec4(-1.0);
     float minDepth = 1.0;
+    bool passedLastPtr = false;
+    uint minPtr = 0;
```
ResearchDaniel, line 124: Why not headPtr?

modules/meshrenderinggl/glsl/oit/sort.glsl
```diff
@@ -105,27 +105,33 @@ void fillFragmentArraySortedUntilDepth(uint pixelIdx, float maxDepth, out int nu
...
     while (headPtr != 0) {
         vec4 val = readPixelStorage(headPtr - 1);
-        if (val.y > lastDepth && val.y < minDepth) {
+        if (headPtr == lastPtr) passedLastPtr = true;
+        else if (val.y < minDepth && (val.y > lastDepth || (passedLastPtr && val.y == lastDepth) )) {
```
ResearchDaniel, line 132: This is a bit hard to understand. Consider adding comments about the reasoning here.

modules/meshrenderinggl/glsl/oit/sort.glsl
```diff
@@ -46,7 +46,7 @@ void fillFragmentArray(uint idx, out int numFrag);
 // Fill local memory array with lowest values
 void fillFragmentArraySortedUntilDepth(uint pixelIdx, float maxDepth, out int numFrag);
 // Selection sort, allowing to get the first elements of a larger series
-vec4 selectionSortNext(uint headPtr, float lastDepth);
+vec4 selectionSortNext(uint headPtr, float lastDepth, inout uint lastPtr);
```
ResearchDaniel, line 50: Should also add documentation about the new variable.

modules/meshrenderinggl/glsl/oit/sort.glsl
```diff
@@ -105,27 +105,33 @@ void fillFragmentArraySortedUntilDepth(uint pixelIdx, float maxDepth, out int nu
...
-vec4 selectionSortNext(uint headPtr, float lastDepth) {
+vec4 selectionSortNext(uint headPtr, float lastDepth, inout uint lastPtr) {
     vec4 closestVal = vec4(-1.0);
     float minDepth = 1.0;
+    bool passedLastPtr = false;
+    uint minPtr = 0;
     while (headPtr != 0) {
         vec4 val = readPixelStorage(headPtr - 1);
-        if (val.y > lastDepth && val.y < minDepth) {
+        if (headPtr == lastPtr) passedLastPtr = true;
```
ResearchDaniel, line 129: No need for the extra bool. Just skip if they are equal

ResearchDaniel (changes requested):

modules/meshrenderinggl/glsl/oit/sort.glsl
AnkeAnke: We need to keep the pointer to the element we return for correct ordering later on. See other comments.

modules/meshrenderinggl/glsl/oit/sort.glsl
AnkeAnke: Comments added in a345371f90a65a87225604c36ef357e9879e292e.
Basically, this all is here to solve the problem of fragments at the exact same depth.
Instead of relying on the depth for order alone, we need to keep tabs on what the last selected fragment was. In case we encounter the same depth (as `lastDepth`), we need to make sure it is not the same fragment (therefor, taking and updating `lastPtr`). Also, we need to make sure that fragment comes behind the last one, in order of list traversal - else, we'd get an endless loop between any two fragments of same depth. Therefor we save `passedLastPtr`.

modules/meshrenderinggl/glsl/oit/sort.glsl
AnkeAnke: Should be clearer now.

modules/meshrenderinggl/glsl/oit/sort.glsl
AnkeAnke: See other comment.

ResearchDaniel: Should not this suffice:
if (val.y < minDepth && (val.y > lastDepth || (headPtr > lastPtr && val.y == lastDepth) ))

petersteneteg:
I still don't understand your hubristics. Is not a stable sort what you are looking for here?

modules/meshrenderinggl/glsl/oit/sort.glsl
AnkeAnke: We can't be sure the pointer is in the same order as the linked list. I guess it would mostly be, but it's not a given.

AnkeAnke:
@petersteneteg Yes, I'm looking for a stable selection sort, where the naïve approach just discards fragments of same depth. Is there an obvious easier solution I'm missing?

petersteneteg:
Would it be possible to use a stable sort algorithm like insertion sort instead? What happens above if more then 2 elements have the same depth?

AnkeAnke:
Insertion sort requires some memory, which would just go to waste. I basically just give a pointer to the last element selected to the selection sort, that's much less overhead.
The changes can handle any number of fragments at the same depth.

ResearchDaniel (approved):

petersteneteg:
Ahh i see, i assumed there was an actual SelectionSort being used but that is not a selection sort, there are no swaps hence it will probably be stable. Maybe it should be named something else. Also can we remove all the other algos in that file that are not used?

PR #1036, opened 2020-09-28
mac linux fixes

petersteneteg:

### DIFF
--- a/cmake/vcpkghelpers.cmake
+++ b/cmake/vcpkghelpers.cmake
@@ -27,6 +27,11 @@
 # 
 #################################################################################
 
+# Retrive the git sha of the VCPKG installation, once for later use in ivw_vcpkg_install
+if(VCPKG_TOOLCHAIN)
+    ivw_git_get_hash(${_VCPKG_ROOT_DIR} ivw_vcpkg_sha)
+endif()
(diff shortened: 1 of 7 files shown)


cmake/vcpkghelpers.cmake
```diff
@@ -76,6 +75,9 @@ function(ivw_vcpkg_paths)
     endif()
 endfunction()
 
+if(VCPKG_TOOLCHAIN)
```
ResearchDaniel: Why outside any function?

ResearchDaniel (changes requested):
Looks good except some free floating stuff

cmake/vcpkghelpers.cmake
petersteneteg: I only need to do it once, then I can just use the ivw_vcpkg_sha var in the function. Think it gets captured kind of.

petersteneteg: hence it is not by accident...

ResearchDaniel: I think that you should put it in the top and state that it is for initialization in that case

ResearchDaniel (approved):

PR #1040, opened 2020-09-30
hdf and opengl fixes

petersteneteg:


ResearchDaniel (dismissed):

modules/opengl/src/shader/shader.cpp
```diff
@@ -169,8 +129,8 @@ Shader::Shader(const char *vertexFilename, const char *geometryFilename,
 
 Shader::Shader(const Shader &rhs) : program_{rhs.program_}, warningLevel_{rhs.warningLevel_} {
     for (auto &elem : rhs.shaderObjects_) {
-        shaderObjects_.emplace(
-            elem.first, ShaderAttachment(this, std::make_unique<ShaderObject>(elem.second.obj())));
+        shaderObjects_.emplace_back(elem);
+        attached_.emplace_back(false);
```
martinfalk, line 136: what about the callbacks here? Isn't the invariant here that `shaderObjects_`, `attached_`, and `callbacks_` have the same lengths?

modules/opengl/src/shader/shader.cpp
```diff
@@ -324,6 +318,13 @@ void Shader::rebuildShader(ShaderObject *obj) {
                   "Id: " + toString(program_.id) + ", resource: " + obj->getFileName() +
                       " successfully reloaded",
                   LogLevel::Info, LogAudience::User);
+                  
+        auto log = utilgl::getProgramInfoLog(program_.id);
+        if (!log.empty()) {
```
martinfalk, line 323: might cause some logging on Intel or AMD

martinfalk (changes requested):
nice refactoring of the shader objects

modules/opengl/src/shader/shader.cpp
petersteneteg: Yes, think I had changes that, but seems I lost some changes...

modules/opengl/src/shader/shader.cpp
petersteneteg: think it is better to log more then less in the case of OpenGL. If someone thinks it is to much then we can try an limit it

martinfalk (approved):

PR #1054, opened 2020-10-19
Camera interpolation for animation module along with various fixes

ResearchDaniel:
First stab at adding camera support in the animation module.
It is now possible to interpolate between two key frames with the camera pointing at the object x, e.g. -> x <-, without crossing x.

Features:
- Orbit/pan/tilt interpolation for the camera
- Easier to use animation manager externally by returning added key frames.
- Added CallbackTrack for programmatically reacting to an animation

Fixes:
- SequenceEditor showed deleted keyframes
- Only show selected keyframe properties
- Added OrdinalRefProperty to AnimationQt

Limitations: Only consider the lookFrom, lookAt, lookUp of a camera. The other properties can be animated through regular PropertyTracks in my opinion.


include/inviwo/core/properties/cameraproperty.h
```diff
@@ -57,6 +57,7 @@ class CameraFactory;
  */
 class IVW_CORE_API CameraProperty : public CompositeProperty, public TrackballObject {
 public:
+    using value_type = Camera;
```
petersteneteg, line 60: Will this work even if Camera is a polymorphic type?

modules/animation/include/modules/animation/datastructures/basekeyframesequence.h
```diff
@@ -168,23 +172,25 @@ void BaseKeyframeSequence<Key>::onKeyframeTimeChanged(Keyframe* key, Seconds /*o
 }
 
 template <typename Key>
-void BaseKeyframeSequence<Key>::add(std::unique_ptr<Keyframe> key) {
+Keyframe* BaseKeyframeSequence<Key>::add(std::unique_ptr<Keyframe> key) {
```
petersteneteg: since you do the dynamic cast you could return a typed KeyFrame?

modules/animation/include/modules/animation/datastructures/basetrack.h
```diff
@@ -89,10 +88,22 @@ class BaseTrack : public Track, public KeyframeSequenceObserver {
...
     virtual void add(Seconds time, bool asNewSequence) override;
-    virtual void add(std::unique_ptr<KeyframeSequence> sequence) override;
-    virtual void add(std::unique_ptr<Seq> sequence);
+    /**
+     * Add KeyframeSequence and call KeyframeSequenceObserverble::notifyKeyframeSequenceAdded
+     * @throw Exception if KeyframeSequence is not compatible with BaseTrack<Seq>
+     * @throw Exception if KeyframeSequence overlaps existing sequences
+     * @see BaseTrack::add(std::unique_ptr<Seq> sequence)
+     */
+    virtual KeyframeSequence* add(std::unique_ptr<KeyframeSequence> sequence) override;
```
petersteneteg: Should be able to return the typed seq here also

modules/animation/include/modules/animation/datastructures/basetrack.h
```diff
@@ -89,10 +88,22 @@ class BaseTrack : public Track, public KeyframeSequenceObserver {
     const_iterator begin() const;
     iterator end();
     const_iterator end() const;
-
+    /**
+     * Add a Keyframe/KeyframeSequence at time
+     */
     virtual void add(Seconds time, bool asNewSequence) override;
```
petersteneteg: Might return the seq here also?

modules/animation/include/modules/animation/datastructures/basetrack.h
```diff
@@ -89,10 +88,22 @@ class BaseTrack : public Track, public KeyframeSequenceObserver {
     const_iterator begin() const;
     iterator end();
     const_iterator end() const;
-
+    /**
+     * Add a Keyframe/KeyframeSequence at time
```
petersteneteg: Adds a default constructed KeyFrame?

modules/animation/include/modules/animation/datastructures/basetrack.h
```diff
@@ -392,6 +410,11 @@ std::unique_ptr<KeyframeSequence> BaseTrack<Seq>::remove(KeyframeSequence* seq)
     }
 }
 
+template <typename Seq>
+std::unique_ptr<typename BaseTrack<Seq>::key_type> BaseTrack<Seq>::createKeyframe(Seconds time) {
```
petersteneteg: I don't see why this function is needed?

modules/animation/include/modules/animation/datastructures/buttonkeyframe.h
```diff
@@ -0,0 +1,69 @@
...
+#include <modules/animation/datastructures/animationstate.h>
+
+#include <inviwo/core/properties/buttonproperty.h>
+
+namespace inviwo {
+
+namespace animation {
+
+/** \class ButtonKeyframe
+ * Keyframe for buttons. Button will be pressed when passing over the keyframe.
```
petersteneteg: it this direction dependent?

modules/animation/include/modules/animation/datastructures/buttonkeyframe.h
```diff
@@ -0,0 +1,69 @@
...
+ * @see Keyframe
+ */
+class IVW_MODULE_ANIMATION_API ButtonKeyframe : public BaseKeyframe {
+public:
+    // using value_type = ButtonProperty;
```
petersteneteg: remove?

modules/animation/include/modules/animation/datastructures/buttonkeyframe.h
petersteneteg: So this is fundamentally different from a PropertyTrack of ButtonProperty? Each keyframe can refer to a different property?

modules/animation/include/modules/animation/datastructures/callbackkeyframesequence.h
```diff
@@ -0,0 +1,62 @@
...
+
+namespace animation {
+
+/** \class CallbackKeyframeSequence
+ * KeyframeSequence for CallbackKeyframe.
+ * Animating over a CallbackKeyframe in the sequence will call its corresponding callbacks.
+ * @see CallbackKeyframe
+ * @see CallbackTrack
+ */
+class IVW_MODULE_ANIMATION_API CallbackKeyframeSequence
```
petersteneteg, line 45: is this classes needed or would not 
```
using CallbackKeyframeSequence = BaseKeyframeSequence<CallbackKeyframe>
```
work? 
Seems this class does not add any functionality over the base class? or am I missing something?

modules/animation/include/modules/animation/datastructures/callbacktrack.h
```diff
@@ -0,0 +1,63 @@
...
+
+/** \class CallbackTrack
+ * A track for executing callbacks when animating forward/backward.
+ * Exposes functions for adding a CallbackKeyFrame and CallbackKeyFrameSequence
+ * This track is intended to be added programatically, i.e., not through the animation user
+ * interface.
+ * @see CallbackKeyframe
+ * @see CallbackKeyframeSequence
+ */
+class IVW_MODULE_ANIMATION_API CallbackTrack : public BaseTrack<CallbackKeyframeSequence> {
```
petersteneteg, line 49: Is it needed to add classes for all these track? could one not just use a common BaseTrackImpl class?

modules/animation/include/modules/animation/datastructures/camerakeyframe.h
```diff
@@ -0,0 +1,90 @@
...
+     * \brief Get unnormalized direction of camera: lookTo - lookFrom
+     */
+    vec3 getDirection() const;
+
+    void updateFrom(const Camera& value);
+
+    virtual void serialize(Serializer& s) const override;
+    virtual void deserialize(Deserializer& d) override;
+
+    static std::string getName() { return "Camera"; }
```
petersteneteg: return const &

modules/animation/include/modules/animation/datastructures/camerakeyframe.h
```diff
@@ -0,0 +1,90 @@
...
+
+private:
+    vec3 lookFrom_ = cameradefaults::lookFrom, lookTo_ = cameradefaults::lookTo,
```
petersteneteg: pls put these on new lines

modules/animation/include/modules/animation/datastructures/propertytrack.h
```diff
@@ -265,75 +293,137 @@ class PropertyTrack : public BaseTrack<KeyframeSequenceTyped<Key>>, public BaseP
...
     }
 
+protected:
+    /*
+     * Helper function for when we know that we are between keyframes within a KeyframeSequence.
+     * Called from operator()(Seconds from, Seconds to, AnimationState state) const
+     * Provide template specialization of this method if you want custom property/sequence
+     * behaviour.
+     */
+    AnimationTimeState animateSequence(const Seq& seq, Seconds from, Seconds to,
```
petersteneteg: this seems like an awkward way of customising the behaviour. I would rather make this an external type:
```
template <typename Prop, typename Seq>
struct AnimateSequence {
    static AnimationTimeState animate(Prop* prop, const Seq& seq, Seconds from, Seconds to, AnimationState state);
}
```
that makes it much cleaner to provide a specialisation.

petersteneteg: also should it be const since it changes the property?

modules/animation/include/modules/animation/datastructures/propertytrack.h
```diff
@@ -393,62 +483,75 @@ void PropertyTrack<Prop, Key>::addKeyFrameUsingPropertyValue(
...
-template <typename Prop, typename Key>
-void PropertyTrack<Prop, Key>::deserialize(Deserializer& d) {
-    BaseTrack<KeyframeSequenceTyped<Key>>::deserialize(d);
+template <typename Prop, typename Key, typename Seq>
+void PropertyTrack<Prop, Key, Seq>::deserialize(Deserializer& d) {
+    BaseTrack<Seq>::deserialize(d);
     d.deserializeAs<Property>("property", property_);
 }
 
+// Template specialization (CameraProperty)
```
petersteneteg: make this into a proper customisation point se comment above

modules/animation/include/modules/animation/datastructures/valuekeyframesequence.h
```diff
@@ -237,6 +237,17 @@ void KeyframeSequenceTyped<Key>::deserialize(Deserializer& d) {
     }
 }
 
+template <>
```
petersteneteg: Why is this needed? did not the out ref fix that? and if not why i the out ref needed then? don't see why both would  be needed. 
Also I really dislike specialisation of class template member functions. I makes the code horrible to read, and scales very poorly, and also exposes internal state in a very bad way. If the specialisations truly are needed then make then into proper customisation points. See above

modules/animation/src/animationmanager.cpp
```diff
@@ -111,47 +111,80 @@ AnimationController& AnimationManager::getAnimationController() { return control
...
-void AnimationManager::addKeyframeCallback(Property* property, Seconds time) {
+Keyframe* AnimationManager::addKeyframe(Property* property, Seconds time) {
     auto it = trackMap_.find(property);
+    std::string interpolationErrMsg;
     try {
-        auto interpolation = getDefaultInterpolation(property);
+        std::unique_ptr<Interpolation> interpolation{nullptr};
+        try {
+            interpolation = getDefaultInterpolation(property);
+        } catch (const Exception& ex) {
```
petersteneteg: this looks a lot like control flow using exceptions, that should not be used. why not return a nullptr if there is no interpolation?

modules/animation/src/animationmanager.cpp
```diff
@@ -111,47 +111,80 @@ AnimationController& AnimationManager::getAnimationController() { return control
...
+KeyframeSequence* AnimationManager::addKeyframeSequence(Property* property, Seconds time) {
     auto it = trackMap_.find(property);
+    std::string interpolationErrMsg;
     try {
-        auto interpolation = getDefaultInterpolation(property);
+        std::unique_ptr<Interpolation> interpolation{nullptr};
+        try {
+            interpolation = getDefaultInterpolation(property);
+        } catch (const Exception& ex) {
+            // No interpolation method registered, but it might not be needed
```
petersteneteg: See comment above

petersteneteg (changes requested):

modules/animation/include/modules/animation/datastructures/basetrack.h
ResearchDaniel: Yes, but can be Track-specific. Updated comments

modules/animation/include/modules/animation/datastructures/basetrack.h
ResearchDaniel: Avoids code duplication if one wants custom Keyframe creation. Updated comment

modules/animation/include/modules/animation/datastructures/buttonkeyframe.h
ResearchDaniel: Not direction dependent. Would making it forward-only make sense?
It is a Keyframe intended to be used for a ButtonTrack. All keyframes in a ButtonTrack (PropertyTrack) must refer to the same property.

if one wants a different behaviour I suggest adding options for the ButtonSequence/ButtonKeyframe for pressing the button in forward/backward/any animation direction

modules/animation/include/modules/animation/datastructures/callbackkeyframesequence.h
ResearchDaniel: The base class do not have `AnimationTimeState operator()(Seconds from, Seconds to, AnimationState state) const;`
I guess it could be moved to the Track or some helper function somewhere.

modules/animation/include/modules/animation/datastructures/callbacktrack.h
ResearchDaniel: Good point, I'll see if I can refactor

modules/animation/include/modules/animation/datastructures/propertytrack.h
ResearchDaniel: Changed according to your suggestion.

ResearchDaniel:
@petersteneteg I think I fixed most of the comments.
Our BaseKeyframeSequence/Basekeyframe do not include an interface or default implementation for `operator()(Seconds from, Seconds to, ...)`. So, currently, we need explicit classes for the sequences.
One option would be to introduce template arguments for a function, like `Ret operator(Seconds from, Seconds to, T val) const` and provide a default implementation, but I thought it was too obtrusive. 
Our ValueKeyframeSequence assumes 
`void operator(Seconds from, Seconds to, Ret& val) const`, 
while the track and ControlKeyframeSequence uses 
`AnimationTimeState operator()(Seconds from, Seconds to, AnimationState state)`.
So, we would need something to cover at least these cases.

modules/animation/include/modules/animation/datastructures/camerakeyframe.h
```diff
@@ -46,7 +46,19 @@ namespace animation {
  */
 class IVW_MODULE_ANIMATION_API CameraKeyframe : public BaseKeyframe {
 public:
-    using value_type = Camera;
+    struct IVW_MODULE_ANIMATION_API CamData {
+        CamData() = default;
+        CamData(vec3 lookFrom, vec3 lookTo, vec3 lookUp)
+            : lookFrom_{lookFrom}, lookTo_{lookTo}, lookUp_{glm::normalize(lookUp)} {}
```
petersteneteg: why normalise here and not in the setter? don't think that should be needed here, incoming data should always be normalised

modules/animation/include/modules/animation/datastructures/camerakeyframe.h
```diff
@@ -46,7 +46,19 @@ namespace animation {
...
+
+        void setLookFrom(vec3 val) { lookFrom_ = val; }
```
petersteneteg: why have setter for public members? why not just make this a POD struct?

modules/animation/include/modules/animation/datastructures/propertytrack.h
```diff
@@ -146,16 +146,87 @@ void setPropertyFromKeyframeHelper(CameraProperty* property, const CameraKeyfram
...
  */
 void setKeyframeFromPropertyHelper(const ButtonProperty* property, ButtonKeyframe* keyframe);
 
+template <typename Prop, typename Seq>
+struct AnimateSequence {
+    /*
+     * Helper function for when we know that we are between keyframes within a KeyframeSequence.
+     * Called from PropertyTrack<Prop, Key, Seq>::operator()(Seconds from, Seconds to,
+     * AnimationState state) const
+     * Provide template specialization of this method if you want custom
```
petersteneteg: the specialisation should be of the struct not the method. I would put this doc string on the struct instead/also

modules/animation/include/modules/animation/datastructures/propertytrack.h
```diff
@@ -146,16 +146,87 @@ void setPropertyFromKeyframeHelper(CameraProperty* property, const CameraKeyfram
...
+                                                       Seconds to, AnimationState state) {
+    typename Prop::value_type v;
+    seq(from, to, v);
+    prop->set(v);
+    return {to, state};
+}
+
+template <>
+inline AnimationTimeState
+AnimateSequence<CameraProperty, KeyframeSequenceTyped<CameraKeyframe>>::animate(
```
petersteneteg: Don't think this belongs in this header file. python in it's own camera track file

petersteneteg: Note: specialising the member function work as long as it is a full specialisation. i.e. you provide all the template arguments, but it is not supported for partial specialisations then you need to do a partial specialisation of the struct. Which is why I wanted you to break out the functionality into its own struct so one can do partial specialisations. I also prefer that we then also to the specialisation on the class even if it is full to avoid having two different kinds of specialisations around to avoid running into problems when someone tries to do a partial specialisation

modules/animation/include/modules/animation/datastructures/propertytrack.h
```diff
@@ -146,16 +146,87 @@ void setPropertyFromKeyframeHelper(CameraProperty* property, const CameraKeyfram
...
+        prop->pressButton();
+    }
+    return {to, state};
+}
+
+/*
+ * Helper for compile-time check if a property, Prop, has a member function 'get' returning Ret
+ * https://stackoverflow.com/questions/87372/check-if-a-class-has-a-member-function-of-a-given-signature
+ */
+template <typename Prop, typename Ret>
```
petersteneteg: this is really old school style use:

```
#include <inviwo/core/util/detected.h>
namespace detail {
template <typename T>
using getterTest = decltype(std::declval<T>().get());
}
template <typename T>
using hasGetMethod = is_detected<detail::getterTest, T>;

template <typename T, typename Ret >
using hasGetMethodOfType = is_detected_exact<Ret, detail::getterTest, T>;
```

modules/animation/include/modules/animation/datastructures/propertytrack.h
```diff
@@ -361,7 +413,11 @@ inline std::unique_ptr<Key> PropertyTrack<Prop, Key, Seq>::createKeyframe(Second
 template <typename Prop, typename Key, typename Seq>
 inline std::unique_ptr<Key> PropertyTrack<Prop, Key, Seq>::createKeyframe(const Prop* property,
                                                                           Seconds time) const {
-    return std::make_unique<Key>(time, property->get());
+    if constexpr (detail::property_has_get<Prop, Key::value_type>::value) {
+        return std::make_unique<Key>(time, property->get());
+    } else {
+        return std::make_unique<Key>(time);
```
petersteneteg, line 309: is this case not an error?

modules/animation/include/modules/animation/datastructures/valuekeyframesequence.h
```diff
@@ -174,7 +177,7 @@ void KeyframeSequenceTyped<Key>::operator()(Seconds from, Seconds to, value_type
     if (interpolation_) {
         (*interpolation_)(this->keyframes_, from, to, easing_, out);
     } else {
-        out = this->keyframes_.front()->getValue();
+        IVW_ASSERT(true, "Interpolation must be set before animating")
```
petersteneteg: If this is the case why not enforce that there always is an interpolation, if it is mandatory it should be passed in or created during construction, not later on.

modules/animation/src/animationmanager.cpp
```diff
@@ -213,18 +191,12 @@ std::unique_ptr<Interpolation> AnimationManager::getDefaultInterpolation(Propert
     if (interpolationIt != propertyToInterpolationMap_.end()) {
         interpolation = interpolationFactory_.create(interpolationIt->second);
         if (!interpolation) {
-            throw Exception("Default interpolation method for " + property->getClassIdentifier() +
-                                " was registered but the interpolation method was not added to the "
-                                "interpolation factory. @Developer: Please follow examples in "
-                                "animationmodule.cpp",
-                            IVW_CONTEXT);
+            LogError("Default interpolation method for " + property->getClassIdentifier() +
```
petersteneteg: should use << for the logging instead of +, or fmt::format

the LogError is redirecting the whole argument into a std::stringstream

modules/animation/include/modules/animation/datastructures/camerakeyframe.h
```diff
@@ -46,7 +46,19 @@ namespace animation {
...
+        void setLookTo(vec3 val) { lookTo_ = val; }
+        void setLookUp(vec3 val) { lookUp_ = val; }
+        vec3 lookFrom_;
+        vec3 lookTo_;
```
petersteneteg: Think we usually skip the _ for POD like things

petersteneteg (changes requested):

modules/animation/include/modules/animation/datastructures/camerakeyframe.h
ResearchDaniel: It was to mimic the Camera interface. But I have removed this now. I thought I committed it. I’ll double check

modules/animation/include/modules/animation/datastructures/propertytrack.h
ResearchDaniel: Replaced this

modules/animation/include/modules/animation/datastructures/valuekeyframesequence.h
ResearchDaniel: Perhaps we should. Ideally, we would then remove the constructors without interpolation. However, this is not possible, or at least I get compile errors then. We could throw exceptions in those constructors. That would mean that one needs to at least now a default interpolation during construction, which might be ok.

modules/animation/include/modules/animation/datastructures/propertytrack.h
ResearchDaniel: No, not all properties have get functions, e.g. ButtonProperty.

petersteneteg: Ahh I see...

modules/animation/include/modules/animation/datastructures/valuekeyframesequence.h
ResearchDaniel: I have enforced it now. Took a while to figure out the needed template magic :D

modules/animation/include/modules/animation/datastructures/buttontrack.h
```diff
@@ -0,0 +1,76 @@
...
+#include <modules/animation/datastructures/buttonkeyframesequence.h>
+#include <modules/animation/datastructures/propertytrack.h>
+
+#include <inviwo/core/properties/buttonproperty.h>
+
+namespace inviwo {
+
+namespace animation {
+
+using ButtonTrack = PropertyTrack<ButtonProperty, ButtonKeyframe, ButtonKeyframeSequence>;
```
petersteneteg, line 41: Nice!

petersteneteg:
You need to rebase this branch on master, but you can do a force push to the same branch after the rebase to keep this PR etc. The CMake people works that way and it seems to work well. (only problem might be with jenkins, but I think it should work)

ResearchDaniel:
@petersteneteg Done with changes now

modules/animation/include/modules/animation/datastructures/basetrack.h
```diff
@@ -264,36 +305,37 @@ auto BaseTrack<Seq>::end() const -> const_iterator {
  *           |-case 2a---|---------case 2b-|
  */
 template <typename Seq>
-void BaseTrack<Seq>::add(Seconds time, bool asNewSequence) {
-    auto addNew = [this](std::unique_ptr<key_type> key) {
+typename BaseTrack<Seq>::key_type* BaseTrack<Seq>::add(Seconds time, bool asNewSequence) {
```
petersteneteg, line 308: use can you trailing return types to avoid having to use the typename thing...
`auto BaseTrack<Seq>::add(Seconds time, bool asNewSequence) -> key_type*`

modules/animation/src/datastructures/valuekeyframesequence.cpp
```diff
@@ -28,6 +28,7 @@
  *********************************************************************************/
 
 #include <modules/animation/datastructures/valuekeyframesequence.h>
+#include <modules/animation/interpolation/camerasphericalinterpolation.h>
```
petersteneteg: why?

petersteneteg (dismissed):

petersteneteg:
Nice Work! Looking forward to trying it out! Do we have any example workspaces with animations?

modules/animation/include/modules/animation/datastructures/basetrack.h
ResearchDaniel: Hmm, true. I both like and dislike that solution. It becomes inconsistent with current code, but on the other side might make it easier for non-template savvy people to understand.

ResearchDaniel: Should not the definition be changed as well in that case?

ResearchDaniel:
No example workspaces. But just drag and drop the camera property into the animation editor

modules/animation/include/modules/animation/datastructures/basetrack.h
petersteneteg: since the definition is inside the class decl you can just write 
```
key_type* add(Seconds time, bool asNewSequence);
```
The using decl is available in that scope. It is only when you do the out of class definition that you run into the issue...

Trailing return types are a new thing and we can't really change to use them everywhere, that would be way too much work. But on the other hand they are way to useful in some cases not to use them, hence I think we have to live with having mixed usage. 

No changes needed here for now, just wanted to point out that using trailing return types simplifies this kind of code...

petersteneteg:
Would be really nice to add python support to the animation module.

modules/animation/include/modules/animation/datastructures/basetrack.h
ResearchDaniel: Yes, but then the def looks like 
ret func(...) 
and the implementation
auto func(...) -> ret

petersteneteg (approved):

PR #1060, opened 2020-11-03
fbo fixes

petersteneteg:
And some bug fixes


ResearchDaniel (approved):
ivw_make_package Is not used anymore?

petersteneteg:
@ResearchDaniel It has not been used for some time. we have a  ivw_register_package that it delegates to but that is not needed for our modules

PR #1063, opened 2020-11-06
Added tracy support, tracy needs to be supplied externally.

petersteneteg:


tools/tracy/include/inviwo/tracy/tracy.h
```diff
@@ -0,0 +1,169 @@
+#pragma once
```
martinfalk, line 1: might want to disable clang format here `// clang-format off`

tools/tracy/include/inviwo/tracy/tracyopengl.h
```diff
@@ -0,0 +1,235 @@
...
+
+#if !defined GL_TIMESTAMP && defined GL_TIMESTAMP_EXT
+#  define GL_TIMESTAMP GL_TIMESTAMP_EXT
+#  define GL_QUERY_COUNTER_BITS GL_QUERY_COUNTER_BITS_EXT
+#  define glGetQueryObjectiv glGetQueryObjectivEXT
+#  define glGetQueryObjectui64v glGetQueryObjectui64vEXT
+#  define glQueryCounter glQueryCounterEXT
+#endif
+
+#define TRACY_GPU_CONTEXT tracy::InitRPMallocThread(); tracy::GetGpuCtx().ptr = (tracy::GpuCtx*)tracy::tracy_malloc( sizeof( tracy::GpuCtx ) ); new(tracy::GetGpuCtx().ptr) tracy::GpuCtx;
```
martinfalk, line 58: `// clang-format off`?

martinfalk (dismissed):

src/core/CMakeLists.txt
```diff
@@ -774,9 +774,9 @@ target_link_libraries(inviwo-core
         inviwo::span
         inviwo::half
         inviwo::flags
+        inviwo::tracy
```
ResearchDaniel, line 777: Should be:
$<$<BOOL:${IVW_USE_TRACY}>:inviwo::tracy>

?

tools/tracy/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
...
+)
+
+add_library(inviwo-tracy INTERFACE)
+add_library(inviwo::tracy ALIAS inviwo-tracy)
+
+target_include_directories(inviwo-tracy INTERFACE include)
+
+option(IVW_USE_TRACY "Enable profiling using Tracy" OFF)
+if(IVW_USE_TRACY)
+	find_package(unofficial-tracy REQUIRED)
```
ResearchDaniel, line 16: Link to this one? 
Way not official one, elaborate...

tools/tracy/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
+cmake_minimum_required(VERSION 3.13...3.18 FATAL_ERROR)
+project(inviwo-tracy DESCRIPTION "Inviwo tracy wrapper" LANGUAGES CXX)
+
+set(HEADER_FILES
+    include/inviwo/tracy/tracy.h
```
ResearchDaniel, line 5: I can’t see the difference between these ones and https://github.com/wolfpld/tracy

Why not submodule?

ResearchDaniel (changes requested):
Does this work even when not having Tracy installed right now?

Might want to add a readme file to the Tracy folder.

src/core/CMakeLists.txt
petersteneteg: No this needs to always be there. We need to add macros in the code. they will come from here. This way it will work even you don't have tracy installed

tools/tracy/CMakeLists.txt
petersteneteg: it does not have an official vcpkg port. We have our own in inviwo

tools/tracy/CMakeLists.txt
petersteneteg: we need to replicate that to avoid having to include tracy all the time. I also want the macro to follow our style.

We obviously don't want to add any more submodule, that is a really bad solution.

tools/tracy/include/inviwo/tracy/tracy.h
petersteneteg: yes...

tools/tracy/CMakeLists.txt
ResearchDaniel: Ok, then please add a comment on where it comes from so that we can update it in the future in case it changes. I saw that they also have opencl support, but it might not be worthwhile to add?

tools/tracy/CMakeLists.txt
ResearchDaniel: Ok, I read in the docs that they use vcpkg, but it looks like that was only for the server side.

ResearchDaniel (approved):

tools/tracy/CMakeLists.txt
petersteneteg: OpenCL would probably be nice, I have not had the need for it yet, but it should be very easy to add

petersteneteg: there is also vulkan i think when that becomes relevant...

tools/tracy/CMakeLists.txt
petersteneteg: Yes, I have created a vcpkg port, but I have not pushed that yet, have yet to decide where is should go...

PR #1065, opened 2020-11-10
Feature/volume voronoi segmentation

sigsid:
New processor for volume voronoi segmentation, with the possibility to use weights.

To do:
- Add a regression test

### DIFF
--- a/modules/base/include/modules/base/algorithm/volume/volumevoronoi.h
+++ b/modules/base/include/modules/base/algorithm/volume/volumevoronoi.h
@@ -37,14 +37,25 @@ namespace inviwo {
 namespace util {
 
 /**
- * \brief VERY_BRIEFLY_DESCRIBE_THE_CLASS
- * DESCRIBE_THE_CLASS_FROM_A_DEVELOPER_PERSPECTIVE
+ * Implementation of Voronoi segmentation.
+ *
+ * The function returns a volume with each voxel containing the index for the closest seed point
+ * (according to the power distance with or without weights).
+ *
+ *     * volumeDimensions is the dimensions for the volume.
+ *     * indexToModelMatrix is the matrix to transform the voxel positions from index to model space.
+ *     * seedPointsWithIndices is a vector containing the seed points for the algorithm together
+ *       with their index number on the form {index, position}. The positions are expected be in
+ *       model space.
+ *     * weigths is an optional vector containing the weights for each seed point.
+ *     * weightedVoronoi is a boolean deciding if the weighted version of voronoi should be used.
+ *       If true, the weights must be provided.
  */
 
 IVW_MODULE_BASE_API std::shared_ptr<Volume> voronoiSegmentation(
     const size3_t volumeDimensions, const mat4& indexToModelMatrix,
     const std::vector<std::pair<uint32_t, vec3>>& seedPointsWithIndices,
-    const const std::optional<std::vector<float>>& weights, bool weightedVoronoi);
+    const std::optional<std::vector<float>>& weights, bool weightedVoronoi);
 
 }  // namespace util
 }  // namespace inviwo
--- a/modules/base/include/modules/base/algorithm/volume/volumevoronoi.h
+++ b/modules/base/include/modules/base/algorithm/volume/volumevoronoi.h
@@ -1,0 +1,45 @@
+/*********************************************************************************
+ *
+ * Inviwo - Interactive Visualization Workshop
+ *
+ * Copyright (c) 2020 Inviwo Foundation
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ *********************************************************************************/
+#pragma once
+
+#include <modules/base/basemoduledefine.h>
+
+namespace inviwo {
+
+/**
+ * \brief VERY_BRIEFLY_DESCRIBE_THE_CLASS
+ * DESCRIBE_THE_CLASS_FROM_A_DEVELOPER_PERSPECTIVE
+ */
+class IVW_MODULE_BASE_API volumevoronoi {
+public:
+    volumevoronoi();
+    virtual ~volumevoronoi() = default;
+};
+
+}  // namespace inviwo
--- a/modules/base/include/modules/base/algorithm/volume/volumevoronoi.h
+++ b/modules/base/include/modules/base/algorithm/volume/volumevoronoi.h
@@ -33,6 +33,8 @@
 #include <inviwo/core/datastructures/volume/volumeramprecision.h>
 #include <inviwo/core/util/indexmapper.h>
 #include <inviwo/core/util/glm.h>
+#include <inviwo/core/util/zip.h>
+#include <inviwo/core/util/volumeramutils.h>
 #include <vector>
 
 namespace inviwo {
--- a/modules/base/include/modules/base/algorithm/volume/volumevoronoi.h
+++ b/modules/base/include/modules/base/algorithm/volume/volumevoronoi.h
@@ -40,8 +40,8 @@ namespace util {
  */
 
 IVW_MODULE_BASE_API std::shared_ptr<Volume> voronoiSegmentation(
-    std::shared_ptr<const Volume> volume,
-    const std::unordered_map<dvec3, uint32_t> seedPointsWithIndices);
+    std::shared_ptr<const Volume> volume, const mat3& voxelTransformation,
+    const std::unordered_map<dvec3, uint32_t>& seedPointsWithIndices);
 
 }  // namespace util
 }  // namespace inviwo
--- a/modules/base/include/modules/base/algorithm/volume/volumevoronoi.h
+++ b/modules/base/include/modules/base/algorithm/volume/volumevoronoi.h
@@ -42,8 +42,7 @@ namespace util {
  */
 
 IVW_MODULE_BASE_API std::shared_ptr<Volume> voronoiSegmentation(
-    const size3_t volumeDimensions, const mat4& modelMatrix, const mat4& worldMatrix,
-    const mat4& indexToModelMatrix, const Wrapping3D& wrapping,
+    const size3_t volumeDimensions, const mat4& indexToModelMatrix,
     const std::vector<std::pair<uint32_t, vec3>>& seedPointsWithIndices,
     const const std::optional<std::vector<float>>& weights, bool weightedVoronoi);
 
--- a/modules/base/include/modules/base/algorithm/volume/volumevoronoi.h
+++ b/modules/base/include/modules/base/algorithm/volume/volumevoronoi.h
@@ -30,6 +30,7 @@
 
 #include <modules/base/basemoduledefine.h>
 #include <inviwo/core/datastructures/volume/volume.h>
+#include <inviwo/core/util/glm.h>
 #include <vector>
 
 namespace inviwo {
--- a/modules/base/include/modules/base/algorithm/volume/volumevoronoi.h
+++ b/modules/base/include/modules/base/algorithm/volume/volumevoronoi.h
@@ -47,7 +47,8 @@ namespace util {
  * (according to the power distance with or without weights).
  *
  *     * volumeDimensions is the dimensions for the volume.
- *     * indexToModelMatrix is the matrix to transform the voxel positions from index to model space.
+ *     * indexToModelMatrix is the matrix to transform the voxel positions from index to model
+ *       space.
  *     * seedPointsWithIndices is a vector containing the seed points for the algorithm together
  *       with their index number on the form {index, position}. The positions are expected be in
  *       model space.
--- a/modules/base/include/modules/base/algorithm/volume/volumevoronoi.h
+++ b/modules/base/include/modules/base/algorithm/volume/volumevoronoi.h
@@ -29,7 +29,6 @@
 #pragma once
 
 #include <modules/base/basemoduledefine.h>
-#include <modules/base/datastructures/kdtree.h>
 #include <inviwo/core/datastructures/volume/volume.h>
 #include <vector>
 
@@ -43,7 +42,7 @@ namespace util {
 
 IVW_MODULE_BASE_API std::shared_ptr<Volume> voronoiSegmentation(
     std::shared_ptr<const Volume> volume, const mat3& voxelTransformation,
-    const std::vector<std::pair<dvec3, uint32_t>>& seedPointsWithIndices);
+    const std::vector<std::pair<dvec3, uint32_t>>& seedPointsWithIndices, bool weightedVoronoi);
 
 }  // namespace util
 }  // namespace inviwo
--- a/modules/base/include/modules/base/algorithm/volume/volumevoronoi.h
+++ b/modules/base/include/modules/base/algorithm/volume/volumevoronoi.h
@@ -30,6 +30,8 @@
 
 #include <modules/base/basemoduledefine.h>
 #include <inviwo/core/datastructures/volume/volume.h>
+#include <inviwo/core/datastructures/volume/volumeramprecision.h>
+#include <inviwo/core/util/indexmapper.h>
 #include <inviwo/core/util/glm.h>
 #include <vector>
 
--- a/modules/base/include/modules/base/algorithm/volume/volumevoronoi.h
+++ b/modules/base/include/modules/base/algorithm/volume/volumevoronoi.h
@@ -42,7 +42,8 @@ namespace util {
 
 IVW_MODULE_BASE_API std::shared_ptr<Volume> voronoiSegmentation(
     std::shared_ptr<const Volume> volume, const mat3& voxelTransformation,
-    const std::vector<std::pair<dvec3, uint32_t>>& seedPointsWithIndices, bool weightedVoronoi);
+    const std::vector<std::pair<uint32_t, vec3>>& seedPointsWithIndices,
+    const std::vector<float> weights, bool weightedVoronoi);
 
 }  // namespace util
 }  // namespace inviwo
--- a/modules/base/include/modules/base/algorithm/volume/volumevoronoi.h
+++ b/modules/base/include/modules/base/algorithm/volume/volumevoronoi.h
@@ -44,7 +44,7 @@ namespace util {
 IVW_MODULE_BASE_API std::shared_ptr<Volume> voronoiSegmentation(
     std::shared_ptr<const Volume> volume, const mat3& voxelTransformation,
     const std::vector<std::pair<uint32_t, vec3>>& seedPointsWithIndices,
-    const std::vector<float> weights, bool weightedVoronoi);
+    const const std::optional<std::vector<float>>& weights, bool weightedVoronoi);
 
 }  // namespace util
 }  // namespace inviwo
--- a/modules/base/include/modules/base/algorithm/volume/volumevoronoi.h
+++ b/modules/base/include/modules/base/algorithm/volume/volumevoronoi.h
@@ -29,17 +29,19 @@
 #pragma once
 
 #include <modules/base/basemoduledefine.h>
+#include <modules/base/datastructures/kdtree.h>
 
 namespace inviwo {
+namespace util {
 
 /**
  * \brief VERY_BRIEFLY_DESCRIBE_THE_CLASS
  * DESCRIBE_THE_CLASS_FROM_A_DEVELOPER_PERSPECTIVE
  */
-class IVW_MODULE_BASE_API volumevoronoi {
-public:
-    volumevoronoi();
-    virtual ~volumevoronoi() = default;
-};
 
+IVW_MODULE_BASE_API std::shared_ptr<Volume> voronoiSegmentation(
+    std::shared_ptr<const Volume> volume,
+    const std::unordered_map<dvec3, uint32_t> seedPointsWithIndices);
+
+}  // namespace util
 }  // namespace inviwo
--- a/modules/base/include/modules/base/algorithm/volume/volumevoronoi.h
+++ b/modules/base/include/modules/base/algorithm/volume/volumevoronoi.h
@@ -42,7 +42,8 @@ namespace util {
  */
 
 IVW_MODULE_BASE_API std::shared_ptr<Volume> voronoiSegmentation(
-    std::shared_ptr<const Volume> volume,
+    const size3_t volumeDimensions, const mat4& modelMatrix, const mat4& worldMatrix,
+    const mat4& indexToModelMatrix, const Wrapping3D& wrapping,
     const std::vector<std::pair<uint32_t, vec3>>& seedPointsWithIndices,
     const const std::optional<std::vector<float>>& weights, bool weightedVoronoi);
 
--- a/modules/base/include/modules/base/algorithm/volume/volumevoronoi.h
+++ b/modules/base/include/modules/base/algorithm/volume/volumevoronoi.h
@@ -42,7 +42,7 @@ namespace util {
  */
 
 IVW_MODULE_BASE_API std::shared_ptr<Volume> voronoiSegmentation(
-    std::shared_ptr<const Volume> volume, const mat3& voxelTransformation,
+    std::shared_ptr<const Volume> volume,
     const std::vector<std::pair<uint32_t, vec3>>& seedPointsWithIndices,
     const const std::optional<std::vector<float>>& weights, bool weightedVoronoi);
 
--- a/modules/base/include/modules/base/algorithm/volume/volumevoronoi.h
+++ b/modules/base/include/modules/base/algorithm/volume/volumevoronoi.h
@@ -30,6 +30,8 @@
 
 #include <modules/base/basemoduledefine.h>
 #include <modules/base/datastructures/kdtree.h>
+#include <inviwo/core/datastructures/volume/volume.h>
+#include <vector>
 
 namespace inviwo {
 namespace util {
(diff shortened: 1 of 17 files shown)


modules/base/include/modules/base/algorithm/volume/volumevoronoi.h
```diff
@@ -0,0 +1,47 @@
...
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ *********************************************************************************/
+#pragma once
+
+#include <modules/base/basemoduledefine.h>
+#include <modules/base/datastructures/kdtree.h>
```
petersteneteg: should include Volume and unordered_map and probably inviwo/core/util/glm.h also

modules/base/include/modules/base/algorithm/volume/volumevoronoi.h
```diff
@@ -0,0 +1,47 @@
...
+namespace inviwo {
+namespace util {
+
+/**
+ * \brief VERY_BRIEFLY_DESCRIBE_THE_CLASS
+ * DESCRIBE_THE_CLASS_FROM_A_DEVELOPER_PERSPECTIVE
+ */
+
+IVW_MODULE_BASE_API std::shared_ptr<Volume> voronoiSegmentation(
+    std::shared_ptr<const Volume> volume, const mat3& voxelTransformation,
```
petersteneteg: voxelTransformation should be drived from the basis in the volume. Only problem is of the seedPoints are in a different space. I think the caller should make sure the seed points are in the model space of the volume.

modules/base/include/modules/base/algorithm/volume/volumevoronoi.h
```diff
@@ -0,0 +1,47 @@
...
+#pragma once
+
+#include <modules/base/basemoduledefine.h>
+#include <modules/base/datastructures/kdtree.h>
+
+namespace inviwo {
+namespace util {
+
+/**
+ * \brief VERY_BRIEFLY_DESCRIBE_THE_CLASS
```
petersteneteg: need to update this template text

modules/base/include/modules/base/processors/volumevoronoisegmentation.h
```diff
@@ -0,0 +1,69 @@
...
+#include <modules/base/basemoduledefine.h>
+#include <inviwo/core/processors/processor.h>
+#include <inviwo/core/properties/ordinalproperty.h>
+#include <inviwo/core/ports/volumeport.h>
+
+namespace inviwo {
+
+/** \docpage{org.inviwo.VolumeVoronoiSegmentation, Volume Voronoi Segmentation}
+ * ![](org.inviwo.VolumeVoronoiSegmentation.png?classIdentifier=org.inviwo.VolumeVoronoiSegmentation)
+ * Explanation of how to use the processor.
```
petersteneteg: should also be updated, can use the Tools/Create Processor Docs/Base/VolumeVoronoiSegementaion tool in inviwo

modules/base/include/modules/base/processors/volumevoronoisegmentation.h
```diff
@@ -0,0 +1,69 @@
...
+    virtual ~VolumeVoronoiSegmentation() = default;
+
+    virtual void process() override;
+
+    virtual const ProcessorInfo getProcessorInfo() const override;
+    static const ProcessorInfo processorInfo_;
+
+private:
+    VolumeInport volume_;
+    MeshInport inputMesh_;
```
petersteneteg: Are the "atom" positions always in a mesh? Or do you also have then in a DataFrame might be more useful / easier to take the DataFrame as input instead of the mesh I think. If you don't have the atoms in a DataFrame maybe we should fix that too

modules/base/src/algorithm/volume/volumevoronoi.cpp
```diff
@@ -0,0 +1,84 @@
...
+    auto newVolumeRep = std::make_shared<VolumeRAMPrecision<float>>(volume->getDimensions());
+    auto newVolume = std::make_shared<Volume>(newVolumeRep);
+    newVolume->setModelMatrix(volume->getModelMatrix());
+    newVolume->setWorldMatrix(volume->getWorldMatrix());
+
+    // Should this really be the same as for the input volume? 
+    newVolume->setInterpolation(volume->getInterpolation());
+    newVolume->setWrapping(volume->getWrapping());
+
+    newVolume->dataMap_.dataRange = dvec2(0.0, (double)seedPointsWithIndices.size());
```
petersteneteg: use static_cast<double>(seedPointsWithIndices.size()) 
(double)seedPointsWithIndices.size() is a c-style cast and should not be used in c++ code, the main reason being that the static_cast will do some sanity checks, like not casting a string to a double, where as the c-style cast will happily do that even though the result will just be garbage...

modules/base/src/algorithm/volume/volumevoronoi.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+    // Need to have this as float (otherwise I could not get it to be visible in the canvas)
+    // Why??
+    auto newVolumeRep = std::make_shared<VolumeRAMPrecision<float>>(volume->getDimensions());
+    auto newVolume = std::make_shared<Volume>(newVolumeRep);
+    newVolume->setModelMatrix(volume->getModelMatrix());
+    newVolume->setWorldMatrix(volume->getWorldMatrix());
+
+    // Should this really be the same as for the input volume? 
+    newVolume->setInterpolation(volume->getInterpolation());
```
petersteneteg: For the index/atlas volume you probably want InterpolationType::Nearest always and not linear interpolation

modules/base/src/algorithm/volume/volumevoronoi.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+std::shared_ptr<Volume> voronoiSegmentation(
+    std::shared_ptr<const Volume> volume, const mat3& voxelTransformation,
+    const std::unordered_map<dvec3, uint32_t>& seedPointsWithIndices) {
+
+    // TODO: check that the seed points are inside the volume (after transforming voxel pos)??
+
+    // Need to have this as float (otherwise I could not get it to be visible in the canvas)
+    // Why??
+    auto newVolumeRep = std::make_shared<VolumeRAMPrecision<float>>(volume->getDimensions());
```
petersteneteg: that should not be needed. I would probably go with `unsigned short` here instead of float. That "should" work in the rendering

modules/base/src/algorithm/volume/volumevoronoi.cpp
```diff
@@ -0,0 +1,84 @@
...
+    newVolume->setWorldMatrix(volume->getWorldMatrix());
+
+    // Should this really be the same as for the input volume? 
+    newVolume->setInterpolation(volume->getInterpolation());
+    newVolume->setWrapping(volume->getWrapping());
+
+    newVolume->dataMap_.dataRange = dvec2(0.0, (double)seedPointsWithIndices.size());
+    newVolume->dataMap_.valueRange = newVolume->dataMap_.dataRange;
+
+    auto newData = newVolumeRep->getDataTyped();
```
petersteneteg: newData is a bit vague as a name use volumeIndices or atlasData or something...

modules/base/src/algorithm/volume/volumevoronoi.cpp
```diff
@@ -0,0 +1,84 @@
...
+ *********************************************************************************/
+
+#include <modules/base/algorithm/volume/volumevoronoi.h>
+
+namespace inviwo {
+namespace util {
+
+std::shared_ptr<Volume> voronoiSegmentation(
+    std::shared_ptr<const Volume> volume, const mat3& voxelTransformation,
+    const std::unordered_map<dvec3, uint32_t>& seedPointsWithIndices) {
```
petersteneteg: if you are not going to use the lookup functionality of the map (below you only loop over it), It would be more efficient to use a std::vector<std::pair<dvec3, uint32_t>> instead

modules/base/src/algorithm/volume/volumevoronoi.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+    volume->getRepresentation<VolumeRAM>()->dispatch<void>([&](auto vrprecision) {
+        util::forEachVoxelParallel(*vrprecision, [&](const size3_t& voxelPos) {
+            const auto transformedVoxelPos = voxelTransformation * voxelPos;
+            auto minDist = std::numeric_limits<double>::max();
+
+            std::for_each(seedPointsWithIndices.cbegin(), seedPointsWithIndices.cend(),
+                          [&](const std::pair<dvec3, uint32_t>& point) {
+                              // Squared distance
+                              auto dist = (point.first.x - transformedVoxelPos.x) *
```
petersteneteg: can use glm::distance2(a,b) here

modules/base/src/algorithm/volume/volumevoronoi.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+    auto newData = newVolumeRep->getDataTyped();
+    util::IndexMapper3D index(volume->getDimensions());
+
+    volume->getRepresentation<VolumeRAM>()->dispatch<void>([&](auto vrprecision) {
+        util::forEachVoxelParallel(*vrprecision, [&](const size3_t& voxelPos) {
+            const auto transformedVoxelPos = voxelTransformation * voxelPos;
+            auto minDist = std::numeric_limits<double>::max();
+
+            std::for_each(seedPointsWithIndices.cbegin(), seedPointsWithIndices.cend(),
```
petersteneteg: use std::min_element

modules/base/src/algorithm/volume/volumevoronoi.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+    newVolume->dataMap_.dataRange = dvec2(0.0, (double)seedPointsWithIndices.size());
+    newVolume->dataMap_.valueRange = newVolume->dataMap_.dataRange;
+
+    auto newData = newVolumeRep->getDataTyped();
+    util::IndexMapper3D index(volume->getDimensions());
+
+    volume->getRepresentation<VolumeRAM>()->dispatch<void>([&](auto vrprecision) {
+        util::forEachVoxelParallel(*vrprecision, [&](const size3_t& voxelPos) {
+            const auto transformedVoxelPos = voxelTransformation * voxelPos;
```
petersteneteg: here you can use either 
`volume.getCoordinateTransformer().getIndexToModelMatrix();`
or
`volume.getCoordinateTransformer().getIndexToWorldMatrix();`
depending on if you want to compare distances in model or world space. 
Just make sure it is the same as for the atom positions. 

or you could to be a bit more efficient transform the atomposition to index space instead using for example
`volume.getCoordinateTransformer().getWorldToIndexMatrix();`
since you probably have way less atoms than voxels.

modules/base/src/processors/volumevoronoisegmentation.cpp
```diff
@@ -0,0 +1,104 @@
...
+    addPort(inputMesh_);
+    addPort(outport_);
+}
+
+void VolumeVoronoiSegmentation::process() {
+    auto inputMesh = inputMesh_.getData();
+
+    // Get seed points
+    // TODO: see if this can be done in a better way
+    auto posIt = util::find_if(inputMesh->getBuffers(), [](const auto& buf) {
```
petersteneteg: think there is a `inputmesh->findBuffer(BufferType::PositionAttrib);`

modules/base/src/processors/volumevoronoisegmentation.cpp
```diff
@@ -0,0 +1,104 @@
...
+    }
+    const auto positionRam = posIt->second->getRepresentation<BufferRAM>();
+    std::vector<dvec3> positions;
+    for (std::size_t i = 0; i < positionRam->getSize(); ++i) {
+        positions.push_back(positionRam->getAsDVec3(i));
+    }
+
+    // Get indices for seed points (if exists??)
+    // TODO: should probably have some fallback solution if there are no indices...?
+    auto indexIt = util::find_if(inputMesh->getBuffers(), [](const auto& buf) {
```
petersteneteg: So the IndexBuffers in the mesh only refer to positions in the position buffer above, that does not really have to have anything to do with atom indices. This is why I think using a `DataFrame` here would make it clearer what indices it is that are being used.

modules/base/src/processors/volumevoronoisegmentation.cpp
```diff
@@ -0,0 +1,104 @@
...
+    if (indexIt == inputMesh->getBuffers().end()) {
+        return;
+    }
+    // TODO: see if this can be done in a better way
+    const auto& indices = static_cast<const BufferRAMPrecision<uint32_t, BufferTarget::Data>*>(
+                              indexIt->second->getRepresentation<BufferRAM>())
+                              ->getDataContainer();
+
+    if (positions.size() != indices.size()) {
+        return;
```
petersteneteg: generally when something goes wrong we would throw and exception with a message
`throw Exception("Unexpected dimension missmatch", IVW_CONTEXT);`
the last argument `IVW_CONTEXT` is a macro that will add the filename, line and function to the error message automatically.

modules/base/src/processors/volumevoronoisegmentation.cpp
```diff
@@ -0,0 +1,104 @@
...
+        seedPointsWithIndices.emplace(positions[i], indices[i]);
+    }
+
+    // In the case of the cube data volume,
+    // the voxel positions needs to be divided with the dimensions of the volume
+    const auto volume = volume_.getData();
+    const auto basis = volume->getBasis();
+    const auto dimensions = volume->getDimensions();
+    const auto scaledBasis =
+        mat3{basis[0] / dimensions.x, basis[1] / dimensions.y, basis[2] / dimensions.z};
```
petersteneteg: see previous comments about the coordinate transformers.

modules/base/src/processors/volumevoronoisegmentation.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+namespace inviwo {
+
+// The Class Identifier has to be globally unique. Use a reverse DNS naming scheme
+const ProcessorInfo VolumeVoronoiSegmentation::processorInfo_{
+    "org.inviwo.VolumeVoronoiSegmentation",  // Class identifier
+    "Volume Voronoi Segmentation",           // Display name
+    "Undefined",                             // Category
+    CodeState::Experimental,                 // Code state
+    Tags::None,                              // Tags
```
petersteneteg: should add the CPU tag

modules/base/src/processors/volumevoronoisegmentation.cpp
```diff
@@ -0,0 +1,104 @@
...
+#include <modules/base/processors/volumevoronoisegmentation.h>
+#include <modules/base/algorithm/volume/volumevoronoi.h>
+
+namespace inviwo {
+
+// The Class Identifier has to be globally unique. Use a reverse DNS naming scheme
+const ProcessorInfo VolumeVoronoiSegmentation::processorInfo_{
+    "org.inviwo.VolumeVoronoiSegmentation",  // Class identifier
+    "Volume Voronoi Segmentation",           // Display name
+    "Undefined",                             // Category
```
petersteneteg: "Volume Operation"

modules/base/src/algorithm/volume/volumevoronoi.cpp
```diff
@@ -0,0 +1,84 @@
...
+ *
+ *********************************************************************************/
+
+#include <modules/base/algorithm/volume/volumevoronoi.h>
+
+namespace inviwo {
+namespace util {
+
+std::shared_ptr<Volume> voronoiSegmentation(
+    std::shared_ptr<const Volume> volume, const mat3& voxelTransformation,
```
petersteneteg: Humm you are never using the "data" in the volume right? only the dimensions and the transformations right? might just pass those instead potentially?

modules/base/src/algorithm/volume/volumevoronoi.cpp
```diff
@@ -0,0 +1,84 @@
...
+    // Need to have this as float (otherwise I could not get it to be visible in the canvas)
+    // Why??
+    auto newVolumeRep = std::make_shared<VolumeRAMPrecision<float>>(volume->getDimensions());
+    auto newVolume = std::make_shared<Volume>(newVolumeRep);
+    newVolume->setModelMatrix(volume->getModelMatrix());
+    newVolume->setWorldMatrix(volume->getWorldMatrix());
+
+    // Should this really be the same as for the input volume? 
+    newVolume->setInterpolation(volume->getInterpolation());
+    newVolume->setWrapping(volume->getWrapping());
```
petersteneteg: If the volume is wrapping (repeat mode) that mean that you have to adjust how you calculate distances in principle
might give an error for cases other then Clamp to begin with

modules/base/tests/unittests/volumevoronoi-test.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+#include <warn/push>
+#include <warn/ignore/all>
+#include <gtest/gtest.h>
+#include <warn/pop>
+
+namespace inviwo {
+
+TEST(Base, volumevoronoi_test) {
+    EXPECT_EQ(1, 1);
```
petersteneteg: could do a a simple test with just 1 point, and maybe 2 points in a 3x3x3 volume for example.
What happens for 0 points?

petersteneteg (approved):
Think this looks good! I have a lot off suggestions and possible improvements though :)

petersteneteg (changes requested):
pressed the wrong button...

modules/base/include/modules/base/processors/volumevoronoisegmentation.h
sigsid: I have now changed to use DataFrame instead (also included the weights in the DataFrame to be able to use weighted Voronoi)

modules/base/src/algorithm/volume/volumevoronoi.cpp
sigsid: I have changed this into a static_cast now!

modules/base/src/algorithm/volume/volumevoronoi.cpp
sigsid: Ok, changed now

modules/base/src/algorithm/volume/volumevoronoi.cpp
sigsid: I'm now using unsigned short and it works :)

modules/base/src/algorithm/volume/volumevoronoi.cpp
sigsid: I agree, changed to volumeIndices

modules/base/src/algorithm/volume/volumevoronoi.cpp
sigsid: True! I changed to use vector instead

modules/base/src/algorithm/volume/volumevoronoi.cpp
sigsid: Ok, fixed

modules/base/src/processors/volumevoronoisegmentation.cpp
sigsid: Not using the mesh anymore since I changed to DataFrame, so resolving this comment

modules/base/include/modules/base/algorithm/volume/volumevoronoi.h
sigsid: Fixed!

modules/base/src/algorithm/volume/volumevoronoi.cpp
sigsid: I'm not sure I understand how it should be used, you mean instead of the for_each?

modules/base/src/processors/volumevoronoisegmentation.cpp
sigsid: Yes, I have changed it to use DataFrame instead!

modules/base/src/processors/volumevoronoisegmentation.cpp
sigsid: Ok thanks! I am now using that pattern instead

modules/base/src/processors/volumevoronoisegmentation.cpp
sigsid: fixed

modules/base/src/processors/volumevoronoisegmentation.cpp
sigsid: fixed

modules/base/src/algorithm/volume/volumevoronoi.cpp
sigsid: Ah ok... I have now set it to wrapping3d::clampAll, do you think it will be correct?

modules/base/tests/unittests/volumevoronoi-test.cpp
sigsid: Yes I will add some tests for that!
And good question, not sure what would be best for 0 points, I guess it shouldn't create any segmentation at all....

modules/base/src/algorithm/volume/volumevoronoi.cpp
sigsid: True! I'm not using the data, but I'm using the voxel positions when I do the "volume->getRepresentation<...>()->dispatch<...>" thing, do you have any suggestion on how to do that without using the volume explicitly?

modules/base/src/algorithm/volume/volumevoronoi.cpp
petersteneteg: I was thinking you would keep this like as is, but add an additional check the the input volume is has wrapping3d::clampAll otherwise you throw an error or at least print a warning

modules/base/src/algorithm/volume/volumevoronoi.cpp
petersteneteg: yes, you are trying to find the shortest distance among all the seedPoints right? min_element will do that for you, something like this:
```
auto it = std::min_element(seedPointsWithIndices.cbegin(), seedPointsWithIndices.cend(), 
    [](const auto& p1, const auto& p2) {
        return glm::distance2(p1.first, transformedVoxelPos) < glm::distance2(p2.first, transformedVoxelPos);
    }); 
newData[index(voxelPos)] = it->second;
```

modules/base/src/algorithm/volume/volumevoronoi.cpp
petersteneteg: Ic, you can use `forEachVoxelParallel(const size3_t dims, C callback)` instead then you don't even need that dispatch :)

modules/base/src/algorithm/volume/volumevoronoi.cpp
```diff
@@ -48,20 +60,26 @@ std::shared_ptr<Volume> voronoiSegmentation(
 
     util::forEachVoxelParallel(volumeDimensions, [&](const size3_t& voxelPos) {
         const auto transformedVoxelPos = mat3(indexToModelMatrix) * voxelPos;
-        auto minDist = std::numeric_limits<double>::max();
 
-        for (size_t i = 0; i < seedPointsWithIndices.size(); i++) {
-            // Squared distance
-            auto dist = glm::distance2(seedPointsWithIndices[i].second, transformedVoxelPos);
+        if (weightedVoronoi && weights.has_value()) {
```
petersteneteg: This is maybe nitpicking, but it is a good thing to consider.
I would move this if statements outside of the forEachVoxelParallel since they will have the same result for all voxels. You generally want to have as little code as possible inside of the "hot loop" i.e. the forEachVoxelParallel in this case.

petersteneteg (dismissed):

modules/base/src/algorithm/volume/volumevoronoi.cpp
sigsid: Alright! I have added a check for that now which throws if the wrapping is not wrapping3d::clampAll

modules/base/src/algorithm/volume/volumevoronoi.cpp
sigsid: Great! I'm using that now instead :)

modules/base/src/algorithm/volume/volumevoronoi.cpp
sigsid: Thanks, I've fixed so that it uses min_element now!

modules/base/src/algorithm/volume/volumevoronoi.cpp
sigsid: Good suggestion! I have fixed it now

modules/base/tests/unittests/volumevoronoi-test.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+#include <warn/push>
+#include <warn/ignore/all>
+#include <gtest/gtest.h>
+#include <warn/pop>
+#include <modules/base/algorithm/volume/volumevoronoi.h>
+
+namespace inviwo {
+
+TEST(VolumeVoronoi, Voronoi_NoSeedPoints_ThrowsException) {
```
sigsid, line 41: Now I've started to add some tests (so far only for when the algorithm should throw an exception), but do you have a suggestion on how to look at the data in the created volume? Should I use the dispatch-pattern in the tests? @petersteneteg

petersteneteg: In the test code you "know" what type of data that should come out. hence you can just do
```
VolumeRAM* ram = volume->getRepresentation<VolumeRAM>(); 
VolumeRAMPrecision<unsigned short>* ramtyped = dynamic_cast<VolumeRAMPrecision<unsigned short>*>(ram)
```
or something like that

sigsid: Thank you! I have now added more tests which uses this pattern :)

petersteneteg (dismissed):
Very nice! I like the tests!

petersteneteg:
jenkins is complaining about some missing include I think http://jenkins.inviwo.org:8080/job/inviwo/job/PR-1065/19/console

petersteneteg:
For the future... , I've probably never mentioned, but we try to start all commit messages with the Module/moduls that it relates to, like
    Base: Added more volonoi tests.
or something like that. It makes it a bit easier to understand the context of a commit

modules/base/include/modules/base/algorithm/volume/volumevoronoi.h
sigsid: Now I'm just sending in the indexToModelMatrix which I use to transform the voxel-positions to model space, so I'm expecting the seedPoints to be in model space....

modules/base/src/algorithm/volume/volumevoronoi.cpp
sigsid: Now I'm using indexToModelMatrix (`volume.getCoordinateTransformer().getIndexToModelMatrix()`) as input because I somehow found that easier... (but if I need to optimize it further in the future I will change to worldToIndexMatrix instead!)

modules/base/src/processors/volumevoronoisegmentation.cpp
sigsid: Yep I hope I have fixed this now to be correct!

modules/base/src/algorithm/volume/volumevoronoi.cpp
```diff
@@ -0,0 +1,97 @@
...
+ *********************************************************************************/
+
+#include <modules/base/algorithm/volume/volumevoronoi.h>
+
+namespace inviwo {
+namespace util {
+
+std::shared_ptr<Volume> voronoiSegmentation(
+    const size3_t volumeDimensions, const mat4& indexToModelMatrix,
+    const std::vector<std::pair<uint32_t, vec3>>& seedPointsWithIndices,
```
sigsid, line 37: I'm thinking if I should add a check to make sure the indices have distinct values... 
What do you think @petersteneteg?
Maybe it's up to the consumer of this function to make sure they put in the correct indices. And I guess there could be use cases where you want to group things together by using the same index..?

modules/base/include/modules/base/algorithm/volume/volumevoronoi.h
sigsid: I have updated it now! Tell me if you want any changes

modules/base/include/modules/base/processors/volumevoronoisegmentation.h
sigsid: Updated! tell me if I should make any changes

modules/base/src/algorithm/volume/volumevoronoi.cpp
petersteneteg: I think that responsibility falls on the caller, and as you said, there could be uses cases for having duplicates also.

sigsid: Ok good, then I leave it as it is now

petersteneteg (approved):

PR #1066, opened 2020-11-10
Serialization bugfixes

petersteneteg:


include/inviwo/core/properties/optionproperty.h
```diff
@@ -193,14 +204,19 @@ class TemplateOptionProperty : public BaseOptionProperty {
...
+    virtual bool isSelectedIdentifier(std::string_view identifier) const override;
+    virtual bool isSelectedDisplayName(std::string_view name) const override;
     bool isSelectedValue(const T& val) const;
 
     const T& get() const;
     const T& operator*() const;
     const T* operator->() const;
-    void set(const T& value);
+
+    bool set(const T& value);
```
ResearchDaniel, line 231: Return true if modified?

src/core/properties/cameraproperty.cpp
```diff
@@ -382,16 +366,50 @@ void CameraProperty::addCamerapProperty(std::unique_ptr<Property> camprop) {
...
+}
+
+bool CameraProperty::needsSerialization() const {
+    return serializationMode_ != PropertySerializationMode::None;
+}
+
 void CameraProperty::serialize(Serializer& s) const {
     CompositeProperty::serialize(s);
-    s.serialize("Camera", camera_);
+    if (serializationMode_ != PropertySerializationMode::None) {
```
ResearchDaniel, line 397: Are we doing this check for other properties?

ResearchDaniel (dismissed):

include/inviwo/core/properties/optionproperty.h
petersteneteg: it returns true if it was actually set, i.e. if the value was found in the list of options

petersteneteg: will add a comment...

src/core/properties/cameraproperty.cpp
petersteneteg: the value wrapper usually handles that automatically

martinfalk (dismissed):

ResearchDaniel (dismissed):
Not sure if some of the changes also affect the json module. The added functions might need to be reflected there as well

petersteneteg:
@ResearchDaniel @martinfalk poke

ResearchDaniel (approved):

PR #1069, opened 2020-11-10
Qt: improved filtering in processor list

martinfalk:
* filtering based on matching substrings (space is the separator)
* filter now similar to "go to files" functionality in Visual Studio

image

Processor tags are also matched
image

For comparison, "go to file" functionality in Visual Studio
image

### DIFF
--- a/src/qt/editor/lineediteventfilter.cpp
+++ b/src/qt/editor/lineediteventfilter.cpp
@@ -1,0 +1,73 @@
+/*********************************************************************************
+ *
+ * Inviwo - Interactive Visualization Workshop
+ *
+ * Copyright (c) 2020 Inviwo Foundation
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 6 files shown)


src/qt/editor/lineediteventfilter.cpp
```diff
@@ -0,0 +1,73 @@
...
+                    if (focusAndDown_) {
+                        // move cursor down one row
+                        QKeyEvent* keydown =
+                            new QKeyEvent(QKeyEvent::KeyPress, Qt::Key_Down, Qt::NoModifier);
+                        QCoreApplication::postEvent(widget_, keydown);
+                    }
+                    break;
+                }
+                case Qt::Key_Escape:
+                    reinterpret_cast<QLineEdit*>(parent())->clear();
```
petersteneteg: This seems nasty, at least make the parent arg a QLineEdit, why not a dynamic_cast?

petersteneteg (dismissed):

petersteneteg (approved):

PR #1070, opened 2020-11-11
Animation: Bug fixes

ResearchDaniel:
- Control/Callback tracks did not operate on all keyframes
- Added tests for control/callback tracks
- Fixed serialization/deserialization for PropertyTrack where the property is not in the network (if loaded outside of workspace).
- Added functionality for playing animation backwards
- Removed unused code (AnimationPlaySettings)
- Fixed a couple warnings

### DIFF
--- a/modules/animation/include/modules/animation/algorithm/animationrange.h
+++ b/modules/animation/include/modules/animation/algorithm/animationrange.h
@@ -30,6 +30,8 @@
 
 #include <modules/animation/animationmoduledefine.h>
 #include <modules/animation/datastructures/animationtime.h>
+#include <modules/animation/datastructures/animationstate.h>
+#include <modules/animation/datastructures/keyframesequence.h>
 
 #include <algorithm>
 #include <iterator>
@@ -38,23 +40,76 @@ namespace inviwo {
 
 namespace animation {
 
+namespace detail {
+template <typename T>
+struct GetTimeHelper {
+
+    static Seconds getTime(const T& elem, PlaybackDirection direction) {
+        if constexpr (std::is_base_of_v<KeyframeSequence, T>) {
+            if (direction == PlaybackDirection::Forward) {
+                return elem.getLast().getTime();
+            } else {
+                return elem.getFirst().getTime();
+            }
+
+        } else {
+            return elem.getTime();
+        };
+    }
+};
+
+}  // namespace detail
 /*
- * Returns iterators for elements in the range [from, to].
- * Deals with cases where from > to.
- * @return One iterator to first item >= min(from, to) and one iterator to item > max(from, to)
+ * Calls operator(Seconds from, Seconds to,
+                      AnimationState state) -> AnimationTimeState for each element [begin, end) in
+ the time range [from, to].
+ * The elements are processed in the order given by from and to, e.g. reverse order if from > to.
+ * The function is currently designed for Dirac delta Keyframe and KeyframeSequence, i.e. no
+ interpolation (such as ControlKeyframe, ControlKeyframeSequence).
+ * @param begin, end iterator range to types std::unique_ptr<Keyframe> or
+ std::unique_ptr<KeyframeSequence>.
+ * @return AnimationTimeState after processing the elements.
  */
 template <typename Iterator>
-auto getRange(Iterator begin, Iterator end, Seconds from, Seconds to)
-    -> std::tuple<decltype(begin), decltype(end)> {
+AnimationTimeState animateRange(Iterator begin, Iterator end, Seconds from, Seconds to,
+                                AnimationState state) {
+    auto animate = [](auto begin, auto end, Seconds from, Seconds to,
+                      AnimationState state) -> AnimationTimeState {
+        auto direction = from <= to ? PlaybackDirection::Forward : PlaybackDirection::Backward;
+        AnimationTimeState res{to, state};
+        while (begin != end && res.state != AnimationState::Paused) {
+            res = (**begin)(from, to, state);
+            if ((direction == PlaybackDirection::Forward && res.time <= (**begin)) ||
+                (direction == PlaybackDirection::Backward && res.time >= (**begin))) {
+                // We jumped in the opposite direction
+                break;
+            }
+            // Use jump-to-time if set, previous keyframe time otherwise
+            from = res.time != to
+                       ? res.time
+                       : detail::GetTimeHelper<
+                             std::remove_reference<decltype(**begin)>::type>::getTime(**begin,
+                                                                                      direction);
+
+            ++begin;
+        }
+        return res;
+    };
     auto first = std::min(from, to);
     auto last = std::max(from, to);
+
     // 'fromIt' will be the first item with a time larger than or equal to 'first'
     auto fromIt = std::lower_bound(begin, end, first,
                                    [](const auto& it, const auto& val) { return *it < val; });
     // 'toIt' will be the first key with a time larger than 'last'
-    auto toIt = std::upper_bound(begin, end, last,
+    auto toIt = std::upper_bound(fromIt, end, last,
                                  [](const auto& val, const auto& it) { return val < *it; });
-    return {fromIt, toIt};
+    if (from <= to) {
+        return animate(fromIt, toIt, from, to, state);
+    } else {
+        return animate(std::make_reverse_iterator(toIt), std::make_reverse_iterator(fromIt), from,
+                       to, state);
+    }
 }
 
 }  // namespace animation
--- a/modules/animation/include/modules/animation/algorithm/animationrange.h
+++ b/modules/animation/include/modules/animation/algorithm/animationrange.h
@@ -26,29 +26,35 @@
  * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  *
  *********************************************************************************/
+#pragma once
 
-#include <modules/animation/datastructures/callbackkeyframe.h>
+#include <modules/animation/animationmoduledefine.h>
+#include <modules/animation/datastructures/animationtime.h>
+
+#include <algorithm>
+#include <iterator>
 
 namespace inviwo {
 
 namespace animation {
 
-CallbackKeyframe::CallbackKeyframe(Seconds time, std::function<void()> doForward,
-                                   std::function<void()> undo)
-    : BaseKeyframe(time), do_(std::move(doForward)), undo_(std::move(undo)) {}
-
-CallbackKeyframe* CallbackKeyframe::clone() const { return new CallbackKeyframe(*this); }
-
-AnimationTimeState CallbackKeyframe::operator()(Seconds from, Seconds to,
-                                                AnimationState state) const {
-    if (do_ && (from <= getTime() && to >= getTime())) {
-        // Animating forward, passing from left to right
-        do_();
-    } else if (undo_ && (to <= getTime() && from >= getTime())) {
-        // Animating backward, passing from right to left
-        undo_();
-    }
-    return {to, state};
+/*
+ * Returns iterators for elements in the range [from, to].
+ * Deals with cases where from > to.
+ * @return One iterator to first item >= min(from, to) and one iterator to item > max(from, to)
+ */
+template <typename Iterator>
+auto getRange(Iterator begin, Iterator end, Seconds from, Seconds to)
+    -> std::tuple<decltype(begin), decltype(end)> {
+    auto first = std::min(from, to);
+    auto last = std::max(from, to);
+    // 'fromIt' will be the first item with a time larger than or equal to 'from'
+    auto fromIt =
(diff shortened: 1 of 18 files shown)


modules/animation/include/modules/animation/algorithm/animationrange.h
```diff
@@ -0,0 +1,62 @@
...
+ * Returns iterators for elements in the range [from, to].
+ * Deals with cases where from > to.
+ * @return One iterator to first item >= min(from, to) and one iterator to item > max(from, to)
+ */
+template <typename Iterator>
+auto getRange(Iterator begin, Iterator end, Seconds from, Seconds to)
+    -> std::tuple<decltype(begin), decltype(end)> {
+    auto first = std::min(from, to);
+    auto last = std::max(from, to);
+    // 'fromIt' will be the first item with a time larger than or equal to 'from'
```
weinkauf: You mean 'first', not 'from', right?

modules/animation/include/modules/animation/algorithm/animationrange.h
```diff
@@ -0,0 +1,62 @@
...
+    auto fromIt =
+        std::lower_bound(begin, end, first, [](const auto& it, const auto& val) { return *it < val; });
+    // 'toIt' will be the first key with a time larger than 'to'
```
weinkauf: Same here, you mean 'last', not 'to'

modules/animation/include/modules/animation/animationcontroller.h
```diff
@@ -95,12 +95,10 @@ class IVW_MODULE_ANIMATION_API AnimationController : public AnimationControllerO
     /// Returns the current state of the controller, whether it is playing, or pausing, and such.
     const AnimationState& getState() const;
 
-    /// Returns playback mode such as loop or swing and such.
-    const AnimationPlaySettings& getPlaybackSettings() const { return settingsPlay_; }
-    AnimationPlaySettings& getPlaybackSettings() { return settingsPlay_; }
-    void setPlaybackSettings(const AnimationPlaySettings& newSettings);
```
weinkauf, line 101: Interesting. We did not use that? I assume that looping and swinging still work, though, right?

modules/animation/src/animationcontroller.cpp
```diff
@@ -362,7 +362,7 @@ void AnimationController::tick() {
     // What to do when network cannot be evaluated in the speed that is given by deltaTime?
     // Initial solution: Don't care about that, and let it evaluate fully in the speed that it can
     // muster.
```
weinkauf: Can this comment be resolved now?

modules/animation/src/datastructures/callbackkeyframesequence.cpp
```diff
@@ -43,20 +43,31 @@ CallbackKeyframeSequence* CallbackKeyframeSequence::clone() const {
...
+            res = (**begin)(from, to, state);
+            ++begin;
         }
-    }
+        return res;
+    };
+    auto direction = from <= to ? PlaybackDirection::Forward : PlaybackDirection::Backward;
+    auto first = direction == PlaybackDirection::Forward ? from : to;
+    auto last = direction == PlaybackDirection::Forward ? to : from;
+    // 'fromIt' will be the first key with a time larger than or equal to 'from'
```
weinkauf: Same as above, you mean 'first'.

modules/animation/src/datastructures/callbackkeyframesequence.cpp
```diff
@@ -43,20 +43,31 @@ CallbackKeyframeSequence* CallbackKeyframeSequence::clone() const {
...
+    auto fromIt = std::lower_bound(keyframes_.begin(), keyframes_.end(), first,
+                                   [](const auto& a, const auto& b) { return *a < b; });
+    // 'toIt' will be the first key with a time larger than 'to'
```
weinkauf: 'last'

modules/animation/include/modules/animation/animationcontroller.h
ResearchDaniel: I guess we intended to use it but never did. Probably better to redo it if we need it.
Everything still works, it was completely unused :)

weinkauf (commented):
This looks fine to me. I added a few comments, please have a look.

modules/animation/src/animationcontroller.cpp
ResearchDaniel: Not really. I think that it might need to be determined from the outside

modules/animation/src/datastructures/callbackkeyframesequence.cpp
ResearchDaniel: Ah, forgot to refactor this. Should use same code as other

modules/animation/include/modules/animation/algorithm/animationrange.h
```diff
@@ -0,0 +1,62 @@
...
+ * Deals with cases where from > to.
+ * @return One iterator to first item >= min(from, to) and one iterator to item > max(from, to)
+ */
+template <typename Iterator>
+auto getRange(Iterator begin, Iterator end, Seconds from, Seconds to)
+    -> std::tuple<decltype(begin), decltype(end)> {
+    auto first = std::min(from, to);
+    auto last = std::max(from, to);
+    // 'fromIt' will be the first item with a time larger than or equal to 'first'
+    auto fromIt = std::lower_bound(begin, end, first,
```
petersteneteg, line 96: can't you just use equal range for this? https://en.cppreference.com/w/cpp/algorithm/equal_range

modules/animation/include/modules/animation/algorithm/animationrange.h
```diff
@@ -0,0 +1,62 @@
...
+
+namespace animation {
+
+/*
+ * Returns iterators for elements in the range [from, to].
+ * Deals with cases where from > to.
+ * @return One iterator to first item >= min(from, to) and one iterator to item > max(from, to)
+ */
+template <typename Iterator>
+auto getRange(Iterator begin, Iterator end, Seconds from, Seconds to)
```
petersteneteg: would it be possible to use a more descriptive name? what range are we talking about?

modules/animation/include/modules/animation/animationcontroller.h
```diff
@@ -115,6 +113,7 @@ class IVW_MODULE_ANIMATION_API AnimationController : public AnimationControllerO
     DoubleMinMaxProperty playWindow;
     DoubleProperty framesPerSecond;
     TemplateOptionProperty<PlaybackMode> playMode;
+    TemplateOptionProperty<PlaybackDirection> playbackDirection;
```
petersteneteg: is the redundant with respect to a negative framesPerSecond? or do we always have a positive fps?

modules/animation/include/modules/animation/datastructures/propertytrack.h
```diff
@@ -468,6 +468,9 @@ Seq* PropertyTrack<Prop, Key, Seq>::addSequenceUsingPropertyValue(
 
 template <typename Prop, typename Key, typename Seq>
 void PropertyTrack<Prop, Key, Seq>::serialize(Serializer& s) const {
+    if (!property_) {
+        throw SerializationException("No property set for the PropertyTrack");
```
petersteneteg: add IVW_CONTEXT

modules/animation/include/modules/animation/datastructures/propertytrack.h
```diff
@@ -481,6 +484,9 @@ void PropertyTrack<Prop, Key, Seq>::deserialize(Deserializer& d) {
 
     IVW_ASSERT(network_, "Property track deserialization requires a ProcessorNetwork");
     property_ = dynamic_cast<Prop*>(network_->getProperty(propertyId));
+    if (!property_) {
+        throw SerializationException("Could not find property " + propertyId);
```
petersteneteg: add IVW_CONTEXT

modules/animation/src/datastructures/callbacktrack.cpp
```diff
@@ -47,32 +48,22 @@ std::string CallbackTrack::getClassIdentifier() const { return classIdentifier()
...
-        } else {  // case 2b
-            if (from < seq1->getLastTime()) {
-                // We came from before the previous key
-                return (*seq1)(from, to, state);
-            } else if (it != sequences_.end() && from > (*it)->getFirstTime()) {
-                // We came form after the next key
-                return (**it)(from, to, state);
-            }
-            // we moved in an unmarked region, do nothing.
+    auto animate = [](auto begin, auto end, Seconds from, Seconds to,
```
petersteneteg: You need to remove the comment above with the cases since you no longer refer to them

modules/animation/src/datastructures/controlkeyframesequence.cpp
```diff
@@ -52,20 +53,31 @@ ControlKeyframeSequence* ControlKeyframeSequence::clone() const {
...
+            ++begin;
         }
+        return res;
+    };
+    auto [fromIt, toIt] = getRange(keyframes_.begin(), keyframes_.end(), from, to);
+    if (from <= to) {
+        return animate(fromIt, toIt, from, to, state);
+    } else {
+        return animate(std::make_reverse_iterator(toIt), std::make_reverse_iterator(fromIt), from,
+                       to, state);
```
petersteneteg: This seems to be a common operation? make it a function?

petersteneteg (changes requested):

modules/animation/include/modules/animation/algorithm/animationrange.h
ResearchDaniel: Hmm, yes. I looked at it but I seem too have disregarded it too quickly

ResearchDaniel: On a closer look it seems as if it is easier to separate the two since we are not comparing with a value as equal_range intends to do.

modules/animation/include/modules/animation/algorithm/animationrange.h
ResearchDaniel: Either sequences or keyframes. 
getTrackRange?

modules/animation/include/modules/animation/animationcontroller.h
ResearchDaniel: Hmm, I wonder if the implementation of the AnimationController was finished. There are a number of unused properties, such as framesPerSecond.
Nevertheless, specifying a negative fps seems a bit too hidden to me. The most natural way would probably be to add a "play backwards" action button instead of having this dropdown.

modules/animation/src/datastructures/controlkeyframesequence.cpp
ResearchDaniel: Wanted to include in the getRange function but were not able to return the different iterator types

modules/animation/include/modules/animation/algorithm/animationrange.h
petersteneteg: getRangeForTimeInterval
getTimeIntervalRange

modules/animation/include/modules/animation/animationcontroller.h
petersteneteg: This in the internal representation It has little to do with the how we present it in the GUI. And the internal representation should have a little redundant information as possible. It here is redundant information then you are also responsible for making sure that that information never get out of sync. Which is often very had to do.

And yes all of the animation suff is still WIP

modules/animation/src/datastructures/controlkeyframesequence.cpp
petersteneteg: just pass in the animate function as an argument

ResearchDaniel: Easier said than done since the pointer types are different. Anyway, I further refactored this part into the function

modules/animation/include/modules/animation/animationcontroller.h
ResearchDaniel: Removed the duplicate state and instead use the sign of deltaTime_

modules/animation/include/modules/animation/algorithm/animationrange.h
```diff
@@ -38,23 +40,76 @@ namespace inviwo {
 
 namespace animation {
 
+namespace detail {
+template <typename T>
+struct GetTimeHelper {
```
petersteneteg: Don't think you need the struct here, jus use a regular templated function. then you also get automatic template argument deduction.
The wrapping struct is mostly needed when one wants to do partial specialisation or provide a customisation point.

petersteneteg (dismissed):

modules/animation/include/modules/animation/algorithm/animationrange.h
ResearchDaniel: Sweet, thanks!

ResearchDaniel:
@petersteneteg ping

petersteneteg (approved):

PR #1072, opened 2020-11-17
Core: Added DataTraits to TransferFunction

ResearchDaniel:


include/inviwo/core/datastructures/transferfunction.h
```diff
@@ -128,4 +129,31 @@ class IVW_CORE_API TransferFunction : public TFPrimitiveSet {
 bool operator==(const TransferFunction& lhs, const TransferFunction& rhs);
 bool operator!=(const TransferFunction& lhs, const TransferFunction& rhs);
 
+template <>
+struct DataTraits<TransferFunction> {
+    static std::string classIdentifier() { return "org.inviwo.transferfunction"; }
```
petersteneteg: one can return const std::string& here to make it a bit more performant

petersteneteg (dismissed):

include/inviwo/core/datastructures/transferfunction.h
ResearchDaniel: Ah, I updated this and the comment in DataTraits to reflect it. Also updated DataTraits<DataFrame> since I "stole" most of the code from there :)

petersteneteg (dismissed):

martinfalk (approved):

PR #1074, opened 2020-11-19
QtWidgets fixes

petersteneteg:


modules/qtwidgets/src/properties/lightpropertywidgetqt.cpp
```diff
@@ -73,20 +123,29 @@ LightPropertyWidgetQt::LightPropertyWidgetQt(FloatVec3Property* property)
         LogWarn("Maximum value is assumed to be positive. Widget may produce values out of range.")
     }
 
-    vec3 maxVal = glm::abs(property_->getMaxValue());
-    radiusSpinBox_->setMinimum(-glm::length(maxVal));
-    radiusSpinBox_->setMaximum(glm::length(maxVal));
+    radius_->setMinValue(0, ConstraintBehavior::Immutable);
```
martinfalk, line 126: immutable?

src/core/properties/ordinalproperty.cpp
```diff
@@ -91,4 +91,14 @@ OrdinalPropertyState<vec3> util::ordinalColor(const vec3& value,
             PropertySemantics::Color};
 }
 
+OrdinalPropertyState<vec3> util::ordinalLight(const vec3& pos, float min, float max,
```
martinfalk, line 94: isn't that even more generic than light? It's a position...

src/qt/editor/welcomewidget.cpp
```diff
@@ -381,13 +323,7 @@ WelcomeWidget::WelcomeWidget(InviwoMainWindow* window, QWidget* parent)
 
     setTabOrder(filterLineEdit_, filetree_);
     setTabOrder(filetree_, loadWorkspaceBtn_);
-    setTabOrder(loadWorkspaceBtn_, newButton);
```
martinfalk, line 384: did you check whether the tab order is still as expected?

martinfalk (dismissed):

modules/qtwidgets/src/properties/lightpropertywidgetqt.cpp
petersteneteg: it's a length can never be less then 0, I've removed the hack with negative distance from the gui at least

src/core/properties/ordinalproperty.cpp
petersteneteg: it has the light semantics, but I was also considering adding a pos something... but did not have a need for it at the moment...

src/qt/editor/welcomewidget.cpp
petersteneteg: I just moved them, but I have not checked yet

petersteneteg: it still works, but we have no focus indication on the buttons...

modules/qtwidgets/src/lightpositionwidgetqt.cpp
```diff
@@ -36,119 +42,100 @@
...
 
-float LightPositionWidgetQt::getX() const { return sin(theta_) * cos(phi_); }
+double LightPositionWidgetQt::getX() const { return pos_.x; }
 
-float LightPositionWidgetQt::getY() const { return -sin(theta_) * sin(phi_); }
+double LightPositionWidgetQt::getY() const { return -pos_.y; }
 
-float LightPositionWidgetQt::getZ() const { return (radius_ < 0.0f ? -1.0f : 1.0f) * cos(theta_); }
+double LightPositionWidgetQt::getZ() const {
+    return (radius_ < 0.0f ? -1.0f : 1.0f) * std::sqrt(1.0 - glm::length(pos_));
```
ResearchDaniel: If you want to nitpick: remove f in -1.0f and 1.0f

ResearchDaniel (dismissed):
Nice!

martinfalk (dismissed):

modules/qtwidgets/src/lightpositionwidgetqt.cpp
petersteneteg: humm missed one... thanks

martinfalk (approved):

PR #1077, opened 2020-11-25
Fix/webbrowser keyevents

ResearchDaniel:
- Upgrade CEF version
- Fix key event injection. Tested on Windows
- Added a test which injects keyboard events from python. Not ideal because they are not platform agnostic, but might catch some errors.


petersteneteg (dismissed):

ResearchDaniel:
@petersteneteg moved regression test since I could not get it to work on Jenkins. It is now in the examples directory instead.  Could check and approve again if things are ok?

martinfalk (approved):

ResearchDaniel:
@petersteneteg @martinfalk  Do you know why the regression runs for the moved workspace?

PR #1085, opened 2020-12-14
OpenGL: bindAttributes location check

martinfalk:


modules/opengl/src/shader/shader.cpp
```diff
@@ -301,10 +301,14 @@ bool Shader::checkLinkStatus() const {
 void Shader::bindAttributes() {
     for (const auto& obj : getShaderObjects()) {
         for (const auto& item : obj.getInDeclarations()) {
-            glBindAttribLocation(program_.id, item.location, item.name.c_str());
+            if (item.location >= 0) {
```
ResearchDaniel, line 304: Should that be checked when setting the location instead?

ResearchDaniel (approved):

modules/opengl/src/shader/shader.cpp
martinfalk: No, since the in/out declaration datatypes also allow for _valid_ declarations without any location. For example:
```glsl
layout (depth_greater) out float gl_FragDepth;
```

which can be achieved by
```c++
ShaderObject::OutDeclaration outdecl;
outdecl.decl = "layout (depth_greater) out float gl_FragDepth;";
shader.getFragmentShaderObject()->addOutDeclaration(outdecl);
```

PR #1086, opened 2020-12-14
Acknowledgements: Added ELLIIT as funding source

ResearchDaniel:


martinfalk (changes requested):
The logos shipped with Inviwo usually come only in sizes 200x200 and 600x600 (see `docs/images/`). The elliit logo is 1579x1578 pixels.

martinfalk:
Good catch!

ResearchDaniel:
> The logos shipped with Inviwo usually come only in sizes 200x200 and 600x600 (see `docs/images/`). The elliit logo is 1579x1578 pixels.

Fixed

martinfalk (dismissed):

petersteneteg (changes requested):
What about the about dialog?

ResearchDaniel:
> What about the about dialog?

Fixed

ResearchDaniel:
Seems like there are some issues with the hdf repository:
fatal: repository 'https://github.com/live-clones/hdf5.git/' not found

src/qt/editor/inviwoaboutwindow.cpp
```diff
@@ -186,6 +187,9 @@ InviwoAboutWindow::InviwoAboutWindow(InviwoMainWindow* mainwindow)
 
         auto uulm = p.append("a", "", {{"href", "http://www.uni-ulm.de/en/"}});
         uulm.append("img", "", makeImg(":/images/uulm.png", 50));
+
+        auto elliit = p.append("a", "", {{"href", "https://old.liu.se/elliit?l=en"}});
+        elliit.append("img", "", makeImg(":/images/elliit.png", 50));
```
petersteneteg, line 192: Have you tested this? I don't see any adding to the .rc file for the image?

petersteneteg (changes requested):

petersteneteg:
you have to update your submodules, we use the official HDFGroup repo now for hdf5

src/qt/editor/inviwoaboutwindow.cpp
ResearchDaniel: Ah, right. Added it and tested it. Changed to transparent background to be consistent with the others

petersteneteg (dismissed):

petersteneteg (approved):

ResearchDaniel:
> you have to update your submodules, we use the official HDFGroup repo now for hdf5

I have rebased onto master and done a clean build on jenkins. Still same error. It seems to run the module sync/update on jenkins but it still produces an error for warn. Any ideas?

petersteneteg:
But the merge one works right? we only care about that one... probably have to delete the source dir in jenkins to properly get it to pull the fresh submodule. You can just go ahead and merge I think

PR #1089, opened 2020-12-17
Add processors for property-based testing

mstegmaier:

**Always**

**New features**

**Updated feature / Fixes** 

Windows and Linux
Microsoft Visual Studio 2019 and GCC


petersteneteg:
@mstegmaier Thanks for contributing!!! Before I start to review the code any further, I would ask to you do some stylistic fixes so we can leave those details out of the review.
We format all or code with clang-format using the .clang-format style file in the root dir. 
For class members we only use a trailing underscore for making members. In the case of a struct will all public members the trailing underscore can be removed. 
We __always__ include a pair of {} for all if/for/while etc even if they are only one line.
We have a very high bar on the use of any macros. pls remove any uses unless there is a very good reason.
Thanks again for very interesting work!

modules/propertybasedtesting/include/inviwo/propertybasedtesting/algorithm/generatingassignments.h
```diff
@@ -0,0 +1,164 @@
...
+namespace inviwo {
+
+/*
+ * Interface to property assignments, enables to apply the assignment
+ * (if it not deactivated).
+ * Is used mainly by TestProperty
+ */
+class PropertyAssignment {
+private:
+    const bool* m_deactivated;
```
petersteneteg: we use a trailing underscore for members

modules/propertybasedtesting/include/inviwo/propertybasedtesting/algorithm/coveringarray.h
```diff
@@ -0,0 +1,53 @@
...
+
+namespace inviwo {
+
+namespace util {
+
+// 2-coverage, randomized discrete SLJ strategy
+std::vector<Test> coveringArray(
+    const Test& init, const std::vector<std::vector<std::shared_ptr<PropertyAssignment>>>& vars);
+
+std::vector<Test> optCoveringArray(
```
petersteneteg: Add documentation

modules/propertybasedtesting/include/inviwo/propertybasedtesting/algorithm/generatingassignments.h
```diff
@@ -0,0 +1,164 @@
...
+
+protected:
+    virtual void m_apply() const = 0;
```
petersteneteg: doApply and doIsApplied

modules/propertybasedtesting/include/inviwo/propertybasedtesting/algorithm/generatingassignments.h
```diff
@@ -0,0 +1,164 @@
...
+ */
+template <typename P, typename T = typename P::value_type>
+class PropertyAssignmentTyped : public PropertyAssignment {
+private:
+    P* const prop;
+    const T value;
+
+    void m_apply() const override {
+        prop->set(value);
+        std::cerr << prop->getDisplayName() << " : " << value << " (" << prop->get() << ")"
```
petersteneteg: we never write to err, use LogError, or preferably add an assert to throw an exception if this is an error.

modules/propertybasedtesting/include/inviwo/propertybasedtesting/algorithm/coveringarray.h
```diff
@@ -0,0 +1,53 @@
...
+
+#include <inviwo/propertybasedtesting/algorithm/generatingassignments.h>
+#include <inviwo/propertybasedtesting/algorithm/propertyanalyzing.h>
+
+namespace inviwo {
+
+namespace util {
+
+// 2-coverage, randomized discrete SLJ strategy
+std::vector<Test> coveringArray(
```
petersteneteg: Missing API macro, IVW_MODULE_[module name]_API

modules/propertybasedtesting/include/inviwo/propertybasedtesting/algorithm/generatingassignments.h
```diff
@@ -0,0 +1,164 @@
...
+    void print(std::ostream& out) const override {
+        out << "PropertyAssignment(" << prop->getIdentifier() << "," << getValue() << ")";
+    }
+};
+
+/*
+ * A test is a set of assignments to properties. Should contain no
+ * property more than once.
+ */
+using Test = std::vector<std::shared_ptr<PropertyAssignment>>;
```
petersteneteg, line 100: This is a bit too generic of a name to be in the main inviwo namespace I think. I would move into a new namespace pbt inside of the inviwo namespace, same goes for many of the other entities here.

modules/propertybasedtesting/include/inviwo/propertybasedtesting/algorithm/generatingassignments.h
```diff
@@ -0,0 +1,164 @@
...
+struct GenerateAssignments<OrdinalProperty<T>> {
+    std::vector<std::shared_ptr<PropertyAssignment>> operator()(OrdinalProperty<T>* const prop,
+                                                                const bool* deactivated) const {
+
+        using P = OrdinalProperty<T>;
+
+        std::vector<std::shared_ptr<PropertyAssignment>> res;
+
+        const auto increment = prop->getIncrement();
+        const size_t maxSteps = (prop->getMaxValue() - prop->getMinValue()) / increment;
```
petersteneteg: this will overflow in cases where min and max is equals to numeric_limits<T>::lowest and numeric_limits<T>::max which it not uncommon.

modules/propertybasedtesting/include/inviwo/propertybasedtesting/algorithm/generatingassignments.h
petersteneteg: why shared_ptr and not unique_ptr?

modules/propertybasedtesting/include/inviwo/propertybasedtesting/algorithm/generatingassignments.h
```diff
@@ -0,0 +1,164 @@
...
+                res.emplace_back(std::make_shared<PropertyAssignmentTyped<P>>(
+                    deactivated, prop, prop->getMinValue() + (rand() % maxSteps) * increment));
+            }
+        }
+
+        return res;
+    }
+};
+
+// NOTE: We assume that a MinMaxProperty<T>* may be casted to
```
petersteneteg: this assumption is false

modules/propertybasedtesting/include/inviwo/propertybasedtesting/algorithm/propertyanalyzing.h
```diff
@@ -0,0 +1,127 @@
...
+#include <inviwo/core/processors/processor.h>
+#include <inviwo/core/properties/property.h>
+#include <inviwo/core/properties/optionproperty.h>
+
+#include <inviwo/propertybasedtesting/algorithm/generatingassignments.h>
+#include <optional>
+
+namespace inviwo {
+
+namespace util {
```
petersteneteg: I would use a module specific namespace like pbt instead.

modules/propertybasedtesting/include/inviwo/propertybasedtesting/algorithm/propertyanalyzing.h
```diff
@@ -0,0 +1,127 @@
...
+    GREATER,
+    GREATER_EQUAL,
+    ANY,
+    NOT_COMPARABLE
+};
+constexpr size_t numPropertyEffects = 1 + static_cast<size_t>(PropertyEffect::NOT_COMPARABLE);
+
+template <typename A, typename B>
+bool propertyEffectComparator(const PropertyEffect& e, const A& a, const B& b) {
+    assert(static_cast<size_t>(e) < numPropertyEffects);
```
petersteneteg: use IVW_ASSERT. and potentially use a default in the switch with the assert or throw an Exception

modules/propertybasedtesting/include/inviwo/propertybasedtesting/algorithm/propertyanalyzing.h
```diff
@@ -0,0 +1,127 @@
...
+            return a >= b;
+    }
+    assert(false);
+}
+
+using AssignmentComparator = std::function<std::optional<util::PropertyEffect>(
+    const std::shared_ptr<PropertyAssignment>& oldVal,
+    const std::shared_ptr<PropertyAssignment>& newVal)>;
+
+std::ostream& operator<<(std::ostream& out, const PropertyEffect& a);
```
petersteneteg: Missing API macro, IVW_MODULE_[module name]_API

modules/propertybasedtesting/include/inviwo/propertybasedtesting/algorithm/propertyanalyzing.h
```diff
@@ -0,0 +1,127 @@
...
+
+template <typename T>
+std::ostream& operator<<(std::ostream& out, const std::optional<T>& a) {
```
petersteneteg: overloading operators for standard library types is a very bad idea. Only ever do this for Ts that you have defined your self.

modules/propertybasedtesting/include/inviwo/propertybasedtesting/algorithm/propertyanalyzing.h
```diff
@@ -0,0 +1,127 @@
...
+    return (!a ? out << "{}" : out << "{" << *a << "}");
+}
+
+std::optional<PropertyEffect> combine(const PropertyEffect& a, const PropertyEffect& b);
```
petersteneteg: Missing API macro, IVW_MODULE_[module name]_API

modules/propertybasedtesting/include/inviwo/propertybasedtesting/algorithm/propertyanalyzing.h
```diff
@@ -0,0 +1,127 @@
...
+
+const PropertyEffect& reverseEffect(const PropertyEffect& pe);
```
petersteneteg: Missing API macro, IVW_MODULE_[module name]_API

modules/propertybasedtesting/include/inviwo/propertybasedtesting/algorithm/propertyanalyzing.h
```diff
@@ -0,0 +1,127 @@
...
+};
+template <typename T>
+struct GetComponent<T, 1> {
+    static T get(const T& v, size_t i) {
+        assert(i == 0);
+        return v;
+    }
+};
+
+std::optional<Processor*> getOwningProcessor(Property* const prop);
```
petersteneteg: Missing API macro, IVW_MODULE_[module name]_API
having a optional pointer is but odd, the pointer is self also has a empty state, ie. nullptr, why and an other empty state with the optional?

modules/propertybasedtesting/include/inviwo/propertybasedtesting/algorithm/propertyanalyzing.h
```diff
@@ -0,0 +1,127 @@
...
+template <typename T>
+PropertyEffect propertyEffect(const PropertyEffect& selectedEffect, const T& newVal,
+                              const T& oldVal) {
+    if (newVal > oldVal) return selectedEffect;
+    if (newVal == oldVal) return PropertyEffect::ANY;
+    return reverseEffect(selectedEffect);
+}
+
+template <typename T, size_t N>
+struct GetComponent {
```
petersteneteg: what is this for? Needs documentation.

modules/propertybasedtesting/include/inviwo/propertybasedtesting/algorithm/reservoirsampling.h
```diff
@@ -0,0 +1,41 @@
...
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ *********************************************************************************/
+#pragma once
+
+#include <inviwo/propertybasedtesting/propertybasedtestingmoduledefine.h>
+#include <inviwo/core/common/inviwo.h>
```
petersteneteg: this include is not needed, include vector.

modules/propertybasedtesting/include/inviwo/propertybasedtesting/html/html.h
```diff
@@ -0,0 +1,197 @@
...
+
+#pragma once
+
+#include <inviwo/propertybasedtesting/propertybasedtestingmoduledefine.h>
+#include <inviwo/propertybasedtesting/testproperty.h>
+
+#include <iostream>
+
+namespace inviwo {
+namespace HTML {
```
petersteneteg: have you considered using inviwo/core/util/document.h?

modules/propertybasedtesting/include/inviwo/propertybasedtesting/html/html.h
```diff
@@ -0,0 +1,197 @@
...
+    }
+    Derived& addAttribute(const std::string& aName, const std::string& aValue) {
+        attributes.push_back({aName, aValue});
+        return *static_cast<Derived*>(this);
+    }
+
+    virtual ~Element() = default;
+};
+
+class HTML : public Element<HTML> {
```
petersteneteg: Missing API macro, IVW_MODULE_[module name]_API
and for the others below...

modules/propertybasedtesting/include/inviwo/propertybasedtesting/html/report.h
```diff
@@ -0,0 +1,53 @@
...
+#include <inviwo/propertybasedtesting/propertybasedtestingmoduledefine.h>
+#include <inviwo/propertybasedtesting/testproperty.h>
+
+#include <inviwo/propertybasedtesting/html/html.h>
+#include <iostream>
+#include <optional>
+
+namespace inviwo {
+
+class PropertyBasedTestingReport {
```
petersteneteg: Missing API macro, IVW_MODULE_[module name]_API

modules/propertybasedtesting/include/inviwo/propertybasedtesting/processors/imagecomparator.h
```diff
@@ -0,0 +1,142 @@
...
+#include <inviwo/core/properties/ordinalproperty.h>
+#include <inviwo/core/properties/boolproperty.h>
+#include <inviwo/core/properties/directoryproperty.h>
+#include <inviwo/core/ports/imageport.h>
+
+#include <filesystem>
+
+namespace inviwo {
+
+/** \docpage{org.inviwo.ImageComparator, Image Comparator}
```
petersteneteg, line 44: fill in template

modules/propertybasedtesting/include/inviwo/propertybasedtesting/processors/imagecomparator.h
```diff
@@ -0,0 +1,142 @@
...
+    double difference(const ComparisonType&, const glm::dvec4&, const glm::dvec4&);
+    double absoluteARGBdifference(const glm::dvec4&, const glm::dvec4&);
+
+    struct Comparison {
+        time_t timestamp;
+        double result;
+        ReductionType reduction;
+        size_t differentPixels;
+        size_t pixelCount;
+        std::filesystem::path img1;
```
petersteneteg, line 151: We have not used std::filesystem yet, for lack of complier support especially Xcode. But I think we will require support for it soonish.

modules/propertybasedtesting/include/inviwo/propertybasedtesting/processors/propertyanalyzer.h
```diff
@@ -0,0 +1,249 @@
...
+ *   Afterwards a new report is generated, which now only contains the tests
+ *   that have failed with the distilled set of properties.
+ */
+class IVW_MODULE_PROPERTYBASEDTESTING_API PropertyAnalyzer : public Processor,
+                                                             private TestPropertyObserver,
+                                                             private ProcessorNetworkObserver {
+public:
+    PropertyAnalyzer(InviwoApplication*);
+    virtual ~PropertyAnalyzer() {
+        const std::lock_guard<std::mutex> lock(mutex_);
```
petersteneteg: What is going on here? Having a lock here on the main thread, seems like a very bad ide.

modules/propertybasedtesting/include/inviwo/propertybasedtesting/processors/propertyanalyzer.h
```diff
@@ -0,0 +1,249 @@
...
+    static std::mutex mutex_;
+    enum TestingState { NONE, GATHERING, SINGLE_COUNT };
+    TestingState testingState;
+    bool currently_condensing = false;
+
+    InviwoApplication* const app_;
+    std::filesystem::path tempDir_;
+
+    ImageInport inport_;
+    std::optional<std::shared_ptr<Image>> outputImage;
```
petersteneteg: no point of having a optional pointer...

modules/propertybasedtesting/include/inviwo/propertybasedtesting/algorithm/reservoirsampling.h
```diff
@@ -0,0 +1,41 @@
...
+
+namespace inviwo {
+
+// randomly select k elements out of n elements (indexed from 0 to n-1)
+// note: if k > n, then all n elements are chosen
+// Time complexity: O(k * (1+log(n/k)))
+std::vector<size_t> reservoirSampling(const size_t n, const size_t k);
```
petersteneteg: Missing API macro, IVW_MODULE_[module name]_API

modules/propertybasedtesting/include/inviwo/propertybasedtesting/processors/propertyanalyzer.h
```diff
@@ -0,0 +1,249 @@
...
+
+    virtual void setNetwork(ProcessorNetwork*) override;
+
+private:
+    static std::set<size_t> curr_alive;
+    size_t m_id;
+    template <typename F>
+    void dispatchFrontAndForget(F);
+
+    static std::mutex mutex_;
```
petersteneteg: is this guarding the curr_alive? then you need to lock on both read and write

modules/propertybasedtesting/include/inviwo/propertybasedtesting/processors/randomvolumegenerator.h
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * ### Outports
+ *   * __outport__ The generated volume.
+ *
+ * ### Properties
+ *   * __Seed__ The seed used for the random number generator.
+ *   * __Number of Points__ The number of points to use for the volume
+ *   generation.
+ */
+class IVW_MODULE_PROPERTYBASEDTESTING_API RandomVolumeGenerator : public Processor {
```
petersteneteg: RandomVolume is quite vague, maybe, RandomDistanceField?

modules/propertybasedtesting/include/inviwo/propertybasedtesting/testproperty.h
```diff
@@ -0,0 +1,491 @@
...
+#include <inviwo/core/properties/ordinalproperty.h>
+#include <inviwo/core/properties/cameraproperty.h>
+#include <inviwo/core/properties/boolcompositeproperty.h>
+
+#include <filesystem>
+#include <fstream>
+
+namespace inviwo {
+
+class TestResult;
```
petersteneteg: put all of this into a pbt namespace

modules/propertybasedtesting/include/inviwo/propertybasedtesting/testproperty.h
```diff
@@ -0,0 +1,491 @@
...
+
+class TestPropertyObservable;
+
+class TestPropertyObserver : public Observer {
+public:
+    friend TestPropertyObservable;
+    TestPropertyObserver() = default;
+    virtual void onTestPropertyChange() {}
```
petersteneteg, line 57: how does this differ from the regular onChange on a property?

modules/propertybasedtesting/include/inviwo/propertybasedtesting/testproperty.h
```diff
@@ -0,0 +1,491 @@
...
+    void notifyObserversAboutChange() {
+        forEachObserver([](TestPropertyObserver* o) { o->onTestPropertyChange(); });
+    }
+    virtual ~TestPropertyObservable() = default;
+};
+
+template <typename T>
+class NetworkPath : public Serializable {
+private:
+    std::unique_ptr<std::string> path_;
```
petersteneteg: this looks very strange, why would you put a string and a T* into a unique_ptr?

modules/propertybasedtesting/include/inviwo/propertybasedtesting/testproperty.h
```diff
@@ -0,0 +1,491 @@
...
+            *ptrPtr_ = nullptr;
+        }
+        pn_ = pn;
+    }
+
+    T* get() const {
+        if (*ptrPtr_ == nullptr) {
+            assert(pn_ != nullptr);
+            if (isProcessor_) {
+                return *ptrPtr_ = reinterpret_cast<T*>(pn_->getProcessorByIdentifier(*path_));
```
petersteneteg: you should never use reinterpret_cast here.

modules/propertybasedtesting/include/inviwo/propertybasedtesting/testproperty.h
```diff
@@ -0,0 +1,491 @@
...
+        if (pn == nullptr) {
+            *path_ = path();
+            *ptrPtr_ = nullptr;
+        }
+        pn_ = pn;
+    }
+
+    T* get() const {
+        if (*ptrPtr_ == nullptr) {
+            assert(pn_ != nullptr);
```
petersteneteg: use IVW_ASSERT

modules/propertybasedtesting/include/inviwo/propertybasedtesting/testproperty.h
```diff
@@ -0,0 +1,491 @@
...
+protected:
+    TestPropertyObservable() = default;
+    void notifyObserversAboutChange() {
+        forEachObserver([](TestPropertyObserver* o) { o->onTestPropertyChange(); });
+    }
+    virtual ~TestPropertyObservable() = default;
+};
+
+template <typename T>
+class NetworkPath : public Serializable {
```
petersteneteg, line 79: this class needs more work, can you explain what you want it to do?

modules/propertybasedtesting/include/inviwo/propertybasedtesting/testproperty.h
```diff
@@ -0,0 +1,491 @@
...
+            {"EQUAL", "EQUAL", 0},     {"NOT_EQUAL", "NOT_EQUAL", 1},
+            {"LESS", "LESS", 2},       {"LESS_EQUAL", "LESS_EQUAL", 3},
+            {"GREATER", "GREATER", 4}, {"GREATER_EQUAL", "GREATER_EQUAL", 5},
+            {"ANY", "ANY", 6},         {"NOT_COMPARABLE", "NOT_COMPARABLE", 7}};
+        return options;
+    }
+
+public:
+    void setNetwork(ProcessorNetwork*);
+    virtual size_t totalCheckedComponents() const = 0;
```
petersteneteg: all these functions need documentation

modules/propertybasedtesting/include/inviwo/propertybasedtesting/testproperty.h
```diff
@@ -0,0 +1,491 @@
...
+ * It is used to provide a general interface to all properties, regardless of
+ * their value_type.
+ * It enables
+ *	- generation of assignments,
+ *	- a textual description of the expected effects on the counted number of pixels,
+ *	- the expected effect on the counted number of pixels given two testcases, and
+ *  - storing and resetting the value of the property befor the tests have started.
+ * Instantiation happes mainly by testableProperty (see below).
+ */
+class TestProperty : public Serializable, public TestPropertyObservable {
```
petersteneteg: The naming here is a but unfortunate, since inviwo uses Property. And this is not a Inviwo Property. Could it be named something else?

modules/propertybasedtesting/include/inviwo/propertybasedtesting/testproperty.h
```diff
@@ -0,0 +1,491 @@
...
+std::optional<util::PropertyEffect> propertyEffect(
+    const T& val_old, const T& val_new,
+    const std::array<util::PropertyEffect, numComp>& selectedEffects);
+
+/*
+ * Creates and returns a TestProperty for an arbitrary property, if possible.
+ * That is, if the given property derives from CompositeProperty or one of
+ * the properties in PropertyTypes (see below).
+ */
+std::optional<std::unique_ptr<TestProperty>> testableProperty(Property* prop);
```
petersteneteg: Missing API macro, IVW_MODULE_[module name]_API
All these free function should be a the pbt namespace and potentially in side of a util namespace inside of pbt
No point in returning a optional pointer again... the pointer it self already has an empty state.

petersteneteg: I world have verb in the name, createTestableProperty maybe

modules/propertybasedtesting/include/inviwo/propertybasedtesting/testproperty.h
```diff
@@ -0,0 +1,491 @@
...
+        std::shared_ptr<TestResult> newTestResult,
+        std::shared_ptr<TestResult> oldTestResult) const override;
+
+    std::ostream& ostr(std::ostream& out, std::shared_ptr<TestResult> testResult) const override;
+
+    std::ostream& ostr(std::ostream& out, std::shared_ptr<TestResult> newTestResult,
+                       std::shared_ptr<TestResult> oldTestResult) const override;
+
+    // Constructor for all Propertyowners with a display name and an identifier
+    template <typename C, decltype(static_cast<PropertyOwner*>(std::declval<C*>()),
```
petersteneteg: Why not just add a regular constructor for CompositeProperty and one for Processor?

modules/propertybasedtesting/include/inviwo/propertybasedtesting/testproperty.h
```diff
@@ -0,0 +1,491 @@
...
+    static constexpr size_t numComponents = DataFormat<value_type>::components();
+
+    NetworkPath<T> typedProperty_;
+    value_type defaultValue_;
+    std::array<NetworkPath<OptionPropertyInt>, numComponents> effectOption_;
+
+    TestPropertyTyped() = default;
+
+    std::array<util::PropertyEffect, numComponents> selectedEffects() const;
+    std::unique_ptr<bool> deactivated_;
```
petersteneteg: why would you put a bool on the heap?

modules/propertybasedtesting/include/inviwo/propertybasedtesting/testproperty.h
```diff
@@ -0,0 +1,491 @@
...
+        std::pair<util::AssignmentComparator, std::vector<std::shared_ptr<PropertyAssignment>>>>
+    generateAssignmentsCmp() const override;
+};
+
+/*
+ * Container for a test result.
+ * Allows access to the tested value of each property, as well as the
+ * resulted number of pixels.
+ */
+class TestResult {
```
petersteneteg: Missing API macro, IVW_MODULE_[module name]_API

should probably more this into a separate file

modules/propertybasedtesting/include/inviwo/propertybasedtesting/testproperty.h
```diff
@@ -0,0 +1,491 @@
...
+    const std::array<util::PropertyEffect, numComp>& selectedEffects) {
+
+    std::optional<util::PropertyEffect> res = {util::PropertyEffect::ANY};
+    for (size_t i = 0; res && i < numComp; i++) {
+        auto compEff = util::propertyEffect(selectedEffects[i],
+                                            util::GetComponent<T, numComp>::get(val_new, i),
+                                            util::GetComponent<T, numComp>::get(val_old, i));
+        res = util::combine(*res, compEff);
+    }
+    return res;
```
petersteneteg, line 461: why return a optional? the result will never be nullopt?

modules/propertybasedtesting/include/inviwo/propertybasedtesting/testproperty.h
```diff
@@ -0,0 +1,491 @@
...
+}
+
+template <typename F>
+void TestPropertyComposite::for_each_checked(F f) const {
+    for (const auto& prop : subProperties) {
+        if (prop->getBoolComp()->isChecked()) f(prop.get());
+    }
+}
+
+class TestPropertyFactory : public Factory<TestProperty> {
```
petersteneteg: Missing API macro, IVW_MODULE_[module name]_API

move into separate file
Why not use the StandardFactory?

modules/propertybasedtesting/include/inviwo/propertybasedtesting/testproperty.h
```diff
@@ -0,0 +1,491 @@
...
+    static const std::unordered_map<std::string, std::function<std::unique_ptr<TestProperty>()>>
```
petersteneteg: don't use statics for this, the factory is not a and should not be a singleton

modules/propertybasedtesting/include/inviwo/propertybasedtesting/testproperty.h
```diff
@@ -0,0 +1,491 @@
...
+
+    for (auto def : defaultValues) {
+        if (auto p = dynamic_cast<TestPropertyTyped<T>*>(def); p != nullptr) {
+            if (p->getTypedProperty() == prop) return p->getDefaultValue();
+        } else if (auto p = dynamic_cast<TestPropertyComposite*>(def); p != nullptr) {
+            if (auto res = p->getDefaultValue(prop); res != std::nullopt) return *res;
+        }
+    }
+    std::cerr << "could not get value for " << prop << " " << typeid(T).name() << std::endl;
+    exit(1);
```
petersteneteg: use a assert or throw an exception. never use std::cerr, use LogError.

modules/propertybasedtesting/src/algorithm/coveringarray.cpp
```diff
@@ -0,0 +1,225 @@
...
+namespace inviwo {
+
+namespace util {
+
+// 2-coverage, randomized discrete SLJ strategy
+std::vector<Test> coveringArray(
+    const Test& init, const std::vector<std::vector<std::shared_ptr<PropertyAssignment>>>& vars) {
+    srand(42);  // deterministic for regression testing
+
+    std::cerr << "coveringArray: vars.size() = " << vars.size() << std::endl;
```
petersteneteg: Use LogInfo if needed, in the standard case there should be minimal logging

modules/propertybasedtesting/src/algorithm/coveringarray.cpp
```diff
@@ -0,0 +1,225 @@
...
+    assert(vars.size() > 0);
+
+    // special case
+    if (vars.size() == 1) {
+        std::vector<Test> res(vars[0].size(), init);
+        for (size_t i = 0; i < vars[0].size(); i++) res[i].emplace_back(vars[0][i]);
```
petersteneteg, line 57: use std::copy

modules/propertybasedtesting/src/algorithm/coveringarray.cpp
```diff
@@ -0,0 +1,225 @@
...
+        std::cerr << "special case : res.size() = " << res.size() << std::endl;
+        return res;
+    }
+
+    const size_t v = [&]() {
+        size_t res = 0;
+        for (const auto& v : vars) res = std::max(res, v.size());
```
petersteneteg: use std::max_element

modules/propertybasedtesting/src/algorithm/coveringarray.cpp
```diff
@@ -0,0 +1,225 @@
...
+
+    // special case
+    if (vars.size() == 1) {
+        std::vector<Test> res(vars[0].size(), init);
+        for (size_t i = 0; i < vars[0].size(); i++) res[i].emplace_back(vars[0][i]);
+        std::cerr << "special case : res.size() = " << res.size() << std::endl;
+        return res;
+    }
+
+    const size_t v = [&]() {
```
petersteneteg: v?

modules/propertybasedtesting/src/algorithm/coveringarray.cpp
```diff
@@ -0,0 +1,225 @@
...
+    for (size_t i = 1; i < vars.size(); i++) {
+        for (size_t j = 0; j < i; j++) {
+            for (size_t ii = 0; ii < vars[i].size(); ii++) {
+                for (size_t ji = 0; ji < vars[j].size(); ji++) {
+                    uncovered.insert(idx.size());
+                    idx[{i, j, ii, ji}] = idx.size();
+                }
+            }
+        }
+    }
```
petersteneteg: maybe a some comment on what happens

modules/propertybasedtesting/src/algorithm/coveringarray.cpp
```diff
@@ -0,0 +1,225 @@
...
+
+    std::vector<std::vector<size_t>> coveringArray;
+    while (!uncovered.empty()) {
+        size_t expectedCoverage =
+            (uncovered.size() + (v * v - 1)) / (v * v);  // expectedCoverage > 0
+        size_t coverage;
+        std::vector<size_t> row(vars.size());
+        do {
+            for (size_t i = 0; i < row.size(); i++) row[i] = rand() % vars[i].size();
```
petersteneteg: never use rand, use the modering random tools in the standard library, also rand() % N is not a uniform distribution if that was what you were going for.

modules/propertybasedtesting/src/algorithm/coveringarray.cpp
```diff
@@ -0,0 +1,225 @@
...
+            for (size_t ii = 0; ii < vars[i].size(); ii++) {
+                for (size_t ji = 0; ji < vars[j].size(); ji++) {
+                    uncovered.insert(idx.size());
+                    idx[{i, j, ii, ji}] = idx.size();
+                }
+            }
+        }
+    }
+
+    std::vector<std::vector<size_t>> coveringArray;
```
petersteneteg: using a vector of a vector us really bad from a cache perspective, if all the rows a the same size, consider using a single vector and a IndexMapper instead.

modules/propertybasedtesting/src/algorithm/coveringarray.cpp
```diff
@@ -0,0 +1,225 @@
...
+                }
+            }
+        } while (coverage < expectedCoverage);
+        for (size_t i = 1; i < vars.size(); i++) {
+            for (size_t j = 0; j < i; j++) {
+                size_t id = idx[{i, j, row[i], row[j]}];
+                uncovered.erase(id);
+            }
+        }
+        coveringArray.emplace_back(row);
```
petersteneteg: `coveringArray.emplace_back(std::move(row));`

modules/propertybasedtesting/src/algorithm/coveringarray.cpp
```diff
@@ -0,0 +1,225 @@
...
+    return res;
+}
+
+std::vector<Test> optCoveringArray(
+    const size_t num, const Test& init,
+    const std::vector<std::pair<std::function<std::optional<util::PropertyEffect>(
+                                    const std::shared_ptr<PropertyAssignment>& oldVal,
+                                    const std::shared_ptr<PropertyAssignment>& newVal)>,
+                                std::vector<std::shared_ptr<PropertyAssignment>>>>& vars) {
+    srand(42);  // deterministic for regression testing
```
petersteneteg: the the modern random tools

modules/propertybasedtesting/src/algorithm/coveringarray.cpp
```diff
@@ -0,0 +1,225 @@
...
+    std::shuffle(unused.begin(), unused.end(), std::default_random_engine(rand()));
+
+    std::vector<std::pair<TestConf, size_t>>
+        finished;  // {TestConf, current num of other finished comparables}
+
+    size_t comparisons = 0;  // just for debugging
+
+    // note: assumes that all keys of the second argument are also present in
+    // the first
+    const std::function<bool(const TestConf&, const TestConf&)> comparable = [&](const auto& ref,
```
petersteneteg: avoid using std::function unless you actually need the type erasure, just use auto comparable = ...

modules/propertybasedtesting/src/algorithm/coveringarray.cpp
```diff
@@ -0,0 +1,225 @@
...
+            const auto& expect = vars[var].first(vars[var].second[i], vars[var].second[j]);
+            comparisons++;
+            if (!expect) return false;
+            res = util::combine(*expect, *res);
+            if (!res) return false;
+        }
+        return true;
+    };
+
+    const std::function<std::vector<size_t>(std::vector<size_t>, const TestConf&)>
```
petersteneteg: See comment above

modules/propertybasedtesting/src/algorithm/coveringarray.cpp
```diff
@@ -0,0 +1,225 @@
...
+                if (!comparable(finished[comparables[i]].first, test)) {
+                    comparables[i] = comparables.back();
+                    comparables.pop_back();
+                    i--;
+                }
+            }
+            return comparables;
+        };
+    // combined score of finished comparables when merging ref into gen
+    const std::function<size_t(const bool, const std::vector<size_t>&, const TestConf&,
```
petersteneteg: See comment above

modules/propertybasedtesting/src/algorithm/propertyanalyzing.cpp
```diff
@@ -0,0 +1,63 @@
+#include <inviwo/propertybasedtesting/algorithm/propertyanalyzing.h>
+
+namespace inviwo {
+
+namespace util {
+
+std::ostream& operator<<(std::ostream& out, const PropertyEffect& a) {
+    static const std::string names[] = {"EQUAL",   "NOT_EQUAL",     "LESS", "LESS_EQUAL",
```
petersteneteg: `static constexpr std::string_view names[] = ...`

modules/propertybasedtesting/src/algorithm/propertyanalyzing.cpp
```diff
@@ -0,0 +1,63 @@
...
+        {false, false, true, false, false},  // EQUAL
+        {true, false, false, false, true},   // NOT_EQUAL
+        {true, false, false, false, false},  // LESS
+        {true, true, false, false, false},   // LESS_EQUAL
+        {false, false, false, false, true},  // GREATER
+        {false, false, false, true, true},   // GREATER_EQUAL
+        {true, true, true, true, true},      // ANY
+        {false, false, false, false, false}  // NOT_COMPARABLE
+    }};
+    auto resAll = compatibility[(size_t)a];
```
petersteneteg: never use c-style casts, use static_cast<size_t>(a)

modules/propertybasedtesting/src/algorithm/propertyanalyzing.cpp
```diff
@@ -0,0 +1,63 @@
...
+    for (size_t i = 0; i < resAll.size(); i++) resAll[i] &= compatibility[(size_t)b][i];
+
+    for (size_t i = 0; i < numPropertyEffects; i++) {
+        if (resAll == compatibility[i]) {
+            PropertyEffect res = PropertyEffect(i);
+            if (res != PropertyEffect::NOT_COMPARABLE) return PropertyEffect(i);
+        }
+    }
+
+    return std::nullopt;
```
petersteneteg: why not return NOT_COMPAREABLE instead of std::nullopt?

modules/propertybasedtesting/src/algorithm/propertyanalyzing.cpp
```diff
@@ -0,0 +1,63 @@
...
+        PropertyEffect::LESS_EQUAL,     // GREATER_EQUAL
+        PropertyEffect::ANY,            // ANY
+        PropertyEffect::NOT_COMPARABLE  // NOT_COMPARABLE
+    };
+    return reverseEffects[(size_t)pe];
+}
+
+std::optional<Processor*> getOwningProcessor(Property* const prop) {
+    PropertyOwner* const owner = prop->getOwner();
+    if (Processor* const proc = dynamic_cast<Processor*>(owner); proc != nullptr) return {proc};
```
petersteneteg: just use the idiomatic 
`if (auto proc = dynamic_cast<Processor*>(owner)) return proc;`

modules/propertybasedtesting/src/algorithm/reservoirsampling.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+namespace inviwo {
+
+std::vector<size_t> reservoirSampling(const size_t n, const size_t k) {
+    if (k > n) return reservoirSampling(n, n);
+
+    std::vector<size_t> result(k);
+    std::iota(result.begin(), result.end(), 0);  // start with the first k values
+
+    double w = exp(log((double(rand()) / RAND_MAX)) / k);
```
petersteneteg: never use rand() use the modern random tools

modules/propertybasedtesting/src/processors/faultycubeproxyprocessor.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+namespace inviwo {
+
+// The Class Identifier has to be globally unique. Use a reverse DNS naming scheme
+const ProcessorInfo FaultyCubePoxyGeometry::processorInfo_{
+    "org.inviwo.FaultyCubePoxyGeometry",  // Class identifier
+    "Faulty Cube Proxy Geometry",         // Display name
+    "Undefined",                          // Category
+    CodeState::Experimental,              // Code state
+    Tags::None,                           // Tags
```
petersteneteg: you can make this hidden so it dosen't show ut in the GUI but you can still use it in the tests.

modules/propertybasedtesting/src/processors/imagecomparator.cpp
```diff
@@ -0,0 +1,262 @@
...
+                << " != " << dim2;
+            util::log(IVW_CONTEXT, str.str(), LogLevel::Info, LogAudience::User);
+            return false;
+        }
+        return true;
+    });
+}
+
+void ImageComparator::setNetwork(ProcessorNetwork* network) {
+    if (network) network->addObserver(this);
```
petersteneteg: Why?

modules/propertybasedtesting/src/processors/imagecomparator.cpp
```diff
@@ -0,0 +1,262 @@
...
+    auto colorLayerDiff = diffRAM->getColorLayerRAM();
+    auto colorLayerMask = maskRAM->getColorLayerRAM();
+
+    const ReductionType reduction = reductionType_.get();
+
+    double result = getUnitForReduction<double>(reduction);
+    size_t diffPixels = 0;
+    for (size_t x = 0; x < dim1.x; x++) {
+        for (size_t y = 0; y < dim1.y; y++) {
+            const auto col1 = colorLayerRAM1->getAsDVec4(size2_t(x, y));
```
petersteneteg: these virtual function as very slow, and should really only be use for checking a few pixels not all of them. Use LayerRAM->dispatch instead.

modules/propertybasedtesting/src/processors/randomvolumegenerator.cpp
```diff
@@ -0,0 +1,82 @@
...
+    const size3_t dimensions(1 << 5);
+    const mat3 basis(1.0);  // identity matrix
+
+    std::vector<dvec3> points(numPoints_.get());
+    for (auto& pos : points) {
+        pos = dvec3(dis(generator), dis(generator), dis(generator));
+    }
+
+    std::shared_ptr<Volume> volume =
+        util::generateVolume(dimensions, basis, [&](const size3_t& ind) {
```
petersteneteg: util::generateVolume is to vague a name. use a more unique one

modules/propertybasedtesting/src/processors/propertyanalyzer.cpp
```diff
@@ -0,0 +1,659 @@
...
+    "org.inviwo.PropertyAnalyzer",  // Class identifier
+    "PropertyAnalyzer",             // Display name
+    "Undefined",                    // Category
+    CodeState::Experimental,        // Code state
+    Tags::None,                     // Tags
+};
+const ProcessorInfo PropertyAnalyzer::getProcessorInfo() const { return processorInfo_; }
+
+void PropertyAnalyzer::resetAllProps() {
+    const std::lock_guard<std::mutex> lock(mutex_);
```
petersteneteg: what does the mutex protect?

modules/propertybasedtesting/src/processors/propertyanalyzer.cpp
```diff
@@ -0,0 +1,659 @@
...
+}
+void PropertyAnalyzer::updateProcessors() {
+    const std::lock_guard<std::mutex> lock(mutex_);
+    std::cerr << "PropertyAnalyzer::updateProcessors() @ " << this << std::endl;
+
+    processors_.insert(std::make_move_iterator(inactiveProcessors_.begin()),
+                       std::make_move_iterator(inactiveProcessors_.end()));
+    inactiveProcessors_.clear();
+
+    // for(const auto&[proc,tp] : processors_) {
```
petersteneteg: remove

modules/propertybasedtesting/src/processors/propertyanalyzer.cpp
```diff
@@ -0,0 +1,659 @@
...
+        i++;
+    }
+    for (const auto& [proc, tp] : inactiveProcessors_) {
+        s.serialize("Proc" + std::to_string(i), proc);
+        s.serialize("Tp" + std::to_string(i), tp);
+        i++;
+    }
+}
+
+void PropertyAnalyzer::deserialize(Deserializer& d) {
```
petersteneteg, line 325: this also has to handle state changes for the undo functionality.

modules/propertybasedtesting/src/processors/propertyanalyzer.cpp
```diff
@@ -0,0 +1,659 @@
...
+                if (pixelCol.x == 1) res++;
+            } else {
+                if (glm::length(pixelCol - col) < 1e-3) res++;
+            }
+        }
+    }
+    return res;
+}
+
+void PropertyAnalyzer::process() {
```
petersteneteg, line 652: Im fairly skeptical to use a processor to do all this, would be much clear to make it a separate thing, like for example how the animation stuff works.

modules/propertybasedtesting/src/testproperty.cpp
```diff
@@ -0,0 +1,559 @@
...
+ *********************************************************************************/
+
+#include <inviwo/propertybasedtesting/testproperty.h>
+
+namespace inviwo {
+
+TestProperty::TestProperty(const std::string& displayName, const std::string& identifier)
+    : displayName_(displayName)
+    , identifier_(identifier)
+    , boolComp_(new BoolCompositeProperty(identifier_ + "Comp", displayName_, false)) {
```
petersteneteg, line 40: why a pointer? why not just have it as a member?

petersteneteg (changes requested):
Interesting work! 
I think you should consider strongly moving the propertyanalyzer out of the network. you can look at the animation modules for how that can be done. 
I have not looked much at the propertyanalyzer.cpp or testproperty.cpp yet.

modules/propertybasedtesting/include/inviwo/propertybasedtesting/testproperty.h
```diff
@@ -0,0 +1,491 @@
...
+                Property* tmp = pn_->getProperty(*path_);
+                assert(tmp != nullptr);
+                *ptrPtr_ = dynamic_cast<T*>(tmp);
+                assert(*ptrPtr_ != nullptr);
+            }
+        }
+        return *ptrPtr_;
+    }
+    operator T*() const { return get(); }
+    T* operator->() const { return get(); }
```
xeTaiz, line 174: Why are these necessary?

modules/propertybasedtesting/include/inviwo/propertybasedtesting/algorithm/generatingassignments.h
xeTaiz: I don't think that is necessarily the case

modules/propertybasedtesting/src/processors/imagecomparator.cpp
```diff
@@ -0,0 +1,262 @@
...
+            const auto col2 = colorLayerRAM2->getAsDVec4(size2_t(x, y));
+            const double diff = difference(comparisonType_.get(), col1, col2);
+            const bool pixelDifferent =
+                (diff / std::max({1.0, glm::length(col1), glm::length(col2)})) <=
+                maxPixelwiseDeviation_.get();
+            const double c = pixelDifferent * 255.0;
+
+            result = combine(reduction, result, diff);
+            colorLayerDiff->setFromDVec3(size2_t(x, y), 128.0 + (col1 - col2) / 2.0);
```
xeTaiz: with col1=255.0 and col2=0.0 that would result in a color value of 255.5.
Changing the constant to 127.5 should solve this

modules/marketplace/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
+ivw_module(Marketplace)
```
petersteneteg: The marketplace module should not be in this PR

petersteneteg (changes requested):

modules/propertybasedtesting/include/inviwo/propertybasedtesting/algorithm/generatingassignments.h
qwertzyuiop: A Property might be assigned the same value in multiple different Tests, hence the use of shared_ptr

modules/propertybasedtesting/include/inviwo/propertybasedtesting/testproperty.h
qwertzyuiop: We use this class for a reference to some object (Processor or Property) of type T in the ProcessorNetwork (which is
necessary since we need to access their properties or values). A normal pointer (T*) is not sufficient,
since serializing two pointers to the same object and then deserializing them yields two separate (but equal) objects, which is not what we want. Instead, we store and serialize the identifier of/path to the object (i.e. a string). Because querying the Network for the object corresponding to the identifier/path each time we want to access the referenced object seemed expensive, we also cache the pointer on the first access after deserialization.

petersteneteg: That seems resonable I think. What values can T have? is that always Processor or Property or can it be different things? is it always the base class or can it be derived classes to? ~~Would be good to add this explanation as documentation of the class~~ did not see the comment first...

modules/propertybasedtesting/include/inviwo/propertybasedtesting/testproperty.h
```diff
@@ -0,0 +1,480 @@
...
+class IVW_MODULE_PROPERTYBASEDTESTING_API TestPropertyObservable : public Observable<TestPropertyObserver> {
+protected:
+    TestPropertyObservable() = default;
+    void notifyObserversAboutChange() {
+        forEachObserver([](TestPropertyObserver* o) { o->onTestPropertyChange(); });
+    }
+    virtual ~TestPropertyObservable() = default;
+};
+
+/** \docpage{org.inviwo.PropertyAnalyzer, PropertyAnalyzer}
```
petersteneteg: The docpage only used for processors, to be able to show the processor help inside of inviwo...

modules/propertybasedtesting/include/inviwo/propertybasedtesting/testproperty.h
```diff
@@ -0,0 +1,480 @@
...
+ * ![](org.inviwo.NetworkPath.png?classIdentifier=org.inviwo.NetworkPath)
+ *
+ * A NetworkPath<T> is a reference to a Processor or Property of type T in the
+ * ProcessorNetwork. If the pointer to the object is not known (i.e.
+ * after deserialization) or one wants to deserialize a NetworkPath, 
+ * the corresponding ProcessorNetwork must be provided via
+ * setNetwork(...) before the deserialization/first access.
+ */
+template <typename T>
+class IVW_MODULE_PROPERTYBASEDTESTING_API NetworkPath : public Serializable {
```
petersteneteg: One can't export templates, neither functiosn or classes, hence there should not be any IVW_MODULE_X_API macros here...

petersteneteg (changes requested):
Just a couple of quick comments, I have not had time to go through all of it yet.
But from a glance it is looks very much improved!

modules/propertybasedtesting/include/inviwo/propertybasedtesting/algorithm/propertyanalyzing.h
qwertzyuiop: This is used to provide uniform access to the components of (glm) vector types (e.g. dvec2) and scalar types (e.g. floats):
GetComponent<float>::get(3,0) returns 3
GetComponent<dvec2>::get(dvec(4,5), 1) returns 5

petersteneteg:
Hi, nice work, starting to look quite nice I think... A few thinks left, can you add the module to this list: https://github.com/inviwo/inviwo/blob/709c00f8bb173b0fe9e73ece74d9054205aa52db/Jenkinsfile#L19
in the PR, then jenkins will build and test that module. Also there was some strange formatting sneaking in, http://jenkins.inviwo.org:8080/job/inviwo/job/PR-1089/Format/ is a diff or the needed updates.

modules/propertybasedtesting/include/inviwo/propertybasedtesting/algorithm/propertyanalyzing.h
```diff
@@ -86,7 +86,7 @@ using AssignmentComparator =
     std::function<PropertyEffect(const std::shared_ptr<PropertyAssignment>& oldVal,
                                  const std::shared_ptr<PropertyAssignment>& newVal)>;
 
-std::ostream& operator<<(std::ostream& out, const PropertyEffect& a);
+std::ostream& IVW_MODULE_PROPERTYBASEDTESTING_API operator<<(std::ostream& out, const PropertyEffect& a);
```
petersteneteg:
```suggestion
IVW_MODULE_PROPERTYBASEDTESTING_API std::ostream& operator<<(std::ostream& out, const PropertyEffect& a);
```

modules/propertybasedtesting/include/inviwo/propertybasedtesting/algorithm/propertyanalyzing.h
```diff
@@ -98,7 +98,7 @@ const PropertyEffect& IVW_MODULE_PROPERTYBASEDTESTING_API reverseEffect(const Pr
  * ANY if no preference
  */
 template <typename T>
-PropertyEffect propertyEffect(const PropertyEffect& selectedEffect, const T& newVal,
+PropertyEffect IVW_MODULE_PROPERTYBASEDTESTING_API propertyEffect(const PropertyEffect& selectedEffect, const T& newVal,
```
petersteneteg:
```suggestion
PropertyEffect propertyEffect(const PropertyEffect& selectedEffect, const T& newVal,
```

petersteneteg: templates should not have any API export pragmas

petersteneteg:
I see you got a merge issue, it is probably a good idea to rebase the branch on the latest master. If you want you can also clean up the history a bit by doing an interactive rebase. Then you can just force push to this branch again

modules/propertybasedtesting/src/testproperty.cpp
qwertzyuiop: The `BoolCompositeProperty` is either in a `Processor` or another `BoolCompositeProperty`, so
if we had this as a member, two identical versions of the property would be created on deserialization.
Also note that `boolComp_` is of type `NetworkPath<BoolCompositeProperty>`

modules/propertybasedtesting/include/inviwo/propertybasedtesting/testproperty.h
qwertzyuiop: They are not strictly necessary, but make accessing the object referenced by the `NetworkPath` much more convenient and allow using it mostly like one would use a pointer

modules/propertybasedtesting/src/algorithm/coveringarray.cpp
qwertzyuiop: I don't see how one would use std::copy here

modules/propertybasedtesting/src/algorithm/coveringarray.cpp
qwertzyuiop: I don't see how one could use `std::max_element` for this. `std::transform_reduce` and the like should work tho, but i don't think that would make the code any better

modules/propertybasedtesting/src/processors/randomvolumegenerator.cpp
qwertzyuiop: `util::generateVolume` is a procedure from the base module

modules/propertybasedtesting/src/algorithm/reservoirsampling.cpp
```diff
@@ -0,0 +1,32 @@
...
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ *********************************************************************************/
+
+#include <inviwo/propertybasedtesting/algorithm/reservoirsampling.h>
+
+namespace inviwo {}  // namespace inviwo
```
xeTaiz, line 32: This file is empty and can be deleted

modules/propertybasedtesting/src/processors/propertyanalyzer.cpp
```diff
@@ -0,0 +1,643 @@
...
+    }
+
+    // Invalidate all predecessors in order to force the update of their output
+    for (Processor* const processor : util::getPredecessors(this)) {
+        if (processor == this) continue;
+        processor->invalidate(InvalidationLevel::InvalidOutput);
+    }
+
+    dispatchPool([this]() {
+        // necessary because of synchronicity issues, TODO: find better solution
```
xeTaiz: This is to wait until the predecessors are evaluated I suppose?
I think by invalidating all predecessors, thus invalidating the input to PropertyAnalyzer, it should be automatically invalidated and evaluated after all its predecessors.

Otherwise `InviwoApplication::waitForPool()` may be a better option to wait for network evaluations. Or would that also be a bad idea from within a processor? @petersteneteg

xeTaiz (commented):
The randomvolumegenerator -> randomdistancefieldgenerator name change is still missing in the CMake.
But I think this looks very good already.

On the topic of having this in a separate GUI instead of a processor:
I think a separate GUI like the Animation editor would be awesome, but from my experience in replicating such a nice GUI, I think this change is out of scope for this already large PR.

modules/propertybasedtesting/src/algorithm/reservoirsampling.cpp
petersteneteg: There should __always__ be a cpp file for all header file even if it just contains a single include of its header.
This is to ensure that the header file can be compiled in isolation and is not depending on any other headers.

modules/propertybasedtesting/src/processors/propertyanalyzer.cpp
petersteneteg: Doing this kind of things from inside of a processor is not something I have considered, and I would generally advise against it.

petersteneteg:
just need to update date format, and fix then copyright year in some files, then this should be in quite good shape.

petersteneteg:
@martinfalk This only needs a format and copyright fix, could you help with that?

PR #1092, opened 2020-12-18
Qt: More options in the tools menu for updating workspaces

petersteneteg:
image

### DIFF
--- a/include/inviwo/core/network/workspaceutils.h
+++ b/include/inviwo/core/network/workspaceutils.h
@@ -31,7 +31,8 @@
 
 #include <inviwo/core/common/inviwocoredefine.h>
 
-#include <string>
+#include <string_view>
+#include <functional>
 
 namespace inviwo {
 
@@ -39,11 +40,16 @@ class InviwoApplication;
 
 namespace util {
 
-void IVW_CORE_API updateWorkspaces(InviwoApplication* app);
+IVW_CORE_API void forEachWorkspaceInDirRecusive(std::string_view path,
+                                                std::function<void(std::string_view)> callback);
 
-void IVW_CORE_API updateRegressionWorkspaces(InviwoApplication* app);
+enum class DryRun { Yes, No };
 
-void IVW_CORE_API updateWorkspaces(InviwoApplication* app, const std::string& path);
+IVW_CORE_API void updateWorkspaces(InviwoApplication* app, std::string_view path, DryRun dryRun);
+
+IVW_CORE_API void updateExampleWorkspaces(InviwoApplication* app, DryRun dryRun);
+
+IVW_CORE_API void updateRegressionWorkspaces(InviwoApplication* app, DryRun dryRun);
 
 }  // namespace util
 
--- a/src/core/network/workspacemanager.cpp
+++ b/src/core/network/workspacemanager.cpp
@@ -93,7 +93,7 @@ private:
 };
 
 struct ErrorHandle {
-    ErrorHandle(const InviwoSetupInfo& info, const std::string& filename)
+    ErrorHandle(const InviwoSetupInfo& info, std::string_view filename)
         : info_(info), filename_(filename) {}
 
     ~ErrorHandle() {
@@ -142,7 +142,7 @@ void WorkspaceManager::clear() {
     clears_.invoke();
 }
 
-void WorkspaceManager::save(std::ostream& stream, const std::string& refPath,
+void WorkspaceManager::save(std::ostream& stream, std::string_view refPath,
                             const ExceptionHandler& exceptionHandler, WorkspaceSaveMode mode) {
     Serializer serializer(refPath);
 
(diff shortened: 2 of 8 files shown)


ResearchDaniel:
How does loading of multiple workspaces work?

src/core/network/workspacemanager.cpp
```diff
@@ -169,22 +169,22 @@ void WorkspaceManager::load(std::istream& stream, const std::string& refPath,
     deserializers_.invoke(deserializer, exceptionHandler);
 }
 
-void WorkspaceManager::save(const std::string& path, const ExceptionHandler& exceptionHandler,
+void WorkspaceManager::save(std::string_view path, const ExceptionHandler& exceptionHandler,
                             WorkspaceSaveMode mode) {
-    auto ostream = filesystem::ofstream(path);
+    auto ostream = filesystem::ofstream(std::string(path));
```
martinfalk, line 174: maybe we should provide an override for string_view for both ofstream and ifstream

src/core/network/workspaceutils.cpp
```diff
@@ -36,26 +36,34 @@
...
-void updateWorkspaces(InviwoApplication* app) {
-    updateWorkspaces(app, filesystem::getPath(PathType::Workspaces));
-
-    for (const auto& m : app->getModules()) {
-        updateWorkspaces(app, m->getPath(ModulePath::Workspaces));
-    }
-}
+void forEachWorkspaceInDirRecusive(std::string_view path,
+                                   std::function<void(std::string_view)> callback) {
+    std::function<void(const std::string&)> traverse = [&](const std::string& path) {
```
martinfalk: why not using `filesystem::getDirectoryContentsRecursively()` here?

src/qt/editor/toolsmenu.cpp
```diff
@@ -181,6 +183,78 @@ ToolsMenu::ToolsMenu(InviwoMainWindow *win) : QMenu(tr("&Tools"), win) {
     auto sourceMenu = addMenu("Create &Sources");
     addInviwoMetaAction(sourceMenu);
 #endif
-}
+
+    auto workspaceMenu = addMenu("Workspaces");
+
+    auto loadExampleWorkspaces = workspaceMenu->addAction("Load Example Workspaces");
```
martinfalk, line 190: could add a tooltip to those actions explaining the difference between load (dryrun::yes) and update (dryrun::no)

src/qt/editor/toolsmenu.cpp
```diff
@@ -181,6 +183,78 @@ ToolsMenu::ToolsMenu(InviwoMainWindow *win) : QMenu(tr("&Tools"), win) {
...
+    connect(updateRegressionWorkspaces, &QAction::triggered, [win]() {
+        try {
+            util::updateRegressionWorkspaces(win->getInviwoApplication(), util::DryRun::No);
+        } catch (const Exception& e) {
+            util::log(e.getContext(), e.getMessage(), LogLevel::Error);
+        }
+    });
+
+    auto loadWorkspaces = workspaceMenu->addAction("Load Workspaces In Folder...");
+    connect(loadWorkspaces, &QAction::triggered, [win]() {
```
martinfalk, line 241: is this also recursive?

martinfalk (changes requested):

martinfalk:
> How does loading of multiple workspaces work?

also, what happens if loading a workspace is not successful?

petersteneteg:
Load just opens them one after an other... update opens and saves them again useful for updating the workspace versions

src/core/network/workspacemanager.cpp
petersteneteg: problem it that ifstream needs a string or ah fs::path. hence Im sticking with string until we can switch to fs::path

src/core/network/workspaceutils.cpp
petersteneteg: guess so... the function is old I just moved it

src/qt/editor/toolsmenu.cpp
petersteneteg: yes

src/qt/editor/toolsmenu.cpp
petersteneteg: mmm

include/inviwo/core/network/workspaceutils.h
```diff
@@ -31,19 +31,25 @@
...
 
-void IVW_CORE_API updateWorkspaces(InviwoApplication* app);
+IVW_CORE_API void forEachWorkspaceInDirRecusive(std::string_view path,
+                                                std::function<void(std::string_view)> callback);
 
-void IVW_CORE_API updateRegressionWorkspaces(InviwoApplication* app);
+enum class DryRun { Yes, No };
 
-void IVW_CORE_API updateWorkspaces(InviwoApplication* app, const std::string& path);
+IVW_CORE_API void updateWorkspaces(InviwoApplication* app, std::string_view path, DryRun dryRun);
```
ResearchDaniel, line 48: Explain what DryRun means in this context.
I guess it just opens the workspaces, updates but does not save them?

ResearchDaniel: Probably also want to mention that update refers to the workspace version.
I know that this is more or less updating existing code but we could mention what happens if there are errors while loading a workspace. Is it still updated and saved?

ResearchDaniel (dismissed):
It might be clearer with a slightly different menu structure and naming (order according to usage frequency):
**Workspace version** 
- Update Examples
- Update Regressions
- Update Folder
- Load Examples
- Load Regressions
- Load Folder

martinfalk (dismissed):

martinfalk (approved):

PR #1093, opened 2020-12-18
Feature/constexpr trim

martinfalk:
making `util::trim(std::string_view)` constexpr

### DIFF
--- a/include/inviwo/core/util/stringconversion.h
+++ b/include/inviwo/core/util/stringconversion.h
@@ -249,7 +249,7 @@ IVW_CORE_API std::vector<std::string_view> splitStringView(std::string_view str,
                                                            char delimiter = ' ');
 
 /**
- * \brief constexpr function for trimming \p str from beginning and end by removing white spaces
+ * \brief trims \p str from beginning and end by removing white spaces
  *
  * @param str   input string
  * @return trimmed stringview without leading/trailing white space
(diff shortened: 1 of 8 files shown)


petersteneteg (dismissed):

modules/dataframe/src/processors/dataframesource.cpp
```diff
@@ -34,7 +34,7 @@ namespace inviwo {
 // The Class Identifier has to be globally unique. Use a reverse DNS naming scheme
 const ProcessorInfo DataFrameSource::processorInfo_{
     "org.inviwo.DataFrameSource",                                // Class identifier
-    "Data Frame Source",                                         // Display name
+    "DataFrame Source",                                          // Display name
```
ResearchDaniel, line 37: I would argue that DataFrame is the technical term and Data frame would be the one to display

include/inviwo/core/util/stringconversion.h
```diff
@@ -248,8 +248,18 @@ IVW_CORE_API std::vector<std::string> splitString(std::string_view str, char del
 IVW_CORE_API std::vector<std::string_view> splitStringView(std::string_view str,
                                                            char delimiter = ' ');
 
-// trim from both ends
-IVW_CORE_API std::string_view trim(std::string_view s);
+/**
+ * \brief constexpr function for trimming \p str from beginning and end by removing white spaces
```
ResearchDaniel: I would leave out "constexpr function for " and just go with Trim ....

ResearchDaniel (dismissed):

modules/dataframe/src/processors/dataframesource.cpp
martinfalk: valid point, but since it refers here to the data structure...

petersteneteg (dismissed):

ResearchDaniel (dismissed):

petersteneteg (approved):

PR #1094, opened 2020-12-18
OpenGL: Raycasting surface normals

petersteneteg:
Base/BaseGL: EntryExitPoint normals

image
vs old
image


modules/base/src/algorithm/meshutils.cpp
```diff
@@ -760,6 +762,79 @@ std::shared_ptr<ColoredMesh> cameraFrustum(const Camera& camera, vec4 color,
...
+    auto calcNnormal = [&](const std::array<size_t, 4>& inds) {
+        const auto a = positions[inds[1]] - positions[inds[0]];
+        const auto b = positions[inds[3]] - positions[inds[0]];
+        return glm::normalize(glm::cross(a, b));
+    };
+
+    auto indexBuffer = mesh->addIndexBuffer(DrawType::Triangles, ConnectivityType::None);
+    auto& indices = indexBuffer->getDataContainer();
+
+    std::vector<vec3> positionVertices;
```
ResearchDaniel, line 799: Could reserve size for these

modules/base/include/modules/base/algorithm/meshutils.h
```diff
@@ -103,6 +103,57 @@ IVW_MODULE_BASE_API std::shared_ptr<ColoredMesh> cameraFrustum(
     const Camera& camera, vec4 color,
     std::shared_ptr<ColoredMesh> mesh = std::make_shared<ColoredMesh>());
 
+enum class IncludeNormals { Yes, No };
```
ResearchDaniel, line 106: No biggie, but to me it seems more like it is usel as ’CreateNormals’ or “AddNormals” than IncludeNormals

modules/base/include/modules/base/algorithm/meshutils.h
```diff
@@ -103,6 +103,57 @@ IVW_MODULE_BASE_API std::shared_ptr<ColoredMesh> cameraFrustum(
...
+ *
+ * @param origin position
+ * @param p1 basis vector 1
+ * @param p2 basis vector 2
+ * @param p3 basis vector 3
+ * @param color at origin
+ * @param c1 color at p1
+ * @param c2 color at p2
+ * @param c3 color at p3
+ * @param includeNormals Include a face normals
```
ResearchDaniel: Remove ”a”
Add face normals to the mesh

modules/base/include/modules/base/algorithm/meshutils.h
```diff
@@ -103,6 +103,57 @@ IVW_MODULE_BASE_API std::shared_ptr<ColoredMesh> cameraFrustum(
...
+ * @return parallelepiped mesh
+ */
+IVW_MODULE_BASE_API std::shared_ptr<Mesh> parallelepiped(glm::vec3 origin, glm::vec3 p1,
```
ResearchDaniel, line 151: Another PR: We might want to think about how auxiliary data such as color, texture coordinates, and scalar attributes can be added to create mesh functions to avoid having hard coded ones.

modules/base/include/modules/base/processors/cubeproxygeometryprocessor.h
```diff
@@ -53,6 +53,7 @@ namespace inviwo {
  *   * __Outport__ Output proxy geometry.
  *
  * ### Properties
+ *   * __Enable Normals__ Face normals are included in the mesh
```
ResearchDaniel, line 56: If we want to expose this I suggest
__Add Normals__ Add face normals to the mesh

modules/base/src/algorithm/cubeproxygeometry.cpp
```diff
@@ -102,47 +98,37 @@ std::shared_ptr<SimpleMesh> createCubeProxyGeometry(const std::shared_ptr<const
...
-
     colOrigin += c1 * clipOrigin.x + c2 * clipOrigin.y + c3 * clipOrigin.z;
     c1 *= clipExtent.x;
     c2 *= clipExtent.y;
     c3 *= clipExtent.z;
 
     // Create parallelepiped and set it to the outport. The outport will own the data.
-    auto geom = SimpleMeshCreator::parallelepiped(origin, e1, e2, e3, texOrigin, t1, t2, t3,
-                                                  colOrigin, c1, c2, c3);
+    auto geom = meshutil::parallelepiped(origin, e1, e2, e3, colOrigin, c1, c2, c3, normals);
```
ResearchDaniel, line 118: Removing texture coordinates? 
I thought 3d tex coords were udde later, but i guess the color is instead?

modules/basegl/include/modules/basegl/algorithm/entryexitpoints.h
```diff
@@ -47,6 +47,9 @@ class Volume;
 
 namespace algorithm {
 
+enum class IncludeNormals { Yes, No };
```
ResearchDaniel, line 50: Can we use one instead of two in different namespaces?

modules/basegl/include/modules/basegl/algorithm/entryexitpoints.h
```diff
@@ -116,31 +123,45 @@ class IVW_MODULE_BASEGL_API EntryExitPointsHelper {
...
+    static std::shared_ptr<Shader> createEntryExitShader(IncludeNormals includeNormals,
+                                                         ApplyTransformation applyTrafo);
+    static std::shared_ptr<Shader> createNearClipShader(IncludeNormals includeNormals);
 
     std::unique_ptr<Image> tmpEntry_;
     ImageGL* tmpEntryGL_ = nullptr;
+
+    IncludeNormals includeNormals_ = IncludeNormals::No;
+    ApplyTransformation applyTrafo_ = ApplyTransformation::No;
+    std::shared_ptr<Shader> entryExitShader_{nullptr};
```
ResearchDaniel: Sharing shaders means that all parameters need to be set before executing them. Probably better than having multiple ones, but perhaps it should be noted in the code in case someone makes changes. I think that it could be a quite hard bug to track down

modules/basegl/src/algorithm/entryexitpoints.cpp
```diff
@@ -197,33 +209,302 @@ void EntryExitPointsHelper::createCappedEntryExitPoints(ImageGL& entryPoints, Im
...
+Shader& EntryExitPointsHelper::getEntryExitShader(IncludeNormals includeNormals,
+                                                  ApplyTransformation applyTrafo) {
+
+    if (!entryExitShader_ || includeNormals != includeNormals_ || applyTrafo != applyTrafo_) {
+        entryExitShader_ = createEntryExitShader(includeNormals, applyTrafo);
+        includeNormals_ = includeNormals;
+        applyTrafo_ = applyTrafo;
+    }
+
+    entryExitCallback_ = entryExitShader_->onReloadScoped([this]() { onReloadCallback_.invoke(); });
```
ResearchDaniel: Every time? Move inside if above?

modules/base/include/modules/base/algorithm/meshutils.h
ResearchDaniel: After going through the whole code I see that it is used in a couple of different places where create/add might not fit as well...

modules/basegl/src/algorithm/entryexitpoints.cpp
```diff
@@ -197,33 +209,302 @@ void EntryExitPointsHelper::createCappedEntryExitPoints(ImageGL& entryPoints, Im
...
+constexpr std::string_view meshEntryExitVert = R"(
+uniform mat4 dataToClip = mat4(1);
+uniform mat4 meshDataToVolData = mat4(1);
+out vec4 color;
+void main() {
+    color = meshDataToVolData * in_Vertex;
+    gl_Position = dataToClip * in_Vertex;
+}
+)";
+constexpr std::string_view meshEntryExitFrag = R"(
```
ResearchDaniel: Equivalent to entryExitFrag.

modules/basegl/src/processors/entryexitpointsprocessor.cpp
```diff
@@ -58,22 +58,40 @@ EntryExitPoints::EntryExitPoints()
     addProperty(camera_);
     addProperty(trackball_);
 
-    for (auto& shader : entryExitHelper_.getShaders()) {
-        shader.get().onReload([this]() { invalidate(InvalidationLevel::InvalidResources); });
-    }
+    callback_ =
```
ResearchDaniel: Would name this onReloadCallback_

modules/basegl/src/processors/entryexitpointsprocessor.cpp
```diff
@@ -58,22 +58,40 @@ EntryExitPoints::EntryExitPoints()
...
 void EntryExitPoints::process() {
-    if (!entryImg_ || !entryImg_->isValid()) {
-        entryImg_ = entryPort_.getEditableData()->getEditableRepresentation<ImageGL>();
+    auto entry = entryPort_.getEditableData();
+
+    if (inport_.getData()->hasBuffer(BufferType::NormalAttrib) &&
+        entry->getNumberOfColorLayers() != 2) {
+
+        entry = std::make_shared<Image>(entry->getDimensions(), DataVec4UInt16::get());
+        entry->addColorLayer(
```
ResearchDaniel, line 75: Add comment that this layer is for the normals. 8-bit integer should be enough when they are axis-aligned. Precisison is perhaps not too important since they are only used once and arbitrary clipping is not too common.

modules/basegl/src/processors/geometryentryexitpoints.cpp
```diff
@@ -59,15 +59,15 @@ GeometryEntryExitPoints::GeometryEntryExitPoints()
     addProperty(camera_);
     addProperty(trackball_);
 
-    for (auto& shader : entryExitHelper_.getShaders()) {
-        shader.get().onReload([this]() { invalidate(InvalidationLevel::InvalidResources); });
-    }
+    callback_ =
```
ResearchDaniel: onReloadCallback_

modules/opengl/glsl/raycasting.frag
```diff
@@ -76,7 +79,7 @@ uniform int channel;
 #endif
 
 
-vec4 rayTraversal(vec3 entryPoint, vec3 exitPoint, vec2 texCoords, float backgroundDepth) {
+vec4 rayTraversal(vec3 entryPoint, vec3 exitPoint, vec2 texCoords, float backgroundDepth, vec3 startNormal) {
```
ResearchDaniel: startNormal -> firstNormal

modules/opengl/glsl/raycasting.frag
```diff
@@ -162,6 +172,8 @@ vec4 rayTraversal(vec3 entryPoint, vec3 exitPoint, vec2 texCoords, float backgro
 #endif // ISOSURFACE_ENABLED
             t += tIncr;
         }
+        first = false;
+        //t = tEnd;
```
ResearchDaniel: remove

modules/opengl/glsl/raycasting.frag
```diff
@@ -135,17 +140,22 @@ vec4 rayTraversal(vec3 entryPoint, vec3 exitPoint, vec2 texCoords, float backgro
 #if defined(INCLUDE_DVR)
         color = APPLY_CHANNEL_CLASSIFICATION(transferFunction, voxel, channel);
         if (color.a > 0) {
-            vec3 gradient =
-                COMPUTE_GRADIENT_FOR_CHANNEL(voxel, volume, volumeParameters, samplePos, channel);
-            gradient = normalize(gradient);
-
+            
+            vec3 gradient;
+            if (first && useNormals) {
```
ResearchDaniel, line 145: Future: By refactoring what happens inside the loop it would be easier to move this first time-only code outside of the loop

modules/opengl/glsl/raycasting.frag
```diff
@@ -201,8 +213,10 @@ void main() {
         discard;
     }
 #endif // BACKGROUND_AVAILABLE
+
+    vec3 normal = useNormals?texture(entryNormal, texCoords).xyz : vec3(0,0,0);
```
ResearchDaniel: Missing space
`useNormals ? texture(entryNormal, texCoords).xyz : vec3(0,0,0);`

ResearchDaniel (changes requested):

modules/base/include/modules/base/algorithm/meshutils.h
petersteneteg: this version has to have unique vertex for each face of the cube sone the normals are not the same at the corners... hence just adding a buffer to the old one would not work.

modules/base/src/algorithm/cubeproxygeometry.cpp
petersteneteg: Nope they are not used

modules/basegl/include/modules/basegl/algorithm/entryexitpoints.h
petersteneteg: I wanted to avoid having to create that dependency.

modules/basegl/src/algorithm/entryexitpoints.cpp
petersteneteg: humm had to be like that in an older version...

modules/basegl/src/processors/entryexitpointsprocessor.cpp
petersteneteg: Since the normals are just just for shading usually I don't think it should matter very much, And in the 4k stereo rendering, we might not want to waste full image textures...

modules/opengl/glsl/raycasting.frag
petersteneteg: yes

ResearchDaniel (approved):

PR #1098, opened 2021-01-19
Various: GCC compile fix

petersteneteg:


ResearchDaniel (approved):
We could also remove the resource manager

petersteneteg:
the old one was removed, this is a new one rickard did, and it does work. can be quite useful to avoid having to reload data in some cases.

PR #1099, opened 2021-01-19
Tools: regression fail only on loglevel error not just the Error: string

petersteneteg:


ResearchDaniel (approved):

petersteneteg:
@ResearchDaniel That was fast!

PR #1102, opened 2021-01-21
QtWidget: TF updates

petersteneteg:
Added colorbrewer presets, and simple range generation. Also scaling tf point interaction.

image

image
image

https://user-images.githubusercontent.com/3638222/105418496-8475b880-5c3d-11eb-87ae-5ea58e37c651.mov

### DIFF
--- a/modules/qtwidgets/src/tf/tfeditor.cpp
+++ b/modules/qtwidgets/src/tf/tfeditor.cpp
@@ -80,6 +80,13 @@ private:
     const TFPrimitive& p_;
 };
 
+template <typename Operator, typename Proj = util::identity>
+constexpr auto derefOperator(Operator&& op, Proj&& proj = {}) {
+    return [o = std::forward<Operator>(op), p = std::forward<Proj>(proj)](auto* a, auto* b) {
+        return std::invoke(o, std::invoke(p, *a), std::invoke(p, *b));
+    };
+}
+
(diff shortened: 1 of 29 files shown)


modules/qtwidgets/src/tf/tfeditor.cpp
```diff
@@ -494,39 +589,107 @@ void TFEditor::contextMenuEvent(QGraphicsSceneContextMenuEvent* e) {
...
     }
 
     auto transformMenu = menu.addMenu("Trans&form");
     menu.addSeparator();
 
     {
-        auto flip = transformMenu->addAction("&Horizontal Flip");
-        auto interpolate = transformMenu->addAction("&Interpolate Alpha");
-        auto equalize = transformMenu->addAction("&Equalize Alpha");
+        auto distributeAlphaEvenly = transformMenu->addAction("&Distribute Alpha Evenly");
```
ResearchDaniel, line 609: Would be nice to have buttons for the different available interactions

ResearchDaniel (approved):

modules/qtwidgets/src/tf/tfeditor.cpp
petersteneteg: that could be added I guess, but there are 10 different transforms, where would you put them all?

ResearchDaniel: Could use a dropdown button

PR #1103, opened 2021-01-27
Feature/region shrink

petersteneteg:


modules/basegl/include/modules/basegl/processors/volumeprocessing/volumeregionshrink.h
```diff
@@ -0,0 +1,79 @@
...
+#include <inviwo/core/ports/volumeport.h>
+#include <modules/opengl/shader/shader.h>
+#include <modules/opengl/buffer/framebufferobject.h>
+#include <modules/opengl/shader/shaderresource.h>
+
+namespace inviwo {
+
+/** \docpage{org.inviwo.VolumeRegionShrink, Volume Region Shrink}
+ * ![](org.inviwo.VolumeRegionShrink.png?classIdentifier=org.inviwo.VolumeRegionShrink)
+ * Shink all regions in the volume
```
martinfalk: Functionality quite unclear. Care to elaborate a bit more? Like preconditions (assuming index volume)

modules/basegl/include/modules/basegl/processors/volumeprocessing/volumeregionshrink.h
```diff
@@ -0,0 +1,79 @@
...
+private:
+    VolumeInport inport_;
+    VolumeOutport outport_;
+    IntProperty iterations_;
+
+    std::shared_ptr<StringShaderResource> fragShader_;
+    Shader shader_;
+
+    std::array<std::shared_ptr<Volume>, 2> out_;
+    std::array<FrameBufferObject, 2> fbo_;
```
martinfalk: necessary to use 2 FBO instead of a single one with two attachments and then switch between attachments (`glDrawBuffer(s)`)?

modules/basegl/src/processors/volumeprocessing/volumeregionshrink.cpp
```diff
@@ -0,0 +1,172 @@
...
+        out_[1]->getDimensions() != volume->getDimensions()) {
+        out_[1] = std::shared_ptr<Volume>(volume->clone());
+    }
+
+    out_[1]->setModelMatrix(volume->getModelMatrix());
+    out_[1]->setWorldMatrix(volume->getWorldMatrix());
+    out_[1]->copyMetaDataFrom(*volume);
+    out_[1]->dataMap_ = volume->dataMap_;
+
+    fbo_[1].activate();
```
martinfalk: better to switch between color attachments using glDrawBuffer

modules/basegl/src/processors/volumeprocessing/volumeregionshrink.cpp
```diff
@@ -0,0 +1,172 @@
...
+    fbo_[1].attachColorTexture(outGL1->getTexture().get(), 0);
+    out_[1]->invalidateHistogram();
+
+    size_t src = 1;
+    size_t dst = 0;
+    for (int i = 2; i < iterations_; ++i) {
+        std::swap(src, dst);
+        TextureUnitContainer cont;
+        utilgl::bindAndSetUniforms(shader_, cont, *out_[src], "volume");
+        fbo_[dst].activate();
```
martinfalk: expensive!

modules/opengl/src/image/imagecompositor.cpp
```diff
@@ -30,88 +30,51 @@
...
 
-void ImageCompositor::composite(const Image& source, Image& target, ImageType type) {
-    copyTextures(target);
-
-    utilgl::activateTarget(target, type);
-    shader_.activate();
-    utilgl::GlBoolState depthTest(GL_DEPTH_TEST, false);
+void ImageCompositor::composite(const Image& source0, const Image& source1, Image& destination,
+                                ImageType type) {
+    utilgl::GlBoolState depthTest(GL_DEPTH_TEST, true);
```
martinfalk, line 48: could add assertion/runtime check that destination is different from source0 and source1

modules/opengl/src/image/imagecompositor.cpp
```diff
@@ -30,88 +30,51 @@
...
-void ImageCompositor::composite(ImageInport& source, ImageOutport& target, ImageType type) {
-    if (source.isReady() && target.hasData()) {
-        copyTextures(*target.getData());
-
-        utilgl::activateTarget(target, type);
-        shader_.activate();
-        utilgl::GlBoolState depthTest(GL_DEPTH_TEST, false);
+void ImageCompositor::composite(const ImageInport& source0, const ImageInport& source1,
+                                ImageOutport& destination, ImageType type) {
+    if (source0.isReady() && source1.isReady()) {
```
martinfalk, line 65: check whether ports are different (write to same texture as reading is undefined behavior)

martinfalk (dismissed):

modules/basegl/include/modules/basegl/processors/volumeprocessing/volumeregionshrink.h
petersteneteg: no

modules/basegl/include/modules/basegl/processors/volumeprocessing/volumeregionshrink.h
```diff
@@ -0,0 +1,81 @@
...
+#include <inviwo/core/ports/volumeport.h>
+#include <modules/opengl/shader/shader.h>
+#include <modules/opengl/buffer/framebufferobject.h>
+#include <modules/opengl/shader/shaderresource.h>
+
+namespace inviwo {
+
+/** \docpage{org.inviwo.VolumeRegionShrink, Volume Region Shrink}
+ * ![](org.inviwo.VolumeRegionShrink.png?classIdentifier=org.inviwo.VolumeRegionShrink)
+ * Shrinks regions of identical values. The processor will assign 0 to each all boarder voxels in
```
ResearchDaniel: Some typos. I think it should say
Shrinks regions of identical values. The processor will assign 0 to each border voxel in
 each iteration. A voxel is considered on the border if the value of any of the 8 closest neighbors is
  different. The procedure is repeated number of __iterations__ times.

ResearchDaniel: Also, is this standard procedure? Might want to add a reference to the algorithm.

modules/basegl/src/processors/volumeprocessing/volumeregionshrink.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+void main() {
+    vec4 value = getVoxel(volume, volumeParameters, texCoord_.xyz);
+
+    bool edge = false;
+    for(int z = -1; z < 1; z++) {
+        for(int y = -1; y < 1; y++) {
+            for(int x = -1; x < 1; x++){
+
+                vec4 neighbor = getVoxel(volume, volumeParameters, texCoord_.xyz + vec3(x,y,z)*volumeParameters.reciprocalDimensions);
```
ResearchDaniel: Need to use the index to texture matrix in case of non-orthogonal volumes?

modules/basegl/src/processors/volumeprocessing/volumeregionshrink.cpp
```diff
@@ -0,0 +1,172 @@
...
+                if (neighbor != value) {
+                    edge = true;
```
ResearchDaniel: Break loops

modules/basegl/src/processors/volumeprocessing/volumeregionshrink.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+    // Iteration 1
+    {
+        glDrawBuffer(GL_COLOR_ATTACHMENT0);
+        TextureUnitContainer cont;
+        utilgl::bindAndSetUniforms(shader_, cont, *volume, "volume");
+        utilgl::multiDrawImagePlaneRect(static_cast<int>(dim.z));
+    }
+    if (iterations_ == 1) {
+        outport_.setData(out_[0]);
```
ResearchDaniel, line 138: Deactivate FBO?

ResearchDaniel: Also remove the second temporary volume in case it exist?

modules/opengl/include/modules/opengl/image/imagecompositor.h
```diff
@@ -46,21 +44,13 @@ namespace inviwo {
 class IVW_MODULE_OPENGL_API ImageCompositor {
 public:
     ImageCompositor(std::string programFileName = "composite.frag");
-    virtual ~ImageCompositor() = default;
+    ~ImageCompositor() = default;
 
-    void composite(const Image& source, Image& target, ImageType type);
-    void composite(ImageInport& source, ImageOutport& target, ImageType type);
+    void composite(const Image& source0, const Image& source1, Image& destination, ImageType type);
```
ResearchDaniel, line 52: Source0 and Source1?

ResearchDaniel (changes requested):

modules/basegl/include/modules/basegl/processors/volumeprocessing/volumeregionshrink.h
petersteneteg: might be, but I just made it up for my current needs...

modules/basegl/src/processors/volumeprocessing/volumeregionshrink.cpp
petersteneteg: since everything is in texture space here, I think it should be fine.

modules/basegl/src/processors/volumeprocessing/volumeregionshrink.cpp
petersteneteg: Don't think that make a difference in glsl.

modules/basegl/src/processors/volumeprocessing/volumeregionshrink.cpp
petersteneteg: I want to avoid having to recreate it if it needs to recompute. guess that is a space / time trade off though.

modules/opengl/include/modules/opengl/image/imagecompositor.h
petersteneteg: the two sources gets composed using depth in to the destination. what that you question?

ResearchDaniel: Ah, so before target was also source1?

petersteneteg: yes, which is undefined behaviour, so we did an internal copy. But in the call site we frist copied one source to the target...

modules/basegl/src/processors/volumeprocessing/volumeregionshrink.cpp
ResearchDaniel: Recreation would only happen if the property changes so it should not be too often. But yeah, a trade off

petersteneteg: of if the input changes, which can happen often in some cases...

ResearchDaniel: It would only deallocate once if iterations_ == 1. Then reallocate when iterations_ changes

petersteneteg: ahh now I understand what you mean, i missed the fact that your comment was in the iterations ==1 block, yes that would make sense.

ResearchDaniel (approved):

PR #1105, opened 2021-01-28
Inviwo Python-scripted execution

xeTaiz:
@petersteneteg as discussed here's what I've implemented for the scripted execution.
As I have tested this with some of my research stuff, there is still a bunch of it in there.

Here's an overview:
```
apps/inviwopyapp/ivw_interface.py   # This class is used to hold an InviwoApplication, modify workspaces easily to get data from Python using the Injection processors. Also a bunch of other handy helpers here
apps/inviwopyapp/*py    # Scripts I used to test. Many of them probably have PyTorch and other python dependencies.
modules/python3/processors  # The Injection processors that expose a setArray method used by the InviwoInterface
data/workspaces   # Has some workspaces. Note that the InviwoInterface can replace processors with Injectors and some of the workspaces in here are before, some after injection.
```

If you're trying this on windows I would not recommend to try install PyTorch etc. If there's any questions let me know


petersteneteg:
some of this was merge in #1106 the rest is in feature/python-scripting

PR #1110, opened 2021-02-09
Fix to issue #1109 regarding Webbrowser

### LINKED ISSUE #1109: Webbrowser crash on loading of new/same workspace
Crash in Webbrowser when reloading of workspace.

eriksunden:
Fixes issue #1109

### DIFF
--- a/modules/webbrowser/src/processors/webbrowserprocessor.cpp
+++ b/modules/webbrowser/src/processors/webbrowserprocessor.cpp
@@ -151,7 +151,10 @@ std::string WebBrowserProcessor::getSource() {
 }
 
 WebBrowserProcessor::~WebBrowserProcessor() {
-    static_cast<WebBrowserClient*>(browser_->GetHost()->GetClient().get())->removeLoadHandler(this);
+    WebBrowserClient* webbrowser =
+        static_cast<WebBrowserClient*>(browser_->GetHost()->GetClient().get());
+    webbrowser->removeLoadHandler(this);
+    webbrowser->OnBeforeClose(browser_);
     // Force close browser
     browser_->GetHost()->CloseBrowser(true);
 }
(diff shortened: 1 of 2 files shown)


modules/webbrowser/src/processors/webbrowserprocessor.cpp
```diff
@@ -151,7 +151,10 @@ std::string WebBrowserProcessor::getSource() {
 }
 
 WebBrowserProcessor::~WebBrowserProcessor() {
-    static_cast<WebBrowserClient*>(browser_->GetHost()->GetClient().get())->removeLoadHandler(this);
+    WebBrowserClient* webbrowser =
+        static_cast<WebBrowserClient*>(browser_->GetHost()->GetClient().get());
+    webbrowser->removeLoadHandler(this);
+    webbrowser->OnBeforeClose(browser_);
```
ResearchDaniel, line 157: I don't think that this is a solution to the problem. From the docs of close_browser:
"Results in a call to cef_life_span_handler_t::do_close() if the event handler allows the close or if |force_close| is true (1)"

ResearchDaniel (changes requested):

modules/webbrowser/src/processors/webbrowserprocessor.cpp
ResearchDaniel: If the code above resolves the issue we might want to look into handling browser destruction in `WebBrowserClient::OnBeforeClose`

modules/webbrowser/src/webbrowserclient.cpp
```diff
@@ -141,13 +141,13 @@ bool WebBrowserClient::DoClose(CefRefPtr<CefBrowser> browser) {
 void WebBrowserClient::OnBeforeClose(CefRefPtr<CefBrowser> browser) {
     CEF_REQUIRE_UI_THREAD();
     auto bdIt = browserParents_.find(browser->GetIdentifier());
-    if (bdIt != browserParents_.end()) {
+    if (bdIt != browserParents_.end() && messageRouter_) {
```
ResearchDaniel, line 144: This is caused by a bug you introduced with your other fix :)

ResearchDaniel:
See PR #1115 instead

PR #1113, opened 2021-02-12
Feature/propwidget

petersteneteg:
Disable and enable layout in the propertylists when doing many changes. 10X perf in some cases.
PropertyOwner got a clear() and empty() method.


ResearchDaniel (dismissed):
What’s the reason for changing the depth format to 32-bit?

petersteneteg:
We use a float32 on the cpu side, so to match that one. And I ran into some precision issues..

modules/python3/bindings/include/inviwopy/vectoridentifierwrapper.h
```diff
@@ -59,11 +59,21 @@ class VectorIdentifierWrapper {
...
+        if (pos >= 0) {
+            const auto ind = static_cast<size_t>(pos);
+            if (ind < vector_.size()) {
+                return *vector_[ind];
+            } else {
+                throw pybind11::index_error();
+            }
         } else {
-            throw pybind11::index_error();
+            const auto ind = static_cast<ptrdiff_t>(vector_.size()) + pos;
```
martinfalk, line 71: that's neat

modules/qtwidgets/src/propertylistwidget.cpp
```diff
@@ -195,6 +205,16 @@ bool PropertyListWidget::event(QEvent* e) {
     }
 }
 
+void PropertyListWidget::onProcessorNetworkEvaluationBegin() { 
+    setUpdatesEnabled(false); 
+    listLayout_->setEnabled(false);
+}
```
martinfalk, line 211: this should take care of most of the issues we encountered so for, shouldn't it? Nice.

martinfalk (dismissed):

martinfalk (approved):

petersteneteg:
@martinfalk I think this merge resulted in a quite substantial perf regression, My guess is that it is related to the move to 32 bit depth, does that make sense to you?

martinfalk:
Not sure. That would need performance testing in particular with different hardware. In some older posts I found that AMD GPUs did not support 24bits at all...

See also https://www.khronos.org/opengl/wiki/Common_Mistakes#Depth_Buffer_Precision

petersteneteg:
this for the ray caster with animation regression test:
image
the drop is when the bitdepth changed. see the last set of runs on http://jenkins.inviwo.org:8080/job/tests/job/hist/Regression/

But maybe we should consider other hardware too?

PR #1114, opened 2021-02-12
OpenGL: BufferGL optimization

martinfalk:
* to avoid unnecessary reallocations, the internal `BufferObject` of a `BufferGL` is only updated when requesting more memory


petersteneteg (dismissed):

petersteneteg:
@ResearchDaniel does this affect OpenCL?

ResearchDaniel (approved):
Checked and it should work with OpenCL sharing as well.

PR #1115, opened 2021-02-23
Fix/web animation

ResearchDaniel:
- Closes issue #1109 (Processor was destroyed before OnBeforeClose(CefRefPtr<CefBrowser> browser) was called)
- Fixes issue when a property changed without the CefFrame being set.
- Animation: Fixes bugs when PropertyTrack has no owner and where AnimationState was set before evaluating animation

### DIFF
--- a/modules/webbrowser/include/modules/webbrowser/webbrowserclient.h
+++ b/modules/webbrowser/include/modules/webbrowser/webbrowserclient.h
@@ -70,8 +70,6 @@ class IVW_MODULE_WEBBROWSER_API WebBrowserClient : public CefClient,
                                                    public CefDisplayHandler,
                                                    public CefResourceRequestHandler {
 public:
-    using BrowserParentHandle = std::shared_ptr<Processor*>;
-
     WebBrowserClient(ModuleManager& moduleManager, const PropertyWidgetCEFFactory* widgetFactory);
 
     virtual CefRefPtr<CefLoadHandler> GetLoadHandler() override { return this; }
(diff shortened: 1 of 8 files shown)


modules/webbrowser/src/webbrowserclient.cpp
```diff
@@ -82,12 +82,15 @@ WebBrowserClient::WebBrowserClient(ModuleManager& moduleManager,
     });
 }
 
-void WebBrowserClient::setBrowserParent(CefRefPtr<CefBrowser> browser, Processor* parent) {
+WebBrowserClient::BrowserParentHandle WebBrowserClient::setBrowserParent(
+    CefRefPtr<CefBrowser> browser, Processor* parent) {
     CEF_REQUIRE_UI_THREAD();
-    BrowserData bd{parent, new ProcessorCefSynchronizer(parent)};
+    auto processor = std::make_shared<Processor*>(parent);
```
petersteneteg: What is going on here? How you are allowed to take ownership if this processor? Processors are owned by the processor network. Unless this processor it not in the network this is UB

petersteneteg (changes requested):

modules/webbrowser/src/webbrowserclient.cpp
ResearchDaniel: Perhaps a bit unclear, but it is only the pointer to the processor that is kept.
I first thought of calling it dummy or something...
Any suggestion on how to make it more expressive?

An alternative that I thought of was to observe the network, but I thought this was more self-contained

petersteneteg: I still don't understand, the processor is not in the network? And in that case What do you need it for?

petersteneteg: Humm looking again, so it is a shared_ptr to a __pointer__ to a processor. Now I'm even more confused, why would you have that?

ResearchDaniel: I need a reference to the processor in order for the browser to access it. It has turned out that the Processor might be deleted before the reference to it is removed (OnBeforeBrowserClose). So I solved it by keeping a weak_ptr around. 
I considered these alternatives:
- Processor explicitly removes itself as "parent" upon destruction
- Let the WebbrowserClient observe the network and remove the parent if the processor was removed
- Use a handle that needs to be kept alive as long as the processor is alive

I chose the third option...

petersteneteg: Ahh so you lock that one before you invalidate? Is that the only place you need to do that? Should that not
Be the same for all property changes?

petersteneteg (dismissed):
Looks good to me, some more comment about the lifetime issue, and point out that you have a shared ptr of a pointer not a processor...

modules/webbrowser/src/webbrowserclient.cpp
ResearchDaniel: Hmm, yes I guess you could get unlucky... 
Perhaps explicitly removing it upon processor destruction is a better option as it would also remove the associated property change callbacks

modules/webbrowser/include/modules/webbrowser/webbrowserclient.h
```diff
@@ -211,7 +220,7 @@ class IVW_MODULE_WEBBROWSER_API WebBrowserClient : public CefClient,
 
 protected:
     struct BrowserData {
-        Processor* processor = nullptr;
+        Processor* processor;
```
petersteneteg, line 223: Is this only use in the main thread or how is access synchronized?

ResearchDaniel: Yes, should only be on the main thread

petersteneteg (approved):

PR #1116, opened 2021-02-25
Core: MSVC fix for debugbreak crash

martinfalk:
* added check if debugger is present for Visual Studio builds

**Reason:** both `__debugbreak` and `SIGTRAP` terminate the program if no debugger is present. 

So dragging a processor into the network which causes an OpenGL error will kill Inviwo when not using a debugger and the "Break on" setting is set to anything else than "off".

### DIFF
--- a/src/core/util/assertion.cpp
+++ b/src/core/util/assertion.cpp
@@ -34,6 +34,8 @@
 
 #ifndef WIN32
 #include <signal.h>
+#else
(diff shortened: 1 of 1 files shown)


martinfalk:
Potentially, we could add also a check for remote debuggers (VS only)

ResearchDaniel (approved):

petersteneteg (approved):

PR #1120, opened 2021-03-02
Qt: Support for Qt6

petersteneteg:
cmake support for qt6 and some warning fixes

### DIFF
--- a/apps/inviwo/CMakeLists.txt
+++ b/apps/inviwo/CMakeLists.txt
@@ -1,8 +1,8 @@
 # Inviwo Qt Application
 project(inviwo)
 
-find_package(Qt5Core QUIET REQUIRED)
-find_package(Qt5Widgets QUIET REQUIRED)
+find_package(QT NAMES Qt6 Qt5 COMPONENTS Core Widgets REQUIRED)
+find_package(Qt${QT_VERSION_MAJOR} COMPONENTS Core Widgets REQUIRED)
 
 # Add MOC files
 set(MOC_FILES  ${IVW_APPLICATION_DIR}/inviwo/inviwosplashscreen.h)
@@ -19,7 +19,7 @@ set(SOURCE_FILES
 ivw_group("Source Files" ${SOURCE_FILES})
 
 # Add moc files
-qt5_wrap_cpp(MOC_FILES ${MOC_FILES})
+qt_wrap_cpp(MOC_FILES ${MOC_FILES})
 source_group("MOC Files" FILES ${MOC_FILES})
 
 # Define libraries that should be linked
@@ -47,8 +47,11 @@ target_link_libraries(inviwo PUBLIC
     inviwo::core
     inviwo::qtapplicationbase
     inviwo::qteditor
-    Qt5::Core
-    Qt5::Widgets
+    Qt${QT_VERSION_MAJOR}::Core
+    Qt${QT_VERSION_MAJOR}::Widgets
+)
+target_compile_definitions(inviwo PUBLIC 
+    "QT_DEPRECATED_WARNINGS_SINCE=QT_VERSION_CHECK(5, 15, 0)"
 )
 set_target_properties(inviwo PROPERTIES OUTPUT_NAME "Inviwo")
 
@@ -74,6 +77,7 @@ if(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
         MACOSX_BUNDLE_SHORT_VERSION_STRING "${IVW_VERSION}"
         MACOSX_BUNDLE_BUNDLE_VERSION "${IVW_VERSION}"
         MACOSX_BUNDLE_COPYRIGHT "C) 2012-${YEAR} Inviwo Foundation"
+        XCODE_ATTRIBUTE_PRODUCT_BUNDLE_IDENTIFIER "org.inviwo.Inviwo"
     )
     #Following http://www.cmake.org/Wiki/BundleUtilitiesExample
     set(plugin_dest_dir Inviwo.app/Contents/MacOS)
@@ -118,8 +122,8 @@ if(WIN32)
         inviwo::core 
         inviwo::qtapplicationbase
         inviwo::qteditor
-        Qt5::Core 
-        Qt5::Widgets
+        Qt${QT_VERSION_MAJOR}::Core 
+        Qt${QT_VERSION_MAJOR}::Widgets
     )
     set_target_properties(inviwo-cli PROPERTIES OUTPUT_NAME "Inviwo-cli")
     ivw_configure_application_module_dependencies(inviwo-cli ${enabled_modules})
--- a/apps/inviwo/CMakeLists.txt
+++ b/apps/inviwo/CMakeLists.txt
@@ -1,14 +1,15 @@
 # Inviwo Qt Application
 project(inviwo)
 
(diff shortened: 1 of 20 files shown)


apps/inviwo/CMakeLists.txt
```diff
@@ -47,8 +47,11 @@ target_link_libraries(inviwo PUBLIC
     inviwo::core
     inviwo::qtapplicationbase
     inviwo::qteditor
-    Qt5::Core
-    Qt5::Widgets
+    Qt${QT_VERSION_MAJOR}::Core
+    Qt${QT_VERSION_MAJOR}::Widgets
+)
+target_compile_definitions(inviwo PUBLIC 
+    "QT_DEPRECATED_WARNINGS_SINCE=QT_VERSION_CHECK(5, 15, 0)"
```
ResearchDaniel, line 59: Could add a comment explaining at which point this should be bumped or removed

modules/openglqt/include/modules/openglqt/canvasqt.h
```diff
@@ -83,8 +88,12 @@ class CanvasQt : public CanvasQOpenGLWidget {
     bool mapPanTriggered(QPanGesture*);
     bool mapPinchTriggered(QPinchGesture* e);
 
+#if (QT_VERSION >= QT_VERSION_CHECK(6, 0, 0))
     //! Links QTouchDevice to inviwo::TouchDevice
+    std::map<const QPointingDevice*, TouchDevice> devices_;
```
ResearchDaniel, line 93: Should be named the same

modules/openglqt/src/canvasqt.cpp
```diff
@@ -315,6 +315,90 @@ bool CanvasQt::mapKeyReleaseEvent(QKeyEvent* keyEvent) {
...
+                   prevTouchPoints_.end())) {
+        prevTouchPoints_ = touchPoints;
+        touch->accept();
+        return true;
+    }
+    // Store previous touch points
+    prevTouchPoints_ = touchPoints;
+
+    // Get the device generating the event (touch screen or touch pad)
+    auto deviceIt = devices_.find(touch->pointingDevice());
```
ResearchDaniel, line 372: It is only this part that differ? No need to duplicate the code above

src/qt/applicationbase/inviwoapplicationqt.cpp
```diff
@@ -329,12 +329,22 @@ bool InviwoApplicationQt::notify(QObject* receiver, QEvent* e) {
         }
         case QEvent::TouchEnd: {
             auto te = static_cast<QTouchEvent*>(e);
+#if (QT_VERSION >= QT_VERSION_CHECK(6, 0, 0))
+            if (util::all_of(te->points(), [](const QTouchEvent::TouchPoint& tp) {
+                    return tp.state() == QEventPoint::Released;
```
ResearchDaniel, line 334: Only this line that differ

ResearchDaniel (changes requested):
Nice, some duplicate code that can/should be removed

modules/openglqt/src/canvasqt.cpp
petersteneteg: there are more changes, the enums...

apps/inviwo/CMakeLists.txt
petersteneteg: I don't know any more then when we move to require qt6. And that is quite obvious from the names...

modules/openglqt/include/modules/openglqt/canvasqt.h
petersteneteg: I want a different name so I don't miss any instance of it

modules/openglqt/src/canvasqt.cpp
petersteneteg: some of the function names...

src/qt/applicationbase/inviwoapplicationqt.cpp
petersteneteg: te->points() vs te->touchPoints() and QEventPoint::Released vs Qt::TouchPointReleased have the ifdef on the whole block makes it much more readable

ResearchDaniel (dismissed):

modules/openglqt/CMakeLists.txt
```diff
@@ -58,6 +61,9 @@ target_link_libraries(inviwo-module-openglqt PRIVATE
     Qt${QT_VERSION_MAJOR}::Svg
     Qt${QT_VERSION_MAJOR}::OpenGLWidgets
 )
+if(QT_VERSION_MAJOR EQUAL 6)
+    target_link_libraries(inviwo-module-openglqt PRIVATE Qt6::OpenGLWidgets)
```
ResearchDaniel, line 65: I don't understand. Is this not equivalent to Qt${QT_VERSION_MAJOR}::OpenGLWidgets above?

modules/openglqt/CMakeLists.txt
```diff
@@ -58,6 +61,9 @@ target_link_libraries(inviwo-module-openglqt PRIVATE
     Qt${QT_VERSION_MAJOR}::Svg
     Qt${QT_VERSION_MAJOR}::OpenGLWidgets
```
ResearchDaniel: Does this work for Qt5 even though there is no find_package for it?

modules/openglqt/CMakeLists.txt
```diff
@@ -2,9 +2,12 @@
...
 ivw_module(OpenGLQt)
 
-find_package(QT NAMES Qt6 Qt5 COMPONENTS Svg OpenGL OpenGLWidgets REQUIRED)
-find_package(Qt${QT_VERSION_MAJOR} COMPONENTS Svg OpenGL OpenGLWidgets REQUIRED)
-
+find_package(QT NAMES Qt6 Qt5 COMPONENTS Svg OpenGL REQUIRED)
+if(QT_VERSION_MAJOR EQUAL 6)
+    find_package(Qt6 COMPONENTS Svg OpenGL OpenGLWidgets REQUIRED)
+else()
+    find_package(Qt5 COMPONENTS Svg OpenGL REQUIRED)
```
ResearchDaniel, line 9: Is this not the same as find_package(QT NAMES Qt6 Qt5 COMPONENTS Svg OpenGL REQUIRED)?

ResearchDaniel (dismissed):
Approved unless the comments need to be addressed :)

modules/openglqt/CMakeLists.txt
petersteneteg: for some reason the find package needs to be repeated with the real name, was some note in the qt docs, some variables are not set with the right name otherwise... 
the first one is mostly used to figure out witch qt that is installed...

modules/openglqt/CMakeLists.txt
petersteneteg: The openglwidgets is new in qt6 and the qopenglwidget it moved to it...

modules/openglqt/CMakeLists.txt
petersteneteg: the one above should not be there...

modules/openglqt/CMakeLists.txt
ResearchDaniel: Ok, would be good if a link to it could be added and state that it is a workaround

modules/openglqt/CMakeLists.txt
petersteneteg: it is gone in the latest version

ResearchDaniel (approved):

PR #1123, opened 2021-03-09
Animation: Fix for constant interpolation when playing the animation …

ResearchDaniel:
…in reverse

Having a keyframe at time 0 and playing from for example t=1 to t=0 would not set the value of the keyframe.
Added unit tests for ConstantInterpolation


petersteneteg (approved):
Nice catch

PR #1127, opened 2021-03-11
Fix/imagerescale events

ResearchDaniel:
Ensure that canvasSize of events matches the size of rescaled images. This behaviour matches that of the ViewManager

### DIFF
--- a/modules/basegl/CMakeLists.txt
+++ b/modules/basegl/CMakeLists.txt
@@ -19,6 +19,7 @@ set(HEADER_FILES
     include/modules/basegl/datavisualizer/meshvisualizer.h
     include/modules/basegl/datavisualizer/volumeraycastvisualizer.h
     include/modules/basegl/datavisualizer/volumeslicevisualizer.h
+    include/modules/basegl/eventscaler.h
     include/modules/basegl/processors/axisalignedcutplane.h
     include/modules/basegl/processors/background.h
     include/modules/basegl/processors/cuberenderer.h
@@ -125,6 +126,7 @@ set(SOURCE_FILES
     src/datavisualizer/meshvisualizer.cpp
     src/datavisualizer/volumeraycastvisualizer.cpp
     src/datavisualizer/volumeslicevisualizer.cpp
+    src/eventscaler.cpp
     src/processors/axisalignedcutplane.cpp
     src/processors/background.cpp
     src/processors/cuberenderer.cpp
--- a/modules/basegl/include/modules/basegl/eventscaler.h
+++ b/modules/basegl/include/modules/basegl/eventscaler.h
@@ -1,0 +1,80 @@
+/*********************************************************************************
+ *
+ * Inviwo - Interactive Visualization Workshop
+ *
+ * Copyright (c) 2021 Inviwo Foundation
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ *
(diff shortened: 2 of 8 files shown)


petersteneteg (changes requested):
Some initial thoughts. Can you explain I more detail how this is a problem? why do we even need to pass on the imageSize? if we do the translations correctly according to this, should we not end up with the original image size in that case? then the consumer could just use that size and the normalized coords or? 
Given that it is needed, should the eventscaler not be in base or core? 
Also does it handle cases where there aspect does not match? then one would need to adjust the positions to or? Should this functionality not go in the imageinport propagate event?

petersteneteg:
Is there a common base for events that hold a canvas size? If not why not add one? instead of repeating the various events...

ResearchDaniel:
> Can you explain I more detail how this is a problem?

It is only a problem if one tries to calculate the integer pixel position of an event and expect it to match the ImageOutport of a processor. 

Assuming that one has an image scaling processor that downsizes by 2 and then use an event that has propagated through that processor you could end up outside of the image outport in your processor “above” the scaling. 
your processor
|
scaling
|
canvas

ResearchDaniel:
> why do we even need to pass on the imageSize?

In case your processor does not have an image outport but need the event and wants to know the integer pixel position

ResearchDaniel:
> Should this functionality not go in the imageinport propagate event?

Hmm, probably. In case the “original” canvas size is needed somewhere we could add that to the event

ResearchDaniel:
> Is there a common base for events that hold a canvas size? If not why not add one? 

we could move it to InteractionEvent. I considered it but never did it.

petersteneteg:
> > why do we even need to pass on the imageSize?
> 
> In case your processor does not have an image outport but need the event and wants to know the integer pixel position

When would you need the pixel position of the event and not have the image in question at hand?

petersteneteg:
> > Is there a common base for events that hold a canvas size? If not why not add one?
> 
> we could move it to InteractionEvent. I considered it but never did it.

Key events don't have a size I think. and those derive InteractionEvent, MouseInteractionEvent has the size, but TouchEvents inherits Interaction Event and also have a size. So Now would need an extra class between InteractionEvent and MouseInteractionEvent that also TOuchEvent can derive from

ResearchDaniel:
> > > why do we even need to pass on the imageSize?
> > 
> > 
> > In case your processor does not have an image outport but need the event and wants to know the integer pixel position
> 
> When would you need the pixel position of the event and not have the image in question at hand?

https://github.com/inviwo/inviwo/blob/8bbd0dc7442eb4a280f6f8819f03889d1559c354/modules/webbrowser/src/interaction/cefinteractionhandler.cpp#L133
For convenience in many cases I guess. One could probably forward the size in most cases

ResearchDaniel:
Did a bit of research.
The ability to get canvasSize in an event is used in a few different places, e.g., BoxSelectionInteractionHandler, ViewManager, CefInteractionHandler. Those could probably be adjusted, but I think that it makes sense that one does not only have access to normalized coordinates in an event. This kind of leaves me with the conclusion that the canvasSize function should be there.

#### Moving this work into the ImageInport
We could override Inport::propagateEvent(Event* event, Outport* target) - create a copy of each event and set the canvas size to match that of the ImageInport. 
**Pros**: No need to consider the canvas size for the programmer.
**Cons**: A copy of the event would need to be created for each ImageInport of each processor (probably no performance impact?). There would be more 'magic' with respect to event propagation (I currently don't know any scenario where this behaviour would be undesirable, but if there are it is very hard to undo the magic).

I'm not sure which solution is best here.

petersteneteg:
Agree I would probably be nice to have the size in the event. Did you think anything more on the aspect ratio mismatch, Im thinking about the cases when we add black padding to an image for example. Also does this interact with the image port resize behavior, OutpotDetermainsSize and so on?

ResearchDaniel:
The ratio-matching cases will probably need to do something similar to the ViewManager. A bit more complicated than the problem I had in mind...

> Also does this interact with the image port resize behavior, OutpotDetermainsSize and so on

It is based on the actual size of the inport so I think that should not be a problem. I tried a couple of different settings in the ImageResampler manually and it seemed to work as intended. One would have to take care that the EventScaler has the correct size set. We could use a lambda here to avoid duplicating state.

PR #1128, opened 2021-03-15
Add addPorts method to processor, see reasoning below

PetiteViking:
This method will implicitly add all ports in the same port group so if a user wants to add his ports to a certain port group, he will have to do that explicitly with `addPort(myPort, "myPortGroup");`.
I would argue though that in 99% of the cases, adding the ports to the default group is the desired behaviour and that is why I think having this method makes sense in terms of convenience.

The behaviour is stated in the comment.


ResearchDaniel (approved):
Sounds good to me!

PR #1129, opened 2021-03-17
AnimationEditorQt: Added functionality for loading/saving animations …

ResearchDaniel:
…to file

Makes it possible to import an Animation from another workspace and save/load them in separate files


modules/animationqt/src/animationeditordockwidgetqt.cpp
```diff
@@ -190,6 +192,35 @@ AnimationEditorDockWidgetQt::AnimationEditorDockWidgetQt(
     toolBar->setFloatable(false);
     toolBar->setMovable(false);
     mainWindow_->addToolBar(toolBar);
+
+    // Menu
+    auto menu = mainWindow_->menuBar();
```
martinfalk: Is it necessary to have a menu bar with just the file menu? I would think just the toolbar buttons are sufficient. Compare with the python/shader editor.

modules/animationqt/src/animationeditordockwidgetqt.cpp
```diff
@@ -190,6 +192,35 @@ AnimationEditorDockWidgetQt::AnimationEditorDockWidgetQt(
...
+
+    auto fileMenuItem = menu->addMenu(tr("&File"));
+    // file menu entries
+    {
+        auto openAction = new QAction(QIcon(":/svgicons/open.svg"), tr("&Open Animation"), this);
+        openAction->setShortcut(QKeySequence::Open);
```
martinfalk: not sure this is a good idea. With this it depends on where you clicked last when trying to open a file. That is either the animation or workspace.
I'd consider this to be very confusing.

modules/animationqt/src/animationeditordockwidgetqt.cpp
```diff
@@ -190,6 +192,35 @@ AnimationEditorDockWidgetQt::AnimationEditorDockWidgetQt(
...
+        addAction(openAction);
+        connect(openAction, &QAction::triggered, this, [this]() { openAnimation(); });
+        fileMenuItem->addAction(openAction);
+        toolBar->addAction(openAction);
+    }
+
+    {
+        auto saveAction =
+            new QAction(QIcon(":/svgicons/save-as.svg"), tr("&Save Animation As"), this);
+        saveAction->setShortcut(QKeySequence::SaveAs);
```
martinfalk: same thought as with open file

modules/animationqt/src/animationeditordockwidgetqt.cpp
```diff
@@ -268,11 +299,69 @@ AnimationEditorDockWidgetQt::AnimationEditorDockWidgetQt(
...
+
     controller_.AnimationControllerObservable::addObserver(this);
 }
 
 AnimationEditorDockWidgetQt::~AnimationEditorDockWidgetQt() = default;
 
+void AnimationEditorDockWidgetQt::openAnimation() {
+    InviwoFileDialog openFileDialog(this, "Import Animation ...", "animation");
+    openFileDialog.addSidebarPath(PathType::Workspaces);
+    openFileDialog.addExtension("inv", "Inviwo File");
```
martinfalk, line 336: add a specific extension for animation files?

martinfalk: otherwise, an animation `.inv` can be confused with a regular workspace. What happens then?

modules/animationqt/src/animationeditordockwidgetqt.cpp
```diff
@@ -268,11 +299,69 @@ AnimationEditorDockWidgetQt::AnimationEditorDockWidgetQt(
...
+    if (openFileDialog.exec()) {
+        QString path = openFileDialog.selectedFiles().at(0);
+        std::string fileName{utilqt::fromQString(path)};
+        fileName = filesystem::cleanupPath(fileName);
+        if (!filesystem::fileExists(fileName)) {
+            LogError("Could not find file: " << fileName);
+            return;
+        }
+        try {
+            std::ifstream anim{std::string(fileName)};
```
martinfalk: use `filesystem::ifstream()` in order to handle unicode paths.

https://github.com/inviwo/inviwo/blob/20222bd6170a488ce8cf18df32f2fff0931c25b9/include/inviwo/core/util/filesystem.h#L107-L108

modules/animationqt/src/animationeditordockwidgetqt.cpp
```diff
@@ -268,11 +299,69 @@ AnimationEditorDockWidgetQt::AnimationEditorDockWidgetQt(
...
+        QString path = saveFileDialog.selectedFiles().at(0);
+        if (!path.endsWith(".inv")) path.append(".inv");
+        std::string fileName{utilqt::fromQString(path)};
+        fileName = filesystem::cleanupPath(fileName);
+        try {
+            Serializer serializer(fileName);
+            controller_.pause();
+            serializer.serialize("Animation", controller_.getAnimation());
+            serializer.serialize("AnimationController", controller_);
+            std::ofstream anim{std::string(fileName)};
```
martinfalk: `filesystem::ofstream()`

martinfalk: why creating a new string of `fileName`?

modules/animationqt/src/animationeditordockwidgetqt.cpp
martinfalk: why `std::string(fileName)` ?

martinfalk (changes requested):
Nice idea.

modules/animationqt/src/animationeditordockwidgetqt.cpp
petersteneteg: There should only be one menubar for the app. How will this work on Mac where the menubar is app global and not window specific.

modules/animationqt/src/animationeditordockwidgetqt.cpp
petersteneteg: Agree, if we do something like this is it has to go through the main app menu similar to how the edit menu works

modules/animationqt/src/animationeditordockwidgetqt.cpp
ResearchDaniel: Ok, I can remove the menubar. I guess the toolbar buttons should be descriptive enough. The only problem is if users think that it saves/loads the entire network since the icons are the same, but I think the context might give enough cues

modules/animationqt/src/animationeditordockwidgetqt.cpp
```diff
@@ -268,11 +299,69 @@ AnimationEditorDockWidgetQt::AnimationEditorDockWidgetQt(
...
+            LogError("Could not find file: " << fileName);
+            return;
+        }
+        try {
+            std::ifstream anim{std::string(fileName)};
+            auto app_ = InviwoApplication::getPtr();
+            auto deserializer = app_->getWorkspaceManager()->createWorkspaceDeserializer(
+                anim, std::string(fileName));
+            controller_.pause();
+            deserializer.deserialize("Animation", controller_.getAnimation());
```
petersteneteg: For this to work the destination network must have the same properties as the source network, which is a tricky precondition I think. I don't think open and save carry the right semantics for this kind of operation, something like import animation. Then it also makes sense to use the same extension. And it is understandable it there are issues...

petersteneteg (changes requested):

modules/animationqt/src/animationeditordockwidgetqt.cpp
martinfalk: I would think so.

modules/animationqt/src/animationeditordockwidgetqt.cpp
ResearchDaniel: It was actually what I first used, but then changed to match the shortcuts. I have switched back to import/export. 

Trying to open the exported animation will produce a bit of a nasty error ("application might be in a corrupt state"), but I think we can live with it for now as this will not be very common.

modules/animationqt/src/animationeditordockwidgetqt.cpp
ResearchDaniel: That is fine, it imports the saved Animation in that workspace.
Export is worse because we cannot open the generated workspace. 
I guess we could add ".ainv"? Might become a bit confusing?

petersteneteg: Should this even be allowed considering hat it is not self contained. I rather think one would just save the whole workspace and then import the tf

ResearchDaniel: If exporting is not allowed we should consider adding support for multiple animations in a workspace. That would be more work, but certainly an option.
Might be a minor issue but the entire workspace is in one of my cases 300 kB and just the animation is 2 kB

petersteneteg: Don't think the size is anything we have to care about at this point...

petersteneteg: One could also do export/import on a track level. that those files would then not refer to any other properties not in the file... copy/paste for track contents would also work here I think

petersteneteg: I generally think saving the animation by them selfs is a very bad idea given the very unclear and invisible requirements for begin able to load them again... would more or less only work when loading them into the same workspace as where is was save from.

modules/animationqt/src/animationeditordockwidgetqt.cpp
petersteneteg: I really don't want to introduce those kind of bugs willingly into the codebase. That should be fixed.

modules/animationqt/src/animationeditordockwidgetqt.cpp
ResearchDaniel: > would more or less only work when loading them into the same workspace as where is was save from

Agree, that is the use case I have :) So being able to export just this information is useful in that case. In general it might not be as useful. Having support for multiple animations in the workspace would solve this I think.

modules/animationqt/src/animationeditordockwidgetqt.cpp
ResearchDaniel: It is more a bug in the workspace deseralization, which is kind of independent of this.

petersteneteg: No in the regular deserialization we know that all properties exist. Since we gave invariants that are guaranteed by the animation editor and the network. In this use case though you break does invariants.

martinfalk (dismissed):

martinfalk (dismissed):

modules/animation/include/modules/animation/datastructures/animation.h
```diff
@@ -74,8 +74,9 @@ class IVW_MODULE_ANIMATION_API Animation : public AnimationObservable,
      * @param animationManager used for creating PropertyTrack/KeyframeSequence/Keyframe.
      */
     Animation(AnimationManager* animationManager = nullptr);
-    Animation(const Animation&) = delete;
-    Animation& operator=(const Animation& that) = delete;
+    Animation(const Animation&);
+    Animation(Animation&&) = default;
+    Animation& operator=(const Animation& that);
```
petersteneteg, line 79: Move assignment?

modules/animation/include/modules/animation/datastructures/basetrack.h
```diff
@@ -73,6 +73,10 @@ class BaseTrack : public Track, public KeyframeSequenceObserver {
                   "Seq has to derive from KeyframeSequence");
 
     BaseTrack(const std::string& identifier, const std::string& name, size_t priority = 0);
+    BaseTrack(const BaseTrack& other);
+    BaseTrack(BaseTrack&& other) = default;
+    BaseTrack& operator=(const BaseTrack&);
```
petersteneteg: Move assignment

modules/animation/include/modules/animation/datastructures/basetrack.h
```diff
@@ -73,6 +73,10 @@ class BaseTrack : public Track, public KeyframeSequenceObserver {
                   "Seq has to derive from KeyframeSequence");
 
     BaseTrack(const std::string& identifier, const std::string& name, size_t priority = 0);
+    BaseTrack(const BaseTrack& other);
```
petersteneteg: Is this supposed to be public? Or do you have to call clone?

modules/animation/include/modules/animation/datastructures/basetrack.h
```diff
@@ -159,6 +163,35 @@ template <typename Seq>
 BaseTrack<Seq>::BaseTrack(const std::string& identifier, const std::string& name, size_t priority)
     : identifier_{identifier}, name_{name}, priority_{priority} {}
 
+template <typename Seq>
+BaseTrack<Seq>::BaseTrack(const BaseTrack<Seq>& other)
+    : KeyframeSequenceObserver(other)
+    , enabled_{other.enabled_}
+    , identifier_{other.identifier_}
```
petersteneteg: Should the identifier not be unique?

modules/animation/include/modules/animation/datastructures/basetrack.h
```diff
@@ -159,6 +163,35 @@ template <typename Seq>
...
+    if (this != &that) {
+        while (!sequences_.empty()) {
+            remove(sequences_.back().get());
+        }
+        enabled_ = that.enabled_;
+        identifier_ = that.identifier_;
+        name_ = that.name_;
+        priority_ = that.priority_;
+        for (const auto& seq : that.sequences_) {
+            add(std::unique_ptr<Seq>(seq->clone()));
```
petersteneteg: Why do we need a clone here when the Sequence type is templates? We should know the concrete type here right?

modules/animation/include/modules/animation/datastructures/track.h
```diff
@@ -64,8 +64,8 @@ class IVW_MODULE_ANIMATION_API Track : public Serializable, public TrackObservab
      * Remove all keyframe sequences and call TrackObserver::notifyKeyframeSequenceRemoved
      */
     virtual ~Track() = default;
-    Track(const Track&) = delete;
-    Track& operator=(const Track&) = delete;
+
+    virtual Track* clone() const = 0;
```
petersteneteg, line 70: Should document the expected semantics with respect to any underlying reference into the network ie a property.

modules/animation/include/modules/animation/animationmodule.h
```diff
@@ -49,6 +50,9 @@ class IVW_MODULE_ANIMATION_API AnimationModule : public InviwoModule,
     virtual int getVersion() const override;
     virtual std::unique_ptr<VersionConverter> getConverter(int version) const override;
 
+    animation::WorkspaceAnimations& getWorkspaceAnimations();
```
petersteneteg, line 57: should probably explain here what's the difference between the main one and this one

modules/animation/include/modules/animation/datastructures/controltrack.h
```diff
@@ -50,6 +50,8 @@ class IVW_MODULE_ANIMATION_API ControlTrack : public BaseTrack<ControlKeyframeSe
     ControlTrack();
     virtual ~ControlTrack();
 
+    ControlTrack* clone() const override;
```
petersteneteg: our convention is to have virtual at the beginning too.

modules/animation/include/modules/animation/datastructures/basetrack.h
petersteneteg: should tracks follow reference semantics or value semantics? Since they are polymorphic I assume they should follow reference semantics, which makes moving them a bit strange perhaps..

modules/animation/include/modules/animation/datastructures/propertytrack.h
```diff
@@ -195,6 +195,8 @@ class PropertyTrack : public BaseTrack<Seq>, public BasePropertyTrack {
     static_assert(std::is_same<Key, typename Seq::key_type>::value,
                   "The KeyframeSequence must match Keyframe 'Key'");
 
+    PropertyTrack(const PropertyTrack& other) = default;
```
petersteneteg: private? clone should be used?

modules/animation/include/modules/animation/workspaceanimations.h
```diff
@@ -0,0 +1,120 @@
...
+    std::string_view getName(size_t index);
+    Animation& add(std::string_view name);
+    Animation& add(std::string_view name, Animation anim);
+    Animation& insert(size_t index, std::string_view name);
+    /**
+     * Deletes animations in the range [from to) and sets the main animation to the next available
+     * index.
+     * An empty animation will be added if all animations are erased.
+     */
+    void erase(size_t from, size_t to);
```
petersteneteg: half open range?

modules/animation/include/modules/animation/workspaceanimations.h
```diff
@@ -0,0 +1,120 @@
...
+
+    Animation& get(size_t index);
+    /**
+     * @return all animations matching name
+     */
+    std::vector<Animation*> get(std::string_view name);
+    const std::vector<Animation>& get() const { return animations_; }
+
+    std::string_view getName(size_t index);
+    Animation& add(std::string_view name);
```
petersteneteg, line 87: are names arbitrary of identifiers? are duplicated allowed? why are names needed?

modules/animation/include/modules/animation/workspaceanimations.h
petersteneteg: why not just erase(size_t index) and then a clear()

modules/animation/include/modules/animation/workspaceanimations.h
```diff
@@ -0,0 +1,120 @@
...
+    using OnChangedDispatcher = Dispatcher<void(size_t, size_t)>;
+    WorkspaceAnimations(InviwoApplication* app, AnimationManager& manager);
+    virtual ~WorkspaceAnimations() = default;
+
+    Animation& get(size_t index);
+    /**
+     * @return all animations matching name
+     */
+    std::vector<Animation*> get(std::string_view name);
+    const std::vector<Animation>& get() const { return animations_; }
```
petersteneteg: better to expose iterators.

modules/animation/src/workspaceanimations.cpp
```diff
@@ -0,0 +1,177 @@
...
+    animationSerializationHandle_ = app->getWorkspaceManager()->onSave([&](Serializer& s) {
+        s.serialize("MainAnimationIndex", mainAnimationIdx_);
+        s.serialize("Animations", animations_, "Animation");
+        s.serialize("AnimationNames", names_);
+    });
+    animationDeserializationHandle_ = app->getWorkspaceManager()->onLoad([&](Deserializer& d) {
+        size_t mainAnimation = 0;
+        erase(0, animations_.size());
+        d.deserialize("MainAnimationIndex", mainAnimation);
+        d.deserialize("AnimationNames", names_);
```
petersteneteg: what about the callbacks for name changes when undoing for example?

modules/animation/src/workspaceanimations.cpp
```diff
@@ -0,0 +1,177 @@
...
+    animationClearHandle_ =
+        app->getWorkspaceManager()->onClear([&]() { erase(0, animations_.size()); });
+    animationSerializationHandle_ = app->getWorkspaceManager()->onSave([&](Serializer& s) {
+        s.serialize("MainAnimationIndex", mainAnimationIdx_);
+        s.serialize("Animations", animations_, "Animation");
+        s.serialize("AnimationNames", names_);
+    });
+    animationDeserializationHandle_ = app->getWorkspaceManager()->onLoad([&](Deserializer& d) {
+        size_t mainAnimation = 0;
+        erase(0, animations_.size());
```
petersteneteg: we don't want to remove the old ones we generally want to update them. the IndexedDeserializer should generally do that.

modules/animation/src/workspaceanimations.cpp
```diff
@@ -0,0 +1,177 @@
...
+        d.deserialize("MainAnimationIndex", mainAnimation);
+        d.deserialize("AnimationNames", names_);
+
+        // Must pass AnimationManager to Animation constructor
+        util::IndexedDeserializer<Animation>("Animations", "Animation").setMakeNew([&]() {
+            return Animation(&animationManager_);
+        })(d, animations_);
+
+        // Failsafe
```
petersteneteg: for what?

modules/animation/src/workspaceanimations.cpp
```diff
@@ -0,0 +1,177 @@
...
+    , names_{{"Animation 1"}}
+    , mainAnimation_(app, animations_.front())
+    , mainAnimationIdx_{0} {
+
+    animationClearHandle_ =
+        app->getWorkspaceManager()->onClear([&]() { erase(0, animations_.size()); });
+    animationSerializationHandle_ = app->getWorkspaceManager()->onSave([&](Serializer& s) {
+        s.serialize("MainAnimationIndex", mainAnimationIdx_);
+        s.serialize("Animations", animations_, "Animation");
+        s.serialize("AnimationNames", names_);
```
petersteneteg: why not make the name part of the animation instead?

modules/animation/src/workspaceanimations.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+        // Failsafe
+        if (animations_.empty()) {
+            names_.clear();
+            add("Animation 1");
+            setMainAnimationIndex(0);
+        } else {
+            setMainAnimationIndex(mainAnimation);
+            auto idx = static_cast<size_t>(animations_.size() - 1);
+            onChanged_.invoke(0, idx);
```
petersteneteg: this should only be called in the main animation changes. Deserialization is called whenever we undo. And in the cases nothing has changes it should also do nothing.

modules/animation/src/workspaceanimations.cpp
```diff
@@ -0,0 +1,177 @@
...
+        }
+    });
+}
+
+Animation& animation::WorkspaceAnimations::get(size_t index) { return animations_.at(index); }
+
+std::vector<Animation*> WorkspaceAnimations::get(std::string_view name) {
+    std::vector<Animation*> animations;
+    for (auto&& [ind, elem] : util::enumerate(names_)) {
```
petersteneteg: use zip(names_, animations_)

modules/animation/src/workspaceanimations.cpp
```diff
@@ -0,0 +1,177 @@
...
+Animation& WorkspaceAnimations::add(std::string_view name) {
+    animations_.emplace_back(Animation(&animationManager_));
+    names_.emplace_back(name);
+    auto idx = static_cast<size_t>(animations_.size() - 1);
+    onChanged_.invoke(idx, idx);
+    return animations_.back();
+}
+
+Animation& WorkspaceAnimations::add(std::string_view name, Animation anim) {
+    animations_.emplace_back(anim);
```
petersteneteg:
```suggestion
    animations_.emplace_back(std::move(anim));
```

modules/animation/src/workspaceanimations.cpp
```diff
@@ -0,0 +1,177 @@
...
+    }
+}
+
+void WorkspaceAnimations::setMainAnimationIndex(size_t index) {
+    Animation& anim = animations_.at(mainAnimationIdx_);
+    mainAnimationIdx_ = index;
+    mainAnimation_.set(anim);
+}
+
+size_t WorkspaceAnimations::getMainAnimationIndex() const { return mainAnimationIdx_; }
```
petersteneteg: why is this index exposed, really seems like an implementation detail. is it even needed?

modules/animation/src/workspaceanimations.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+MainAnimation& WorkspaceAnimations::getMainAnimation() { return mainAnimation_; }
+
+const MainAnimation& WorkspaceAnimations::getMainAnimation() const { return mainAnimation_; }
+
+size_t WorkspaceAnimations::import(Deserializer& d) {
```
petersteneteg: think it makes more sense to have this as a free function.

modules/animationqt/src/animationeditordockwidgetqt.cpp
```diff
@@ -190,6 +194,54 @@ AnimationEditorDockWidgetQt::AnimationEditorDockWidgetQt(
...
+        connect(newAction, &QAction::triggered, this, [this]() {
+            animationsList_->addItem(
+                utilqt::toQString(fmt::format("Animation {}", animations_.get().size() + 1)));
+            animationsList_->setCurrentIndex(animationsList_->count() - 1);
+        });
+        toolBar->addAction(newAction);
+    }
+
+    {
+        auto openAction = new QAction(QIcon(":/svgicons/open.svg"), tr("&Import Animation"), this);
```
petersteneteg: I really wound not call this open. this is an import if even that. The animation is only transferable from workspace to workspace in very special cases. This should not be treated as a open button.

modules/animationqt/src/workspaceanimationsmodel.cpp
```diff
@@ -0,0 +1,118 @@
...
+    : QAbstractListModel(parent), animations_{animations} {
+
+    onChangedHandle_ = animations_.onChanged_.add([this](size_t from, size_t to) {
+        auto fromI = createIndex(static_cast<int>(from), 0);
+        auto toI = createIndex(static_cast<int>(to), 0);
+        emit dataChanged(fromI, toI);
+    });
+};
+
+inline Qt::ItemFlags AnimationsModel::flags(const QModelIndex& index) const {
```
petersteneteg:
```suggestion
Qt::ItemFlags AnimationsModel::flags(const QModelIndex& index) const {
```

resources/inviwo.qrc
```diff
@@ -100,6 +100,7 @@
         <file>svgicons/file-warning.svg</file>
 
         <file>svgicons/haptics.svg</file>
+        <file>svgicons/trashcan.svg</file>
```
petersteneteg: should this not be in the animationqt resources?

petersteneteg (changes requested):

modules/animation/include/modules/animation/datastructures/basetrack.h
ResearchDaniel: Good point. It would need to be unique within an animation. 
I have removed the identifier since it was basically not used and it would have been more complicated to enforce the uniqueness.

modules/animation/include/modules/animation/datastructures/basetrack.h
ResearchDaniel: True, changed to use copy constructor instead

modules/animation/include/modules/animation/workspaceanimations.h
ResearchDaniel: Sure, simplifies a bit.

modules/animation/src/workspaceanimations.cpp
ResearchDaniel: It is used by the GUI. I guess it could be made private, but would make the GUI part more complicated.

modules/animation/src/workspaceanimations.cpp
ResearchDaniel: I considered it, but right now it is based on also having names for the animations. So it is a bit more special case I guess.

modules/animationqt/src/animationeditordockwidgetqt.cpp
ResearchDaniel: It says "Import Animation" (also in the file dialog) and we currently do not have an icon for importing. I have renamed it to importAction in the code at least

resources/inviwo.qrc
ResearchDaniel: True, did not know that it had its own

modules/animation/src/workspaceanimations.cpp
ResearchDaniel: I guess we could. Would make it a bit easier even though it is only used for display purposes

modules/animation/include/modules/animation/animationmodule.h
ResearchDaniel: Added documentation

modules/animation/include/modules/animation/datastructures/propertytrack.h
ResearchDaniel: Made them protected now

modules/animation/include/modules/animation/datastructures/track.h
ResearchDaniel: Added documentation

ResearchDaniel:
@petersteneteg I think that your comments have been addressed now. Please check again. Undo/redo should now work with in the  Animation Editor.

modules/animation/src/workspaceanimations.cpp
```diff
@@ -37,137 +38,228 @@ namespace animation {
...
 
 const MainAnimation& WorkspaceAnimations::getMainAnimation() const { return mainAnimation_; }
 
+void WorkspaceAnimations::onAnimationChanged(AnimationController*, Animation*, Animation* newAnim) {
+    if (find(newAnim) == end()) {
+        // WorkspaceAnimations must own the Animation
+        throw Exception("MainAnimation must be set through WorkspaceAnimations");
+    }
+    // Enable undo/redo of currently selected Animation
+    app_->getProcessorNetwork()->notifyObserversProcessorNetworkChanged();
```
petersteneteg, line 186: humm I don't understand why we have to trigger this notification here?

modules/animation/src/workspaceanimations.cpp
petersteneteg: no names anymore though?

modules/animation/src/workspaceanimations.cpp
petersteneteg: ahh this it to make the undo state dirty so it will save the state... ic... I guess we don't have a proper way of doing that, we have not enabled it for anything else before.

petersteneteg (dismissed):
Nice work!

would like a cleaner way for the undo, but maybe that can be an other PR...

modules/animation/src/workspaceanimations.cpp
petersteneteg: ideally this should only observe animations, and the animaition should observe the track and so on. observing the multiple "levels" is a bit of a code smell also I think...

ResearchDaniel: One option would be to move the undo/redo state tracking to the UI. I could not see a good way to mark a state change though

ResearchDaniel: I guess that would involve partially moving the UndoManager to core?

ResearchDaniel:
@petersteneteg moved the import function to AnimationManager

modules/animation/src/workspaceanimations.cpp
petersteneteg: I think the workspace manager in core could be notified when something changes and also notify the undo manager for example

petersteneteg: Be we would need to extend the current api of course

petersteneteg (dismissed):

modules/animation/src/workspaceanimations.cpp
ResearchDaniel: > I think the workspace manager in core could be notified when something changes and also notify the undo manager for example

Sounds reasonable.

petersteneteg (approved):

PR #1130, opened 2021-03-17
Nifti: Add support for vector data (up to vec4)

ResearchDaniel:


modules/nifti/src/niftireader.cpp
```diff
@@ -112,8 +156,17 @@ const DataFormatBase* NiftiReader::niftiDataTypeToInviwoDataFormat(int niftiData
...
     }
     // clang-format on
-}
+    // In NIFTI-1 files, dimensions 1,2,3 are for space, dimension 4 is for time,
+    // and dimension 5 is for storing multiple values at each spatiotemporal voxel
+    if (niftiImage->dim[5] > 0) {
+        components = niftiImage->dim[5];
+    }
+    if (components > 4) {
+        return nullptr;
```
martinfalk, line 161: would it make sense to only consider the first 4 components?

modules/nifti/src/niftireader.cpp
```diff
@@ -86,22 +86,66 @@ NiftiReader::NiftiReader() : DataReaderType<VolumeSequence>() {
...
+            break;
+        case DT_INT16:      
+            type = NumericType::SignedInteger;
+            precision = 16;
+            break;
+        case DT_UINT16:     
+            type = NumericType::UnsignedInteger;
+            precision = 16;
+            break;
+            return DataUInt16::get();
```
martinfalk: remove

modules/nifti/src/niftireader.cpp
```diff
@@ -86,22 +86,66 @@ NiftiReader::NiftiReader() : DataReaderType<VolumeSequence>() {
...
+        case DT_INT32:      
+            type = NumericType::SignedInteger;
+            precision = 32;
+            break;
+        case DT_UINT32:     
+            type = NumericType::UnsignedInteger;
+            precision = 32;
+            break;
+            return DataUInt32::get();
```
martinfalk: remove

modules/nifti/src/niftireader.cpp
```diff
@@ -86,22 +86,66 @@ NiftiReader::NiftiReader() : DataReaderType<VolumeSequence>() {
...
+        case DT_INT64:      
+            type = NumericType::SignedInteger;
+            precision = 64;
+            break;
+            return DataInt64::get();
```
martinfalk: remove

modules/nifti/src/niftireader.cpp
```diff
@@ -86,22 +86,66 @@ NiftiReader::NiftiReader() : DataReaderType<VolumeSequence>() {
...
+        case DT_UINT64:     
+            type = NumericType::UnsignedInteger;
+            precision = 64;
+            break;
+            return DataUInt64::get();
```
martinfalk: remove

modules/nifti/src/niftireader.cpp
```diff
@@ -86,22 +86,66 @@ NiftiReader::NiftiReader() : DataReaderType<VolumeSequence>() {
...
+        case DT_FLOAT32:    
+            type = NumericType::Float;
+            precision = 32;
+            break;
+            return DataFloat32::get();
```
martinfalk, line 140: remove

modules/nifti/src/niftireader.cpp
```diff
@@ -112,8 +156,17 @@ const DataFormatBase* NiftiReader::niftiDataTypeToInviwoDataFormat(int niftiData
             return nullptr;
     }
     // clang-format on
-}
```
martinfalk, line 115: `DT_RGB24` and `DT_RGBA32` should also only set type and precision.

modules/nifti/src/niftireader.cpp
```diff
@@ -225,8 +278,13 @@ std::shared_ptr<NiftiReader::VolumeSequence> NiftiReader::readData(const std::st
     volume->setModelMatrix(basisAndOffset);
 
     std::array<int, 7> start_index = {0, 0, 0, 0, 0, 0, 0};
-    std::array<int, 7> region_size = {
-        niftiImage->dim[1], niftiImage->dim[2], niftiImage->dim[3], 1, 1, 1, 1};
+    std::array<int, 7> region_size = {niftiImage->dim[1],
+                                      niftiImage->dim[2],
+                                      niftiImage->dim[3],
+                                      1,
+                                      std::max(niftiImage->dim[5], 1),
```
martinfalk: use the DataFormat here instead

modules/nifti/src/niftireader.cpp
petersteneteg: Return after break?

martinfalk (changes requested):

petersteneteg (changes requested):

modules/nifti/src/niftireader.cpp
ResearchDaniel: I would rather give an error here at this point. More than 4 can mean matrices and then it will look like it works even though it is bogus data.

modules/nifti/src/niftireader.cpp
ResearchDaniel: I think that components will be 1 for these types, which do not translate to components in our DataFormat

ResearchDaniel: I do not have a data set to test this format though.

petersteneteg (approved):

modules/nifti/src/niftireader.cpp
martinfalk: I see

martinfalk (approved):

PR #1133, opened 2021-03-19
Add volume normalization algorithm

PetiteViking:
Adds a GL algorithm implementation for volume normalization.

I thought this might be useful to others. If not, I'll add it to my module, let me know.

### DIFF
--- a/modules/basegl/glsl/volumenormalization.frag
+++ b/modules/basegl/glsl/volumenormalization.frag
@@ -1,4 +1,33 @@
-#include "utils/sampler3d.glsl"
+/*********************************************************************************
+ *
+ * Inviwo - Interactive Visualization Workshop
+ *
+ * Copyright (c) 2014-2021 Inviwo Foundation
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ * 
+ *********************************************************************************/
+ 
+ #include "utils/sampler3d.glsl"
 
 uniform VolumeParameters volumeParameters;
 uniform sampler3D volume;
@@ -7,8 +36,8 @@ in vec4 texCoord_;
 
 void main() {
 
-vec4 original = texture(volume, texCoord_.xyz);
-float l, m, n, o;
+    vec4 original = texture(volume, texCoord_.xyz);
+    float l, m, n, o;
 
 #ifdef NORMALIZE_CHANNEL_0
     l = getNormalizedVoxelChannel(volume, volumeParameters, texCoord_.xyz, 0);
--- a/modules/basegl/glsl/volumenormalization.frag
+++ b/modules/basegl/glsl/volumenormalization.frag
@@ -1,0 +1,13 @@
+#include "utils/sampler3d.glsl"
+
+uniform VolumeParameters volumeParameters;
+uniform sampler3D volume;
+uniform int channel = 0;
+
+in vec4 texCoord_;
+
+void main() {
+    float v = getNormalizedVoxelChannel(volume, volumeParameters, texCoord_.xyz, channel);
+
+    FragData0 = vec4(v);
+}
--- a/modules/basegl/glsl/volumenormalization.frag
+++ b/modules/basegl/glsl/volumenormalization.frag
@@ -2,12 +2,37 @@
 
 uniform VolumeParameters volumeParameters;
 uniform sampler3D volume;
-uniform int channel = 0;
 
 in vec4 texCoord_;
 
 void main() {
-    float v = getNormalizedVoxelChannel(volume, volumeParameters, texCoord_.xyz, channel);
 
-    FragData0 = vec4(v);
+vec4 original = texture(volume, texCoord_.xyz);
+float l, m, n, o;
+
+#ifdef NORMALIZE_CHANNEL_0
+    l = getNormalizedVoxelChannel(volume, volumeParameters, texCoord_.xyz, 0);
+#else
+    l = original[0];
+#endif
+
+#ifdef NORMALIZE_CHANNEL_1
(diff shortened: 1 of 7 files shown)


modules/basegl/src/algorithm/volumenormalization.cpp
```diff
@@ -0,0 +1,71 @@
...
+#include <modules/basegl/algorithm/volumenormalization.h>
+
+#include <modules/opengl/volume/volumegl.h>
+#include <modules/opengl/shader/shaderutils.h>
+#include <modules/opengl/volume/volumeutils.h>
+
+namespace inviwo {
+
+std::shared_ptr<Volume> VolumeNormalization::normalize(const Volume& volume, const int channel) {
+    auto outVolume = std::make_shared<Volume>(volume.getDimensions(), volume.getDataFormat(),
```
ResearchDaniel: Output should be float with only one channel.

ResearchDaniel (changes requested):
I think it could be useful. No processor for it?

It would be possible to consider all channels in case there are using something like:
Is it possible to generalize to consider all channels like 
https://github.com/inviwo/inviwo/blob/267fe93e89ed77e8a28309cc62461b0e46e9e44a/modules/basegl/src/processors/multichannelraycaster.cpp#L104

modules/basegl/src/algorithm/volumenormalization.cpp
JochenJankowai: Not sure I agree. What I would rather do is to let the user specify which channels s/he wants to normalize. One could have a processor with a checkbox for every available channel. We could work with shader defines for control-free code.

I would argue against normalizing the data in sense of a norm since that would be more of a vectorfield operation.

I have added a corresponding processor in my module. Mostly because a volume processor that doesn't inherit from `VolumeGLProcessor` wouldn't fit in with the rest of the volume processors.

petersteneteg: There is no need for volume processors to inherit VolumeGlProcessor

JochenJankowai:
I've enabled the channel selection now and added a processor to go along with the algorithm.

modules/basegl/glsl/volumenormalization.frag
```diff
@@ -0,0 +1,38 @@
+#include "utils/sampler3d.glsl"
```
martinfalk: please add the Inviwo copyright notice.

modules/basegl/glsl/volumenormalization.frag
```diff
@@ -0,0 +1,38 @@
...
+
+uniform VolumeParameters volumeParameters;
+uniform sampler3D volume;
+
+in vec4 texCoord_;
+
+void main() {
+
+vec4 original = texture(volume, texCoord_.xyz);
```
martinfalk: indentation ;-)

modules/basegl/src/algorithm/volumenormalization.cpp
```diff
@@ -0,0 +1,104 @@
...
+               {ShaderType::Geometry, utilgl::findShaderResource("volume_gpu.geom")},
+               {ShaderType::Fragment, utilgl::findShaderResource("volumenormalization.frag")}},
+              Shader::Build::No)
+    , fbo_()
+    , needsCompilation_(false) {
+    shader_.getFragmentShaderObject()->addShaderDefine("NORMALIZE_CHANNEL_0");
+    shader_.build();
+}
+
+void VolumeNormalization::setNormalizeChannel(const size_t channel, const bool normalize) {
```
martinfalk, line 53: should there be a convenience function to set all 4 at once? e.g. with a `bvec4` or 4x bool

modules/basegl/include/modules/basegl/algorithm/volumenormalization.h
```diff
@@ -0,0 +1,69 @@
...
+    }
+    VolumeNormalization();
+
+    virtual ~VolumeNormalization() {}
+
+    std::shared_ptr<Volume> normalize(const Volume& volume);
+
+    void setNormalizeChannel(const size_t channel, const bool normalize);
+
+    void reset();
```
martinfalk, line 87: document what it does.

modules/basegl/include/modules/basegl/algorithm/volumenormalization.h
```diff
@@ -0,0 +1,69 @@
...
+#include <inviwo/core/common/inviwo.h>
+#include <modules/opengl/shader/shader.h>
+#include <inviwo/core/datastructures/volume/volume.h>
+#include <modules/opengl/buffer/framebufferobject.h>
+#include <array>
+#include <string_view>
+
+namespace inviwo {
+
+class IVW_MODULE_BASEGL_API VolumeNormalization {
```
martinfalk, line 48: add doxygen docs, at least that by default it only normalizes the first channel (for whatever reason)

modules/basegl/src/processors/volumeprocessing/volumenormalizationprocessor.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+            originalDataRange_ = volume->dataMap_.dataRange;
+            originalValueRange_ = volume->dataMap_.valueRange;
+
+            const auto channels = static_cast<int>(volume->getDataFormat()->getComponents());
+            if (channels == static_cast<int>(channels_.getProperties().size())) return;
+
+            NetworkLock l;
+
+            channels_.clear();
```
martinfalk: is it necessary to add/remove these properties dynamically? Why not just use 4 and hide them when not needed.

martinfalk (changes requested):
Nice. Only some minor details like minimal doxygen docs missing.

modules/basegl/include/modules/basegl/algorithm/volumenormalization.h
JochenJankowai: The default is only channel 0 because at the time of construction we do not know if the input volume will have more than 1 channel. It's just a failsafe.

I was actually working on the documentation while you made this review :)

modules/basegl/src/algorithm/volumenormalization.cpp
JochenJankowai: I though about it. If you think it would be useful, I'll add it. Didn't want to overload the whole thing.

modules/basegl/src/processors/volumeprocessing/volumenormalizationprocessor.cpp
JochenJankowai: That is indeed a better idea. I'll change that.

modules/basegl/include/modules/basegl/algorithm/volumenormalization.h
martinfalk: Ok, good point.

martinfalk (dismissed):

JochenJankowai:
I think I have addressed all comments and ironed out a few quirks. Have another look @martinfalk @ResearchDaniel

martinfalk (dismissed):

modules/basegl/src/algorithm/volumenormalization.cpp
ResearchDaniel: Output format must be float. Otherwise your data range below will not work

modules/basegl/include/modules/basegl/processors/volumeprocessing/volumenormalizationprocessor.h
```diff
@@ -0,0 +1,85 @@
...
+ * Normalizes the selected channels of the input volume to range [0,1].
+ *
+ * ### Inputs
+ *   * __Volume inport__ Input Volume
+ *
+ * ### Outports
+ *   * __Volume outport__ Normalized volume (if so selected)
+ *
+ * ### Properties
+ *   * __Channels__ Check the boxes for those channels you wish to normalize to range [0,1}
```
ResearchDaniel: Typo } -> ]

ResearchDaniel (changes requested):
Data format of the output needs to be changed otherwise I don’t think this will work on integer data.

After thinking about it we should probably call it VolumeChannelNormalization? I would most likely expect volume normalization to do vector normalization. What do you think @JochenJankowai

martinfalk:
@ResearchDaniel 
>Data format of the output needs to be changed otherwise I don’t think this will work on integer data.

That is a valid point. So the output format should be something like:
```c++
using ValueType = util::PrecisionValueType<decltype(repr)>;
using P = typename util::same_extent<ValueType, float>::type;

auto newVolume = std::make_shared<Volume>(volume->getDimensions(), DataFormat<P>::get());
```

> After thinking about it we should probably call it VolumeChannelNormalization? I would most likely expect volume normalization to do vector normalization. What do you think @JochenJankowai

That would be fitting if it normalizes only a single channel, but since you can choose all of them... Also, in the context of Inviwo a volume can have anything between one and four channels. I don't see a particular connection to vector fields there (neither 2D nor 3D).

martinfalk:
potentially do this specific cast to float volume only for integral types.

JochenJankowai:
What I think is that it's tricky. I personally would expect vector field normalization to be (a) called exactly that and (b) to be located in the vector field visualization module. But that's probably because I'm familiar with the system and its notations but I could totally see how it would be confusing to an outside user.

I'd vote for keeping the name and adding a hint to the documentation. Something along the lines of "If you're looking to normalize a vector field, please refer to..."

ResearchDaniel (dismissed):

martinfalk (dismissed):

ResearchDaniel (dismissed):

martinfalk (approved):

PR #1141, opened 2021-04-07
Tracking the last performed action of a list property

PetiteViking:


martinfalk:
Not quite sure I understand the reason for this additional property state. Wouldn't it be possible to observe the list property instead? See https://github.com/inviwo/inviwo/blob/master/include/inviwo/core/properties/propertyownerobserver.h

JochenJankowai:
Wasn't aware of the observers, closed

PR #1142, opened 2021-04-19
Added an optional canvas fullscreen command line option to qtminimum-app.

eriksunden:
In the effort to use Inviwo for public exhibition, the qtminimum app is preferred as the end application running the preferred workspace.
To make it even easier to use for this, I added a the an optional fullscreen command line option to qtminimum-app to occur if a single canvas is already visible, which should be the case in almost every use case as the end result.

### DIFF
--- a/apps/minimals/qt/qtminimum.cpp
+++ b/apps/minimals/qt/qtminimum.cpp
@@ -48,7 +48,6 @@
 #include <warn/push>
 #include <warn/ignore/all>
 #include <QSurfaceFormat>
-#include <QTimer>
 #include <warn/pop>
 
 using namespace inviwo;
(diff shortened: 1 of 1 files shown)


martinfalk:
What is the reason for the timer?

eriksunden:
Good question Martin :).
The timer is there to have an async delay on the fullscreen event. And the delay is actually introduced for two reasons:
1 (And the main one) : It seems Inviwo do not really support command line arguments without parameters. I could not just write "-f", then i t would crash. That's why it seemed logical to have a command line parameter that actually utilize some sort of value, hence the delay for launching it in fullscreen.
2: Why not ;), maybe something occurs during loading of an advanced scene or some other issue, that requires a delay of the fullscreen initialization. Just a feature :). It is easy to set "-f 0" to get instant fullscreen.

eriksunden:
> What is the reason for the timer?

Good question Martin :).
The timer is there to have an async delay on the fullscreen event. And the delay is actually introduced for two reasons:
1 (And the main one) : It seems Inviwo do not really support command line arguments without parameters. I could not just write "-f", then i t would crash. That's why it seemed logical to have a command line parameter that actually utilize some sort of value, hence the delay for launching it in fullscreen.
2: Why not ;), maybe something occurs during loading of an advanced scene or some other issue, that requires a delay of the fullscreen initialization. Just a feature :). It is easy to set "-f 0" to get instant fullscreen.

petersteneteg:
Inviwo does support switches without arguments

apps/minimals/qt/qtminimum.cpp
```diff
@@ -88,6 +92,38 @@ int main(int argc, char** argv) {
         },
         1000);
 
+    TCLAP::ValueArg<int> fullscreenArg(
```
martinfalk: Should be a `TCLAP::SwitchArg`

See for example
https://github.com/inviwo/inviwo/blob/5e6f46832db04a4f160e464d7de29b5be8f9d70c/include/inviwo/qt/editor/inviwomainwindow.h#L251-L252

martinfalk (changes requested):
Remove the delay. (Doesn't really have a purpose)

apps/minimals/qt/qtminimum.cpp
eriksunden: OK, forgot this :). All fixed now.

martinfalk (approved):

PR #1143, opened 2021-04-19
BaseGL: SplitImage GL fix

martinfalk:
* ImageGL representations potentially were only copied partially due to the scissor test

### DIFF
--- a/modules/basegl/src/processors/splitimage.cpp
+++ b/modules/basegl/src/processors/splitimage.cpp
@@ -132,8 +132,6 @@ void SplitImage::process() {
         utilgl::singleDrawImagePlaneRect();
         shader->deactivate();
     } else {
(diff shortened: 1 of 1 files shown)


ResearchDaniel (approved):

modules/basegl/src/processors/splitimage.cpp
```diff
@@ -181,9 +177,14 @@ void SplitImage::process() {
...
             }
 
+            glEnable(GL_SCISSOR_TEST);
+            utilgl::ScissorState scissorviewport(viewport);
+
             shader->activate();
             utilgl::singleDrawImagePlaneRect();
             shader->deactivate();
+
+            glDisable(GL_SCISSOR_TEST);
```
petersteneteg, line 187: If the shader calls would throw, the GL state will be left in the wrong state...

PR #1144, opened 2021-04-19
Data: Paraview color maps

martinfalk:
* transfer functions based on Paraview's color maps


petersteneteg (approved):
copyright?

martinfalk:
Those colormaps were obtained via screenshots. The original ones appear to come partially from IDL and matplotlib (https://www.paraview.org/Wiki/Colormaps).

ResearchDaniel (commented):
Is it better to have them as images than itf?
Would be good with a readme saying something about the TFs as with the ones from matplotlib, is it all the ones paraview has for example?
They seem to overlap with the existing presets as well and it would be good to say something about if and how they differ

PR #1145, opened 2021-04-19
Feature/fonts

martinfalk:
* changing the default font from Montserrat to OpenSans semibold (nicer numbers)
* added new font: Lato

image


modules/fontrendering/include/modules/fontrendering/textrenderer.h
```diff
@@ -109,7 +109,7 @@ struct TextTextureObject {
  */
 class IVW_MODULE_FONTRENDERING_API TextRenderer {
 public:
-    TextRenderer(const std::string& fontPath = util::getDefaultFontPath() + "/arial.ttf");
+    TextRenderer(const std::string &fontPath = util::getDefaultFontPath() + "/OpenSans-semibold.ttf");
```
petersteneteg: maybe we should have a util::getDefaultFont()? could even be backed by a setting a some point...

petersteneteg (dismissed):
Use a defaultFont Function throughout...

modules/fontrendering/include/modules/fontrendering/textrenderer.h
martinfalk: good idea. Will add one.

modules/fontrendering/include/modules/fontrendering/util/fontutils.h
```diff
@@ -57,12 +55,42 @@ getAvailableFonts(const std::string& fontPath = std::string());
...
 std::string IVW_MODULE_FONTRENDERING_API getDefaultFontPath();
 
+/**
+ * \brief returns the default typeface of Inviwo along with the default font path and extension if
+ * \p fullPath is true
+ *
+ * @param fullPath    the function returns the full path of the default font if true
+ * @return default font name
+ */
+std::string IVW_MODULE_FONTRENDERING_API getDefaultFont(bool fullPath = false);
```
petersteneteg: Nice! But I avoid using a book it reads really badly at the call sites. Either add two functions or use an enum, could also use a enum for the different kind of fonts, bold, caption, etc.

petersteneteg: Also all four functions should have the path option I think...

modules/fontrendering/include/modules/fontrendering/util/fontutils.h
```diff
@@ -57,12 +58,28 @@ getAvailableFonts(const std::string& fontPath = std::string());
...
+/**
+ * \brief returns the default typeface of Inviwo for \p type along with the default font path and
+ * extension if \p fullPath is true
+ *
+ * @param type    requested font type
+ * @param path    if equal to FullPath::Yes, then the full path of the font including extension is
+ *                returned, only font name otherwise
+ * @return font name
+ */
+std::string IVW_MODULE_FONTRENDERING_API getFont(FontType type, FullPath path = FullPath::No);
```
ResearchDaniel, line 73: Is there are reason we are not returning a font?
Would make more sense instead of a path to a font.

petersteneteg (approved):

ResearchDaniel (approved):
I think that it would make sense to use a font instead of the path to a font if possible.

modules/fontrendering/include/modules/fontrendering/util/fontutils.h
martinfalk: Font names (i.e. no path, no extension) are used in various properties, most often OptionProperties for font selection. Paths are used to refer to the actual font. And since Inviwo potentially supports multiple font directories... In addition, FreeType has support for otf, ttf, cff, and pcf fonts. Hence, tricky to figure out the extension.
What I would like to have is select a font family and then the style (bold, italics, semibold, etc.). But naming schemes are all over the place to do that. Some kind of heuristics needed.

ResearchDaniel: Font names can/could be retrieved from the font object if that is needed?

PR #1149, opened 2021-04-21
Feature/splitter

martinfalk:
* refactoring of the splitter handle from `SplitImage` into a separate renderer and property
* support to change the mouse cursor in mouse events, for example during picking or hovering

https://user-images.githubusercontent.com/9251300/115586958-6a478700-a2cd-11eb-9b24-6377a08c8a55.mp4


modules/basegl/glsl/splitter.geom
```diff
@@ -0,0 +1,228 @@
...
+
+vec2 convertNDCToScreen(vec2 v) {
+    return (v + 1.0) * 0.5 * screenDim;
+}
+
+vec4 convertScreenToNDC(vec2 v, float z) {
+    return vec4(v / screenDim * 2.0 - 1.0, z, 1.0);
+}
+
+void main() {
```
petersteneteg, line 85: alot of this has to be duplicate from other line renderers? should it not be shared?

modules/basegl/glsl/splitter.vert
```diff
@@ -27,27 +27,4 @@
...
-flat out uint pickID_;
- 
-void main() {
-    vertexColor_ = (pickingEnabled ? vec4(0) : color);
-
-    worldPosition_ = geometry.dataToWorld * in_Vertex;
-    gl_Position = camera.worldToClip * worldPosition_;
-    pickID_ = pickingEnabled ? pickId : 0;
-}
+void main() {}
```
petersteneteg, line 30: I think passing in a line for each splitter to be drawn would be a better method. Then multiple splitters could be drawn at the same time. Or make trafo a list and loop over it, but that seems strange...

modules/basegl/glsl/splitter.geom
petersteneteg: also is all this needed here since it will only be 1 line and it will be axis alined?

modules/basegl/glsl/splittertriangle.geom
```diff
@@ -0,0 +1,62 @@
...
+out vec4 color_;
+
+void emit(in vec2 pos) {
+    gl_Position = trafo * vec4(pos * screenDimInv, 0, 1);
+    color_ = color;
+    EmitVertex();
+}
+
+void main() {
+    const float sqrt3half = 0.8660254037844386;
```
petersteneteg, line 48: could use this one for the lines to...

modules/basegl/include/modules/basegl/rendering/splitterrenderer.h
```diff
@@ -0,0 +1,93 @@
...
+ */
+class IVW_MODULE_BASEGL_API SplitterRenderer {
+public:
+    using Callback = std::function<void(float)>;
+
+    SplitterRenderer(const SplitterSettings& settings, Processor* processor);
+    SplitterRenderer(const SplitterRenderer& rhs);
+    SplitterRenderer(SplitterRenderer&& rhs) = delete;
+    SplitterRenderer& operator=(const SplitterRenderer& rhs) = delete;
+    SplitterRenderer& operator=(SplitterRenderer&& rhs) = default;
```
petersteneteg: so you allow copy construct but not move construct, and allow move assign but not copy assign? seems very strange

modules/basegl/include/modules/basegl/rendering/splitterrenderer.h
```diff
@@ -0,0 +1,93 @@
...
+/**
+ * \brief utility class for rendering a vertical or horizontal splitter and providing drag
+ * interactions. This class will invalidate the processor for hover and drag events in order to
+ * trigger a redraw.
+ */
+class IVW_MODULE_BASEGL_API SplitterRenderer {
+public:
+    using Callback = std::function<void(float)>;
+
+    SplitterRenderer(const SplitterSettings& settings, Processor* processor);
```
petersteneteg: is there a good reason for passing in the settings here and not in the render call?
We might have done it like this before. But Im not convinced it is the best way. The semantics of how the settings will he handled get a bit unclear? will the renderer copy the settings and store them. So you have to recreate the renderer to change a setting. Or does it store a reference to the settings, meaning you have to make sure to keep the settings alive while the renderer is a live.

petersteneteg: What is the processor needed for?

modules/basegl/src/properties/splitterproperty.cpp
```diff
@@ -0,0 +1,92 @@
...
+              {"line", "Line", splitter::Style::Line},
+              {"invisible", "Invisible", splitter::Style::Invisible}})
+    , color_("color", "Color", util::ordinalColor(0.71f, 0.81f, 0.85f, 1.0f))
+    , bgColor_("bgColor", "Background Color", util::ordinalColor(0.27f, 0.3f, 0.31f, 1.0f))
+    , hoverColor_("hoverColor", "Hover Color", util::ordinalColor(1.0f, 0.666f, 0.0f, 1.0f))
+    , length_("length", "Length (pixel)", 75.0f, 0.0f, 200.0f, 1.0f)
+    , width_("width", "Width (pixel)", 9.0f, 0.0f, 50.0f, 0.25f)
+    , triSize_("triSize", "Triangle Size", 10.0f, 0.0f, 50.0f, 0.5f) {
+
+    length_.setSemantics(PropertySemantics("SpinBox"));
```
petersteneteg: move to constructor. need to make it default otherwise

modules/basegl/src/properties/splitterproperty.cpp
```diff
@@ -0,0 +1,92 @@
...
+    width_.setSemantics(PropertySemantics("SpinBox"));
+    triSize_.setSemantics(PropertySemantics("SpinBox"));
+
+    style_.setSelectedValue(style);
```
petersteneteg, line 56: set as default, or move to constructor...

modules/basegl/src/rendering/splitterrenderer.cpp
```diff
@@ -0,0 +1,263 @@
...
+namespace inviwo {
+
+SplitterRenderer::SplitterRenderer(const SplitterSettings& settings, Processor* processor)
+    : settings_(settings)
+    , processor_(processor)
+    , shader_("splitter.vert", "splitter.geom", "linerenderer.frag")
+    , triShader_("splitter.vert", "splittertriangle.geom", "standard.frag")
+    , pickingMapper_(processor, 1, [&](PickingEvent* e) { handlePickingEvent(e); }) {
+
+    mesh_ = std::make_shared<Mesh>(DrawType::Points, ConnectivityType::None);
```
petersteneteg: Why a shared_ptr, just us a Mesh

modules/basegl/src/rendering/splitterrenderer.cpp
```diff
@@ -0,0 +1,263 @@
...
+#include <inviwo/core/datastructures/geometry/mesh.h>
+
+namespace inviwo {
+
+SplitterRenderer::SplitterRenderer(const SplitterSettings& settings, Processor* processor)
+    : settings_(settings)
+    , processor_(processor)
+    , shader_("splitter.vert", "splitter.geom", "linerenderer.frag")
+    , triShader_("splitter.vert", "splittertriangle.geom", "standard.frag")
+    , pickingMapper_(processor, 1, [&](PickingEvent* e) { handlePickingEvent(e); }) {
```
petersteneteg, line 51: ahh that's where the processor is used. We should change the picking at some point to remove that convoluted dependency.

modules/basegl/src/rendering/splitterrenderer.cpp
```diff
@@ -0,0 +1,263 @@
...
+            break;
+        case splitter::Style::Line:
+            lineWidth = width;
+            drawHandle = false;
+            break;
+        case splitter::Style::Invisible:
+            break;
+        default:
+            break;
+    }
```
petersteneteg:
```suggestion
const auto [lineWidth, handleWidth, drawHandle, drawBorder] = [&]() -> std::tuple<float, float, bool, bool> {
    const float width = settings_.get().getWidth();
    switch (style) {
        case splitter::Style::Handle:
            return {width / 3.0f, width, true, true};
        case splitter::Style::Divider:
            return {widtf, width * 0.6f, true, false};
        case splitter::Style::Line:
            return {width, 0.0f, false, false};
        case splitter::Style::Invisible:
            return {0.0f, 0.0f, true, false};
        default:
            return {0.0f, 0.0f, true, false};
    }
}();
```

modules/basegl/src/rendering/splitterrenderer.cpp
```diff
@@ -0,0 +1,263 @@
...
+
+    mat4 m(1.0f);
+    m[0][0] = 2.0f;
+    m[1][1] = 2.0f;
+    m[3] = vec4(pos * 2.0f - 1.0f, -1.0f, 0.0f, 1.0f);
```
petersteneteg: make into a function with a name to explain what it is...

modules/basegl/src/rendering/splitterrenderer.cpp
```diff
@@ -0,0 +1,263 @@
...
+
+    if (e->getEvent()->hash() == MouseEvent::chash()) {
+        auto mouseEvent = e->getEventAs<MouseEvent>();
+
+        const bool leftMouseBtn = (mouseEvent->button() == MouseButton::Left);
+        const bool mousePress = (mouseEvent->state() == MouseState::Press);
+        const bool mouseMove = (mouseEvent->state() == MouseState::Move);
+
+        if (e->getState() == PickingState::Started) {
+            // start hover
```
petersteneteg:
```suggestion
```

modules/basegl/src/rendering/splitterrenderer.cpp
```diff
@@ -0,0 +1,263 @@
...
+            hover_ = true;
+
+            const auto cursor = (currentDirection_ == splitter::Direction::Vertical)
+                                    ? MouseCursor::SplitH
+                                    : MouseCursor::SplitV;
+            mouseEvent->setMouseCursor(cursor);
+            triggerUpdate = true;
+        } else if (e->getState() == PickingState::Finished) {
+            // end hover
```
petersteneteg, line 204:
```suggestion
```

modules/basegl/src/rendering/splitterrenderer.cpp
```diff
@@ -0,0 +1,263 @@
...
+            dragAction_(static_cast<float>(screenPos.x));
+        } else {
+            // invert vertical position
+            screenPos.y = 1.0 - screenPos.y;
+            dragAction_(static_cast<float>(screenPos.y));
+        }
+    };
+
+    if (e->getEvent()->hash() == MouseEvent::chash()) {
+        auto mouseEvent = e->getEventAs<MouseEvent>();
```
petersteneteg:
```suggestion
    if (auto mouseEvent = e->getEventAs<MouseEvent>()) {
```

modules/basegl/src/rendering/splitterrenderer.cpp
```diff
@@ -0,0 +1,263 @@
...
+                triggerUpdate = true;
+            } else if (touchPoint.state() == TouchState::Updated) {
+                moveAction(touchPoint.pos());
+                triggerUpdate = true;
+                e->markAsUsed();
+            }
+        }
+    }
+    if (triggerUpdate) {
+        processor_->invalidate(InvalidationLevel::InvalidOutput);
```
petersteneteg: might be better with a callback here similar to moveAction

petersteneteg (changes requested):

modules/basegl/glsl/splittertriangle.geom
martinfalk: what do you mean?

modules/basegl/include/modules/basegl/rendering/splitterrenderer.h
martinfalk: ~The settings are kept via a reference similar to the AxisRenderer and its settings.~
Settings are now provided via the `render()` call

The processor is required for the picking mapper inside the renderer needed to handle interaction event.

modules/basegl/src/rendering/splitterrenderer.cpp
martinfalk: looks neater.

modules/basegl/include/modules/basegl/rendering/splitterrenderer.h
martinfalk: Multiple splitters will reuse the same settings.

modules/basegl/glsl/splittertriangle.geom
petersteneteg: That shader, the splitter line also only need a quad, or?

martinfalk: A single point becomes two triangles. But I guess the line shader could be simplified, since there are only orthogonal lines.

modules/basegl/include/modules/basegl/rendering/splitterrenderer.h
martinfalk: inspired by AxisRenderer, I think.

modules/basegl/glsl/splitter.geom
martinfalk: The reason is to render geometry-less lines in normalized device coordinates (transformation contains rotation and position). This avoids recreating the mesh when extending it for multiple splitters.

modules/basegl/glsl/splitter.vert
martinfalk: trafo list is possible in the geometry shader. The main idea was to avoid recreating the mesh.

modules/basegl/include/modules/basegl/rendering/splitterrenderer.h
martinfalk: move constructor should be default.
Picking mapper prevents copy-assignment

modules/basegl/src/properties/splitterproperty.cpp
martinfalk: not possible to move to constructor since that one only uses an index to select the initial option.

modules/basegl/src/rendering/splitterrenderer.cpp
martinfalk: agree. Also annoying: the picking mapper holds a processor, but no way to query it. Hence the additional `processor_` member.

modules/basegl/src/properties/splitterproperty.cpp
petersteneteg: just use a lambda.

petersteneteg: or add the constructor

modules/basegl/src/rendering/splitterrenderer.cpp
petersteneteg: well since you can't query it is easier to remove at some point...

petersteneteg (approved):

PR #1152, opened 2021-04-28
Feature/layout

martinfalk:
`Column Layout` and `Row Layout` processors for interactive layouts using splitters.

image


modules/basegl/src/processors/imageprocessing/columnrowlayout.cpp
```diff
@@ -0,0 +1,325 @@
...
+
+    splitterSettings_.setCollapsed(true);
+    splitterSettings_.width_.set(4.0f);
+    splitterSettings_.triSize_.set(0.0f);
+    splitterSettings_.setCurrentStateAsDefault();
+
+    addProperties(splitterSettings_, minWidth_, splitters_);
+
+    inport_.setIsReadyUpdater([this]() {
+        // Ports with dimensions (1,1) or less are considered to be inactive
```
petersteneteg, line 83: you say less here but use equal 1 below?

petersteneteg (dismissed):

modules/basegl/src/processors/imageprocessing/columnrowlayout.cpp
martinfalk: good catch. Should be addressed now.

petersteneteg (approved):

PR #1154, opened 2021-05-04
Fix/webbrowser xcode

ResearchDaniel:
Mac fixes.
Webbrowser
- Add support for mac arm64
- Remove XCode warnings (tested on XCode 12.5).
- Added entitlements for helpers

CMake
- XCode (version 11 and later?, tested on version 12.5) will fail on build when the Webbrowser module is enabled because of the added CEF framework

### DIFF
--- a/cmake/compileoptions.cmake
+++ b/cmake/compileoptions.cmake
@@ -33,6 +33,13 @@
 
 option(IVW_CFG_TREAT_WARNINGS_AS_ERRORS "Treat compiler warnings as errors" OFF)
 option(IVW_CFG_FORCE_ASSERTIONS "Force use of assertions when not in debug mode" OFF)
+if(CMAKE_GENERATOR STREQUAL "Xcode")
+    # Prevent Xcode 11 from doing automatic codesigning because it will fail the build. 
+    # Causes build to fail if Webbrowser module is enabled due to the added CEF framework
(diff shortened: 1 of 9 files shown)


petersteneteg:
Needs some more fixes for the arm stuff. Have some patches I will push as soon as I get some time

petersteneteg:
image
@ResearchDaniel as soon as I load the webrowser processor I crash here, any ideas? had to update to a newer CEF to get arm  support using 90.6.3+gc53c523+chromium-90.0.4430.93 now

ResearchDaniel:
Good question. Related to this?
https://bitbucket.org/chromiumembedded/cef/issues/3115/cef-crash-on-mac-on-createbrowsersync

petersteneteg:
Do you know if we do wait for that callback otherwise that seems likely

ResearchDaniel:
> Do you know if we do wait for that callback otherwise that seems likely

We are using a public function that is supposed to wait... I could reproduce on Windows. I noticed that the CEF example project has not updataed to version 90. Perhaps there are some issues.

ResearchDaniel:
Works on the version before 90, "89.0.12+g2b76680+chromium-89.0.4389.90"

petersteneteg (approved):
this is with version 89, we should probably need to figure out why 90 does not work at some point... but that can be a different PR...

Issue #1136, opened 2021-03-22
Property Link Editor does not reflect dynamic property changes


martinfalk:
When adding/removing dynamic properties, e.g. with a `ListProperty`, the Property Link Editor does not update while open.

![image](https://user-images.githubusercontent.com/9251300/112006862-68e73b00-8b24-11eb-9f9c-e98d84b6b2a5.png)

martinfalk:
Closing the dialog occasionally results in a crash.

PR #1156, opened 2021-05-07
Feature/rasterizer

martinfalk:
basic rasterizer for sphere glyphs


modules/meshrenderinggl/glsl/fancymeshrenderer.frag
```diff
@@ -334,6 +334,7 @@ vec4 performShading() {
 }
 
 void main() {
+    discard;
```
petersteneteg: Always?

martinfalk: oops. Not supposed to be there.

modules/meshrenderinggl/glsl/oit-sphereglyph.frag
```diff
@@ -0,0 +1,146 @@
...
+#endif
+
+uniform CameraParameters camera;
+uniform LightParameters lighting;
+uniform float uniformAlpha = 1.0;
+
+uniform vec4 viewport; // holds viewport offset x, offset y, 2 / viewport width, 2 / viewport height
+uniform float clipShadingFactor = 0.9;
+
+in vec4 color_;
```
petersteneteg: we have decided not to use _ for uniforms.

modules/meshrenderinggl/include/modules/meshrenderinggl/processors/sphererasterizer.h
```diff
@@ -0,0 +1,173 @@
...
+
+/** \docpage{org.inviwo.SphereRasterizer, Sphere Rasterizer}
+ * ![](org.inviwo.SphereRasterizer.png?classIdentifier=org.inviwo.SphereRasterizer)
+ * Create a rasterization object to render one or more point meshes using spherical glyphs in
+ * OpenGL. The glyphs are resolution independent and consist only of a single point.
+ *
+ * ### Inports
+ *   * __geometry__ Input meshes
+ *       The input mesh uses the following buffers:
+ * PositionAttrib vec3
```
petersteneteg: indent?

modules/meshrenderinggl/src/processors/sphererasterizer.cpp
```diff
@@ -0,0 +1,264 @@
...
+    , clipShadingFactor_("clipShadingFactor", "Clip Surface Adjustment", 0.9f, 0.0f, 2.0f)
+    , shadeClippedArea_("shadeClippedArea", "Shade Clipped Area", false,
+                        InvalidationLevel::InvalidResources)
+    , sphereProperties_("sphereProperties", "Sphere Properties")
+    , forceRadius_("forceRadius", "Force Radius", false, InvalidationLevel::InvalidResources)
+    , defaultRadius_("defaultRadius", "Default Radius", 0.05f, 0.00001f, 2.0f, 0.01f)
+    , forceColor_("forceColor", "Force Color", false, InvalidationLevel::InvalidResources)
+    , defaultColor_("defaultColor", "Default Color", vec4(0.7f, 0.7f, 0.7f, 1.0f), vec4(0.0f),
+                    vec4(1.0f), vec4(0.01f), InvalidationLevel::InvalidOutput,
+                    PropertySemantics::Color)
```
petersteneteg: use util::ordinalColor and set the constraintbehaviowurs

modules/meshrenderinggl/src/processors/sphererasterizer.cpp
```diff
@@ -0,0 +1,264 @@
...
+    clipping_.addProperties(clipMode_, clipShadingFactor_, shadeClippedArea_);
+
+    sphereProperties_.addProperties(forceRadius_, defaultRadius_, forceColor_, defaultColor_,
+                                    useMetaColor_, metaColor_);
+
+    addProperties(renderMode_, forceOpaque_, useUniformAlpha_, uniformAlpha_, sphereProperties_,
+                  clipping_, transformSetting_, camera_, lighting_, trackball_);
+
+    clipMode_.onChange([&]() {
+        clipShadingFactor_.setReadOnly(clipMode_.get() == GlyphClippingMode::Discard);
```
petersteneteg: use readOnlyDependsOn

modules/meshrenderinggl/src/processors/sphererasterizer.cpp
```diff
@@ -0,0 +1,264 @@
...
+        shader.setUniform("metaColor", texUnit.getUnitNumber());
+        shader.setUniform("defaultColor", defaultColor_);
+        shader.setUniform("defaultRadius", defaultRadius_);
+        shader.setUniform("clipShadingFactor", clipShadingFactor_);
+        shader.setUniform("uniformAlpha", uniformAlpha_);
+        utilgl::setShaderUniforms(shader, lighting_, "lighting");
+        shader.setUniform("viewport", vec4(0.0f, 0.0f, 2.0f / imageSize.x, 2.0f / imageSize.y));
+        MeshDrawerGL::DrawObject drawer(mesh->getRepresentation<MeshGL>(),
+                                        mesh->getDefaultMeshInfo());
+        setUniforms(shader);
```
petersteneteg, line 221: again?

martinfalk:
@ResearchDaniel could you please have another look at the changes regarding OpenCL?

modules/meshrenderinggl/glsl/oit-sphereglyph.frag
```diff
@@ -0,0 +1,146 @@
...
+in float radius_;
+in vec3 camPos_;
+in vec4 center_;
+flat in vec4 pickColor_;
+
+
+void clipToSolid(in vec4 coord, in vec3 srcColor, out vec4 dstColor, out float dstDepth) {
```
ResearchDaniel: Would be good to describe this function. Is it only used here or should it be extracted?

modules/meshrenderinggl/glsl/oit-sphereglyph.frag
```diff
@@ -0,0 +1,146 @@
...
+                               normal, normalize(camPos_ - intersection));
+#if defined(UNIFORM_ALPHA)
+    glyphColor.a = uniformAlpha;
+#else
+    glyphColor.a = color_.a;
+#endif // UNIFORM_ALPHA
+    if (glyphColor.a < 0.01) discard;
+
+    // depth correction for glyph
+    mat4 mvpTranspose = transpose(camera.worldToClip);
```
ResearchDaniel: Might want to extract this to a function

modules/meshrenderinggl/include/modules/meshrenderinggl/processors/sphererasterizer.h
```diff
@@ -0,0 +1,173 @@
...
+ *
+ * ### Outports
+ *   * __rasterization__ rasterization functor rendering  the molecule either opaquely or into
+ *                    fragment buffer
+ *
+ * ### Properties
+ *   * __Render Mode__               render only input meshes marked as points or everything
+ *   * __Clip Mode__                 defines the handling of spheres clipped at the camera
+ *   * __Clip Surface Adjustment__   brighten/darken glyph color on clip surface
+ *   * __Shaded Clipped Area__       enable illumination computations for the clipped surface
```
ResearchDaniel: Shaded -> Shade

modules/meshrenderinggl/src/processors/sphererasterizer.cpp
```diff
@@ -0,0 +1,264 @@
...
+
+    addPort(inport_);
+    addPort(outport_);
+
+    clipping_.addProperties(clipMode_, clipShadingFactor_, shadeClippedArea_);
+
+    sphereProperties_.addProperties(forceRadius_, defaultRadius_, forceColor_, defaultColor_,
+                                    useMetaColor_, metaColor_);
+
+    addProperties(renderMode_, forceOpaque_, useUniformAlpha_, uniformAlpha_, sphereProperties_,
```
ResearchDaniel, line 120: Should properties be hidden if corresponding forceX is false? Do they still have any effect if force is false?

modules/opencl/src/utilcl.cpp
```diff
@@ -0,0 +1,48 @@
...
+namespace utilcl {
+
+LightParameters fromLightingState(const LightingState& state) {
+    return {vec4(state.position, 1.f),
+            vec4(state.ambient, 1.f),
+            vec4(state.diffuse, 1.f),
+            vec4(state.specular, 1.f),
+            state.exponent,
+            state.shadingMode,
+            0};
```
ResearchDaniel, line 43: 0 is the padding? Might want to add an explanation

ResearchDaniel (dismissed):
Looks good. Some minor comments

modules/opencl/src/utilcl.cpp
martinfalk: yes, `0` is the initialization of the padding, which is defined as
```c++
    char padding[56];  // Align to power of two bytes (128)
```
The initialization of a struct using `{}` requires all members to be set. And since the padding is not really used it is set to `0`.

modules/meshrenderinggl/src/processors/sphererasterizer.cpp
martinfalk: I really dislike properties appearing and disappearing. Occasionally, I also prefer to be able to modify those in advance (radius, default color, ...) before enforcing them.

modules/meshrenderinggl/glsl/oit-sphereglyph.frag
martinfalk: dates back four years...
https://github.com/inviwo/inviwo/blob/99505b5159a6e5e44ef99f132d685ae6e9ded6e3/modules/basegl/glsl/sphereglyph.frag#L41-L45

modules/meshrenderinggl/src/processors/sphererasterizer.cpp
ResearchDaniel: I really dislike changing properties that do not have any effect :)
I think that ease of use is most important and hiding irrelevant information helps to reach that goal.

modules/meshrenderinggl/src/processors/sphererasterizer.cpp
martinfalk: good catch

modules/meshrenderinggl/glsl/oit-sphereglyph.frag
petersteneteg: No reason to stick to the "wrong" way in new code

modules/meshrenderinggl/src/processors/sphererasterizer.cpp
martinfalk: Hidden properties have the quirk of not being immediately shown in the link editor though.
Alternative: mark them as read only. As far as I know we have all three approaches in place, which should be unified at some point ;-)

ResearchDaniel: Hidden in the Link editor sounds like a bug with an easy fix

martinfalk: invisible == hidden (and not updated while the link editor is open #1136)

petersteneteg: I also think hiding and showing properties is a bad practice. It makes it quite confusing when a property at one point is there and the next it is gone. If you want to hide some properties out then in a composite so there is a consistent and logical way of finding them all the time. One might maybe make them read only when not used. But Martin's usecase of configuring before enabling is also very useful. I also think the boolcomposite can be used in many cases like this

ResearchDaniel (dismissed):
Nice!

modules/meshrenderinggl/src/processors/sphererasterizer.cpp
```diff
@@ -0,0 +1,261 @@
...
+                                                  clipMode_.get() == GlyphClippingMode::Discard);
+
+    configureOITShader(shader);
+    shader.build();
+}
+
+void SphereRasterizer::configureOITShader(Shader& shader) {
+    auto fso = shader.getFragmentShaderObject();
+
+    fso->addShaderExtension("GL_NV_gpu_shader5", true);
```
petersteneteg: Lots of extensions here, what
If they are not supported what if it's not on nvidia? Should one not test if the extension is available first?

martinfalk: As far as I understood it, the rasterizer has an internal state whether the A-buffer stuff is supported.

petersteneteg: But It's your code that's setting the extensions, that does not make any sense

martinfalk: should be addressed now

petersteneteg (dismissed):

petersteneteg (approved):

martinfalk:
rebased into #1159 to get rid of superfluous regression test.

PR #1157, opened 2021-05-09
Fix/nifti compressed files

ResearchDaniel:
Fix bug causing .nii.gz files to not find its associated data reader due to their two dots in the extension. 
We now support extensions with two dots, .eg. filename.xyz.zip, which seem to occur for compressed files sometimes.

Because it is hard to find a general solution to finding extensions with multiple dots we instead pass in the entire file and compare its ending with the supported file extensions. The changes are backwards compatible.

### DIFF
--- a/include/inviwo/core/io/datareaderfactory.h
+++ b/include/inviwo/core/io/datareaderfactory.h
@@ -51,8 +51,8 @@ public:
     bool registerObject(DataReader* reader);
     bool unRegisterObject(DataReader* reader);
     virtual std::unique_ptr<DataReader> create(const FileExtension& key) const override;
-    virtual std::unique_ptr<DataReader> create(const std::string& key) const;
-    virtual bool hasKey(const std::string& key) const;
+    virtual std::unique_ptr<DataReader> create(std::string_view key) const;
+    virtual bool hasKey(std::string_view key) const;
     virtual bool hasKey(const FileExtension& key) const override;
 
     template <typename T>
--- a/include/inviwo/core/io/datareaderfactory.h
+++ b/include/inviwo/core/io/datareaderfactory.h
@@ -58,7 +58,8 @@ public:
     /**
      * \brief Return a reader matching the file extension of DataReader of type T.
      * Does case insensitive comparison between the last part of filePathOrExtension and each
-     * registered extension.
+     * registered extension. Does not check for "." before the extension, so it is valid to pass for
+     * example "png".
      * @param filePathOrExtension Path to file, or simply the extension.
      * @return First available DataReaderType<T> if found, nullptr otherwise.
      */
@@ -115,7 +116,7 @@ template <typename T>
 std::unique_ptr<DataReaderType<T>> DataReaderFactory::getReaderForTypeAndExtension(
     std::string_view path) const {
     for (auto& elem : map_) {
-        if (iCaseEndsWith(path, elem.first.extension_)) {
+        if (util::iCaseEndsWith(path, elem.first.extension_)) {
             if (auto r = dynamic_cast<DataReaderType<T>*>(elem.second)) {
                 return std::unique_ptr<DataReaderType<T>>(r->clone());
             }
--- a/include/inviwo/core/io/datareaderfactory.h
+++ b/include/inviwo/core/io/datareaderfactory.h
@@ -42,7 +42,8 @@ namespace inviwo {
 
 class IVW_CORE_API DataReaderFactory : public Factory<DataReader, const FileExtension&> {
 public:
-    using Map = std::unordered_map<FileExtension, DataReader*>;
+    using Map = std::map<FileExtension, DataReader*,
+                         std::function<bool(const FileExtension&, const FileExtension&)>>;
 
     DataReaderFactory() = default;
     virtual ~DataReaderFactory() = default;
@@ -57,29 +58,46 @@ public:
     template <typename T>
     std::vector<FileExtension> getExtensionsForType() const;
 
+    /**
+     * \brief Return a reader matching the file extension of DataReader of type T.
+     * Does case insensive comparison between the last part of filePathOrExtension and each
+     * registered extension.
+     * @param filePathOrExtension Path to file, or simply the extension.
+     * @return First available DataReaderType<T> if found, nullptr otherwise.
+     */
     template <typename T>
-    std::unique_ptr<DataReaderType<T>> getReaderForTypeAndExtension(const std::string& ext) const;
+    std::unique_ptr<DataReaderType<T>> getReaderForTypeAndExtension(
+        std::string_view filePathOrExtension) const;
     template <typename T>
     std::unique_ptr<DataReaderType<T>> getReaderForTypeAndExtension(const FileExtension& ext) const;
 
     /**
      * First look for a reader using the FileExtension ext, and if no reader was found look for a
-     * reader using the fallbackExt. This is often used with a file open dialog, where the dialog
-     * will have a selectedFileExtension that will be used as ext, and the fallbackExt is taken from
-     * the file to be opened. This way any selected reader will have priority over the file
-     * extension.
+     * reader using the fallbackFilePathOrExtension. This is often used with a file open dialog,
+     * where the dialog will have a selectedFileExtension that will be used as ext, and the
+     * fallbackFilePathOrExtension is taken from the file to be opened. This way any selected reader
+     * will have priority over the file extension.
      */
     template <typename T>
     std::unique_ptr<DataReaderType<T>> getReaderForTypeAndExtension(
-        const FileExtension& ext, const std::string& fallbackExt) const;
+        const FileExtension& ext, std::string_view fallbackFilePathOrExtension) const;
 
     template <typename T>
-    bool hasReaderForTypeAndExtension(const std::string& ext) const;
+    bool hasReaderForTypeAndExtension(std::string_view filePathOrExtension) const;
     template <typename T>
     bool hasReaderForTypeAndExtension(const FileExtension& ext) const;
 
 protected:
-    Map map_;
+    // Sort extensions by length to first compare long extensions. Avoids selecting extension xyz in
+    // case xyzw exist, see getReaderForTypeAndExtension
+    Map map_ = Map([](const auto& a, const auto& b) -> bool {
+        if (a.extension_.size() != b.extension_.size()) {
+            return a.extension_.size() > b.extension_.size();
+        } else {
+            // Order does not matter
+            return a < b;
+        }
+    });
 };
 
 template <typename T>
@@ -96,11 +114,12 @@ std::vector<FileExtension> DataReaderFactory::getExtensionsForType() const {
 
 template <typename T>
 std::unique_ptr<DataReaderType<T>> DataReaderFactory::getReaderForTypeAndExtension(
-    const std::string& ext) const {
-
-    auto lkey = toLower(ext);
+    std::string_view path) const {
     for (auto& elem : map_) {
-        if (toLower(elem.first.extension_) == lkey) {
+        if (path.size() >= elem.first.extension_.size() &&
+            // Compare last part of path with the extension
+            iCaseCmp(path.substr(path.size() - elem.first.extension_.size()),
+                     elem.first.extension_)) {
             if (auto r = dynamic_cast<DataReaderType<T>*>(elem.second)) {
                 return std::unique_ptr<DataReaderType<T>>(r->clone());
             }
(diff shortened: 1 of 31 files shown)


include/inviwo/core/io/datareaderfactory.h
```diff
@@ -42,44 +42,62 @@ namespace inviwo {
...
+    virtual std::unique_ptr<DataReader> create(std::string_view key) const;
+    virtual bool hasKey(std::string_view key) const;
     virtual bool hasKey(const FileExtension& key) const override;
 
     template <typename T>
     std::vector<FileExtension> getExtensionsForType() const;
 
+    /**
+     * \brief Return a reader matching the file extension of DataReader of type T.
+     * Does case insensive comparison between the last part of filePathOrExtension and each
```
martinfalk: typo "insensitive"

martinfalk (dismissed):
Looking good from what I can see.

include/inviwo/core/io/datareaderfactory.h
```diff
@@ -42,44 +42,62 @@ namespace inviwo {
...
-    bool hasReaderForTypeAndExtension(const std::string& ext) const;
+    bool hasReaderForTypeAndExtension(std::string_view filePathOrExtension) const;
     template <typename T>
     bool hasReaderForTypeAndExtension(const FileExtension& ext) const;
 
 protected:
-    Map map_;
+    // Sort extensions by length to first compare long extensions. Avoids selecting extension xyz in
+    // case xyzw exist, see getReaderForTypeAndExtension
+    Map map_ = Map([](const auto& a, const auto& b) -> bool {
```
petersteneteg: Will this not change the order of all the readers in the dropdown in Inviwo?

petersteneteg: Also I find this solution a bit weird, I think it would be much cleaner to handle all this at the match stage.

include/inviwo/core/io/datareaderfactory.h
```diff
@@ -96,11 +114,12 @@ std::vector<FileExtension> DataReaderFactory::getExtensionsForType() const {
...
-    const std::string& ext) const {
-
-    auto lkey = toLower(ext);
+    std::string_view path) const {
     for (auto& elem : map_) {
-        if (toLower(elem.first.extension_) == lkey) {
+        if (path.size() >= elem.first.extension_.size() &&
+            // Compare last part of path with the extension
+            iCaseCmp(path.substr(path.size() - elem.first.extension_.size()),
+                     elem.first.extension_)) {
```
petersteneteg: This should probably be a free function called something like ends_with

petersteneteg: Also think you should handle finding the longest match here explicitly

include/inviwo/core/io/datawriterfactory.h
```diff
@@ -45,45 +45,62 @@ class DataWriterType;
 
 class IVW_CORE_API DataWriterFactory : public Factory<DataWriter, const FileExtension&> {
 public:
-    using Map = std::unordered_map<FileExtension, DataWriter*>;
+    using Map = std::map<FileExtension, DataWriter*,
+                         std::function<bool(const FileExtension&, const FileExtension&)>>;
```
petersteneteg: Not that it matters here. But in principle you should really avoid using std::function here since it will be quite slow. just use decltype of the lambda you want to pass in instead. Then it can get inlined and will not have to use virtual functions...

include/inviwo/core/io/datawriterfactory.h
```diff
@@ -45,45 +45,62 @@ class DataWriterType;
...
-    std::unique_ptr<DataWriterType<T>> getWriterForTypeAndExtension(const FileExtension& ext,
-                                                                    const std::string& fallbackExt);
+    std::unique_ptr<DataWriterType<T>> getWriterForTypeAndExtension(
+        const FileExtension& ext, std::string_view fallbackFilePathOrExtension) const;
 
 protected:
-    Map map_;
+    // Sort extensions by length to first compare long extensions. Avoids selecting extension xyz in
+    // case xyzw exist, see getReaderForTypeAndExtension
+    Map map_ = Map([](const auto& a, const auto& b) -> bool {
```
petersteneteg: same as above...

include/inviwo/core/io/datawriterfactory.h
```diff
@@ -96,22 +113,23 @@ std::vector<FileExtension> DataWriterFactory::getExtensionsForType() {
 
 template <typename T>
 std::unique_ptr<DataWriterType<T>> DataWriterFactory::getWriterForTypeAndExtension(
-    const std::string& ext) {
-    auto lkey = toLower(ext);
+    std::string_view path) const {
     for (auto& elem : map_) {
-        if (toLower(elem.first.extension_) == lkey) {
+        if (path.size() >= elem.first.extension_.size() &&
```
petersteneteg: same as above.

include/inviwo/core/rendering/datavisualizermanager.h
```diff
@@ -72,9 +72,11 @@ class IVW_CORE_API DataVisualizerManager {
     std::vector<FileExtension> getSupportedFileExtensions() const;
 
     /**
-     * Get a list of Data Visualizers supporting the supplied extension.
+     * \brief Get a list of Data Visualizers supporting the file extension of the supplied string.
+     * @param filePathOrExtension Path to file, or simply the extension.
      */
-    std::vector<DataVisualizer*> getDataVisualizersForExtension(const std::string& ext) const;
+    std::vector<DataVisualizer*> getDataVisualizersForExtension(
```
petersteneteg: NIT: the function name is a bit of now... But I guess that is a bit harder to change. You could consider just adding a new function that does the same think but with a better name? getDataVisualizersForFile or so

modules/base/src/io/datvolumesequencereader.cpp
```diff
@@ -73,7 +73,6 @@ std::shared_ptr<DatVolumeSequenceReader::VolumeSequence> DatVolumeSequenceReader
     }
 
     std::string fileDirectory = filesystem::getFileDirectory(fileName);
-    std::string fileExtension = filesystem::getFileExtension(fileName);
```
petersteneteg, line 76: unused?

src/core/util/fileextension.cpp
```diff
@@ -98,8 +98,12 @@ bool FileExtension::matches(std::string_view str) const {
     if (extension_.empty()) {
         return true;  // wildcard '*' matches everything
     }
-
-    return iCaseCmp(filesystem::getFileExtension(str), extension_);
+    if (str.size() >= extension_.size()) {
+        // Compare last part of path with the extension
+        return iCaseCmp(str.substr(str.size() - extension_.size()), extension_);
```
petersteneteg: use ends_with, does extension contain the "." or will this match abcivw the the extension ivw?

petersteneteg (changes requested):

include/inviwo/core/io/datareaderfactory.h
ResearchDaniel: > Will this not change the order of all the readers in the dropdown in Inviwo?

No, it was an unordered map before. 

> Also I find this solution a bit weird, I think it would be much cleaner to handle all this at the match stage.

Agree, it would be cleaner but it would also be unnecessarily expensive. Would need to go through all readers during retrieval or sort them each time. Not sure it is worth the extra implementation time

include/inviwo/core/io/datawriterfactory.h
ResearchDaniel: Tricky to use decltype because the lambda needs to be defined before. Hopefully, the compiler can optimize this.
Refactored to not expose the map publicly.

include/inviwo/core/rendering/datavisualizermanager.h
ResearchDaniel: I think that it is only used in one or two places. I changed started changing it at one point to exactly your suggestion, but thought that it might be good to keep backwardscompatibility and consistency with the DataReader/DataWriter

modules/base/src/io/datvolumesequencereader.cpp
ResearchDaniel: yes

src/core/util/fileextension.cpp
ResearchDaniel: Extension does not contain "." (same as before) unless there is a dot in the extension (nii.gz).

>  will this match abcivw the the extension ivw?

It will. I can add that as a caveat. because we never included the dot it is not possible (without major changes) to separate these. I don't see this as an issue as basically all files have an extension

include/inviwo/core/io/datareaderfactory.h
ResearchDaniel: Added new function iCaseEndsWith and refactored code

include/inviwo/core/io/datareaderfactory.h
```diff
@@ -42,44 +42,62 @@ namespace inviwo {
...
+    // Sort extensions by length to first compare long extensions. Avoids selecting extension xyz in
+    // case xyzw exist, see getReaderForTypeAndExtension
+    Map map_ = Map([](const auto& a, const auto& b) -> bool {
+        if (a.extension_.size() != b.extension_.size()) {
+            return a.extension_.size() > b.extension_.size();
+        } else {
+            // Order does not matter
+            return a < b;
+        }
+    });
```
petersteneteg:
```suggestion
    static constexpr comp = [](const auto& a, const auto& b) -> bool {
        if (a.extension_.size() != b.extension_.size()) {
            return a.extension_.size() > b.extension_.size();
        } else {
            // Order does not matter
            return a < b;
        }
    };
    std::map<FileExtension, DataReader*, decltype(comp)> map_{comp};
```

that should be enough, the compiler will have a very hard time optimizing away the std::function...

include/inviwo/core/io/datareaderfactory.h
petersteneteg: you only have to select the longest matching one, and there should be very few of those..  but fine as long as the order is not mixed up. Do we sort the list in the GUI instead then?

include/inviwo/core/util/stringconversion.h
```diff
@@ -329,6 +329,15 @@ struct IVW_CORE_API CaseInsensitiveCompare {
...
 };
 
+
+/*
+ * \brief Checks if provided string ends with suffix using case insensitive equal comparison.
+ * @param str string to check last part of. Allowed to be smaller than suffix.
+ * @param suffix Ending to match.
+ * @return True if last part of str is equal to suffix, false otherwise.
+ */
+IVW_CORE_API bool iCaseEndsWith(std::string_view str, std::string_view suffix);
```
petersteneteg: I try to put all the new functions here into the util namespace.

include/inviwo/core/io/datareaderfactory.h
ResearchDaniel: Yes, the list is sorted in the GUI. I checked that because I had the same thought.

include/inviwo/core/io/datareaderfactory.h
ResearchDaniel: Tried variations of this but could not get it to compile

include/inviwo/core/util/stringconversion.h
ResearchDaniel: makes sense, moving the function

include/inviwo/core/io/datareaderfactory.h
martinfalk: why not just using a regular functor for comparing `FileExtension`?

src/core/util/fileextension.cpp
petersteneteg: Buy the old match required the dot right? You can just check that char...

ResearchDaniel: No, the old one is the same

petersteneteg: How? Get file extension will require there to be a dot or else it will be empty, hence no match, what am I missing?

petersteneteg (dismissed):
Nice refactoring, handling file extensions in a generic fashion is hard...

src/core/util/fileextension.cpp
ResearchDaniel: One could say that the answer is yes and no. The function is used in two different ways. The first case is getting all readers by passing in only the extension (e.g. png) without knowledge about the actual file. The second case has access to the file path and here we extracted the extension before. I removed the extension extraction part.

Come to think of it, a solution would be to split into two functions. getReadersForFile and getReadersForExtension

petersteneteg: Ahh ic, yes that seems reasonable

martinfalk (approved):

PR #1159, opened 2021-05-25
Feature/rasterizer2

martinfalk:


petersteneteg (dismissed):

modules/webbrowser/tests/regression/web-property-sync/web-property-sync.inv
```diff
@@ -517,69 +606,24 @@
...
-                                <Properties>
-                                    <Property type="org.inviwo.OptionPropertyEnumInt" identifier="stippleMode" />
-                                    <Property type="org.inviwo.FloatProperty" identifier="stippleLen" />
-                                    <Property type="org.inviwo.FloatProperty" identifier="stippleSpacing" />
-                                    <Property type="org.inviwo.FloatProperty" identifier="stippleOffset" />
-                                    <Property type="org.inviwo.FloatProperty" identifier="stippleWorldScale">
-                                        <visible content="0" />
-                                    </Property>
-                                </Properties>
+                                <value content="2" />
```
ResearchDaniel: I think that this is what causes regression to fail. The script expects the value to be != 2 at startup. 
```
# Line width will be set to two when the webpage has loaded
while (lineWidth.value != 2): 
    inviwopy.qt.update();
```

martinfalk: How would this apply to the other test with a static page? There is no dependency on any property.

ResearchDaniel: I tried to get it to work with the isLoading_ property, but it was too complicated (it goes from not loading | loading | not loading, e.g. false | true | false)

martinfalk: That solved it for the property sync test, but not the other one.

martinfalk:
Rendering a plain website without any properties captures only a non-initialized canvas.
image

ResearchDaniel:
Unfortunately no way of handling that right now because the page is rendered in parallel in a separate process and we generally don’t know when it is done.

Currently, we need a script that makes sure that the regression test saves its canvas when it the page has been loaded and finished rendering

modules/webbrowser/tests/regression/web-property-sync/web-property-sync.inv
martinfalk: `CefLoadHandler` has a `OnLoadEnd` function when done loading the frame. Could we use this mechanism?

[Cef docs](https://magpcss.org/ceforum/apidocs3/projects/(default)/CefLoadHandler.html#OnLoadEnd(CefRefPtr<CefBrowser>,CefRefPtr<CefFrame>,int))

martinfalk:
It appears that during regression tests, we need to give CEF enough time to render the page before triggering another network evaluation.
This order seems to work for me:
```python
time.sleep(5);
inviwopy.qt.update();
canvas = network.Canvas;
ivw.regression.saveCanvas(canvas, "Canvas");
```

This works only on my machine when testing locally. I am considering to remove that test case once again...

petersteneteg (approved):

modules/webbrowser/src/processors/webbrowserprocessor.cpp
```diff
@@ -86,7 +86,10 @@ WebBrowserProcessor::WebBrowserProcessor(InviwoApplication* app)
     };
     updateVisibility();
 
-    auto reload = [this]() { browser_->GetMainFrame()->LoadURL(getSource()); };
+    auto reload = [this]() {
+        isLoading_ = true;
```
ResearchDaniel, line 90: If the state changing does not work as expected we should add a comment that this is a workaround

modules/webbrowser/tests/regression/web-property-sync/script.py
```diff
@@ -9,7 +9,7 @@
 while (lineWidth.value != 2): 
     inviwopy.qt.update();
 # HACK: Wait for as little as possible while ensuring that the webpage has re-rendered. 
-time.sleep(0.1);
+time.sleep(1.5);
```
ResearchDaniel, line 12: Don’t forget to update this script as well :)

modules/webbrowser/tests/regression/webbrowser/script.py
```diff
@@ -0,0 +1,13 @@
+# Inviwo Python script 
+import inviwopy
+import ivw.regression
+import time
+
+network = inviwopy.app.network;
+isLoading = network.Webbrowser.isLoading;
+# HACK: Wait for as little as possible while ensuring that the webpage has re-rendered. 
+while isLoading.value:
+	time.sleep(0.5);
```
ResearchDaniel, line 10: Could probably decrease this interval now

ResearchDaniel (changes requested):
Looks good, but one of the regression test scripts should be updated

modules/webbrowser/tests/regression/web-property-sync/script.py
martinfalk: I thought the purpose of this test is to check whether the property actually changes. So I would not change this one to wait for `isLoading`

modules/webbrowser/tests/regression/webbrowser/script.py
martinfalk: This was chosen to be on the safe side. Redraw interval of CEF (1/30s) + one or more network evaluations.

ResearchDaniel (approved):

modules/webbrowser/src/processors/webbrowserprocessor.cpp
martinfalk: Another thing coming to my mind: CEF handles that stuff asynchronously meaning that the OnLoadingStateChange callback can potentially be very late in contrast to setting the property directly.

ResearchDaniel: As long as it can be guaranteed that the onloadingstatechanged function is eventually called it should be fine. Otherwise, this change might make the processor indicate that it is loading even though it never was

martinfalk: why? `reload()` is triggered, hence the state change. If CEF decides that's not worth a reload the loading state change callback should reflect this. And in case it doesn't then the processor is still officially in a "loading" state.

PR #1160, opened 2021-05-25
Added an option in ImageMixer to blend images based on depth information

eriksunden:
Off by default, so processor works as before by default.

**Always**

**New features**
- [x] Code / Processors follows the [Inviwo documentation guidelines](https://github.com/inviwo/inviwo/wiki/Documentation-guide)
- [x] Code / processors are tested in [unittest and/or regresion tests](https://github.com/inviwo/inviwo/wiki/Manual#testing)

**Updated feature / Fixes** 
- [x] Relevant documentation has been updated

- Operating system
- Compiler/IDE
- QT/CMake/Python versions

### DIFF
--- a/modules/basegl/glsl/img_mix.frag
+++ b/modules/basegl/glsl/img_mix.frag
@@ -114,11 +114,40 @@ vec4 brightenOnly(vec4 colorA, vec4 colorB) {
     return vec4(max(colorA.rgb, colorB.rgb), max(colorA.a,colorB.a));
 }
 
-
 void main() {
     vec2 texCoords = gl_FragCoord.xy * outportParameters.reciprocalDimensions;
+
     vec4 color0 = texture(inport0Color, texCoords);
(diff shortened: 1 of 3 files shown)


petersteneteg:
Don't we hade a different processer for this like, image composite or
something @martinfalk probably knows which one

ResearchDaniel:
Yes, the composite blends based on depth and also deals with some corner cases. However, the composite only supports one type of blending. Not sure if the other blend types in ImageMixer need to be sorted based on depth though. 
"Blend based on depth" is a bit vague. Perhaps "Order by depth" in case composite is not enough. Also, in case this is needed we need to update the docs in ImageMixer.h

eriksunden:
Using ImageComposite is fine.... just very confusing to have to processors which "mix" or "composite" images, and no it seems there is no documentation on what ImageComposite does...

PR #1162, opened 2021-05-28
README: Add links to repositories using Inviwo

ResearchDaniel:


petersteneteg (dismissed):
Nice! Should add the same for the at least some of the core modules I think. Might also include some images if possible. Also the same or more should be on the website I think

martinfalk (approved):

PR #1166, opened 2021-07-13
Fix/nifti imageslicegl

ResearchDaniel:
Fixes for out-of-bounds in ImageSliceGL and reading of Nifti sequences with varying data ranges

### DIFF
--- a/modules/nifti/src/niftireader.cpp
+++ b/modules/nifti/src/niftireader.cpp
@@ -366,35 +366,35 @@ std::shared_ptr<NiftiReader::VolumeSequence> NiftiReader::readData(
         // In other words:
         // The minimum value will correspond to 0 and the maximum will correspond to 1 in the
         // Transfer Function
-        auto volRAM = volumes->front()->getRepresentation<VolumeRAM>();
-        auto minmax = util::volumeMinMax(volRAM);
-        // minmax always have four components, unused components are set to zero.
-        // Hence, only consider components used by the data format
-        dvec2 dataRange(minmax.first[0], minmax.second[0]);
-        // min/max of all components
-        for (size_t component = 1; component < format->getComponents(); ++component) {
-            dataRange = dvec2(glm::min(dataRange[0], minmax.first[component]),
-                              glm::max(dataRange[1], minmax.second[component]));
+        for (auto& vol : *volumes) {
+
+            auto volRAM = vol->getRepresentation<VolumeRAM>();
+            auto minmax = util::volumeMinMax(volRAM);
+            // minmax always have four components, unused components are set to zero.
+            // Hence, only consider components used by the data format
+            dvec2 dataRange(minmax.first[0], minmax.second[0]);
+            // min/max of all components
+            for (size_t component = 1; component < format->getComponents(); ++component) {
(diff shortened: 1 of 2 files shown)


modules/nifti/src/niftireader.cpp
```diff
@@ -343,37 +343,37 @@ std::shared_ptr<NiftiReader::VolumeSequence> NiftiReader::readData(const std::st
...
+                // where x = stored voxel value and y = "true" voxel value
+                dm.valueRange.x = static_cast<double>(niftiImage->scl_slope) * dm.dataRange.x +
+                                  static_cast<double>(niftiImage->scl_inter);
+                dm.valueRange.y = static_cast<double>(niftiImage->scl_slope) * dm.dataRange.y +
+                                  static_cast<double>(niftiImage->scl_inter);
+
+            } else {
+                dm.valueRange = dm.dataRange;
+            }
+            vol->dataMap_ = dm;
```
petersteneteg, line 396: Setting different min max for each volume in the sequence will make it impossible to compare volumes in the sequence. And will probably lead to a lot of confusion... unless we have absolute tf this is problematic

petersteneteg (changes requested):
If you want this behavior we need to explain that more I think... would be better to use the global min max in the sequence then. This  can also be very very memory expensive to calculate. You will probably run out of memory in many cases.

ResearchDaniel:
Good points. If I recall correctly I had issues with getting normalized values outside of [0 1] and thus strange visual results. A global min/max would resolve the issues as well and as you say make it possible to compare them when mapped through a TF.


For future record I made a small numerical example to check if it is correct to use different data/value ranges than the min/max of a specific volume.

 The following mapping should be applied to the nifti-input value ‘x’ to get what we refer to as the value range:

`y = scl_slope  x + scl_inter`

Let’s assume that you have two volumes with values in ranges [min1 max1], [min2, max2], ‘slc_slope = 0.5’, and that min1 and max1 are used for the ranges. Assume min1 = 1 and max1 = 2. 
Then value range will be [0.5 1].

we have that:
```
v = static_cast<double>(val) - dataRange.x) /
                                  (dataRange.y - dataRange.x) * (valueRange.y - valueRange.x) +
                              valueRange.x)
```

For value min2=-1 we should have v = -0.5

`v’ = (-1 - 1) / (2-1) * (1 - 0.5) + 0.5 = -0.5 `

petersteneteg:
I don't understand your example. Are you saying it is correct or not? 

Our dataRange is used to construct a linear map, that maps the interval [dataRange.min, dataRange.max] -> [0, 1]
This transform is what is used in the ray caster to map a value in the texture to a value in [0,1] to lookup in the TF. 

The value range is then used to map the range [0,1] -> [valueRange.min, valueRange.max] which I usually just shown to the user, in the TF dialog for example. By the fault this mapping will be the inverse of the dataRange mapping. Hence the user will be presented with the same numbers that is in the volumeRAM array/ texture.

I don't know what the interpretation of scl_slope and scl_inter is at what point do we expect to apply that mapping? 
Or do we expect that transformation to bring the data into some specific range?

PR #1170, opened 2021-08-21
Cleanup fixing includes and include guards

petersteneteg:


ResearchDaniel (approved):
Looks good from what I can see!
Would be good to update the changelog with hints on how to apply these changes to your own repository

petersteneteg:
Thanks for the review! so many files...
The include is just a global search and replace with nothing. and then fixing the include errors that pop up when building
for the guards there is a script in the tools refactoring folder, but it is not perfect and require some handholding. It will not be able to remove all the guards if they are not formated as expected. But it will print a message.

PR #1174, opened 2021-08-31
Fix dependency settings for freetype

PetiteViking:
CMake's `set` and `option`methods don't seem to play ball to the effect that setting i.e. `FT_WITH_BZIP2` to false in the GUI had no effect in the configuration process.
The changes seems to fix that.
I have additionally disabled bzip2 by default since it has not been a required dependency before either.


petersteneteg:
Hi
Sorry, for the late reply, been busy.
The interaction with regular variables and cache variables in cmake is a bit complex, and managed by some policies. 
In this case CMP0077 comes into play. We don't want to introduce any options for there settings, because we don't handle changing their values properly anyway. Annoying things is the the freetype submodule uses a ancient CMake version. I have a workaround. Hope to push that soon...

petersteneteg:
This should be fixed in master now. If not let me know. Thanks for spotting the issue!

PR #1179, opened 2021-09-06
Animation: Follow best practices and return const std::string& instead of string_view

ResearchDaniel:
Better to keep information about std::string...

### DIFF
--- a/modules/animation/include/modules/animation/datastructures/animation.h
+++ b/modules/animation/include/modules/animation/datastructures/animation.h
@@ -186,7 +186,7 @@ public:
     /**
      * Return the name of the Animation. Used for display in the GUI.
      */
-    std::string_view getName() const;
+    const std::string& getName() const;
     /**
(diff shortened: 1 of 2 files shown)


petersteneteg:
I have been thinking about this one a bit but could not find any recommendations. Do you have any ref for that best practice?

ResearchDaniel:
Converting to string_view looses info about ensured end line, which may be used by some algorithms. I read this on stack overflow (best practice then being avoid loosing information when there are no benefits)

petersteneteg:
That sounds like a good argument. The downside is of course that it forces the use of a std::string to return. For string view one can return a literal without having to create any string. In this case there is a string anyway, so it does not matter, but in some virtual functions the overriding will just return a literal. Hence for virtual functions I think a string view is better

petersteneteg (approved):

ResearchDaniel:
Agree. The recommendation was to return a string in case that was used internally anyway.

PR #1187, opened 2021-09-15
Core: Make ListProperty invalidate when adding/removing properties

ResearchDaniel:
Calling propertyModified in the ListProperty will have the intended effect because new sub properties will be in a valid state. Calling invalidate with the Property invalidation level will cause the network to be re-evaluated.


martinfalk (commented):
From my side OK. Waiting for @petersteneteg's comments.

petersteneteg (approved):
LGTM

PR #1193, opened 2021-09-21
Feature/webbrowser update

martinfalk:
* added system settings for adjusting the refresh rate (default 30Hz)
* fixed occasional crashes when using the mouse wheel
* updated CEF which required minor fixes (https://bitbucket.org/chromiumembedded/cef/issues/3185/inconsistent-namespace-usage-in)

### DIFF
--- a/modules/webbrowser/include/modules/webbrowser/webbrowsersettings.h
+++ b/modules/webbrowser/include/modules/webbrowser/webbrowsersettings.h
@@ -38,7 +38,7 @@ class IVW_MODULE_WEBBROWSER_API WebBrowserSettings : public Settings {
 public:
     WebBrowserSettings();
 
-    IntProperty refreshrate_;
+    IntProperty refreshRate_;
 };
 
 }  // namespace inviwo
--- a/modules/webbrowser/include/modules/webbrowser/webbrowsersettings.h
+++ b/modules/webbrowser/include/modules/webbrowser/webbrowsersettings.h
@@ -1,0 +1,44 @@
+/*********************************************************************************
+ *
+ * Inviwo - Interactive Visualization Workshop
+ *
+ * Copyright (c) 2021 Inviwo Foundation
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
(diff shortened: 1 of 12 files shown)


petersteneteg (dismissed):
lgtm

modules/webbrowser/include/modules/webbrowser/webbrowsersettings.h
```diff
@@ -0,0 +1,44 @@
...
+#include <inviwo/core/util/settings/settings.h>
+#include <inviwo/core/properties/ordinalproperty.h>
+
+namespace inviwo {
+
+class IVW_MODULE_WEBBROWSER_API WebBrowserSettings : public Settings {
+public:
+    WebBrowserSettings();
+
+    IntProperty refreshrate_;
```
ResearchDaniel: Camel case typo
refreshRate_

modules/webbrowser/src/webbrowsermodule.cpp
```diff
@@ -112,6 +113,14 @@ WebBrowserModule::WebBrowserModule(InviwoApplication* app)
     // Call 60 times per second
     , doChromiumWork_(Timer::Milliseconds(1000 / 60), []() { CefDoMessageLoopWork(); }) {
```
ResearchDaniel: Don’t you need to use the WebbrowserSettings default value here?

modules/webbrowser/src/webbrowsersettings.cpp
```diff
@@ -0,0 +1,45 @@
...
+ *
+ *********************************************************************************/
+
+#include <modules/webbrowser/webbrowsersettings.h>
+
+namespace inviwo {
+
+WebBrowserSettings::WebBrowserSettings()
+    : Settings("Web Browser")
+    , refreshrate_("refreshrate", "Refresh Rate (Hz)", 30, {0, ConstraintBehavior::Immutable},
```
ResearchDaniel: Either you need to treat 0 as a special case or set minimum to 1. 0 result in infinite value in the Webbrowser module

ResearchDaniel (changes requested):
Nice! Some minor things that need to be addressed. I would also like to test the branch before it is merged. Will do that ASAP

modules/webbrowser/src/webbrowsermodule.cpp
martinfalk: sure, forgot about to change it here.

modules/webbrowser/src/webbrowsersettings.cpp
martinfalk: Good catch. Was also considering refresh rates below 1 second, but then one would need to use a float property. Not sure how useful that would be.

ResearchDaniel (approved):
Also tried it on my computer without issues

PR #1195, opened 2021-09-23
Feature/animation fixes

martinfalk:
* some fixes for the animation module


ResearchDaniel (dismissed):
lgtm

modules/animation/src/animationcontroller.cpp
```diff
@@ -129,6 +129,9 @@ AnimationController::AnimationController(Animation& animation, InviwoApplication
     playOptions.setCollapsed(true);
     addProperty(playOptions);
 
+    framesPerSecond.onChange([this]() { setPlaySpeed(framesPerSecond); });
```
petersteneteg: Should not setPlaySpeed update frames per second too? Don't know why there is two members with same state...

petersteneteg (changes requested):

modules/animation/src/animationcontroller.cpp
martinfalk: `deltaTime_` is the inverse of framesPerSecond and also holds the playback direction (signed delta).

martinfalk: going for OrdinalRefProperty to hold the state only once.

modules/animation/src/animationcontroller.cpp
```diff
@@ -65,15 +65,20 @@ size_t imageExtIndex(InviwoApplication* app, std::string_view ext) {
...
                      0)
     , playWindow("PlayFirstLastTime", "Window", 0, 10, 0, 1e5, 1, 0.0,
                  InvalidationLevel::InvalidOutput, PropertySemantics::Text)
-    , framesPerSecond("PlayFramesPerSecond", "Frames per Second", 25.0, 000.1, 1000.0, 1.0,
-                      InvalidationLevel::InvalidOutput, PropertySemantics::Text)
+    , framesPerSecond(
+          "PlayFramesPerSecond", "Frames per Second",
+          [this]() { return 1.0 / std::abs(deltaTime_.count()); },
+          [this](const double& value) { deltaTime_ = Seconds(1.0 / value); },
+          {000.1, ConstraintBehavior::Immutable}, {1000.0, ConstraintBehavior::Immutable}, 1.0,
```
ResearchDaniel: 000.1 -> 0.001

petersteneteg (dismissed):

ResearchDaniel (approved):

petersteneteg (approved):

PR #1197, opened 2021-09-28
QtWidgets: dpi-aware processor widget

martinfalk:
* changes based on CanvasProcessorWidgetQt


petersteneteg (changes requested):
I need to look in more detail

PR #1198, opened 2021-10-01
BaseGL: Fix to ensure that regular event is propagated if picking eve…

ResearchDaniel:
…nt is not used

CanvasOpengGLWidget::propagateEvent will first propagate the picking event and then the regular event (if the picking event was not used). The regular event will not be propagated past the viewmanager if it is already is marked as visited when passed along with the picking event. This causes problems for processors not handling the picking, but does handle the original event. E.g., when using EventMatcher since it does not match the picking event, but matches other events. Therefore, the original event must also be propagated to processors up the chain.

### DIFF
--- a/modules/basegl/src/viewmanager.cpp
+++ b/modules/basegl/src/viewmanager.cpp
@@ -62,8 +62,8 @@ bool ViewManager::propagatePickingEvent(PickingEvent* pe, Propagator propagator)
                                pe->getPreviousGlobalPickingId(), pressNDC, previousNDC);
 
             propagator(&newPe, ind);
-            if (newPe.hasBeenUsed()) newEvent->markAsUsed();
-            for (auto p : newPe.getVisitedProcessors()) newEvent->markAsVisited(p);
+            if (newPe.hasBeenUsed()) pe->markAsUsed();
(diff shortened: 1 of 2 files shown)


modules/basegl/src/viewmanager.cpp
```diff
@@ -104,7 +102,6 @@ bool ViewManager::propagateMouseEvent(MouseEvent* me, Propagator propagator) {
         newEvent.setPosNormalized(scale * (newEvent.posNormalized() - offset));
         propagator(&newEvent, selectedView_.second);
         if (newEvent.hasBeenUsed()) me->markAsUsed();
-        for (auto p : newEvent.getVisitedProcessors()) me->markAsVisited(p);
```
petersteneteg, line 107: But If it is not a picking event but a regular mouseEvent, that will also use this function, should that not copy the visited processors in that case?

petersteneteg (changes requested):
I think i understand the issue, but what happens with regular mouse event now?

modules/basegl/src/viewmanager.cpp
ResearchDaniel: Good catch! I committed a fix for it

petersteneteg (approved):
Think one could refactor this a bit now. to make it a bit easier to understand...

PR #1199, opened 2021-10-04
Feature/bitset

martinfalk:
Added CRoaring bitmaps and exposed them in Inviwo as `inviwo::BitSet` including unit tests. 

This bitset will be used for brushing & linking coming soon to this PR.


include/inviwo/core/datastructures/bitset.h
```diff
@@ -0,0 +1,399 @@
...
+     * Reallocates the memory needed.
+     *
+     * @return number of bytes saved
+     */
+    size_t shrinkToFit();
+
+private:
+    void addMany(size_t size, const uint32_t* data);
+
+    struct IVW_CORE_API RoaringDeleter {
```
ResearchDaniel, line 389: No need to export this private struct?

src/core/datastructures/bitset.cpp
```diff
@@ -0,0 +1,268 @@
...
+#include <warn/ignore/all>
+#include <roaring.hh>
+#include <warn/pop>
+
+namespace inviwo {
+
+BitSet::BitSetIterator::~BitSetIterator() = default;
+
+BitSet::BitSetIterator::BitSetIterator(const BitSet& b, bool exhausted)
+    : it_(std::make_unique<RoaringIt>(*b.roaring_, exhausted)) {}
```
ResearchDaniel, line 42: Why does it_ need to be a unique_ptr? Can it not be a regular variable?

src/core/tests/unittests/bitset-test.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+TEST(bitset, constructors) {
+    BitSet();
+    BitSet(5);
+    BitSet(1, 2);
+    BitSet(2, 3, 5);
+
+    std::vector<uint32_t> indices = getIndices(20);
+    // BitSet(indices);
+    BitSet(indices.begin(), indices.end());
```
ResearchDaniel, line 71: Would make sense to check if correct indices have been added after construction

ResearchDaniel (dismissed):
Looks ok to me. 
Should it not be added as a submodule though?

src/core/datastructures/bitset.cpp
petersteneteg: then the roaring header would have to be in the header, now it is only forward declared

src/core/CMakeLists.txt
```diff
@@ -791,6 +792,7 @@ target_link_libraries(inviwo-core
         $<$<BOOL:${WIN32}>:inviwo::stackwalker>
         $<$<BOOL:${APPLE}>:${CORESERVICES_LIBRARY}>
         inviwo::tinydir
+        roaring
```
petersteneteg: you should link roaring::roaring

include/inviwo/core/datastructures/bitset.h
```diff
@@ -0,0 +1,399 @@
...
+class Roaring;
+class RoaringSetBitForwardIterator;
+}  // namespace roaring
+
+namespace inviwo {
+
+namespace detail {
+
+template <typename...>
+using void_t = void;
```
petersteneteg: void_t is in the std...  why not use detected?

include/inviwo/core/datastructures/bitset.h
```diff
@@ -0,0 +1,399 @@
...
+
+        std::unique_ptr<RoaringIt> it_;
+    };
+
+    BitSet();
+
+    BitSet(const std::vector<uint32_t>& values);
+
+    template <size_t N>
+    BitSet(const std::array<uint32_t, N>& values) : BitSet() {
```
petersteneteg: Why not use span<uint32_t> constructor instead?

include/inviwo/core/datastructures/bitset.h
```diff
@@ -0,0 +1,399 @@
...
+    template <typename InputIt,
+              class = typename std::enable_if_t<detail::is_iterator<InputIt>::value>>
+    explicit BitSet(InputIt begin, InputIt end) : BitSet() {
+        add(begin, end);
+    }
+
+    template <typename... uint32_ts>
+    explicit BitSet(uint32_t val, uint32_ts&&... values) : BitSet() {
+        add(val, values...);
+    }
```
petersteneteg:
```suggestion
    template <typename... Ts, 
                      typename = std::enable_if<std::conjunction<std::is_convertible<Ts, uint32_t>...>::value>::type;
    explicit BitSet(Ts&&... values) : BitSet() {
         (add(values), ...);
    }
```

include/inviwo/core/datastructures/bitset.h
```diff
@@ -0,0 +1,399 @@
...
+template <class T>
+struct is_iterator<T, void_t<typename std::iterator_traits<T>::iterator_category>>
+    : std::true_type {};
+
+}  // namespace detail
+
+/**
+ * \brief represents a bitset based on roaring bitmaps provided by the CRoaring library
+ */
+class IVW_CORE_API BitSet {
```
petersteneteg, line 55: Should we make this serializable?

include/inviwo/core/datastructures/bitset.h
```diff
@@ -0,0 +1,399 @@
...
+     * @param v  value
+     */
+    void add(uint32_t v);
+
+    void add(const std::vector<uint32_t>& values);
+
+    template <size_t N>
+    void add(const std::array<uint32_t, N>& values) {
+        addMany(values.size(), values.data());
+    }
```
petersteneteg: just use span, no need for any other containers

include/inviwo/core/datastructures/bitset.h
```diff
@@ -0,0 +1,399 @@
...
+              class = typename std::enable_if_t<detail::is_iterator<InputIt>::value>>
+    void add(InputIt begin, InputIt end) {
+        while (begin != end) {
+            add(*begin);
+            ++begin;
+        }
+    }
+
+    template <typename... Ts>
+    void add(uint32_t val, Ts&&... values) {
```
petersteneteg: see constructor

include/inviwo/core/datastructures/bitset.h
```diff
@@ -0,0 +1,399 @@
...
+
+    /**
+     * Flip all bits in the open range [\p min, \p max)
+     */
+    void flipRange(uint32_t min, uint32_t max);
+
+    /**
+     * Return the number of values that are less or equal to \p v
+     */
+    size_t rank(uint32_t v) const;
```
petersteneteg, line 243: could one use this for a faster std::distance(it1, it2) -> rank(*it2) - rank(*it1) or something like that?

src/core/datastructures/bitset.cpp
```diff
@@ -0,0 +1,268 @@
...
+
+BitSet::BitSetIterator& BitSet::BitSetIterator::operator++() {
+    it_->operator++();
+    return *this;
+}
+
+BitSet::BitSetIterator BitSet::BitSetIterator::operator++(int) {
+    BitSet::BitSetIterator it(*it_);
+    it_->operator++();
+    return it;  // <- attempting to reference deleted function BitSetIterator(const BitSetIterator&)
```
petersteneteg: ?

src/core/datastructures/bitset.cpp
```diff
@@ -0,0 +1,268 @@
...
+
+bool BitSet::BitSetIterator::operator==(const BitSetIterator& rhs) const {
+    return it_->operator==(*rhs.it_);
+}
+
+bool BitSet::BitSetIterator::operator!=(const BitSetIterator& rhs) const {
+    return it_->operator!=(*rhs.it_);
+}
+
+BitSet::BitSet()
```
petersteneteg, line 89: could also add a private conversion constructor from roaring::Roaring, would make some of the functions easier to implement I think, would avoid having to create some empty bitsets

petersteneteg (changes requested):

include/inviwo/core/datastructures/bitset.h
martinfalk: Wrong :-) There is a need for it. Otherwise the linker complains about it when trying to destruct a unique_ptr of Roaring (that one is only forward declared).

martinfalk:
> Looks ok to me. Should it not be added) as a submodule though?

This was discussed at some length and we decided against it (see also explanation in [README.md](https://github.com/inviwo/inviwo/blob/6f296062c5112d7e888698b6b59e30c3bfcc4f75/ext/CRoaring/CRoaring/README.md)

include/inviwo/core/datastructures/bitset.h
martinfalk: As it is, it is just a bitset implementation more in the sense of a data source which is not yet tied to Inviwo. Might be sensible to make it also serializable in addition to readData/writeData.

include/inviwo/core/datastructures/bitset.h
martinfalk: does span then use implicit converters for containers?

include/inviwo/core/datastructures/bitset.h
martinfalk: There is some internal function doing the computation
```c++
uint64_t Roaring::rank(uint32_t x) const {
    return api::roaring_bitmap_rank(&roaring, x);
}
```

martinfalk: Potentially yes.

src/core/datastructures/bitset.cpp
martinfalk: forgot to remove the comment.

src/core/datastructures/bitset.cpp
martinfalk: However, internally Roaring always creates an empty bitmap beforehand. Compare
```c++
    Roaring() : roaring{} {
        // The empty constructor roaring{} silences warnings from pedantic static analyzers.
        api::roaring_bitmap_init_cleared(&roaring);
    }

    /**
     * Construct a bitmap from a list of integer values.
     */
    Roaring(size_t n, const uint32_t *data) : Roaring() {
        api::roaring_bitmap_add_many(&roaring, n, data);
    }
```

src/core/CMakeLists.txt
petersteneteg: we should also make it private now that we have hidden the headers...

include/inviwo/core/datastructures/bitset.h
petersteneteg: yes, https://en.cppreference.com/w/cpp/container/span/span overload 7
our span does the some thing with without fancy concepts...

src/core/datastructures/bitset.cpp
petersteneteg: I was mostly thinking it might simplify our side a bit.

include/inviwo/core/datastructures/bitset.h
petersteneteg: otherwise we must add it to serialize.h/deserialize.h I would rather not do that if possible.

src/core/tests/unittests/bitset-test.cpp
martinfalk: This test's function was to check whether the constructors are working, independent of what the bitset internally does. See test case for `contains`.

src/core/CMakeLists.txt
martinfalk: had to add an alias for `roaring::roaring`

include/inviwo/core/datastructures/bitset.h
martinfalk: because for consecutive data like array and vector one can use `addMany` to add multiple values at once in contrast to adding each value individually.

src/core/CMakeLists.txt
petersteneteg: no, it should be there according the the docs?

include/inviwo/core/datastructures/bitset.h
petersteneteg: A span is always consecutive. It is the __perfect__ parameter for this case. It provides just the data and size methods you need.

src/core/CMakeLists.txt
martinfalk: But it isn't for some reason...

martinfalk:
superseded by #1203, #1204, and #1205

PR #1200, opened 2021-10-04
Feature/various fixes

martinfalk:


tools/regression.py
```diff
@@ -162,6 +162,10 @@ def execonf(file, build):
 	logging.getLogger().setLevel(getattr(logging, str(args.log).upper(), logging.WARN))
 
 	if args.inviwo:
+		if not os.path.exists(args.inviwo):
```
petersteneteg, line 165: Does this on also work with relative paths?

petersteneteg (approved):

tools/regression.py
martinfalk: it should.
https://docs.python.org/3/library/os.path.html#os.path.exists

PR #1204, opened 2021-10-05
Core: added base64 utility functions

martinfalk:
* https://github.com/ReneNyffenegger/cpp-base64

### DIFF
--- a/include/inviwo/core/algorithm/base64.h
+++ b/include/inviwo/core/algorithm/base64.h
@@ -33,55 +33,18 @@
 #include <string>
 #include <string_view>
 
+#include <tcb/span.hpp>
+
 namespace inviwo {
 
-namespace base64util {
+namespace util {
 
-// Copyright © 2004-2017 by René Nyffenegger
+IVW_CORE_API std::string base64_encode(util::span<char> s, bool url = false);
+IVW_CORE_API std::string base64_encode_pem(util::span<char> s);
+IVW_CORE_API std::string base64_encode_mime(util::span<char> s);
 
-// This source code is provided 'as-is', without any express or implied
-// warranty. In no event will the author be held liable for any damages
-// arising from the use of this software.
+IVW_CORE_API util::span<char> base64_decode(std::string_view s, bool remove_linebreaks = false);
 
-// Permission is granted to anyone to use this software for any purpose,
-// including commercial applications, and to alter it and redistribute it
-// freely, subject to the following restrictions:
-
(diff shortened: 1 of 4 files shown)


include/inviwo/core/algorithm/base64.h
```diff
@@ -0,0 +1,87 @@
...
+ *
+ *********************************************************************************/
+#pragma once
+
+#include <inviwo/core/common/inviwocoredefine.h>
+
+#include <string>
+#include <string_view>
+
+namespace inviwo {
```
ResearchDaniel: Might not be necessary to wrap in the inviwo namespace?

ResearchDaniel (dismissed):
LGTM

include/inviwo/core/algorithm/base64.h
petersteneteg: I think it should be in the Inviwo namespace, but don't use base64util, just util.
base64util::base64_encode is a bit too much base64. also we usually name it utilbase64.
If you want it that way I would drop the base64 from the function names.

include/inviwo/core/algorithm/base64.h
```diff
@@ -0,0 +1,87 @@
...
+// 3. This notice may not be removed or altered from any source distribution.
+
+//
+//  base64 encoding and decoding with C++.
+//  Version: 2.rc.08 (release candidate)
+//
+//  https://github.com/ReneNyffenegger/cpp-base64
+//
+
+IVW_CORE_API std::string base64_encode(std::string const& s, bool url = false);
```
petersteneteg: don't think we need the const string& versions when we have string_view ones...

include/inviwo/core/algorithm/base64.h
```diff
@@ -0,0 +1,87 @@
...
+
+IVW_CORE_API std::string base64_decode(std::string const& s, bool remove_linebreaks = false);
+IVW_CORE_API std::string base64_encode(unsigned char const*, size_t len, bool url = false);
+
+//
+// Interface with std::string_view rather than const std::string&
+// Requires C++17
+// Provided by Yannic Bonenberger (https://github.com/Yannic)
+//
+IVW_CORE_API std::string base64_encode(std::string_view s, bool url = false);
```
petersteneteg: how is the input byte array given? as a string_view? that is a bit odd... span<unsigned char> would have been better probably...

petersteneteg (changes requested):

petersteneteg (approved):
Nice! like it!

PR #1206, opened 2021-10-08
Feature/brushing highlight

martinfalk:
Highlight events for brushing and linking can be used to temporarily highlight objects, for example on hovering. Example implementation added to scatter plot and parallel coordinate plot. 

Completely reworked Brushing and Linking. Managers are now owned by BrushingAndLinkingInports and BrushingAndLinkingOutports. Brushing actions are propagated directly between managers without needing events.


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 Here we document changes that affect the public API or changes that needs to be communicated to other developers. 
 
+## 2021-10-06 Updated Brushing & Linking
```
petersteneteg: we want pictures :)

modules/brushingandlinking/include/modules/brushingandlinking/datastructures/indexlist.h
```diff
@@ -30,38 +30,45 @@
 #pragma once
 
 #include <modules/brushingandlinking/brushingandlinkingmoduledefine.h>
+#include <inviwo/core/datastructures/bitset.h>
 #include <inviwo/core/util/dispatcher.h>
 
 #include <unordered_map>
 #include <unordered_set>
```
petersteneteg: remove

modules/brushingandlinking/src/datastructures/indexlist.cpp
```diff
@@ -71,4 +74,27 @@ void IndexList::clear() {
...
+void IndexList::serialize(Serializer& s) const {
+    s.serialize(
+        "ports", indicesBySource_, "indices", {},
+        [](const BrushingAndLinkingInport* p) { return p->getPath(); }, util::identity());
+}
+
+void IndexList::deserialize(Deserializer& d, const BrushingAndLinkingOutport& port) {
+    indicesBySource_.clear();
+
+    std::unordered_map<std::string, BitSet> map;
```
petersteneteg: using IdentifiedDeserializer I think you could do this in one step...

modules/brushingandlinking/src/ports/brushingandlinkingports.cpp
```diff
@@ -38,21 +43,21 @@ BrushingAndLinkingInport::BrushingAndLinkingInport(std::string identifier)
     onConnect([&]() {
         sendFilterEvent(filterCache_);
         sendSelectionEvent(selectionCache_);
+        sendHighlightEvent(highlightCache_);
```
petersteneteg: this call probably happens after the B&L manager is deserialized, would it make more sense to serialize all the state in each inport?

modules/brushingandlinking/src/processors/brushingandlinkingprocessor.cpp
```diff
@@ -71,6 +76,9 @@ void BrushingAndLinkingProcessor::invokeEvent(Event* event) {
         } else if (dynamic_cast<SelectionEvent*>(event)) {
             manager_->setSelected(brushingEvent->getSource(), brushingEvent->getIndices());
             event->markAsUsed();
+        } else if (dynamic_cast<HighlightEvent*>(event)) {
```
petersteneteg: we have special tag in the event to avoid having to dynamic_cast here...

modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
```diff
@@ -622,13 +620,17 @@ void ParallelCoordinates::drawLines(size2_t size) {
...
             lineShader_.setUniform("fallofPower", 0.5f * falllofPower_.get());
-
-            glDrawElements(GL_LINE_STRIP, lines_.sizes[hoveredLine_], GL_UNSIGNED_INT,
-                           reinterpret_cast<GLvoid*>(
-                               Lines::indexToOffset(hoveredLine_, lines_.sizes[hoveredLine_])));
+            for (auto index : brushingAndLinking_.getHighlightedIndices()) {
+                if (index >= numLines || brushingAndLinking_.isFiltered(index)) {
+                    continue;
+                }
+                glDrawElements(
```
petersteneteg: if this can also be a set of lines just as will move it into the loop above? or are you trying to avoid the partition here?

modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
```diff
@@ -640,12 +642,13 @@ void ParallelCoordinates::linePicked(PickingEvent* p) {
         if (p->getHoverState() == PickingHoverState::Move ||
             p->getHoverState() == PickingHoverState::Enter) {
             p->setToolTip(dataframe::createToolTipForRow(*df, p->getPickedId()));
-            hoveredLine_ = static_cast<int>(p->getPickedId());
+            BitSet b;
```
petersteneteg: can't you add the indices to the constructor?

modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
```diff
@@ -659,11 +662,7 @@ void ParallelCoordinates::linePicked(PickingEvent* p) {
         auto id = p->getPickedId();
 
         auto selection = brushingAndLinking_.getSelectedIndices();
-        if (brushingAndLinking_.isSelected(indexCol[id])) {
-            selection.erase(indexCol[id]);
-        } else {
-            selection.insert(indexCol[id]);
-        }
+        selection.flip(indexCol[id]);
```
petersteneteg, line 660: nice!

petersteneteg (changes requested):
Nice work, some minor comments

modules/brushingandlinking/include/modules/brushingandlinking/ports/brushingandlinkingports.h
```diff
@@ -47,26 +45,31 @@ class IVW_MODULE_BRUSHINGANDLINKING_API BrushingAndLinkingInport
...
+    const BitSet& getFilteredIndices() const;
+    const BitSet& getHighlightedIndices() const;
+    const BitSet& getSelectedColumns() const;
 
     virtual std::string getClassIdentifier() const override;
 
-    std::unordered_set<size_t> filterCache_;
-    std::unordered_set<size_t> selectionCache_;
-    std::unordered_set<size_t> selectionColumnCache_;
+    BitSet filterCache_;
```
ResearchDaniel: Perhaps not part of this PR, but these should have been private from the beginning because they must be changed through the sendXEvent functions?

modules/brushingandlinking/src/brushingandlinkingmanager.cpp
```diff
@@ -93,16 +109,30 @@ void BrushingAndLinkingManager::clearColumns() {
...
+    s.serialize("selectedColumns", selectedColumns_);
+    filtered_.serialize(s);
+}
+
+void BrushingAndLinkingManager::deserialize(Deserializer& d,
+                                            const BrushingAndLinkingOutport& port) {
+    d.deserialize("selected", selected_);
+    d.deserialize("highlighted", highlighted_);
+    d.deserialize("selectedColumns", selectedColumns_);
+    filtered_.deserialize(d, port);  // TODO: check for double invalidation caused by filtered_
```
ResearchDaniel: Don’t forget to fix

modules/brushingandlinking/src/datastructures/indexlist.cpp
```diff
@@ -71,4 +74,27 @@ void IndexList::clear() {
     onUpdate_.invoke();
 }
 
+void IndexList::serialize(Serializer& s) const {
+    s.serialize(
+        "ports", indicesBySource_, "indices", {},
```
ResearchDaniel: Why ports here?

modules/brushingandlinking/include/modules/brushingandlinking/ports/brushingandlinkingports.h
```diff
@@ -47,26 +45,31 @@ class IVW_MODULE_BRUSHINGANDLINKING_API BrushingAndLinkingInport
...
     virtual ~BrushingAndLinkingInport() = default;
 
-    void sendFilterEvent(const std::unordered_set<size_t>& indices);
+    void sendFilterEvent(const BitSet& indices);
 
-    void sendSelectionEvent(const std::unordered_set<size_t>& indices);
+    void sendSelectionEvent(const BitSet& indices);
 
-    void sendColumnSelectionEvent(const std::unordered_set<size_t>& indices);
+    void sendHighlightEvent(const BitSet& indices);
```
ResearchDaniel: Probably a good idea to state that these should be sent when hovering

modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
```diff
@@ -57,46 +61,63 @@ class IVW_MODULE_BRUSHINGANDLINKING_API BrushingAndLinkingManager {
...
+    void serialize(Serializer& s) const;
+    void deserialize(Deserializer& d, const BrushingAndLinkingOutport& port);
 
 private:
-    std::unordered_set<size_t> selected_;
-    std::unordered_set<size_t> selectedColumns_;
+    BitSet selected_;
+    BitSet highlighted_;
+    BitSet selectedColumns_;
     IndexList filtered_;  // Use IndexList to be able to remove filtered rows on port disconnection
```
ResearchDaniel: Not completely related to this PR, but should we not treat all in the same way

modules/plottinggl/include/modules/plottinggl/plotters/persistencediagramplotgl.h
```diff
@@ -176,7 +177,7 @@ class IVW_MODULE_PLOTTINGGL_API PersistenceDiagramPlotGL {
     std::array<AxisRenderer, 2> axisRenderers_;
 
     PickingMapper picking_;
-    std::unordered_set<size_t> selectedIndices_;
+    BitSet selectedIndices_;
```
ResearchDaniel, line 180: Not part of PR, but would be nice if we could avoid having a copy of the state and use the brushing directly.

modules/plottinggl/include/modules/plottinggl/plotters/scatterplotgl.h
```diff
@@ -185,22 +190,23 @@ class IVW_MODULE_PLOTTINGGL_API ScatterPlotGL : public InteractionHandler {
     PickingMapper picking_;
     std::vector<bool> filtered_;
```
ResearchDaniel, line 191: Should not these be replaced with BitSet?

modules/plottinggl/src/plotters/scatterplotgl.cpp
```diff
@@ -429,19 +430,21 @@ void ScatterPlotGL::plot(const size2_t& dims, IndexBuffer* indexBuffer, bool use
...
+            highlightIndexGL_.upload(static_cast<const void*>(highlighted_.toVector().data()),
+                                     sizeof(uint32_t) * highlighted_.size());
+            highlightDirty_ = false;
         } else {
-            hoverIndexGL_.bind();
+            highlightIndexGL_.bind();
         }
-        glDrawElements(GL_POINTS, 1, hoverIndexGL_.getFormatType(), nullptr);
-        hoverIndexGL_.unbind();
+        glDrawElements(GL_POINTS, highlighted_.cardinality(), highlightIndexGL_.getFormatType(),
```
ResearchDaniel, line 445: Is it possible to upload the bitsets and do the logic in the vertex shader?
Would simplify the code a lot I think

modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
```diff
@@ -806,9 +803,9 @@ void ParallelCoordinates::updateBrushing() {
         }
     }
 
-    std::unordered_set<size_t> brushedID;
-    for (size_t i = 0; i < nRows; ++i) {
-        if (brushed[i]) brushedID.insert(indexCol[i]);
+    BitSet brushedID;
+    for (auto&& [selected, index] : util::zip(brushed, indexCol)) {
+        if (selected) brushedID.add(index);
```
ResearchDaniel: Not really selected…

modules/plottinggl/src/processors/scatterplotprocessor.cpp
```diff
@@ -72,14 +72,28 @@ ScatterPlotProcessor::ScatterPlotProcessor()
             p->setToolTip(dataframe::createToolTipForRow(*dataFramePort_.getData(), rowId));
         }
     });
+    highlightChangedCallBack_ =
+        scatterPlot_.addHighlightChangedCallback([this](const BitSet& highlighted) {
+            if (brushingPort_.isConnected()) {
+                BitSet indices;
+                auto iCol = dataFramePort_.getData()->getIndexColumn();
```
ResearchDaniel, line 80: Finding the indices is a pretty common operation. Should it perhaps be extracted to the data frame  or something?

ResearchDaniel (changes requested):
Nice work! There is a todo that should be fixed and some refactoring related to extracting indices from a data frame that would be good to do.


A general comment/question about brushing and linking ids:
I might be wrong, but it seems as if the id is sometimes used as a row/item index and sometimes mapped through the IndexColumn. Especially when used with picking. We probably need to double check that we are consistent and make it easy and/or clear how to use to the “right” way. This might not be related to this PR though

modules/brushingandlinking/include/modules/brushingandlinking/ports/brushingandlinkingports.h
petersteneteg: I would not say Should more like could or that it's common. There might be other scenarios where it might be useful too

modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
petersteneteg: We can't do that since they have different semantics

modules/plottinggl/include/modules/plottinggl/plotters/persistencediagramplotgl.h
petersteneteg: Can one not just direct access the cache in the inport?

modules/plottinggl/src/plotters/scatterplotgl.cpp
petersteneteg: The bitsets are compressed, so unpacking them in the shader is probably not that easy, or efficient

modules/brushingandlinking/src/datastructures/indexlist.cpp
martinfalk: for the same reason they are stored in the map of this IndexList. This has to do with how filtering works. A selection consists of just the indices set in the latest selection event. Filtering, however, restricts indices and two filter events might overlap.
For example: indices 1 to 10
- filter event A filters 1-5 (i.e. removes those indices)
- filter event B filter 3-6
This results in indices {7, .., 10} remaining.

If you now disconnect the port for event A, the filtered set should be {1, 2, 7-10}. Hence, we need to keep track of where a filter event came from. The actual filtered set is computed through a union of still active filter connections.

modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
martinfalk: see comment above on filter events

modules/plottinggl/include/modules/plottinggl/plotters/scatterplotgl.h
martinfalk: Different can of worms and independent of the brushing and linking mechanism.

modules/plottinggl/src/plotters/scatterplotgl.cpp
martinfalk: @petersteneteg is correct, unless you plan to implement the decoding from CRoaring in shaders. One alternative would be to either encode the set bits as a bit mask (selected, filtered, highlighted) in a vertex buffer (requires mesh update on brushing event) or a Shader Storage Buffer Object (requires OpenGL 4.2 but the mesh doesn't need to be updated).

modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
martinfalk: you could say it is a selected index. Open for name suggestions.

martinfalk:
> A general comment/question about brushing and linking ids:
I might be wrong, but it seems as if the id is sometimes used as a row/item index and sometimes mapped through the IndexColumn. Especially when used with picking. We probably need to double check that we are consistent and make it easy and/or clear how to use to the “right” way. This might not be related to this PR though

Not related but this is a general problem.  For example, things will go nasty once a dataframe is filtered and all of a sudden the row indices no longer match with the index column.

petersteneteg:
From a discussion we had:
> all the B&L events, could they just be one event/call with a Action {Filter, Select, Highlight} and a Target {Row, Column, Arb string}?

modules/brushingandlinking/include/modules/brushingandlinking/datastructures/indexlist.h
martinfalk: why? There is a BitSet member in the class.

ResearchDaniel: He is referring to map/set :)

modules/plottinggl/include/modules/plottinggl/plotters/scatterplotgl.h
ResearchDaniel: Ok

modules/plottinggl/src/plotters/scatterplotgl.cpp
ResearchDaniel: ok, too bad

modules/brushingandlinking/include/modules/brushingandlinking/ports/brushingandlinkingports.h
ResearchDaniel: Agree

modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
ResearchDaniel: filtered?

modules/brushingandlinking/include/modules/brushingandlinking/ports/brushingandlinkingports.h
martinfalk: no longer applicable. Caches replaced with B&L managers.

modules/plottinggl/include/modules/plottinggl/plotters/persistencediagramplotgl.h
martinfalk: Should be possible now after the refactoring.

modules/plottinggl/src/processors/scatterplotprocessor.cpp
martinfalk: Needs more work, potentially a bidirectional map being part of the index column.

modules/brushingandlinking/include/modules/brushingandlinking/datastructures/brushingaction.h
```diff
@@ -0,0 +1,172 @@
...
+
+/**
+ * Represents a target for brushing and linking actions.
+ *
+ * Commonly used brushing targets are BrushingTarget::Row and BrushingTarget::Column
+ *
+ * \see BrushingAndLinkingManager
+ */
+struct IVW_MODULE_BRUSHINGANDLINKING_API BrushingTarget {
+    BrushingTarget() = default;
```
petersteneteg: Hmm should this be allowed? should it default to Row?

modules/brushingandlinking/include/modules/brushingandlinking/datastructures/brushingaction.h
```diff
@@ -0,0 +1,172 @@
...
+    std::string str;
+    ss >> str;
+
+    if (str == toString(BrushingTarget::Row.target)) {
+        bt = BrushingTarget::Row;
+    } else if (str == toString(BrushingTarget::Column)) {
+        bt = BrushingTarget::Column;
+    } else {
+        bt = BrushingTarget(str);
+    }
```
petersteneteg: why not just use the else case?

modules/brushingandlinking/include/modules/brushingandlinking/datastructures/brushingaction.h
```diff
@@ -0,0 +1,172 @@
...
+                                                                      const BrushingTarget& bt);
+
+    template <class Elem, class Traits>
+    friend std::basic_istream<Elem, Traits>& operator>>(std::basic_istream<Elem, Traits>& ss,
+                                                        BrushingTarget& bt);
+
+    static const BrushingTarget Row;
+    static const BrushingTarget Column;
+
+    std::string target;
```
petersteneteg: I would like to avoid having to copy this string around all the time seems expensive to me. Could use a std::shared_ptr<const std::string> instead, then you only have to create it one, and all copies will just copy the pointer not reallocating the string. Also you only have to compare the pointers to check for equality etc...

petersteneteg: Humm maybe that last part is not true... you know they are equal if the pointers are the same, if they are nog you sill have to compare them :(

petersteneteg: One way would be to have a static vector of week_ptr to all the targets, and check if a string matches an existing one... that way you would avoid any duplicates...

modules/brushingandlinking/src/brushingandlinkingmanager.cpp
```diff
@@ -31,78 +31,481 @@
...
+        } else if (std::holds_alternative<IndexListTargets>(targetmap)) {
+            auto& map = std::get<BitSetTargets>(targetmap);
+            if (map.empty()) continue;
+
+            s.serialize(toString(action), map, "selection", {},
+                        [](BrushingTarget t) { return t.target; }, {});
+        }
+    }
+}
+/*
```
ResearchDaniel: Remove commented code

modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
```diff
@@ -30,73 +30,214 @@
...
+     *   + filter action A filters 1-5
+     *   + filter action B filters 3-6
+     * This results in indices {7, 8, 9, 10} remaining. If then action A is updated to filter out
+     * indices 2-4, the overall result for filtering should be {1, 7, ..., 10}. Without different
+     * sources it would not be possible to update the result accordingly.
+     *
+     * @param action   type of brushing action
+     * @param target   target of the action, determines which brushing and linking state to update
+     * @param indices  set of selected/filtered indices
+     * @param source   must be provided if action is equal to BrushingAction::Filter
```
ResearchDaniel, line 83: source should always be provided in my opinion. Otherwise it will not be possible to do union/intersect between sources. 
Usually source == processor identifier?

modules/brushingandlinking/src/brushingandlinkingmanager.cpp
```diff
@@ -31,78 +31,481 @@
...
+
+void BrushingAndLinkingManager::brush(BrushingAction action, BrushingTarget target,
+                                      const BitSet& indices, std::string_view source) {
+    if ((action == BrushingAction::Filter) && source.empty()) {
+        throw Exception("BrushingAction::Filter requires a source", IVW_CONTEXT);
+    }
+
+    const int actionIdx = getActionIndex(action);
+
+    std::visit(util::overloaded{[&](BitSetTargets& map) { map[target] = indices; },
```
ResearchDaniel, line 71: Don’t we want to have one set per source so that we can do union/intersect/latest over the selection? At least do it if user wants union/intersect

modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
```diff
@@ -30,73 +30,214 @@
...
+
+    //! convenience function for action(BrushingAction::Filter, target, idx)
+    void filter(std::string_view src, const BitSet& idx,
+                BrushingTarget target = BrushingTarget::Row);
+    //! convenience function for action(BrushingAction::Select, target, idx)
+    void select(const BitSet& idx, BrushingTarget target = BrushingTarget::Row);
+    //! convenience function for action(BrushingAction::Highlight, target, idx)
+    void highlight(const BitSet& idx, BrushingTarget target = BrushingTarget::Row);
+
+    [[deprecated("clearing filtered indices is no longer supported")]] void clearFiltered();
```
ResearchDaniel: Add:
Use filter(…) with empty BitSet

modules/brushingandlinking/include/modules/brushingandlinking/datastructures/indexlist.h
```diff
@@ -30,38 +30,41 @@
...
 
 private:
-    std::unordered_map<const BrushingAndLinkingInport*, std::unordered_set<size_t>>
-        indicesBySource_;
-    std::unordered_set<size_t> indices_;
-    Dispatcher<void()> onUpdate_;
-};
+    void update() const;
+
+    using PortIndexMap = std::unordered_map<std::string, BitSet>;
```
ResearchDaniel: SourceIndexMap

modules/brushingandlinking/include/modules/brushingandlinking/processors/brushingandlinkingprocessor.h
```diff
@@ -55,22 +55,15 @@ class IVW_MODULE_BRUSHINGANDLINKING_API BrushingAndLinkingProcessor : public Pro
...
-
     static const ProcessorInfo processorInfo_;
 
-    BrushingAndLinkingOutport& getOutport() { return outport_; }
-
 private:
     BrushingAndLinkingOutport outport_;
 
     ButtonProperty clearSelection_;
-    ButtonProperty clearFilter_;
```
ResearchDaniel, line 69: Not possible to clear filter anymore?
What does clearAll_ do in that case?

modules/brushingandlinking/src/datastructures/indexlist.cpp
```diff
@@ -30,45 +30,74 @@
...
-    indices_.clear();
+void IndexList::update() const {
+    if (!indicesDirty_) return;
+
+    using T = PortIndexMap::value_type;
+    util::map_erase_remove_if(indicesBySource_, [](const T& p) { return p.second.empty(); });
+
+    auto bitsets =
+        util::transform(indicesBySource_, [](auto& p) -> const BitSet* { return &p.second; });
+    indices_ = BitSet::fastUnion(bitsets);
```
ResearchDaniel, line 84: We want options for this to cover all cases?

modules/brushingandlinking/src/datastructures/indexlist.cpp
```diff
@@ -30,45 +30,74 @@
...
-    indices_.clear();
-
-    using T =
-        std::unordered_map<const BrushingAndLinkingInport*, std::unordered_set<size_t>>::value_type;
-    util::map_erase_remove_if(indicesBySource_, [](const T& p) {
-        return !p.first->isConnected() ||
-               p.second.empty();  // remove if port is disconnected or if the set is empty
-    });
+bool IndexList::contains(uint32_t idx) const {
+    if (indicesDirty_) {
```
ResearchDaniel: This check is already done in update…

modules/brushingandlinking/src/datastructures/indexlist.cpp
```diff
@@ -30,45 +30,74 @@
...
+    indices_.clear();
+    indicesBySource_.clear();
+    indicesDirty_ = false;
 }
 
-void IndexList::remove(const BrushingAndLinkingInport* src) {
-    indicesBySource_.erase(src);
-    update();
+const BitSet& IndexList::getIndices() const {
+    if (indicesDirty_) {
```
ResearchDaniel: Checked in update

modules/brushingandlinking/src/processors/brushingandlinkingprocessor.cpp
```diff
@@ -47,36 +44,22 @@ const ProcessorInfo BrushingAndLinkingProcessor::getProcessorInfo() const { retu
...
-                    manager_->clearColumns();
-                })
-    , manager_(std::make_shared<BrushingAndLinkingManager>(this)) {
+    , clearSelection_("clearSelection", "Clear Selection",
+                      [&]() { outport_.getManager().clearSelected(); })
+    , clearHighlight_("clearHighlight", "Clear Highlighting",
+                      [&]() { outport_.getManager().clearHighlighted(); })
+    , clearCols_("clearCols", "Clear Columns",
+                 [&]() { outport_.getManager().clearSelected(BrushingTarget::Column); })
+    , clearAll_("clearAll", "Clear All", [&]() {
```
ResearchDaniel: This is not all anymore

modules/brushingandlinking/src/processors/brushingandlinkingprocessor.cpp
```diff
@@ -47,36 +44,22 @@ const ProcessorInfo BrushingAndLinkingProcessor::getProcessorInfo() const { retu
...
+        outport_.getManager().clearSelected();
```
ResearchDaniel, line 54: Is it possible to specify BrushingTarget::Row | BrushingTarget::Column

modules/brushingandlinking/src/brushingandlinkingmanager.cpp
```diff
@@ -31,78 +31,481 @@
...
+        return it->second.getIndices();
     }
-    onFilteringChangeCallback_ =
-        filtered_.onChange([p, validationLevel]() { p->invalidate(validationLevel); });
 }
 
-BrushingAndLinkingManager::~BrushingAndLinkingManager() {}
+void BrushingAndLinkingManager::clearIndices(BrushingAction action, BrushingTarget target) {
+    if (action == BrushingAction::Filter) {
+        throw Exception(fmt::format("Clearing indices for action '{}' is not supported", action),
```
ResearchDaniel, line 128: Supporting this would mean that we propagate other types of information/actions with the manager. 
By delegating these tasks (clear) to the implementations connected to the brushing and linking we could more easily support the actions with cached sources (which might need to be all if we want to support union/intersect for select/highlight as well).
Could for example register an onClearRequest with the manager so that we can support clearing from the manager?

ResearchDaniel: On the other hand it might not be worth the effort to support such concepts?

modules/dataframeqt/src/processors/dataframetable.cpp
```diff
@@ -86,14 +86,18 @@ DataFrameTable::~DataFrameTable() {
...
             w->setDataFrame(inport_.getData(), vectorCompAsColumn_, showCategoryIndices_);
-            w->updateSelection(brushLinkPort_.getSelectedColumns(),
-                               brushLinkPort_.getSelectedIndices());
-        } else if (brushLinkPort_.isChanged()) {
-            w->updateSelection(brushLinkPort_.getSelectedColumns(),
-                               brushLinkPort_.getSelectedIndices());
+            dataUpdated = true;
+        }
+        if (dataUpdated || brushLinkPort_.modifiedSelection()) {
+            w->updateColumnSelection(brushLinkPort_.getSelectedIndices(BrushingTarget::Column));
```
ResearchDaniel, line 96: Do you want to check for isModified for column/row?

modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
```diff
@@ -600,11 +604,14 @@ void ParallelCoordinates::drawLines(size2_t size) {
         utilgl::Enable<MeshGL> enable{meshGL};
         lines_.indices.getRepresentation<BufferGL>()->bind();
 
-        std::array<float, 3> width = {lineWidth_, lineWidth_, selectedLineWidth_};
-        std::array<float, 3> mixColor = {filterIntensity_, 0.0f, 0.0f};
-        std::array<float, 3> mixSelection = {0.0f, 0.0f,
-                                             selectedLineColorOverride_.isChecked() ? 1.0f : 0.0f};
-        std::array<float, 3> mixAlpha = {1.0, 0.0f, 0.0f};
+        std::array<float, 4> width = {lineWidth_, lineWidth_, selectedLineWidth_,
+                                      selectedLineWidth_};
```
ResearchDaniel, line 608: Highlight width is same as selection?

modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
ResearchDaniel: Also, should we use an interface for source instead of a string?
That way the manager could communicate with the source to for example clear filter.

ResearchDaniel (changes requested):
Very nice work!

**Some thoughts**

- BrushingTarget::Row may be confused with the row number, while it actually referes to the id in the IndexColumn. 
BrushingTarget::Index?
BrushingTarget::Id? 

- Do we want to be able to propagate information from the manager to the sources (e.g. for clearing)? Then it would be better to use a BrushingSource interface instead of a string. 

- For columns we are using the direct index, but not for rows. This is a bit inconsistent...

- Perhaps a too big design change now, but using a function interface instead of a Bitset for the brush interface could potentially save memory. For example: Each plot in a scattermatrix has its own selection/highlight/filter. In these cases it is not necessary to store the selection, just the the ranges associated with the two axes.



Only using IndexList would make the code clearer (avoid the overload in each std::visit call.

modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
martinfalk: at the moment, selection and highlight are not requiring a source since they are replacing the currently selected indices. If that is going to change, then sure. But that is beyond the refactoring in this PR.

martinfalk: > Usually source == processor identifier?

If the action is propagated by a port, it is `port->getPath()` otherwise it could be anything. In case of the parallel coords, it uses the names of the axes.

modules/brushingandlinking/src/brushingandlinkingmanager.cpp
martinfalk: union/intersect for highlight and selection -> future PR

modules/brushingandlinking/include/modules/brushingandlinking/processors/brushingandlinkingprocessor.h
martinfalk: "clear filter" in the manager will lead to an inconsistent state since the filter is still active in the source. Clear All will reset all selections but the filtered one.

modules/brushingandlinking/src/datastructures/indexlist.cpp
martinfalk: future PR

modules/brushingandlinking/src/datastructures/indexlist.cpp
martinfalk: good catch

modules/brushingandlinking/src/processors/brushingandlinkingprocessor.cpp
martinfalk: should it state "clear selection, highlight, column selection" instead? ;-)

martinfalk: renamed to `clearAllAbove_("clearAllAbove", "Clear All Above"`

modules/brushingandlinking/src/processors/brushingandlinkingprocessor.cpp
martinfalk: No, since targets are no flags.

modules/dataframeqt/src/processors/dataframetable.cpp
martinfalk: The modified state does not differentiate between columns and rows. We could add that later if we think it is necessary.

modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
martinfalk: Yes, the difference being the color and the line rendered on top.

martinfalk:
> **Some thoughts**
> 
> * BrushingTarget::Row may be confused with the row number, while it actually referes to the id in the IndexColumn.
>   BrushingTarget::Index?
>   BrushingTarget::Id?

Decided to use row as being orthogonal to column and our brushing and linking often relies on a DataFrame as underlying data source.

> * For columns we are using the direct index, but not for rows. This is a bit inconsistent...

That is because we don't have a special mapping for columns besides their index. The current DataFrame implementation _always_ creates an index column even when creating an empty DataFrame (which cause other problems...). We might want to consider getting rid of the index column entirely. However, bear in mind that filtering a DataFrame and outputting a new one would affect those direct indices and thus breaking brushing and linking.

> * Do we want to be able to propagate information from the manager to the sources (e.g. for clearing)? Then it would be better to use a BrushingSource interface instead of a string.

Maybe in a follow-up PR.

> * Perhaps a too big design change now, but using a function interface instead of a Bitset for the brush interface could potentially save memory. For example: Each plot in a scattermatrix has its own selection/highlight/filter. In these cases it is not necessary to store the selection, just the the ranges associated with the two axes.

Agreed, using ranges instead of explicit index lists would change things quite a bit.

modules/brushingandlinking/src/processors/brushingandlinkingprocessor.cpp
ResearchDaniel: I guess clear selection would be fine.

ResearchDaniel: Ah, that is the name of one of the above :D

ResearchDaniel: Clear selections then :)

CHANGELOG.md
```diff
@@ -1,5 +1,37 @@
...
+        const BitSet& highlighted = brushingPort_.getHighlightedIndices();
+        ...
+    }
+    // or check for individual modifications
+    if (brushingPort_.modifiedActions() & BrushingModification::Filtered) {
+        ...
+    }
+}
+```
+**Update Note:** In case you are using `port.isChanged()` on a `BrushingAndLinkingInport`, you **must** replace this with one of the calls above. Otherwise your selection might no longer be updated.
```
petersteneteg: what about the onChange callback is that also not usable anymore? Could we make them work? having both isModified and isChanged is a bit confusing...

include/inviwo/core/io/serialization/serializebase.h
```diff
@@ -154,7 +154,7 @@ class IVW_CORE_API SerializeBase {
                std::is_same_v<T, size_t> || std::is_same_v<T, long long> ||
                std::is_same_v<T, unsigned long long> || std::is_same_v<T, float> ||
                std::is_same_v<T, double> || std::is_same_v<T, long double> ||
-               std::is_same_v<T, std::string>;
+               std::is_same_v<T, std::string> || std::is_same_v<T, std::string_view>;
```
petersteneteg, line 157: Note: In principle we could use is_stream_insertable here, since we handle that on the deserialization side. 
But in principle we should also have something like has_unique_string_representation. Or support_iostream_roundtrip or something like that...

modules/basegl/include/modules/basegl/processors/sphererenderer.h
```diff
@@ -93,7 +93,7 @@ class IVW_MODULE_BASEGL_API SphereRendererSelection {
                     const auto seq = util::make_sequence(
                         uint32_t{0}, static_cast<uint32_t>(indexBuffer.size()), uint32_t{1});
                     std::copy_if(seq.begin(), seq.end(), std::back_inserter(indices),
-                                 [&](uint32_t i) { return selection.count(indexBuffer[i]) > 0; });
```
petersteneteg, line 96: you use brushLinkPort_.isChanged() above?

modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
petersteneteg: Don't have a strong opinion either way here, but always requiring the source would make the interface more consistent, and even if the source is not required, it might be useful for debugging logging, the manager could for example show the source of the current selection.

modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
```diff
@@ -30,73 +30,216 @@
...
+     *
+     * @return modifiedActions() & BrushingModification::Filtered
+     */
+    bool modifiedFiltering() const;
+    /**
+     * return whether there was a select action since the last network evaluation
+     *
+     * @return modifiedActions() & BrushingModification::Selected
+     */
+    bool modifiedSelection() const;
```
petersteneteg, line 123: can you see which target that is changed?

modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
```diff
@@ -30,73 +30,216 @@
...
+    /**
+     * update the internal state with respect to port connections. This function should be called
+     * whenever a connection is removed from a brushing and linking outport. This will only effect
+     * filter actions and has no effect if the manager is owned by an inport.
+     *
+     * \see action
+     */
+    void updatePortConnections();
+
+    //! convenience function for action(BrushingAction::Filter, target, idx)
```
petersteneteg:
```suggestion
    //! convenience function for brush(BrushingAction::Filter, target, idx)
```

modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
```diff
@@ -30,73 +30,216 @@
...
+    void filter(std::string_view src, const BitSet& idx,
+                BrushingTarget target = BrushingTarget::Row);
+    //! convenience function for action(BrushingAction::Select, target, idx)
```
petersteneteg:
```suggestion
    //! convenience function for brush(BrushingAction::Select, target, idx)
```

modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
```diff
@@ -30,73 +30,216 @@
...
+    void select(const BitSet& idx, BrushingTarget target = BrushingTarget::Row);
+    //! convenience function for action(BrushingAction::Highlight, target, idx)
```
petersteneteg:
```suggestion
    //! convenience function for brush(BrushingAction::Highlight, target, idx)
```

modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
```diff
@@ -30,73 +30,216 @@
...
+     * whenever a connection is removed from a brushing and linking outport. This will only effect
+     * filter actions and has no effect if the manager is owned by an inport.
+     *
+     * \see action
+     */
+    void updatePortConnections();
+
+    //! convenience function for action(BrushingAction::Filter, target, idx)
+    void filter(std::string_view src, const BitSet& idx,
+                BrushingTarget target = BrushingTarget::Row);
```
petersteneteg:
```suggestion
    void filter(const BitSet& idx, BrushingTarget target, std::string_view src);
```
To be consistent with brush

modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
petersteneteg: I would put these together with brush

modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
```diff
@@ -30,73 +30,216 @@
...
+    // clang-format off
+    [[deprecated("clearing filtered indices is no longer supported. Use filter() with an empty BitSet")]] void clearFiltered();
+    // clang-format on
+    //! convenience function for clearIndices(BrushingAction::Select, target)
+    void clearSelected(BrushingTarget target = BrushingTarget::Row);
+    //! convenience function for clearIndices(BrushingAction::Highlight, target)
+    void clearHighlighted(BrushingTarget target = BrushingTarget::Row);
+
+    //! convenience function for getIndices(BrushingAction::Filter, target).size()
+    size_t getNumberOfFiltered(BrushingTarget target = BrushingTarget::Row) const;
```
petersteneteg: might add `size_t getNumber(BrushingAction action, BrushingTarget target = BrushingTarget::Row) const;`

modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
petersteneteg: same goes for the other convenience functions

modules/brushingandlinking/include/modules/brushingandlinking/datastructures/brushingaction.h
```diff
@@ -0,0 +1,184 @@
...
+    friend std::basic_istream<Elem, Traits>& operator>>(std::basic_istream<Elem, Traits>& ss,
+                                                        BrushingTarget& bt);
+
+    std::string_view getString() const { return target_; }
+
+    static const BrushingTarget Row;
+    static const BrushingTarget Column;
+
+private:
+    std::string_view findOrAdd(std::string_view target) {
```
petersteneteg: I would move this to the cpp with all the headers that are needed for this...

modules/brushingandlinking/include/modules/brushingandlinking/datastructures/brushingaction.h
```diff
@@ -0,0 +1,184 @@
...
+};
+
+template <class Elem, class Traits>
+std::basic_istream<Elem, Traits>& operator>>(std::basic_istream<Elem, Traits>& ss,
+                                             BrushingTarget& bt) {
+    std::string str;
+    ss >> str;
+    bt = BrushingTarget(str);
+    return ss;
+}
```
petersteneteg: Move this inline to the friend declaration, to make it a "hidden friend"

modules/brushingandlinking/include/modules/brushingandlinking/datastructures/indexlist.h
petersteneteg: unordered_set is not needed?

modules/brushingandlinking/include/modules/brushingandlinking/ports/brushingandlinkingports.h
```diff
@@ -29,53 +29,152 @@
...
 
 namespace inviwo {
 
-class IVW_MODULE_BRUSHINGANDLINKING_API BrushingAndLinkingInport
-    : public DataInport<BrushingAndLinkingManager> {
+class BrushingAndLinkingOutport;
+
+class IVW_MODULE_BRUSHINGANDLINKING_API BrushingAndLinkingInport : public Inport {
 public:
+    using type = void;
```
petersteneteg, line 46: might also change that error message to remove the need for this?

modules/brushingandlinking/include/modules/brushingandlinking/ports/brushingandlinkingports.h
```diff
@@ -29,53 +29,152 @@
...
-    const std::unordered_set<size_t>& getFilteredIndices() const;
-    const std::unordered_set<size_t>& getSelectedColumns() const;
+    /**
+     * check if the state of the manager was changed since the last network evaluation
+     *
+     * @return true if there have been recent changes
+     *
+     * \see modifiedActions
+     */
+    bool isModified() const;
```
petersteneteg: overload isChanged?

modules/brushingandlinking/include/modules/brushingandlinking/processors/brushingandlinkingprocessor.h
petersteneteg: Might be nice to be able to do this somehow. question is how to communicate that to the filter sources... outside fo this PR though...

modules/brushingandlinking/src/brushingandlinkingmanager.cpp
petersteneteg: I think the union/intersect/latest semantics is encoded in the action type. i.e. selection and highlight is always lates, filter is always union. if you want a different semantics, we should add a new action type.

modules/brushingandlinking/src/brushingandlinkingmanager.cpp
```diff
@@ -31,78 +31,422 @@
...
+}
+
+bool BrushingAndLinkingManager::isModified() const { return !modifiedActions_.empty(); }
+
+BrushingModifications BrushingAndLinkingManager::modifiedActions() const {
+    return modifiedActions_;
+}
+
+bool BrushingAndLinkingManager::modifiedFiltering() const {
+    return bool(modifiedActions_ & BrushingModification::Filtered);
```
petersteneteg:
```suggestion
    return modifiedActions_.contains(BrushingModification::Filtered);
```
I think that is much easier to read

modules/brushingandlinking/src/brushingandlinkingmanager.cpp
```diff
@@ -31,78 +31,422 @@
...
+        selections_[actionIdx]);
+}
+
+const BitSet& BrushingAndLinkingManager::getIndices(BrushingAction action,
+                                                    BrushingTarget target) const {
+    if (parent_) {
+        return parent_->getIndices(action, target);
+    }
+
+    static BitSet empty;
```
petersteneteg:
```suggestion
    static const BitSet empty;
```

modules/brushingandlinking/src/brushingandlinkingmanager.cpp
```diff
@@ -31,78 +31,422 @@
...
+
+    const int actionIdx = getActionIndex(action);
+
+    if (std::holds_alternative<BitSetTargets>(selections_[actionIdx])) {
```
petersteneteg: why not use visit here? (just to be consistent?)

modules/brushingandlinking/src/brushingandlinkingmanager.cpp
```diff
@@ -31,78 +31,422 @@
...
+                                    it->second.clear();
+                                },
+                                [&](IndexListTargets& map) {
+                                    auto it = map.find(target);
+                                    if (it == map.end()) {
+                                        return;
+                                    }
+                                    it->second.clear();
+                                }},
+               selections_[actionIdx]);
```
petersteneteg:
```suggestion
    std::visit([&](auto& map) {
        const auto it = map.find(target);
        if (it != map.end()) {
            it->second.clear();
        },
        selections_[actionIdx]);
```

modules/brushingandlinking/src/brushingandlinkingmanager.cpp
```diff
@@ -31,78 +31,422 @@
...
+}
+
+bool BrushingAndLinkingManager::contains(uint32_t idx, BrushingAction action,
+                                         BrushingTarget target) const {
+    if (parent_) {
+        return parent_->contains(idx, action, target);
+    }
+
+    const int actionIdx = getActionIndex(action);
+    return std::visit(util::overloaded{[&](const BitSetTargets& map) {
```
petersteneteg: could use one lambda with auto here to

modules/brushingandlinking/src/brushingandlinkingmanager.cpp
```diff
@@ -31,78 +31,422 @@
...
+    auto extractTargets = [&](BrushingAction action, const auto& map) {
+        std::transform(map.begin(), map.end(), std::back_inserter(targets),
+                       [action](auto elem) -> ActionTarget {
+                           return {action, elem.first};
+                       });
+    };
+
+    for (auto&& [action, map] : util::zip(BrushingActions, selections_)) {
+        std::visit(
+            util::overloaded{
```
petersteneteg, line 181: could just use one lambda with auto...

modules/brushingandlinking/src/brushingandlinkingmanager.cpp
```diff
@@ -31,78 +31,422 @@
...
+    return targets;
+}
+
+std::vector<BrushingTarget> BrushingAndLinkingManager::getTargets(BrushingAction action) const {
+    if (parent_) {
+        return parent_->getTargets(action);
+    }
+
+    const int actionIdx = getActionIndex(action);
+    return std::visit(util::overloaded{[&](const BitSetTargets& map) {
```
petersteneteg: could just use one lambda with auto...

modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
```diff
@@ -30,73 +30,216 @@
...
-    std::unordered_set<size_t> selected_;
-    std::unordered_set<size_t> selectedColumns_;
-    IndexList filtered_;  // Use IndexList to be able to remove filtered rows on port disconnection
-    std::shared_ptr<std::function<void()>> onFilteringChangeCallback_;
+    SelectionMap selections_{{IndexListTargets(), BitSetTargets(), BitSetTargets()}};
 
-    Processor* owner_;  // Non-owning reference
+    std::variant<BrushingAndLinkingInport*, BrushingAndLinkingOutport*> owner_;
+    BrushingAndLinkingManager* parent_ = nullptr;
+    std::unordered_set<BrushingAndLinkingManager*> children_;
```
petersteneteg, line 233: why a set?

modules/brushingandlinking/src/brushingandlinkingmanager.cpp
```diff
@@ -31,78 +31,422 @@
...
-    selected_ = indices;
-    owner_->invalidate(invalidationLevel_);
+void BrushingAndLinkingManager::setParent(BrushingAndLinkingManager* parent) {
+    if (parent_) {
+        parent_->removeChild(this);
+
+        // set manager as modified since filter actions might have changed when removing the parent
+        modifiedActions_ =
+            BrushingModifications(BrushingModification::Filtered, BrushingModification::Selected,
+                                  BrushingModification::Highlighted);
```
petersteneteg:
```suggestion
        modifiedActions_ = BrushingModifications(flags::any);
```
If my memory serves me right

modules/brushingandlinking/src/brushingandlinkingmanager.cpp
```diff
@@ -31,78 +31,422 @@
...
+    modifiedActions_ |= fromAction(action);
+
+    if (std::holds_alternative<BrushingAndLinkingOutport*>(owner_)) {
+        auto outport = std::get<BrushingAndLinkingOutport*>(owner_);
+        outport->getProcessor()->invalidate(invalidationLevel_);
+    }
+
+    if (parent_) {
+        const std::string source =
+            std::visit(util::overloaded{[](BrushingAndLinkingInport* p) { return p->getPath(); },
```
petersteneteg: use auto for the parameter

modules/brushingandlinking/src/brushingandlinkingmanager.cpp
```diff
@@ -31,78 +31,422 @@
...
+                                            return &it->second;
+                                        },
+                                        [&](const IndexListTargets& map) -> const BitSet* {
+                                            auto it = map.find(target);
+                                            if (it == map.end()) {
+                                                return nullptr;
+                                            }
+                                            return &it->second.getIndices();
+                                        }},
+                       selections_[actionIdx]);
```
petersteneteg: this is used in a few places, might make it a private member... getBitset(action, taget)?

modules/brushingandlinking/src/brushingandlinkingmanager.cpp
```diff
@@ -31,78 +31,422 @@
...
+                                       },
+                                       [&](const IndexListTargets& map) {
+                                           return util::transform(
+                                               map, [](const auto& elem) { return elem.first; });
+                                       }},
+                      selections_[actionIdx]);
+}
+
+void BrushingAndLinkingManager::updatePortConnections() {
+    if (std::holds_alternative<BrushingAndLinkingOutport*>(owner_)) {
```
petersteneteg: could all of this not be managed in the removeChild method? why not?

modules/brushingandlinking/src/ports/brushingandlinkingports.cpp
```diff
@@ -29,89 +29,178 @@
...
+    return manager_.getIndices(BrushingAction::Select, BrushingTarget::Column);
+}
+
+BrushingAndLinkingManager& BrushingAndLinkingInport::getManager() { return manager_; }
+
+const BrushingAndLinkingManager& BrushingAndLinkingInport::getManager() const { return manager_; }
+
+void BrushingAndLinkingInport::serialize(Serializer& s) const {
+    Inport::serialize(s);
+    manager_.serialize(s);
```
petersteneteg: why not d.serialize("manager", manager); ?

petersteneteg (changes requested):
Nice work, nice we finally got this done! 
just some minor comments

modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
ResearchDaniel: Would be quite useful when trying to prevent loops in the brushing

CHANGELOG.md
martinfalk: I did not want to overload onChange() originally. But it would be an option, yes.

modules/basegl/include/modules/basegl/processors/sphererenderer.h
martinfalk: overlooked that one.

modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
martinfalk: That would then be a list of targets if we want that.

modules/brushingandlinking/src/brushingandlinkingmanager.cpp
ResearchDaniel: > I think the union/intersect/latest semantics is encoded in the action type. i.e. selection and highlight is always lates, filter is always union. if you want a different semantics, we should add a new action type.

I disagree that it is encoded in the action type, but the ones we have now is probably the most commonly used and covers many use cases

CHANGELOG.md
petersteneteg: having both might be confusing...

modules/brushingandlinking/include/modules/brushingandlinking/ports/brushingandlinkingports.h
martinfalk: connected to the check for type traits of a port

https://github.com/inviwo/inviwo/blob/be406b657492879ea60a3f2fc8adb610b08b738d/include/inviwo/core/common/inviwomodule.h#L408-L416

modules/brushingandlinking/src/brushingandlinkingmanager.cpp
martinfalk: note the distinction here with using `nullptr` while in the other case an empty BitSet is used.

modules/brushingandlinking/src/ports/brushingandlinkingports.cpp
martinfalk: because without that it was not a virtual class. Will change to inherit from Serializable.

modules/brushingandlinking/src/brushingandlinkingmanager.cpp
```diff
@@ -144,22 +129,14 @@ void BrushingAndLinkingManager::clearIndices(BrushingAction action, BrushingTarg
...
-                                },
-                                [&](IndexListTargets& map) {
-                                    auto it = map.find(target);
-                                    if (it == map.end()) {
-                                        return;
-                                    }
-                                    it->second.clear();
-                                }},
-               selections_[actionIdx]);
+    std::visit(util::overloaded{[&](auto& map) {
```
petersteneteg: nit: no need for util::overloaded if there is only one lambda

modules/brushingandlinking/src/brushingandlinkingmanager.cpp
petersteneteg: What I mean is that we can just define that to be the case, if you need selection with an intersect semantics, just create a new action for that with a new name IntersectionSelection or something.

petersteneteg (dismissed):

modules/brushingandlinking/src/ports/brushingandlinkingports.cpp
petersteneteg: Ahh.. I see. Im fine with either, Its just we usually do it that way...

petersteneteg:
This is looking really nice I think!

modules/brushingandlinking/src/brushingandlinkingmanager.cpp
ResearchDaniel: You don’t want/need to change the plots to support that behavior so I think that the logic belongs to the manager

petersteneteg (approved):

modules/brushingandlinking/include/modules/brushingandlinking/ports/brushingandlinkingports.h
ResearchDaniel: Still should be private?

ResearchDaniel (approved):
Great work!

modules/brushingandlinking/src/brushingandlinkingmanager.cpp
petersteneteg: Ahh I think I understand how you mean, you want that semantic to be configurable by the user at runtime for a specific network / b&l setup? Like for these two plots that are linked to a manager I want a to use use an intersection selection, but for this other one I want a global one.

petersteneteg:
@ResearchDaniel that has to be an old version see https://github.com/inviwo/inviwo/blob/feature/brushing-highlight/modules/brushingandlinking/include/modules/brushingandlinking/ports/brushingandlinkingports.h

ResearchDaniel:
Ah, it seems so!

modules/brushingandlinking/src/brushingandlinkingmanager.cpp
ResearchDaniel: Exactly

modules/brushingandlinking/include/modules/brushingandlinking/ports/brushingandlinkingports.h
martinfalk: the managers are private members in both ports.

PR #1208, opened 2021-10-10
Webbrowser: Add support for get/set/subsrcribe to system and module settings

ResearchDaniel:
Ended up in a bit of a rabbit hole, but fixed some other bugs in the process:
- Remove workaround that tried to find the onChange callback in the html file, but meant that the onChange function could not be included in external file (had to be in the html file).
- Clear property-CEF synchronization when reloading a page <- this works better than the workaround above
- Added missing delegation of OnLoadStart in WebbrowserClient
- Update property path in test workspace and make sure a message is logged if a property is not found.
- Use one PropertyCEFSynchronizer per browser to simplify handling


modules/webbrowser/include/modules/webbrowser/properties/propertycefsynchronizer.h
```diff
@@ -76,8 +79,16 @@ class IVW_MODULE_WEBBROWSER_API PropertyCefSynchronizer
...
       public PropertyOwnerObserver {
 public:
-    explicit PropertyCefSynchronizer(const PropertyWidgetCEFFactory* htmlWidgetFactory);
+    /* 
+     * Only handles events from the provided browser
+     */
+    explicit PropertyCefSynchronizer(CefRefPtr<CefBrowser> browser, const PropertyWidgetCEFFactory* htmlWidgetFactory);
     virtual ~PropertyCefSynchronizer() = default;
+    /**
+     * Removes all property synchronizations.
```
martinfalk, line 89: when do you call this function?

martinfalk (dismissed):

martinfalk:
Could you also look at the timeout mechanism used in the web regression tests? The browser test fails quite often and returns an empty canvas.

modules/webbrowser/include/modules/webbrowser/properties/propertycefsynchronizer.h
ResearchDaniel: It is called by the WebBrowserClient (part of CefLoadHandler interface).

It is called every time the webpage loads, e.g. page navigation

ResearchDaniel:
> Could you also look at the timeout mechanism used in the web regression tests? The browser test fails quite often and returns an empty canvas.

now it fails because I fixed an error... Does it fail randomly still?

petersteneteg:
it is know to happen 
image

petersteneteg:
this is for the daily build of master, to there should be 0 fails...

ResearchDaniel:
Unfortunately I don't think that these changes will improve the random build failure situation

martinfalk (approved):

PR #1211, opened 2021-10-12
Platform-independent file and directory check

PetiteViking:


petersteneteg:
I would prefer to do a comprehensive more to using std:filesystem instead of our filesystem in one go. Deprecating all our functions and replacing them with standard one. 
Note: One also has to link to the filesystem lib in gcc and clang...

JochenJankowai:
Alright, fair enough.

PR #1216, opened 2021-10-26
Feature/dataframecolumnproperty

martinfalk:
cleaned up the column property for DataFrame.


modules/dataframe/include/inviwo/dataframe/properties/dataframecolumnproperty.h
```diff
@@ -37,36 +37,69 @@ namespace inviwo {
 
 class IVW_MODULE_DATAFRAME_API DataFrameColumnProperty : public OptionPropertyInt {
 public:
+    enum class EmptySelection { No, Yes };
```
ResearchDaniel: Would prefer the following to make it easier to understand what it is for:
AllowEmptySelection
or as it was before:
AllowNone
or what it actually does:
AddNoneOption
:)

modules/dataframe/include/inviwo/dataframe/properties/dataframecolumnproperty.h
```diff
@@ -37,36 +37,69 @@ namespace inviwo {
...
+    enum class FlattenedView { No, Yes };
```
petersteneteg: remove?

modules/dataframe/include/inviwo/dataframe/properties/dataframecolumnproperty.h
```diff
@@ -37,36 +37,69 @@ namespace inviwo {
...
+                            size_t defaultIndex = 0);
+    /**
+     * Constructor using an inport \p port to popuplate the options. The onChange event of \p port
+     * will trigger an update of the options.
+     *
+     * @param port
+     * @param emptySelection  if equal to EmptySelection::Yes, "None" will be added as option
+     * @param defaultIndex    index of the selected column when the property options are updated
+     */
+    DataFrameColumnProperty(std::string identifier, std::string displayName, DataFrameInport& port,
```
petersteneteg: use string_view?

modules/dataframe/src/properties/dataframecolumnproperty.cpp
```diff
@@ -27,117 +27,116 @@
...
+    if (emptySelection_ == EmptySelection::Yes) {
         options.emplace_back("none", "None", -1);
     }
 
-    int idx = 0;
-    for (const auto& col : *dataframe) {
-        auto header = col->getHeader();
-        auto identifier = header;
-        util::erase_remove_if(identifier, [](char cc) {
+    auto createIdentifier = [](std::string str) {
```
petersteneteg: use stripIdentifier?

petersteneteg (dismissed):
A couple minor comments...

modules/dataframe/include/inviwo/dataframe/properties/dataframecolumnproperty.h
```diff
@@ -37,36 +37,69 @@ namespace inviwo {
...
+    enum class EmptySelection { No, Yes };
+    enum class FlattenedView { No, Yes };
+
     virtual std::string getClassIdentifier() const override;
     static const std::string classIdentifier;
 
-    DataFrameColumnProperty(std::string identifier, std::string displayName, bool allowNone = false,
-                            size_t firstIndex = 0);
+    /**
+     * Constructor, no options will be created
```
ResearchDaniel: This is not true if EmptySelection=Yes?
We could also change the name to AddNoneOption to make it more explicit

modules/dataframe/include/inviwo/dataframe/properties/dataframecolumnproperty.h
```diff
@@ -37,36 +37,69 @@ namespace inviwo {
...
+                            EmptySelection emptySelection = EmptySelection::No,
+                            size_t defaultIndex = 0);
 
     DataFrameColumnProperty(const DataFrameColumnProperty& rhs);
     virtual DataFrameColumnProperty* clone() const override;
 
     virtual ~DataFrameColumnProperty() = default;
 
+    /**
+     * populate the options with the data owned by \p port. The onChange event of \p port
```
ResearchDaniel: capital P -> Populate

modules/dataframe/include/inviwo/dataframe/properties/dataframecolumnproperty.h
```diff
@@ -37,36 +37,69 @@ namespace inviwo {
...
+     * will trigger an update of the options.
+     */
+    void setPort(DataFrameInport& port);
+    /**
+     * update the options based on the columns in \p dataframe
+     */
     void setOptions(std::shared_ptr<const DataFrame> dataframe);
```
ResearchDaniel: Depending on how many breaking changes this PR has you might also what to consider 
```suggestion
    void setOptions(const DataFrame& dataframe);
```

modules/dataframe/include/inviwo/dataframe/properties/dataframecolumnproperty.h
```diff
@@ -37,36 +37,69 @@ namespace inviwo {
...
 
-    std::string getColumnHeader() const;
-    std::shared_ptr<const Column> getColumn() const;
-    std::shared_ptr<const BufferBase> getBuffer() const;
+    void setDefaultIndex(int index);
```
ResearchDaniel: ?
setDefaultSelectedIndex?
Can it be negative?

modules/dataframe/include/inviwo/dataframe/properties/dataframecolumnproperty.h
```diff
@@ -37,36 +37,69 @@ namespace inviwo {
...
+
+    /**
+     * return the header of the currently selected column
```
ResearchDaniel: Return

modules/dataframe/include/inviwo/dataframe/properties/dataframecolumnproperty.h
```diff
@@ -37,36 +37,69 @@ namespace inviwo {
...
 
     virtual ~DataFrameColumnProperty() = default;
 
+    /**
+     * populate the options with the data owned by \p port. The onChange event of \p port
+     * will trigger an update of the options.
+     */
+    void setPort(DataFrameInport& port);
+    /**
+     * update the options based on the columns in \p dataframe
```
ResearchDaniel: update -> Set?
It does remove the other options?

modules/dataframe/include/inviwo/dataframe/properties/dataframecolumnproperty.h
```diff
@@ -37,36 +37,69 @@ namespace inviwo {
...
+     * return the header of the currently selected column
+     *
+     * @return column header. If "None" is selected, an empty string is returned
+     */
+    const std::string& getColumnHeader() const;
 
-    virtual std::string getClassIdentifierForWidget() const override {
-        return TemplateOptionProperty<int>::getClassIdentifier();
-    }
+    // virtual std::string getClassIdentifierForWidget() const override {
```
ResearchDaniel: Remove

modules/dataframe/include/inviwo/dataframe/properties/dataframecolumnproperty.h
ResearchDaniel:
```
/*
 *  @param index    index of the selected column when the property options are updated
 */
```

modules/dataframe/src/properties/dataframecolumnproperty.cpp
```diff
@@ -27,117 +27,116 @@
...
     : OptionPropertyInt(rhs)
-    , dataframe_{nullptr}
-    , allowNone_{rhs.allowNone_}
-    , firstIndex_{rhs.firstIndex_} {}
+    , inport_(rhs.inport_)
+    , emptySelection_(rhs.emptySelection_)
+    , defaultIndex_(rhs.defaultIndex_) {
+
+    if (inport_) {
+        onChangeCallback_ = inport_->onChangeScoped([&]() {
```
ResearchDaniel:
```suggestion
        setPort(inport_);
```

modules/dataframe/src/properties/dataframecolumnproperty.cpp
```diff
@@ -27,117 +27,116 @@
...
-        if (src->options_.size() == 0) return;
-
-        if (options_.size() == 0) return;
+        if ((src->options_.size() == 0) || (options_.size() == 0)) {
+            return;
+        }
 
-        auto i = src->getSelectedValue();
-        if (i == -1 && !allowNone_) {
+        if ((src->getSelectedValue() < 0) && (emptySelection_ == EmptySelection::No)) {
```
ResearchDaniel: Maybe would be nice to add a function for isNoneSelected() 
It is used in at least two places and one kind of has to know that the index of None is -1

modules/dataframe/src/properties/dataframecolumnproperty.cpp
```diff
@@ -27,117 +27,116 @@
...
-        return col->getHeader();
-    }
-    return {};
-}
+void DataFrameColumnProperty::setDefaultIndex(int index) { defaultIndex_ = index; }
 
-std::shared_ptr<const Column> DataFrameColumnProperty::getColumn() const {
-    if (!dataframe_) {
-        return nullptr;
+const std::string& DataFrameColumnProperty::getColumnHeader() const {
```
ResearchDaniel: maybe getSelectedColumnHeader

modules/dataframe/include/inviwo/dataframe/properties/dataframecolumnproperty.h
```diff
@@ -37,36 +37,69 @@ namespace inviwo {
...
 
     DataFrameColumnProperty(const DataFrameColumnProperty& rhs);
     virtual DataFrameColumnProperty* clone() const override;
 
     virtual ~DataFrameColumnProperty() = default;
 
+    /**
+     * populate the options with the data owned by \p port. The onChange event of \p port
+     * will trigger an update of the options.
+     */
```
ResearchDaniel:
```suggestion
     * @note A reference to the port will be kept so the port must outlive the scope of the property.
     */
```

modules/dataframe/include/inviwo/dataframe/properties/dataframecolumnproperty.h
```diff
@@ -37,36 +37,69 @@ namespace inviwo {
 
 class IVW_MODULE_DATAFRAME_API DataFrameColumnProperty : public OptionPropertyInt {
```
ResearchDaniel: While you are at it you might want to consider changing it to 
ColumnOptionProperty
or 
DataFrameColumnOptionProperty

ResearchDaniel (changes requested):
Nice!

modules/dataframe/include/inviwo/dataframe/properties/dataframecolumnproperty.h
```diff
@@ -37,36 +37,69 @@ namespace inviwo {
...
 
     virtual void set(const Property* p) override;
 
 private:
-    std::shared_ptr<const DataFrame> dataframe_;
-    bool allowNone_;
-    size_t firstIndex_;
+    DataFrameInport* inport_ = nullptr;
+    const EmptySelection emptySelection_;
+    size_t defaultIndex_;
```
ResearchDaniel:
```suggestion
    int defaultIndex_;
```

ResearchDaniel (changes requested):

modules/dataframe/include/inviwo/dataframe/properties/dataframecolumnproperty.h
martinfalk: what would be the opposite of `AddNoneOption` then?

modules/dataframe/include/inviwo/dataframe/properties/dataframecolumnproperty.h
martinfalk: That is correct. Previously the options where only populated when a dataframe was set...

modules/dataframe/include/inviwo/dataframe/properties/dataframecolumnproperty.h
martinfalk: -1 would be the none option (if enabled)

modules/dataframe/include/inviwo/dataframe/properties/dataframecolumnproperty.h
martinfalk: it replaces all options.

modules/dataframe/src/properties/dataframecolumnproperty.cpp
martinfalk: calling `setPort()` causes some overhead by replacing the options which are already copied by the copy constructor of OptionPropertyInt()

modules/dataframe/src/properties/columnoptionproperty.cpp
```diff
@@ -0,0 +1,152 @@
...
+    return size() && (getSelectedValue() == -1);
+}
+
+void ColumnOptionProperty::set(const Property* srcProperty) {
+    if (auto src = dynamic_cast<const ColumnOptionProperty*>(srcProperty)) {
+        if ((src->options_.size() == 0) || (options_.size() == 0)) {
+            return;
+        }
+
+        if ((src->getSelectedValue() < 0) && (noneOption_ == AddNoneOption::No)) {
```
ResearchDaniel:
```suggestion
        if (isNoneSelected() && (noneOption_ == AddNoneOption::No)) {
```

ResearchDaniel (dismissed):

modules/dataframe/include/inviwo/dataframe/datastructures/dataframe.h
```diff
@@ -255,16 +255,24 @@ struct DataTraits<DataFrame> {
         tb(H("Number of Rows: "), rowCount);
 
         utildoc::TableBuilder tb2(doc.handle(), P::end());
-        tb2(H("Col"), H("Format"), H("Rows"), H("Name"));
+        tb2(H("Col"), H("Format"), H("Rows"), H("Name"), H("Column Range"));
         // abbreviate list of columns if there are more than 20
         const size_t ncols = (data.getNumberOfColumns() > 20) ? 10 : data.getNumberOfColumns();
 
+        auto range = [](const Column* col) {
+            if (col->hasMetaData<DoubleVec2MetaData>("DataRange")) {
```
petersteneteg: I was thinking about is a bit, is there a good reason not to just have a optional<dvec2> getRange() method in the column, instead of the metadata stuff?

modules/dataframe/include/inviwo/dataframe/properties/columnmetadatalistproperty.h
```diff
@@ -0,0 +1,90 @@
...
+
+namespace inviwo {
+
+class MetaDataOwner;
+
+/**
+ * \ingroup properties
+ * List and override column metadata of a DataFrame
+ */
+class IVW_MODULE_DATAFRAME_API ColumnMetaDataListProperty : public ListProperty {
```
petersteneteg, line 46: Should this be a ListProperty or a composite property?
Does this property handle other metadata then the range? Maybe ColumeRange(List)Property would be better?

modules/dataframe/src/properties/columnmetadatalistproperty.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+const std::string ColumnMetaDataListProperty::classIdentifier =
+    "org.inviwo.ColumnMetaDataListProperty";
+std::string ColumnMetaDataListProperty::getClassIdentifier() const { return classIdentifier; }
+
+ColumnMetaDataListProperty::ColumnMetaDataListProperty(std::string_view identifier,
+                                                       std::string_view displayName,
+                                                       InvalidationLevel invalidationLevel,
+                                                       PropertySemantics semantics)
+    : ListProperty(std::string(identifier), std::string(displayName),
```
petersteneteg, line 50: Why do you cast to string here? ListProperty should take string_views.

modules/dataframe/src/properties/columnmetadatalistproperty.cpp
```diff
@@ -0,0 +1,145 @@
...
+        if (it != toRemove.end()) {
+            (*it)->setColumnIndex(idx);
+            (*it)->setRange(columnRange(col));
+            util::erase_remove(toRemove, *it);
+        } else {
+            auto p = static_cast<ColumnMetaDataProperty*>(constructProperty(0));
+            p->setDisplayName(col->getHeader());
+            p->setColumnIndex(idx);
+            p->setRange(columnRange(col));
+        }
```
petersteneteg: You need to handle setting the default state here to I think. Otherwise we will deserialize an old range if the input data changes, which could be bad...

modules/dataframe/src/properties/columnmetadataproperty.cpp
```diff
@@ -0,0 +1,113 @@
...
+                   0.0, 0.0, InvalidationLevel::InvalidOutput, PropertySemantics("Text"))
+    , columnIndex_("columnIndex", 0u) {
+
+    addProperties(dataRange_, overrideRange_);
+    overrideRange_.addProperty(customRange_);
+
+    setCollapsed(true);
+
+    dataRange_.setReadOnly(true);
+    dataRange_.setSerializationMode(PropertySerializationMode::All);
```
petersteneteg: are you sure this will give you the right operation with respect to the comment above?

petersteneteg (changes requested):

modules/dataframe/src/properties/columnmetadatalistproperty.cpp
petersteneteg: forget this, seems I have not change this, at least not in master, and I can not figure out which brach It is in right now...

modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
```diff
@@ -110,23 +110,31 @@ class IVW_MODULE_BRUSHINGANDLINKING_API BrushingAndLinkingManager : public Seria
...
-    bool modifiedHighlight() const;
+    bool modifiedHighlight(BrushingTarget target = BrushingTarget::Row) const;
+    /**
+     * return which targets were changed since the last network evaluation
+     */
+    const std::vector<BrushingTarget>& modifiedTargets() const;
+    /**
+     * return whether \p target was modified since the last network evaluation
+     */
+    bool modifiedTarget(BrushingTarget target) const;
```
petersteneteg: For any action?

modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
```diff
@@ -110,23 +110,31 @@ class IVW_MODULE_BRUSHINGANDLINKING_API BrushingAndLinkingManager : public Seria
...
     BrushingModifications modifiedActions() const;
     /**
-     * return whether there was a filter action since the last network evaluation
+     * return whether there was a filter action on \p target since the last network evaluation
      *
-     * @return modifiedActions() & BrushingModification::Filtered
+     * @return modifiedActions() & BrushingModification::Filtered && modifiedTarget(target)
      */
-    bool modifiedFiltering() const;
+    bool modifiedFiltering(BrushingTarget target = BrushingTarget::Row) const;
```
petersteneteg: Nit: hasModifiedFiltering or isFilteringModfied would be better grammatically... 
I think the convention of having is/has prefixes for functions returning bool is quite nice. Only downside is that stl uses empty not isEmtpy, which makes it harder to be consistent. And I like to follow stl naming conventions when possible to make it easier to do generic code...

modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
```diff
@@ -234,6 +242,7 @@ class IVW_MODULE_BRUSHINGANDLINKING_API BrushingAndLinkingManager : public Seria
     InvalidationLevel invalidationLevel_;
 
     BrushingModifications modifiedActions_{flags::empty};
+    std::vector<BrushingTarget> modifiedTargets_;
```
petersteneteg: does this work would you not need something like std::map<BrushingTarget, BrushingModifications> here?

modules/brushingandlinking/src/brushingandlinkingmanager.cpp
```diff
@@ -271,6 +279,8 @@ void BrushingAndLinkingManager::setParent(BrushingAndLinkingManager* parent) {
 
         // set manager as modified since filter actions might have changed when removing the parent
         modifiedActions_ = BrushingModifications(flags::any);
+        util::push_back_unique(modifiedTargets_, BrushingTarget::Column);
```
petersteneteg: Why are these two targets hardcoded here? what about other targets?

modules/dataframe/include/inviwo/dataframe/datastructures/column.h
```diff
@@ -51,6 +51,8 @@ class IVW_MODULE_DATAFRAME_API InvalidConversion : public Exception {
     virtual ~InvalidConversion() throw() {}
 };
 
+enum class ColumnType { Unspecified, Ordinal, Categorical };
```
petersteneteg: Why Unspecified? what about Index?

modules/dataframeqt/include/inviwo/dataframeqt/processors/dataframetable.h
```diff
@@ -93,12 +90,9 @@ class IVW_MODULE_DATAFRAMEQT_API DataFrameTable : public Processor,
     IntVec2Property position_;
     BoolProperty showIndexColumn_;
     BoolProperty showCategoryIndices_;
-    BoolProperty vectorCompAsColumn_;
+    BoolProperty showFilteredRowCols_;
```
petersteneteg, line 93: might consider filtering of columns too in the future I think

petersteneteg (changes requested):
sweet, nice work on that model, Might also mention that you can now sort the table in the the changelog!

modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
martinfalk: modified actions and targets are kept separate at the moment. Basically, both modifications are sets of the recent changes.

modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
martinfalk: I am quite often struggling in those cases with `is` and `get` for singular/plural cases. `getIndicesModified()` vs `isIndexModified()`, then you would want to use `areIndicesModified()` to be grammatically correct...

modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
martinfalk: depends how fine-granular we want it to be. At the moment, both are just the superset of modifications. Also not sure we need so much detail.

modules/brushingandlinking/src/brushingandlinkingmanager.cpp
martinfalk: Then I would need to know about all used targets in this branch of network managers. Open for suggestions.

modules/dataframe/include/inviwo/dataframe/datastructures/column.h
martinfalk: good idea.

modules/dataframeqt/include/inviwo/dataframeqt/processors/dataframetable.h
martinfalk: true, in case of the TableView that would just be hiding/showing arbitrary columns.

modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
petersteneteg: all of the above would still be better then modifiedFiltering :)

modules/brushingandlinking/src/brushingandlinkingmanager.cpp
petersteneteg: having a hardcoded set seems like a very hard to find future bug waiting to happen. So I don't think that is an option unless I misunderstand what this function needs to do. Would not be enough to ask the parent/this about the used targets?

modules/dataframe/include/inviwo/dataframe/properties/columnmetadatalistproperty.h
martinfalk: Can be potentially a CompositeProperty. It can handle other metadata as well such as units.

modules/dataframe/include/inviwo/dataframe/datastructures/dataframe.h
martinfalk: will move the range to `Column` but also keep the metadata stuff for other things like units.

petersteneteg: The unit will also be a member not metadata

modules/dataframe/src/datastructures/column.cpp
```diff
@@ -31,11 +31,45 @@
...
+void Column::setRange(dvec2 range) { columnRange_ = range; }
+
+void Column::unsetRange() { columnRange_ = std::nullopt; }
+
+std::optional<dvec2> Column::getRange() const {
+    if (columnRange_.has_value()) {
+        return columnRange_;
+    }
+    return std::nullopt;
+}
```
petersteneteg, line 49:
```suggestion
std::optional<dvec2> Column::getRange() const {
    return columnRange_;
}
```

modules/brushingandlinking/src/brushingandlinkingmanager.cpp
```diff
@@ -78,30 +78,46 @@ void BrushingAndLinkingManager::brush(BrushingAction action, BrushingTarget targ
...
+                                                 BrushingAction action) const {
+    return isTargetModified(target, fromAction(action));
+}
+
+bool BrushingAndLinkingManager::isTargetModified(BrushingTarget target,
+                                                 BrushingModifications modifications) const {
+    auto it = modifications_.find(target);
+    if (it != modifications_.end()) {
+        return !(it->second | modifications).empty();
+    }
```
petersteneteg, line 118: Nit: you can do it it like this now :)
```suggestion
    if (auto it = modifications_.find(target); it != modifications_.end()) {
        return !(it->second | modifications).empty();
    }
```

petersteneteg (dismissed):

petersteneteg (dismissed):

petersteneteg (approved):

modules/brushingandlinking/src/brushingandlinkingmanager.cpp
```diff
@@ -363,6 +401,30 @@ void BrushingAndLinkingManager::propagate(BrushingAction action, BrushingTarget
...
 
+void BrushingAndLinkingManager::propagate(BrushingAction action,
+                                          const std::vector<BrushingTarget>& targets) {
+    for (auto t : targets) {
+        modifications_[t] |= fromAction(action);
+    }
+
+    if (std::holds_alternative<BrushingAndLinkingOutport*>(owner_)) {
+        auto outport = std::get<BrushingAndLinkingOutport*>(owner_);
+        outport->getProcessor()->invalidate(invalidationLevel_);
```
ResearchDaniel, line 412: Is there a need for locking the network during propagation?

modules/dataframe/include/inviwo/dataframe/datastructures/column.h
```diff
@@ -52,15 +54,19 @@ class IVW_MODULE_DATAFRAME_API InvalidConversion : public Exception {
     virtual ~InvalidConversion() throw() {}
 };
 
+enum class ColumnType { Index, Ordinal, Categorical };
+
 /**
  * @brief pure interface for representing a data column, i.e. a Buffer with a name
  */
-class IVW_MODULE_DATAFRAME_API Column {
+class IVW_MODULE_DATAFRAME_API Column : public MetaDataOwner {
```
ResearchDaniel, line 62: Does not have too much to do with this PR, but we should use metadata as a member variable instead of using inheritance in my opinion

modules/dataframe/src/io/csvreader.cpp
```diff
@@ -76,6 +82,33 @@ std::shared_ptr<DataFrame> CSVReader::readData(const std::string& fileName) {
...
+    } else if (auto* doublePrecision = std::any_cast<bool>(&value);
+               doublePrecision && key == "DoublePrecision") {
+        setEnableDoublePrecision(*doublePrecision);
+        return true;
+    }
+    return false;
+}
+
+std::any CSVReader::getOption(std::string_view key) {
+    if (key == "Delimiters") {
```
ResearchDaniel, line 102: lower case compare might be a good idea

modules/dataframe/src/processors/dataframemetadata.cpp
```diff
@@ -0,0 +1,66 @@
...
+    , outport_("outport")
+    , columns_("columns", "Column MetaData", inport_) {
+
+    addPort(inport_);
+    addPort(outport_);
+    addProperty(columns_);
+}
+
+void DataFrameMetaData::process() {
+    auto dataframe = std::make_shared<DataFrame>(*inport_.getData().get());
```
ResearchDaniel, line 57: If we did not use inheritance it would be easier to use a wrapped dataframe with meta data (using delegates) :)

modules/dataframe/include/inviwo/dataframe/properties/columnmetadatalistproperty.h
```diff
@@ -0,0 +1,96 @@
...
+     * property.
+     */
+    void setPort(DataFrameInport& inport);
+
+    void updateColumnProperties(const DataFrame& dataframe);
+
+    virtual ColumnMetaDataListProperty& resetToDefaultState() override;
+
+    /**
+     * Return the currently set column range corresponding to either min/max values of the column or
```
ResearchDaniel, line 79: This returns 0,0 in some circumstances. Should be noted

modules/dataframe/src/properties/columnmetadatalistproperty.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+dvec2 ColumnMetaDataListProperty::getRange(size_t columnIndex) const {
+    auto p = util::find_if_or_null(getProperties(), [columnIndex](const Property* p) {
+        return static_cast<const ColumnMetaDataProperty*>(p)->getColumnIndex() == columnIndex;
+    });
+    if (p) {
+        auto colprop = static_cast<const ColumnMetaDataProperty*>(p);
+        return colprop->getRange();
+    }
+    return dvec2(0.0, 0.0);
```
ResearchDaniel, line 99: Is it a good idea to return 0?

ResearchDaniel (approved):
Nice, some things that you might want to consider changing.

Also, can't we use "XXXChanged" instead of modified now that we don't have the isModified?

petersteneteg:
@ResearchDaniel might address those point in a future PR.

modules/brushingandlinking/src/brushingandlinkingmanager.cpp
martinfalk: from a manager perspective I don't think so since it will only invalidate ports/processors.

PR #1219, opened 2021-11-08
Webbrowser: Only handle clicks originating from itself, closes #1218

### LINKED ISSUE #1218: Camera interaction hickup when pressing non-picking area + releasing on (web) picking area
**Steps to reproduce**
1. Load Test workspaces->WebBrowser->web-property-sync.inv
2. Press volume, rotate, and release button on top of any picking area (GUI)
3. Press to rotate the volume again -> resulting in a jerky motion

**What happens** 
After pressing the volume its corresponding trackball expects a mouse release event to arrive at some point. 
However, when releaseing the mouse over a picking-area it will take presedence over the regular events (see line 192 in canvasopenglwidget.cpp, `CanvasQOpenGLWidget::propagateEvent(Event* e, Outport* source)`). 
Thus, if the picking event is handled it will not be forwarded as a regular event and the trackball will never get the mouse button release event. 

The image below shows event logs when pressing a non-picking area and then releasing the button on a web component. Press and release events are highlighted in yellow. 
picking

**Some thoughts**
In some way one could think that the picking object should not mark the event as used, but the picked object could also do something with the event that made it valid to mark it as used (e.g. indicating a hovering).

It would make sense if the object that handled the press also got the release event. However, I do not think we have a mechanism for that currently?

Can we handle this special case in a graceful way?
Do we want to incorporate a special case similar to when the user releases the mouse outside of the window?

ResearchDaniel:


modules/webbrowser/src/interaction/cefinteractionhandler.cpp
```diff
@@ -101,8 +101,10 @@ void CEFInteractionHandler::handlePickingEvent(PickingEvent* p) {
             }
             bool mouseUp = MouseState::Release & mouseEvent->state() ? true : false;
             int clickCount = MouseState::DoubleClick & mouseEvent->state() ? 2 : 1;
```
martinfalk, line 104: should move those two lines into the if statement

martinfalk (changes requested):

modules/webbrowser/src/interaction/cefinteractionhandler.cpp
ResearchDaniel: Good, catch. See fix

martinfalk (approved):

PR #1221, opened 2021-11-21
Core: Tracking of picking id for touch events using similar state man…

ResearchDaniel:
…ager as for mouse

Tested with touch table, not with touchpad

### DIFF
--- a/include/inviwo/core/interaction/pickingcontroller.h
+++ b/include/inviwo/core/interaction/pickingcontroller.h
@@ -66,12 +66,15 @@ private:
     void propagateEvent(TouchEvent*, EventPropagator*);
     void propagateEvent(GestureEvent*, EventPropagator*);
 
-    PickingManager::Result findPickingAction(const uvec2& coord);
+    size_t pickId(const uvec2& coord);
     std::weak_ptr<const Image> src_;
 
     PickingControllerMouseState mouseState_;
-    std::unordered_map<size_t, PickingControllerTouchState> touchStates_;
-    std::unordered_map<int, size_t> touchPointStartPickingId_;  /// {TouchPointId, PickingId}
+
+    using TouchPointIdToPickingIdMap = std::unordered_map<int, size_t>;
+    using PickingIdToTochStateMap = std::unordered_map<size_t, PickingControllerTouchState>;
+    PickingIdToTochStateMap touchStates_;
+    TouchPointIdToPickingIdMap touchPointStartPickingId_;  ///< {TouchPointId, PickingId}
 };
 
 }  // namespace inviwo
--- a/include/inviwo/core/interaction/pickingcontroller.h
+++ b/include/inviwo/core/interaction/pickingcontroller.h
@@ -70,8 +70,8 @@ private:
     std::weak_ptr<const Image> src_;
 
     PickingControllerMouseState mouseState_;
-    std::unordered_map<size_t, PickingControllerTouchState> touchStates_; 
-    std::unordered_map<int, size_t> touchPointStartPickingId_; /// {TouchPointId, PickingId}
+    std::unordered_map<size_t, PickingControllerTouchState> touchStates_;
+    std::unordered_map<int, size_t> touchPointStartPickingId_;  /// {TouchPointId, PickingId}
 };
 
 }  // namespace inviwo
--- a/include/inviwo/core/interaction/pickingcontrollertouchstate.h
+++ b/include/inviwo/core/interaction/pickingcontrollertouchstate.h
@@ -39,9 +39,14 @@ class TouchEvent;
 class EventPropagator;
 class PickingManager;
 
+/*
+ * \brief Keeps track of when touch events enter/move/exit picking objects.
+ * @see PickingController
+ */
 class IVW_CORE_API PickingControllerTouchState {
 public:
     PickingControllerTouchState(PickingManager* pickingManager = nullptr);
+    PickingControllerTouchState(PickingControllerTouchState&&);
     PickingControllerTouchState(const PickingControllerTouchState&) = delete;
     PickingControllerTouchState& operator=(const PickingControllerTouchState&) = delete;
     PickingControllerTouchState& operator=(PickingControllerTouchState&&);
--- a/include/inviwo/core/interaction/pickingcontrollertouchstate.h
+++ b/include/inviwo/core/interaction/pickingcontrollertouchstate.h
@@ -2,7 +2,7 @@
  *
  * Inviwo - Interactive Visualization Workshop
  *
- * Copyright (c) 2016-2021 Inviwo Foundation
+ * Copyright (c) 2021 Inviwo Foundation
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -26,52 +26,31 @@
  * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  *
  *********************************************************************************/
-
 #pragma once
 
 #include <inviwo/core/common/inviwocoredefine.h>
-#include <inviwo/core/interaction/pickingmanager.h>
-#include <inviwo/core/interaction/pickingcontrollermousestate.h>
-#include <inviwo/core/interaction/pickingcontrollertouchstate.h>
 
 #include <memory>
-#include <unordered_map>
 
 namespace inviwo {
 
-class Image;
-class EventPropagator;
-class Event;
-class PickingAction;
-class MouseInteractionEvent;
-class GestureEvent;
+
+struct PickingControllerTouchStateSM;
 class TouchEvent;
-class WheelEvent;
+class EventPropagator;
+class PickingManager;
 
-/**
- * \class PickingController
- * Handle mapping of interaction events into picking events and propagation of them
- * if the index of the color in the picking buffer of the src image if found by the PickingManager
- */
-class IVW_CORE_API PickingController {
+class IVW_CORE_API PickingControllerTouchState {
 public:
-    PickingController();
-    ~PickingController();
+    PickingControllerTouchState(PickingManager* pickingManager = nullptr);
+    PickingControllerTouchState(const PickingControllerTouchState&) = delete;
+    PickingControllerTouchState& operator=(const PickingControllerTouchState&) = delete;
+    PickingControllerTouchState& operator=(PickingControllerTouchState&&);
+    virtual ~PickingControllerTouchState();
 
-    void propagateEvent(Event*, EventPropagator*);
-    void setPickingSource(const std::shared_ptr<const Image>& src);
-    bool pickingEnabled() const;
+    void propagateEvent(TouchEvent* e, EventPropagator* propagator, size_t globalId);
 
-private:
-    void propagateEvent(TouchEvent*, EventPropagator*);
-    void propagateEvent(GestureEvent*, EventPropagator*);
-
-    PickingManager::Result findPickingAction(const uvec2& coord);
-    std::weak_ptr<const Image> src_;
-
-    PickingControllerMouseState mouseState_;
-    std::unordered_map<size_t, PickingControllerTouchState> touchStates_; 
-    std::unordered_map<int, size_t> touchPointStartPickingId_; /// {TouchPointId, PickingId}
+    std::unique_ptr<PickingControllerTouchStateSM> tsm;
 };
 
 }  // namespace inviwo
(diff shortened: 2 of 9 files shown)


include/inviwo/core/interaction/pickingcontroller.h
```diff
@@ -65,18 +66,12 @@ class IVW_CORE_API PickingController {
...
-    };
-
     PickingManager::Result findPickingAction(const uvec2& coord);
     std::weak_ptr<const Image> src_;
 
     PickingControllerMouseState mouseState_;
-
-    PCTouchState tstate_;
+    std::unordered_map<size_t, PickingControllerTouchState> touchStates_;
+    std::unordered_map<int, size_t> touchPointStartPickingId_;  /// {TouchPointId, PickingId}
```
martinfalk: doxygen comment should be either `///<` or `//!<`
see [docs](https://www.doxygen.nl/manual/docblocks.html#memberdoc)

include/inviwo/core/interaction/pickingcontrollertouchstate.h
```diff
@@ -0,0 +1,55 @@
...
+class TouchEvent;
+class EventPropagator;
+class PickingManager;
+
+class IVW_CORE_API PickingControllerTouchState {
+public:
+    PickingControllerTouchState(PickingManager* pickingManager = nullptr);
+    PickingControllerTouchState(const PickingControllerTouchState&) = delete;
+    PickingControllerTouchState& operator=(const PickingControllerTouchState&) = delete;
+    PickingControllerTouchState& operator=(PickingControllerTouchState&&);
```
martinfalk, line 52: if you implement this one, there should be a move constructor, too.

src/core/interaction/pickingcontroller.cpp
```diff
@@ -91,100 +91,106 @@ void PickingController::propagateEvent(Event* event, EventPropagator* propagator
...
+    //  + Single picking event per object
+    //  + Picked objects will have access to multiple touch points
+    //  - Points initially picking an object may split to multiple object: tricky to maintain state
+    //  (ignored here)
+    //
+    // Other alternative considered:
+    // Treat each touch point individually.
+    //  + Easy to track state
+    //  - Multiple picking events (Enter/Leave/Press) per object possible
+    //  - Picked objects will not have access to multiple touch points
```
martinfalk: why not? Two touch points with same ID?

src/core/interaction/pickingcontroller.cpp
```diff
@@ -91,100 +91,106 @@ void PickingController::propagateEvent(Event* event, EventPropagator* propagator
...
+    };
 
-        auto it = tstate_.pointIdToPickingId.find(point.id());
-        auto pa = it != tstate_.pointIdToPickingId.end() ? it->second : findPickingAction(coord);
+    // Strategy for treating touch points for picking objects
+    //
+    // One touch event per initially picked object.
+    //  + Single picking event per object
+    //  + Picked objects will have access to multiple touch points
+    //  - Points initially picking an object may split to multiple object: tricky to maintain state
```
martinfalk: you mean like the zoom gesture on top of an object? The start point would be the same ID in both case, independent of still on top or not.

martinfalk: Splits could be interesting for hover events when leaving & entering another object.

martinfalk: How are picking coordinates handled?

src/core/interaction/pickingcontroller.cpp
```diff
@@ -91,100 +91,106 @@ void PickingController::propagateEvent(Event* event, EventPropagator* propagator
...
+    //
+    // Other alternative considered:
+    // Treat each touch point individually.
+    //  + Easy to track state
+    //  - Multiple picking events (Enter/Leave/Press) per object possible
+    //  - Picked objects will not have access to multiple touch points
+    //
+
+    // Strategy: One touch event per initially picked object.
+    std::unordered_map<size_t, std::vector<TouchPoint>> pickingIdToTouchPoints;
```
martinfalk, line 105: Consider `using` in the class definition with a name for the map type.

src/core/interaction/pickingcontroller.cpp
```diff
@@ -91,100 +91,106 @@ void PickingController::propagateEvent(Event* event, EventPropagator* propagator
...
-        if (pa.action) {
-            pickngIdToTouchPoints[pa.index].push_back(point);
-            pickingIdToAction[pa.index] = pa.action;
+    for (auto& point : touchPoints) {
+        if (auto it = touchPointStartPickingId_.find(point.id());
+            it == touchPointStartPickingId_.end()) {
+            // Should mean that point.state() == TouchState::Started
+            const auto coord =
+                glm::clamp(point.pos(), dvec2(0.0), dvec2(e->canvasSize() - uvec2(1)));
+            touchPointStartPickingId_[point.id()] = pickId(coord);
```
martinfalk: Why do you clamp here? In case of leaving the window, not clamping will allow you to continue dragging etc. In particular if you are interested in the delta movement since the last event.

src/core/interaction/pickingcontroller.cpp
```diff
@@ -91,100 +91,106 @@ void PickingController::propagateEvent(Event* event, EventPropagator* propagator
...
-                case PickingState::Finished:
-                    return PickingPressState::Release;
-                default:
-                    return PickingPressState::None;
+        // Heuristic for obtaining picked object among the touch points:
+        // Select first non-background object
+        size_t pickedId = 0;
+        for (const auto& point : points) {
+            const auto coord =
+                glm::clamp(point.pos(), dvec2(0.0), dvec2(e->canvasSize() - uvec2(1)));
```
martinfalk: see comment above re clamping

src/core/interaction/pickingcontroller.cpp
```diff
@@ -91,100 +91,106 @@ void PickingController::propagateEvent(Event* event, EventPropagator* propagator
...
-                case PickingState::Started:
-                    return PickingPressState::Press;
-                case PickingState::Updated:
-                    return PickingPressState::Move;
-                case PickingState::Finished:
-                    return PickingPressState::Release;
-                default:
-                    return PickingPressState::None;
+        // Heuristic for obtaining picked object among the touch points:
+        // Select first non-background object
```
martinfalk: which "first"? Are the points sorted?

src/core/interaction/pickingcontroller.cpp
```diff
@@ -91,100 +91,106 @@ void PickingController::propagateEvent(Event* event, EventPropagator* propagator
...
     }
 
-    util::map_erase_remove_if(tstate_.pickingIdToPressNDC, [&](const auto& item) {
-        return pickingIdToAction.find(item.first) == pickingIdToAction.end();
-    });
-    util::map_erase_remove_if(tstate_.pickingIdToPreviousNDC, [&](const auto& item) {
-        return pickingIdToAction.find(item.first) == pickingIdToAction.end();
-    });
+    // Strategy: Treat each touch point separetly
+    // for (auto& point : touchPoints) {
```
martinfalk: remove if not needed before final PR

martinfalk (commented):
some thoughts

src/core/interaction/pickingcontroller.cpp
ResearchDaniel: The PickingEvent will only contain a single touch point if they are treated individually. Makes it quite hard to deal with multi-touch

ResearchDaniel: (in case one wants to do that with picking)

src/core/interaction/pickingcontroller.cpp
ResearchDaniel: This is mostly only relevant for touch pad where a picking object might not be pressed.
Would be nice to handle this kind of splitting, but not easy to do I think

src/core/interaction/pickingcontroller.cpp
ResearchDaniel: I think that it is just to make sure that we do not sample the picking id outside of the image. 
@petersteneteg  Can you verify?

A better way would be to return 0 picking id in that case

src/core/interaction/pickingcontroller.cpp
ResearchDaniel: Generally, yes :)

ResearchDaniel: I have added a comment

martinfalk (dismissed):
looking good to me.

src/core/interaction/pickingcontroller.cpp
petersteneteg: I'm not too fond of hiding types like that with using.. especially not at the declaration site of the member. 
I really want to be able to easily see all the types of all the members in a class. I says _alot_ about a class. And if I have to go through an indirection with some using, that's often not even at the same place in the code, that is a huge pain...

src/core/interaction/pickingcontroller.cpp
petersteneteg: Guess you can handle that in different ways, you just have to make sure not to sample outside of the picking texture...
But I guess using VoidId instead of pickId(clamped) might be more correct.

src/core/interaction/pickingcontroller.cpp
```diff
@@ -89,103 +79,87 @@ void PickingController::propagateEvent(Event* event, EventPropagator* propagator
...
-
-        auto it = tstate_.pointIdToPickingId.find(point.id());
-        auto pa = it != tstate_.pointIdToPickingId.end() ? it->second : findPickingAction(coord);
-
-        if (pa.action) {
-            pickngIdToTouchPoints[pa.index].push_back(point);
-            pickingIdToAction[pa.index] = pa.action;
+        // Not sure if this check is necessary anymore (I recall that there previously have been
+        // instabilities)
+        auto it = touchPointStartPickingId_.find(point.id());
```
petersteneteg: use `it` from line 112

src/core/interaction/pickingcontroller.cpp
```diff
@@ -198,14 +172,14 @@ void PickingController::propagateEvent(GestureEvent*, EventPropagator*) {
 
 void PickingController::setPickingSource(const std::shared_ptr<const Image>& src) { src_ = src; }
 
-PickingManager::Result PickingController::findPickingAction(const uvec2& coord) {
```
petersteneteg, line 201: is PickingManager::Result used any where else now? remove?

src/core/tests/unittests/pickingcontroller-test.cpp
```diff
@@ -564,4 +965,106 @@ TEST(PickingControllerTest, MouseDrag0to1) {
...
+    //{
+    //    SCOPED_TRACE("Press 1,1");
+    //    // uvec2 pre{1, 1};  // id = 0
+    //    uvec2 pos{1, 1};  // id = 0
+
+    //    auto event = touchEvent(TouchState::Stationary, pos, dim, device);
+    //    TestPropagator tp;
+    //    ms.propagateEvent(&event, &tp, canvas[dim.y - pos.y][pos.x]);
+    //    ASSERT_EQ(tp.events.size(), 0);
+    //}
```
petersteneteg: ?

petersteneteg (dismissed):
Looks good I think, just a couple minor questions

src/core/tests/unittests/pickingcontroller-test.cpp
ResearchDaniel: First touch on TouchScreen will always press. Removed the commented part

src/core/interaction/pickingcontroller.cpp
ResearchDaniel: Refactored to us it if possible (have to do it twice if not first found)

src/core/interaction/pickingcontroller.cpp
ResearchDaniel: Just to double check: 
We clamp based on `e->canvasSize()`, which may not be the same as `PickingController::src_`?

petersteneteg: the relevant size it that of the texture we sample I guess...

ResearchDaniel: Ok, just wanted to make sure that it did not have anything to do with layout and such

martinfalk (dismissed):

martinfalk (dismissed):

martinfalk (approved):

ResearchDaniel:
Solved in other PR

PR #1223, opened 2021-11-23
PlottingGL: scatterplot sorting

martinfalk:
* enable custom sorting based on column


modules/plottinggl/src/plotters/scatterplotgl.cpp
```diff
@@ -517,6 +528,15 @@ void ScatterPlotGL::setRadiusData(std::shared_ptr<const Column> col) {
     properties_.minRadius_.setVisible(radius_ != nullptr);
 }
 
+void ScatterPlotGL::setSortingData(std::shared_ptr<const Column> col) {
```
ResearchDaniel:
```suggestion
void ScatterPlotGL::setSortingData(const Column* col) {
```

ResearchDaniel (dismissed):
This kind of functionality is applicable to most plots. A nice to have feature would be to extract the functionality and include it in more places :)

ResearchDaniel (approved):

martinfalk:
> This kind of functionality is applicable to most plots. A nice to have feature would be to extract the functionality and include it in more places :)

Agreed. Unfortunately, the currently available plots in Inviwo take different approaches for rendering the contents. So one would have to unify that before using a global approach for sorting.

PR #1224, opened 2021-11-23
Feature/composite collapse

martinfalk:
More options for collapsing and expanding composite properties.

image


ResearchDaniel (approved):
Nice

petersteneteg (approved):
Looks good, what about those modifiers for clicking the chevron ;)

PR #1227, opened 2021-11-26
Feature/plotting cosmetic

martinfalk:


petersteneteg (dismissed):
LGTM

modules/plottinggl/src/plotters/scatterplotgl.cpp
```diff
@@ -614,32 +605,16 @@ void ScatterPlotGL::renderAxis(const size2_t& dims) {
...
 
     // Show tooltip for current item
-    if (rowIndex) {
-        if (p->getHoverState() == PickingHoverState::Move ||
-            p->getHoverState() == PickingHoverState::Enter) {
+    if (p->getHoverState() == PickingHoverState::Move ||
+        p->getHoverState() == PickingHoverState::Enter) {
 
-            tooltipCallback_.invoke(p, rowIndex.value());
+        tooltipCallback_.invoke(p, id);
```
ResearchDaniel, line 614: How do you know that pickingId = dataframe row id?

petersteneteg (dismissed):

modules/plottinggl/src/plotters/scatterplotgl.cpp
```diff
@@ -409,19 +410,9 @@ void ScatterPlotGL::plot(const size2_t& dims, IndexBuffer* indexBuffer, bool use
...
-                    if (selected && !filtered) {
-                        selectedIndices.push_back(indexCol[ind]);
-                    }
-                }
-            } else {
-                for (auto [ind, selected, filtered] : util::enumerate(selected_, filtered_)) {
-                    if (selected && !filtered) {
-                        selectedIndices.push_back(static_cast<uint32_t>(ind));
-                    }
+            for (auto [ind, selected, filtered] : util::enumerate(selected_, filtered_)) {
```
ResearchDaniel, line 413: brushing and linking contain the identifier (not the row)?

modules/plottinggl/src/processors/scatterplotprocessor.cpp
```diff
@@ -198,7 +198,7 @@ void ScatterPlotProcessor::process() {
 
     BitSet b(brushingPort_.getFilteredIndices());
     // invert the bitset to obtain the ids remaining after the filtering
-    b.flipRange(0, static_cast<uint32_t>(dataframe->getNumberOfRows()));
+    b.flipRange(0, static_cast<uint32_t>(dataframe->getIndexColumn()->getRange().y));
```
ResearchDaniel: What about min range?

ResearchDaniel (changes requested):
Not sure if the new use of brushing id is correct

modules/plottinggl/src/plotters/scatterplotgl.cpp
martinfalk: ScatterPlotGL only uses zero-based indices, i.e. row numbers. The translation to IDs is handled by the ScatterPlot processor.

modules/plottinggl/src/processors/scatterplotprocessor.cpp
martinfalk: Hm, could potentially use that one as well. However, there will hardly be any performance difference I think.

modules/plottinggl/src/plotters/scatterplotgl.cpp
martinfalk: row ids as well as picking ids are zero-based.
See https://github.com/inviwo/inviwo/blob/7fdcf8a8e65dc9c996f15e5c0649545171fcb7af/modules/plottinggl/src/processors/scatterplotprocessor.cpp#L77-L82
and 
https://github.com/inviwo/inviwo/blob/7fdcf8a8e65dc9c996f15e5c0649545171fcb7af/modules/dataframe/src/util/dataframeutil.cpp#L543-L554

modules/plottinggl/src/plotters/scatterplotgl.cpp
martinfalk: They always have. Compare for example with the PCP
https://github.com/inviwo/inviwo/blob/7fdcf8a8e65dc9c996f15e5c0649545171fcb7af/modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp#L490-L502

modules/plottinggl/src/plotters/scatterplotgl.cpp
ResearchDaniel: But this means that you cannot use brushing and linking for different data frames linked via their index column?
E.g., where the identifiers are the same, but the order is different

martinfalk: On the contrary. `ScatterPlotGL` is just concerned with the actual plotting. Indices are provided from the "outside". The **processor** for a scatterplot performs this mapping. 

https://github.com/inviwo/inviwo/blob/62f9ea434b6a44f7ea76808aa7b108dc6829fe74/modules/plottinggl/src/processors/scatterplotprocessor.cpp#L97-L107

https://github.com/inviwo/inviwo/blob/62f9ea434b6a44f7ea76808aa7b108dc6829fe74/modules/plottinggl/src/processors/scatterplotprocessor.cpp#L180-L192

ResearchDaniel (dismissed):

modules/dataframe/src/processors/dataframeexporter.cpp
```diff
@@ -148,7 +148,7 @@ void DataFrameExporter::exportAsCSV(bool separateVectorTypesIntoColumns) {
         const auto components = col->getBuffer()->getDataFormat()->getComponents();
         if (components > 1 && separateVectorTypesIntoColumns) {
             for (size_t k = 0; k < components; k++) {
-                oj = citation + col->getHeader() + ' ' + componentNames[k] + citation;
+                oj = fmt::format("{0}{} {}{0}", citation, col->getHeader(), componentNames[k]);
```
petersteneteg, line 151: Might add units too

petersteneteg: not that we parse then though....

petersteneteg (approved):

modules/dataframe/src/processors/dataframeexporter.cpp
martinfalk: That's the thing. I would expect that a DataFrame which was exported to a file and then read again results in the same DataFrame, units in the header would change that. Unless we have support for parsing, as pointed out.
I'd move that to a future PR.

PR #1230, opened 2021-12-03
Added support for structured bindings for glm types

PetiteViking:
Now you can do something like `const auto [min, max] = inputVolume->dataMap_.dataRange;`

### DIFF
--- a/include/inviwo/core/util/glm.h
+++ b/include/inviwo/core/util/glm.h
@@ -62,6 +62,7 @@
 
 #include <limits>
 #include <type_traits>
+#include <utility>
 
 namespace inviwo {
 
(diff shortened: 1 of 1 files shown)


include/inviwo/core/util/glm.h
```diff
@@ -829,7 +830,36 @@ bool almostEqual(const T& x, const T& y, int ulp = 2) {
...
+
+template <std::size_t I, auto N, class T, auto Q>
+constexpr auto&& get(glm::vec<N, T, Q>&& v) noexcept {
+    return std::move(v[I]);
+}
+
+template <std::size_t I, auto N, class T, auto Q>
+constexpr const auto&& get(const glm::vec<N, T, Q>&& v) noexcept {
+    return std::move(v[I]);
+}
```
petersteneteg: Not 100% sure about this one, since v is just pod data, using move in the return statement might mess with RVO potentially. Those cases are always tricky. 
Had forgotten that you could use auto for non-type template parameters in 17...

petersteneteg (dismissed):

include/inviwo/core/util/glm.h
JochenJankowai: According to https://coderedirect.com/questions/560354/stdmove-and-rvo-optimizations you are right. Apparently, using move in return statements is generally a bad idea.

JochenJankowai:
Looking at
https://developers.redhat.com/blog/2019/04/12/understanding-when-not-to-stdmove-in-c
and
https://coderedirect.com/questions/560354/stdmove-and-rvo-optimizations
I removed the move statements.

petersteneteg (approved):

JochenJankowai:
We might want to reopen this since the code does not compile with VS without the move statement. @petersteneteg

petersteneteg:
Dropped, the rvalues ones in the unit PR, should not make a difference for POD types I think

PR #1231, opened 2021-12-03
Feature/export

martinfalk:


modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
```diff
@@ -218,6 +219,19 @@ class IVW_MODULE_BRUSHINGANDLINKING_API BrushingAndLinkingManager : public Seria
      */
     void setParent(BrushingAndLinkingManager* parent);
 
+    /**
+     * Add a \p callback to the manager that gets called when a brushing action is triggered.
+     *
```
petersteneteg: I would not add so many empty lines...

modules/brushingandlinking/src/processors/brushingandlinkingprocessor.cpp
```diff
@@ -43,21 +49,59 @@ const ProcessorInfo BrushingAndLinkingProcessor::getProcessorInfo() const { retu
...
+        int i = maxVisibleIndices_;
+        while (it != indices.end() && i > 0) {
+            str += std::to_string(*it++) + ',';
+            --i;
+        }
+        if (it != indices.end()) {
+            str += "...";
+        } else if (!str.empty()) {
+            str.pop_back();
+        }
```
petersteneteg:
```suggestion
       auto const str = fmt::format("{}{}", fmt::join(indices.begin(),
            indices.begin() + std::min(indices.size(), maxVisibleIndices_), ", "),
            indices.size() > maxVisibleIndices_ ? "..." : "");

```

modules/brushingandlinking/src/processors/brushingandlinkingprocessor.cpp
```diff
@@ -43,21 +49,59 @@ const ProcessorInfo BrushingAndLinkingProcessor::getProcessorInfo() const { retu
...
 
-    addPort(outport_);
-    addProperties(clearSelection_, clearHighlight_, clearCols_, clearAll_);
+        LogProcessorInfo(fmt::format(
+            "{:<20} action: {}, target: {}, source: {}, indices: [{}] ({})", getDisplayName(),
```
petersteneteg: Do you need to put the display name here, don't we add that to the log anyway? 
Might also try and add some width to the other ones to align them?

src/core/properties/fileproperty.cpp
```diff
@@ -261,6 +262,12 @@ void FileProperty::requestFile() {
     }
 }
 
+void FileProperty::updateWidgets() const {
```
petersteneteg: this is quite confusing the the same function in the base class Property::updateWidgets() which is protected though. Might be more clear to make that one virtual and overriding it...

petersteneteg (changes requested):

src/core/properties/fileproperty.cpp
martinfalk: ah, was not aware of that function. Can we make Property::updateWidgets() public?

modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
martinfalk: I wanted this to be a separate paragraph so it sticks out more.

martinfalk: Probably better to use `\remark{}` https://www.doxygen.nl/manual/commands.html#cmdremark

modules/brushingandlinking/src/processors/brushingandlinkingprocessor.cpp
martinfalk: This was based on the event logger and old brushing events.

`LogProcessorInfo` will only print the identifier of the processor. Is this more useful than the display name?
https://github.com/inviwo/inviwo/blob/4567a9750a1efc3a0416ceb0bcb25ba0e96620a1/src/core/util/logcentral.cpp#L72-L76

petersteneteg: both is to much I think... the lines are already very long...

martinfalk: I opted for additional line break. In particular since target and source can be arbitrarily long. Could remove the identifier if you insist.

src/core/properties/fileproperty.cpp
petersteneteg: Probably, you could already do for(auto w : p.getWidgets()) w.updateFromProperty();
which would be the same more or less. So there is not much reason for hiding it...

include/inviwo/core/datastructures/bitset.h
```diff
@@ -64,7 +64,7 @@ class IVW_CORE_API BitSet : public Serializable {
     public:
         using iterator_category = std::bidirectional_iterator_tag;
         using value_type = uint32_t;
-        using difference_type = uint32_t;
+        using difference_type = std::ptrdiff_t;
```
petersteneteg, line 67: not int32_t?

petersteneteg (approved):

PR #1234, opened 2021-12-10
Fix/pickingcontroller touchstate2

ResearchDaniel:
Rebase of #1221

### DIFF
--- a/src/core/interaction/pickingcontroller.cpp
+++ b/src/core/interaction/pickingcontroller.cpp
@@ -55,16 +55,12 @@ void PickingController::propagateEvent(Event* event, EventPropagator* propagator
     switch (event->hash()) {
         case MouseEvent::chash(): {
             auto me = static_cast<MouseEvent*>(event);
-            const auto coord =
-                glm::clamp(me->pos(), dvec2(0.0), dvec2(me->canvasSize() - uvec2(1)));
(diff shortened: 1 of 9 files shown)


petersteneteg (approved):
LGTM

src/core/interaction/pickingcontroller.cpp
```diff
@@ -52,29 +52,15 @@ void PickingController::propagateEvent(Event* event, EventPropagator* propagator
...
-                return PickingManager::colorToIndex(uvec3(value));
-            }
-        }
-        return 0;
-    };
-
     switch (event->hash()) {
         case MouseEvent::chash(): {
             auto me = static_cast<MouseEvent*>(event);
-            const auto coord =
```
petersteneteg, line 68: @ResearchDaniel Why was this clamping removed? 
This will cause UB in the conversion from dvec2 to uvec2 for numbers <= -1.0

ResearchDaniel: The intension was to move this check to the pickId function and also avoid getting pickId on the border when it is actually outside. 
Changing the input type to dvec2 should fix it?

petersteneteg: yes, did that.

PR #1235, opened 2021-12-14
Feature/plotting fix

martinfalk:


petersteneteg (dismissed):

modules/plottinggl/src/plotters/scatterplotgl.cpp
```diff
@@ -437,6 +437,10 @@ void ScatterPlotGL::plot(const size2_t& dims, IndexBuffer* indexBuffer, bool use
         shader_.setUniform("default_color", properties_.hoverColor_.get());
 
         if (highlightDirty_) {
+            if (highlightIndexGL_.getSizeInBytes() <
```
ResearchDaniel: We might want to refactor this into a function that expands the buffer if necessary since it is used in some places and is quite verbose.

ResearchDaniel (dismissed):

modules/plottinggl/src/plotters/scatterplotgl.cpp
martinfalk: Could make this part of `BufferObject::upload()`

https://github.com/inviwo/inviwo/blob/45234589ccb2293d7d6e48db770dec4c49ac5dd6/modules/opengl/src/buffer/bufferobject.cpp#L236-L239

martinfalk: with the caveat that `upload()` then potentially informs observers on buffer re-initialization.

ResearchDaniel: Good question, perhaps too much responsibilty on the upload function? On the other hand it would simplify the code a bit and would not prevent one from doing the resizing beforehand anyway

martinfalk: Don't think so. This also avoids potential bugs, like missing to resize the buffer before uploading (hence this PR).

ResearchDaniel: +1 then :)

ResearchDaniel (approved):

PR #1236, opened 2021-12-15
BrushingAndLinking: Set BrushingAndLinkingOutport isReady to make it …

ResearchDaniel:
…ready by default. Added documentation to Outport

### DIFF
--- a/include/inviwo/core/ports/outport.h
+++ b/include/inviwo/core/ports/outport.h
@@ -49,8 +49,6 @@ class Processor;
  * \brief Abstract base class for all outports
  * The Outport can be connected to an arbitrary number of Inports.
  *
- * @note The internal isReady_ lambda function must be set by derived class, e.g.,
- * perform isReady_.setUpdate(...) in the constructor of your derived class as in DataOutport.
  */
 class IVW_CORE_API Outport : public Port {
     friend class Inport;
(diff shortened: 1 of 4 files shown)


include/inviwo/core/ports/outport.h
```diff
@@ -60,12 +63,16 @@ class IVW_CORE_API Outport : public Port {
...
      * An outport is ready if it has data and is valid. The port is set valid after its
      * processor successfully finishes processing.
+     * @pre isReady_ must be set by derived the class.
      * @see invalidate @see setValid
      */
     virtual bool isReady() const override;
 
     /**
      * Called by Processor::invalidate, will invalidate its connected inports.
+     * @pre isReady_ must be set by the derived class.
```
petersteneteg: How is isReady relevant for the invalidation/setValid? They have very little to do with each other I think

petersteneteg (changes requested):

include/inviwo/core/ports/outport.h
ResearchDaniel: isReady_ is updated on invalidate
```
void Outport::invalidate(InvalidationLevel invalidationLevel) {
    invalidationLevel_ = invalidationLevel;
    for (auto inport : connectedInports_) inport->invalidate(invalidationLevel);
    isReady_.update();
}
```

petersteneteg: true, but we call isReady_.update() in most non-const functions. I feel like that information does not belong in the documentation of each public member. It is an implementation detail. And only people that derive from outport needs to know about it. Not users of ports in general. Here I only feel that it will add more confusion.

ResearchDaniel: Sure, probably enough with the note in the class documentation

petersteneteg (dismissed):

petersteneteg (dismissed):

martinfalk (dismissed):

petersteneteg (dismissed):

ResearchDaniel:
@petersteneteg just rebased on master

petersteneteg (approved):

PR #1238, opened 2021-12-17
volume region stats, csv reader, and cleanup / refactoring

petersteneteg:
Data Reader/Writer uses string_view
Exceptions has a new constructor for Eception(IVW_CONTEXT, <format_string>. args....)
MetaDataMap uses string_view
Added a Exporter base for processor that should export that to the regression, instead of just using CanvasProcessors
Added a LayerRamResize base, to handle layerRam resizing instead of abusing the data writer impl in Cimg
Filesystem more string_view uses
Raycasters: some documentation
Rewrite of the CSVReader, added CSVWriter
Added support for txt files, to the regression tests
Removed screenshots from the the regression since they consume loads of space
Made the DataFrameTable an Exporter to check CSV files in the regression
Added regression tests for the volume region stats


modules/volume/include/inviwo/volume/processors/volumeregionstatistics.h
```diff
@@ -0,0 +1,70 @@
...
+#include <inviwo/core/processors/poolprocessor.h>
+#include <inviwo/core/properties/ordinalproperty.h>
+#include <inviwo/core/ports/volumeport.h>
+#include <inviwo/dataframe/datastructures/dataframe.h>
+
+namespace inviwo {
+
+/** \docpage{org.inviwo.VolumeRegionStatistics, Volume Region Statistics}
+ * ![](org.inviwo.VolumeRegionStatistics.png?classIdentifier=org.inviwo.VolumeRegionStatistics)
+ * Explanation of how to use the processor.
```
martinfalk: TODO

modules/volume/src/processors/volumeregionstatistics.cpp
```diff
@@ -0,0 +1,385 @@
...
+
+namespace inviwo {
+
+// The Class Identifier has to be globally unique. Use a reverse DNS naming scheme
+const ProcessorInfo VolumeRegionStatistics::processorInfo_{
+    "org.inviwo.VolumeRegionStatistics",  // Class identifier
+    "Volume Region Statistics",           // Display name
+    "Undefined",                          // Category
+    CodeState::Experimental,              // Code state
+    Tags::None,                           // Tags
```
martinfalk: add category and tags

modules/volume/src/processors/volumeregionstatistics.cpp
```diff
@@ -0,0 +1,385 @@
...
+        const auto defaultRanges =
+            util::make_array<4>([&](auto) -> std::optional<dvec2> { return {}; });
+
+        const auto volumeUnit = axes[0].unit * axes[1].unit * axes[2].unit;
+        const auto sumUnits =
+            util::make_array<4>([&](auto) { return volumeUnit * map.valueAxis.unit; });
+
+        const auto posMin = dvec3{data2world * dvec4{0.0, 0.0, 0.0, 1.0}};
+        const auto posMax = dvec3{data2world * dvec4{1.0, 1.0, 1.0, 1.0}};
+        std::array<std::optional<dvec2>, 3> sizeRange = {{dvec2{posMin[0], posMax[0]},
```
martinfalk, line 307: unused?

modules/volume/src/processors/volumeregionstatistics.cpp
```diff
@@ -0,0 +1,385 @@
...
+            addColumns(*df, channels, "Min", nRegions, valueUnits, defaultRanges, channelLabels);
+        regionMax =
+            addColumns(*df, channels, "Max", nRegions, valueUnits, defaultRanges, channelLabels);
+        regionCenter = addColumns(*df, 3, "Center", nRegions, axesUnits, sizeRange, axesNames);
+        regionCoM = addColumns(*df, channels, 3, "CoM", nRegions, axesUnits, sizeRange,
+                               channelLabels, util::make_span(axesNames));
+    }
+
+    size_t getRegion(size_t index) const {
+        return atlasRep->dispatch<size_t, dispatching::filter::UnsignedIntegerScalars>(
```
martinfalk, line 326: is that type guaranteed?

modules/volume/src/processors/volumeregionstatistics.cpp
```diff
@@ -0,0 +1,385 @@
...
+        std::vector<TStats> stats(nRegions * channels, TStats{dvec3{1.0}});
+
+        const util::IndexMapper2D regionMapper(size2_t{nRegions, channels});
+
+        const util::IndexMapper3D indexMapper(dim);
+        util::forEachVoxel(*volumeRep, [&](const size3_t& pos) {
+            const auto p = indexMapper(pos);
+            for (size_t c = 0; c < channels; ++c) {
+                const auto region = regionMapper(getRegion(p) - 1, c);
+                const auto value = getValue(p, c);
```
martinfalk, line 346: Is it sensible to query the value n times instead of querying the entire vector once before the loop?

martinfalk (commented):
neat

modules/volume/src/processors/volumeregionstatistics.cpp
petersteneteg: no

modules/volume/src/processors/volumeregionstatistics.cpp
petersteneteg: added a check

modules/volume/src/processors/volumeregionstatistics.cpp
petersteneteg: not sure if you mean the glm or the std vector... but In either case it seems it all gets inlined...

martinfalk: Doesn't this call `getvalue()` in line 323, which uses a dispatch for each channel c. Maybe better to get the values for all channels and access them in the channels loop.

petersteneteg: Had something like that at one point, but did not see a big difference, But have ot testet it for anything byt scalar volumes... if I do that I have to return a dvec4 from getValue even for scalar which might go the other way perf wise...

include/inviwo/core/io/datareader.h
```diff
@@ -81,7 +82,11 @@ class IVW_CORE_API DataReader {
      */
     virtual std::any getOption([[maybe_unused]] std::string_view key) { return std::any{}; }
 
-private:
+protected:
```
ResearchDaniel, line 85: This will make extensions_ open to inherited classes. Should add the private part below the functions

include/inviwo/core/processors/exporter.h
```diff
@@ -0,0 +1,71 @@
...
+ * \brief A base class for a Processor that might export a file. For example a CanvasProcessor
+ */
+class IVW_CORE_API Exporter {
+public:
+    virtual ~Exporter() = default;
+
+    /**
+     * Export some content to `path/name.ext` where ext is the first ext on candidateExtensions that
+     * is supported.
+     * @returns a string the the path to the exported file, of std::nullopt if no matching
```
martinfalk: -> @returns a string to the path of the exported file, or std::nullopt if no matching

include/inviwo/core/processors/exporter.h
```diff
@@ -0,0 +1,71 @@
...
+     * extensions were found
+     */
+    virtual std::optional<std::string> exportFile(
+        std::string_view path, std::string_view name,
+        const std::vector<FileExtension>& candidateExtensions, Overwrite overwrite) const = 0;
+};
+
+namespace util {
+
+std::vector<std::string> exportAllFiles(ProcessorNetwork& network, std::string_view dir,
```
martinfalk, line 69:
```suggestion
/**
 * Exports the data from all export processors in \p network into the directory \p dir using a \p nameTemplate 
 * and candidate extensions.
 * @return names of exported files
 */
std::vector<std::string> exportAllFiles(ProcessorNetwork& network, std::string_view dir,
```

modules/base/include/modules/base/io/datvolumesequencereader.h
```diff
@@ -40,20 +40,34 @@ namespace inviwo {
...
+ *                              the data in world space. (Optional, default: IdentityMatrix)
+ *   - __DatFile__      Relative path to other file to create a VolumeSequence from.
+ *   - __DataRange__    DataRange of volume (Optional, defaults to the min/max of the first volume)
+ *   - __ValueRange__   ValueRange of volume (Optional, defaults to the DataRange)
+ *   - __ValueName__    Name of Value domain (Optional, defaults to "")
+ *   - __ValueUnit__    Value Unit (Optional, defaults to Unit{})
+ *   - __SwizzleMask__  Data Channel swizzle mask @see SwizzleMask (Optional, defaults to rgba)
+ *   - __Interpolation__ Interpolation mode @see InterpolationType (Optional, defaults to Linear)
+ *   - __Wrapping__     Wrapping mode @see Wrapping3D (Optional defaults to clampAll
+ *   - __AxesNames__    Space separated axes names (Optional, defaults to "")
```
martinfalk: __AxisNames__ to be in line with __Axis1Name__?

modules/base/include/modules/base/io/datvolumesequencereader.h
```diff
@@ -40,20 +40,34 @@ namespace inviwo {
...
+ *   - __Axis(1|2|3)Name__ Axis name (Optional, defaults to "")
+ *   - __AxesUnits__    Space separated axes units (Optional, defaults to Unit{})
```
martinfalk: be consistent with Axis/Axes Names

modules/base/include/modules/base/io/ivfsequencevolumewriter.h
```diff
@@ -117,11 +118,10 @@ namespace util {
  * @see inviwo::IvfSequenceVolumeWriter
  * @see inviwo::IvfSequenceVolumeReader
  */
-
 IVW_MODULE_BASE_API std::string writeIvfVolumeSequence(const VolumeSequence& volumes,
-                                                       std::string name, std::string path,
-                                                       std::string reltivePathToElements = "",
-                                                       bool overwrite = true);
+                                                       std::string_view name, std::string_view path,
+                                                       std::string_view reltivePathToElements = "",
```
martinfalk:
```suggestion
                                                       std::string_view relativePathToElements = "",
```

modules/base/include/modules/base/io/datvolumesequencereader.h
martinfalk: "Axis names" seems to be more common, see e.g. https://www.bigmountainanalytics.com/spotfire-axis-names-bar-charts/

modules/base/include/modules/base/io/datvolumesequencereader.h
```diff
@@ -40,20 +40,34 @@ namespace inviwo {
...
- *   - __Resolution | Dimension__ The size of the data grid: nx,ny,nz (Mandatory).
- *   - __Format__ The type of values in the raw file. (Mandatory)
+ *   - __RawFile__      The name of the raw data file, should be in the same directory (Mandatory).
+ *   - __Sequences__    Number of Volume in the raw file (Optional: default 1)
+ *   - __ByteOrder__    The byte order in the raw data file. (Optional, LittleEndian|BigEndian,
+ *                      default: LittleEndian).
+ *   - __ByteOffset__   Offset in to the raw file (Optional, default: 0)
+ *   - __DataOffset__   Offset in byte to where the data starts
+ *   - __Resolution | Dimension__ the size of the data grid: nx,ny,nz (Mandatory).
+ *   - __Format__       The type of values in the raw file. (Mandatory)
```
martinfalk: mention the valid formats as mentioned in the exception in datvolumesequencereader.cpp

modules/basegl/glsl/raycasting/raycaster-template.frag
```diff
@@ -60,19 +59,27 @@ uniform float samplingRate = 2.0;
...
 void main() {
-    vec4 result = vec4(0.0);  // The accumulated color along the ray;
-    vec4 picking = vec4(0.0); // The picking color of the ray
-    float rayDepth = -1.0;    // The ray depth value (0 to ray length, -1 mean "no" depth
-    float depth = 1.0;        // The image depth
+    vec4 result = vec4(0.0);   // The accumulated color along the ray;
+    vec4 picking = vec4(0.0);  // The picking color of the ray
+    float rayDepth = -1.0;     // The ray depth value (0 to ray length, -1 mean "no" depth.
+                               // The same space as the rayPosition. Usually used to track
+                               // the depth of the "first" hit in DVR
```
martinfalk:
```suggestion
    float rayDepth = -1.0;     // The ray depth value [0, ray length], -1 means "no" depth.
                               // Uses the same space as rayPosition. Usually used to track
                               // the depth of the "first" hit in DVR.
```

modules/basegl/glsl/raycasting/raycaster-template.frag
```diff
@@ -60,19 +59,27 @@ uniform float samplingRate = 2.0;
...
+    float depth = 1.0;         // The image depth, form far (0.0) to near (1.0).
+                               // Will be overridden by rayDepth if != -1 and
+                               // written to gl_FragDepth
```
martinfalk:
```suggestion
    float depth = 1.0;         // The image depth, from far to near [0, 1].
                               // Will be overridden by rayDepth if != -1 and then
                               // written to gl_FragDepth
```

modules/basegl/glsl/raycasting/raycaster-template.frag
```diff
@@ -60,19 +59,27 @@ uniform float samplingRate = 2.0;
...
-    // entryPoint, exitPoint, entryPointDepth, exitPointDepth, rayLength, rayDirection
 
-    #pragma IVW_SHADER_SEGMENT_PLACEHOLDER_SETUP
+    // The setup placeholder is expected to define:
+    //  * entryPoint the ray start point in the volume in texture coordinates
+    //  * exitPoint the ray exit point in the volume in texture coordinates
+    //  * entryPointDepth the image depth of the entry point
+    //  * exitPointDepth the image depth of the exit point
+    //  * rayLength the distance from the start to the exit point in texture space
+    //  * rayDirection the direction of the ray in texture space, normalized.
```
martinfalk:
```suggestion
    // The setup placeholder is expected to define:
    //  * entryPoint       the ray start point in the volume in texture coordinates
    //  * exitPoint        the ray exit point in the volume in texture coordinates
    //  * entryPointDepth  the image depth of the entry point
    //  * exitPointDepth   the image depth of the exit point
    //  * rayLength        the distance from the start to the exit point in texture space
    //  * rayDirection     the direction of the ray in texture space, normalized.
```

modules/cimg/src/cimgutils.cpp
```diff
@@ -241,14 +244,14 @@ struct CImgNormalizedLayerDispatcher {
...
-    void* operator()(void* dst, const std::string& filePath, uvec2& dimensions,
+    void* operator()(void* dst, std::string_view filePath, uvec2& dimensions,
                      DataFormatId& formatId, bool rescaleToDim) {
         using P = typename DF::primitive;
 
         const DataFormatBase* dataFormat = DF::get();
 
         try {
-            cimg_library::CImg<P> img(filePath.c_str());
+            cimg_library::CImg<P> img(SafeCStr{filePath}.c_str());
```
martinfalk, line 254:
```suggestion
            cimg_library::CImg<P> img(SafeCStr<255>{filePath}.c_str());
```
file paths can be longer than the default 120 chars.

modules/dataframe/src/datastructures/column.cpp
```diff
@@ -143,22 +126,23 @@ void CategoricalColumn::append(const Column& col) {
...
 
 glm::uint32_t CategoricalColumn::addOrGetID(std::string_view str) {
-    auto it = std::find(lookUpTable_.begin(), lookUpTable_.end(), str);
-    if (it != lookUpTable_.end()) {
-        return static_cast<glm::uint32_t>(std::distance(lookUpTable_.begin(), it));
+    if (auto it = lookupMap_.find(str); it != lookupMap_.end()) {
+        return it->second;
+    } else {
+        lookUpTable_.emplace_back(str);
+        auto ind = static_cast<glm::uint32_t>(lookUpTable_.size() - 1);
```
martinfalk:
```suggestion
        auto ind = static_cast<std::uint32_t>(lookUpTable_.size() - 1);
```

modules/dataframe/src/io/csvreader.cpp
```diff
@@ -33,262 +33,519 @@
...
+const std::string& CSVReader::getLocale() const { return locale_; }
 
-    return readData(file);
+CSVReader& CSVReader::setHandleEmptyFields(EmptyField emptyField) {
+    emptyField_ = emptyField;
+    return *this;
 }
+CSVReader::EmptyField CSVReader::getHandleEmptyFields() const { return emptyField_; }
 
 bool CSVReader::setOption(std::string_view key, std::any value) {
```
martinfalk, line 124: this is a neat use case for `std::any`

modules/dataframe/src/io/csvreader.cpp
```diff
@@ -33,262 +33,519 @@
...
+#if defined(__cpp_lib_to_chars) && __cpp_lib_to_chars >= 201611L
+constexpr bool charconv = true;
+#else
+constexpr bool charconv = false;
+#endif
+}  // namespace config
+
+template <typename T>
+std::optional<T> toNumberLocale(std::string_view str) {
+    SafeCStr cstr{str};
```
martinfalk, line 282: are the default 120 chars enough? Probably better to put a size here.

modules/dataframe/src/io/csvreader.cpp
```diff
@@ -33,262 +33,519 @@
...
-            // mismatch in the number of columns
-            throw CSVDataReaderException("Column counts do not match (line " +
-                                         std::to_string(lineNumber) + ": " +
-                                         std::to_string(values.size()) + " fields; DataFrame has " +
-                                         std::to_string(maxColCount) + " columns)");
+    } else if constexpr (std::is_same_v<T, int>) {
+        char* end = nullptr;
+        typename std::remove_reference<decltype(errno)>::type errno_save = errno;
+        errno = 0;
+        int val = strtol(cstr.c_str(), &end, 10);
```
martinfalk, line 310: strtol casts to int without a warning?

modules/dataframe/src/io/csvreader.cpp
```diff
@@ -33,262 +33,519 @@
...
+
+}  // namespace util
+
+std::vector<CSVReader::CellType> CSVReader::findCellTypes(
+    size_t nCol, const std::vector<std::pair<std::string_view, size_t>>& rows,
+    size_t sampleRows) const {
+    size_t sampledRows = sampleRows;
+    std::vector<CellType> types(nCol);
+
+    bool cLocale = locale_ == "C";
```
martinfalk:
```suggestion
    const bool cLocale = locale_ == "C";
```

modules/dataframe/src/io/csvreader.cpp
```diff
@@ -33,262 +33,519 @@
...
+                                      "Invalid format on line {}, column {}", line, col);
+        }
+    };
+}
+
+std::vector<std::function<void(std::string_view, size_t, size_t)>> CSVReader::addColumns(
+    DataFrame& df, const std::vector<CellType>& types,
+    const std::vector<std::string>& headers) const {
+
+    bool cLocale = locale_ == "C";
```
martinfalk:
```suggestion
    const bool cLocale = locale_ == "C";
```

modules/dataframe/src/io/csvwriter.cpp
```diff
@@ -0,0 +1,138 @@
...
+}
+
+void CSVWriter::writeData(const DataFrame* dataFrame, std::ostream& file) const {
+
+    const std::string citation = quoteStrings ? "\"" : "";
+    const char lineTerminator = '\n';
+    const std::array<char, 4> componentNames = {'X', 'Y', 'Z', 'W'};
+
+    // headers
+    auto oj = util::make_ostream_joiner(file, delimiter);
```
martinfalk, line 65: won't this always add a separator at the back?

modules/dataframe/src/io/csvwriter.cpp
```diff
@@ -0,0 +1,138 @@
...
+            }
+        } else {
+            oj = fmt::format("{0}{1}{2: [}{0}", citation, col->getHeader(), col->getUnit());
+        }
+    }
+    file << lineTerminator;
+
+    std::vector<std::function<void(std::ostream&, size_t)>> printers;
+    for (const auto& col : *dataFrame) {
+        if ((col == dataFrame->getIndexColumn()) && !exportIndexCol) {
```
martinfalk:
```suggestion
        if ((col->getColumnType() == ColumnType::Index) && !exportIndexCol) {
```

modules/dataframe/src/io/csvwriter.cpp
```diff
@@ -0,0 +1,138 @@
...
+            continue;
+        }
+        auto df = col->getBuffer()->getDataFormat();
+        if (auto cc = dynamic_cast<const CategoricalColumn*>(col.get())) {
```
martinfalk, line 88:
```suggestion
        if (col->getColumnType() == ColumnType::Categorical) {
           auto cc = static_cast<const CategoricalColumn*>(col.get())
```

modules/dataframe/src/processors/csvsource.cpp
```diff
@@ -52,23 +50,40 @@ CSVSource::CSVSource(const std::string& file)
...
-    , doublePrecision_("doublePrecision", "Double Precision", false)
+    , firstRowIsHeaders_("firstRowIsHeaders", "First Row Contains Column Headers",
+                         CSVReader::defaultFirstRowHeader)
+    , unitsInHeaders_("unitsInHeaders", "Look for units in headers",
+                      CSVReader::defaultUnitInHeaders)
+    , unitRegexp_("unitRegexp", "Unit regexp", CSVReader::defaultUnitRegexp)
+    , delimiters_("delimiters", "Delimiters", CSVReader::defaultDelimiters)
+    , stripQuotes_("stripQuotes", "Strip surrounding quotes", CSVReader::defaultStripQuotes)
+    , doublePrecision_("doublePrecision", "Double Precision", CSVReader::defaultDoublePrecision)
+    , exampleRows_("exampleRows", "Example Rows", CSVReader::defaultNumberOfExampleRows, 0, 10000)
```
martinfalk, line 61: what's the initial value? 100?

modules/dataframe/tests/unittests/csvreader-test.cpp
```diff
@@ -347,11 +348,29 @@ TEST(CSVquotes, singlequotes) {
     EXPECT_EQ("\"b\"", col3);
 }
 
-TEST(CVSquotes, headerquotes) {
+TEST(CSVquotes, StripSingleQuotes) {
```
martinfalk:
```suggestion
TEST(CSVquotes, stripSingleQuotes) {
```

modules/dataframeqt/src/processors/dataframetable.cpp
```diff
@@ -154,4 +155,15 @@ void DataFrameTable::setWidgetSize(size2_t dim) {
...
 
+std::optional<std::string> DataFrameTable::exportFile(
+    std::string_view path, std::string_view name,
+    const std::vector<FileExtension>& candidateExtensions, Overwrite overwrite) const {
+
+    if (auto data = inport_.getData()) {
+        return util::saveData(*data, path, name, candidateExtensions, overwrite);
+    }
+
+    throw Exception("Could not find data frame", IVW_CONTEXT);
```
martinfalk:
```suggestion
    throw Exception("Could not find DataFrame", IVW_CONTEXT);
```
alternative:
```suggestion
    throw Exception("Inport has no data", IVW_CONTEXT);
```

modules/nifti/src/niftireader.cpp
```diff
@@ -162,14 +163,15 @@ const DataFormatBase* niftiDataTypeToInviwoDataFormat(const nifti_image* niftiIm
 
     return DataFormatBase::get(type, components, precision);
 }
-std::shared_ptr<NiftiReader::VolumeSequence> NiftiReader::readData(const std::string& filePath) {
+std::shared_ptr<NiftiReader::VolumeSequence> NiftiReader::readData(std::string_view filePath) {
+    checkExists(filePath);
 
     /* read input dataset, but not data */
-    std::shared_ptr<nifti_image> niftiImage(nifti_image_read(filePath.c_str(), 0),
+    std::shared_ptr<nifti_image> niftiImage(nifti_image_read(SafeCStr{filePath}.c_str(), 0),
```
martinfalk, line 170: file paths can be longer than 120
```suggestion
    std::shared_ptr<nifti_image> niftiImage(nifti_image_read(SafeCStr<255>{filePath}.c_str(), 0),
```

modules/plotting/src/utils/axisutils.cpp
```diff
@@ -464,6 +464,10 @@ std::vector<std::pair<double, vec2>> getLabelPositions(const AxisSettings& setti
     const auto axisLength = axisRange.y - axisRange.x;
     std::vector<std::pair<double, vec2>> labelPositions(tickmarks.size());
 
+    if (axisLength > 1.0E10) {
+        LogInfoCustom("labelpos", "WTF" << axisLength << " " << settings.getCaption());
```
martinfalk: :-D
```suggestion
        LogInfoCustom("labelpos", "WTF " << axisLength << " " << settings.getCaption());
```

modules/png/src/pngreader.cpp
```diff
@@ -42,46 +42,50 @@
...
 
-std::shared_ptr<inviwo::Layer> PNGLayerReader::readData(const std::string& filePath) {
-    if (!filesystem::fileExists(filePath)) throw PNGLayerReaderException(filePath);
+std::shared_ptr<inviwo::Layer> PNGLayerReader::readData(std::string_view filePath) {
+    if (!filesystem::fileExists(filePath)) throw DataReaderException(IVW_CONTEXT, filePath);
 
     auto* fp = filesystem::fopen(filePath, "rb");
-    if (!fp) throw PNGLayerReaderException("Failed to open file for reading, " + filePath);
+    if (!fp)
+        throw DataReaderException(IVW_CONTEXT, "Failed to open file for reading, {}", filePath);
```
martinfalk:
```suggestion
    if (!fp) {
        throw DataReaderException(IVW_CONTEXT, "Failed to open file for reading, {}", filePath);
    }
```

modules/png/src/pngreader.cpp
```diff
@@ -42,46 +42,50 @@
...
     }
 
     auto png_ptr = png_create_read_struct(PNG_LIBPNG_VER_STRING, NULL, NULL, NULL);
     if (!png_ptr) {
-        throw PNGLayerReaderException("Internal PNG Error: Failed to create read struct");
+        throw DataReaderException(IVW_CONTEXT,
+                                  "Internal PNG Error: Failed to create read struct: {}", name);
     }
-    util::OnScopeExit cleanup([&]() { png_destroy_read_struct(&png_ptr, NULL, NULL); });
+    util::OnScopeExit cleanup1([&]() { png_destroy_read_struct(&png_ptr, NULL, NULL); });
```
martinfalk, line 76: why `cleanup1` ?

modules/pvm/src/pvmvolumereader.cpp
```diff
@@ -77,19 +76,19 @@ std::shared_ptr<Volume> PVMVolumeReader::readPVMData(std::string filePath) {
     unsigned char* parameter = nullptr;
     unsigned char* comment = nullptr;
 
-    unsigned char* pvmdata =
-        readPVMvolume(filePath.c_str(), &udim.x, &udim.y, &udim.z, &bytesPerVoxel, &spacing.x,
-                      &spacing.y, &spacing.z, &description, &courtesy, &parameter, &comment);
+    unsigned char* pvmdata = readPVMvolume(SafeCStr{filePath}.c_str(), &udim.x, &udim.y, &udim.z,
```
martinfalk, line 79:
```suggestion
    unsigned char* pvmdata = readPVMvolume(SafeCStr<255>{filePath}.c_str(), &udim.x, &udim.y, &udim.z,
```

modules/pvm/src/pvmvolumewriter.cpp
```diff
@@ -98,38 +97,34 @@ void PVMVolumeWriter::writeData(const Volume* data, const std::string filePath)
...
-    auto commMetaData = data->getMetaData<StringMetaData>("comment");
-    if (commMetaData) {
+    if (auto commMetaData = data->getMetaData<StringMetaData>("comment")) {
         comment = new unsigned char[commMetaData->get().size() + 1];
         strncpy((char*)comment, commMetaData->get().c_str(), commMetaData->get().size());
         comment[commMetaData->get().size()] = '\0';
     }
 
-    writePVMvolume(filePath.c_str(), dataPtr, static_cast<unsigned int>(dim.x),
+    writePVMvolume(SafeCStr{filePath}.c_str(), dataPtr, static_cast<unsigned int>(dim.x),
```
martinfalk, line 127:
```suggestion
    writePVMvolume(SafeCStr<255>{filePath}.c_str(), dataPtr, static_cast<unsigned int>(dim.x),
```

modules/volume/include/inviwo/volume/processors/volumeregionstatistics.h
```diff
@@ -0,0 +1,86 @@
...
+ *   * Min for each channel, given in "Value" range
+ *   * Max for each channel, given in "Value" range
+ *   * Center (x,y,z) mean position in each region, given in `Result Space` coordinates
+ *   * Center of Mass for each channel (x, y, z), given in `Result Space` coordinates
+ *
+ * ### Inports
+ *   * __volume__ Segmented input volume
+ *   * __atlas__  Index volume, of unsigned integer type, assigning a region index to each voxel.
+ *                Has to have the same dimensions as volume. The index range is assumed to
+ *                be [0, dataMap.dataRange.y] and without gaps.
```
martinfalk, line 58: including 0 or not?

modules/volume/include/inviwo/volume/processors/volumeregionstatistics.h
```diff
@@ -0,0 +1,86 @@
...
+ *
+ * ### Outports
+ *   * __statistics__ Data Frame with the statistics for each region.
+ *
+ * ### Properties
+ *   * __Result Space__ The spacial domain of the resulting statistics. Data, Model, World, or
```
martinfalk:
```suggestion
 *   * __Result Space__ The spatial domain of the resulting statistics. Data, Model, World, or
```

modules/volume/src/processors/volumeregionstatistics.cpp
```diff
@@ -0,0 +1,402 @@
...
+#include <inviwo/core/datastructures/volume/volumeramprecision.h>
+#include <inviwo/core/datastructures/unitsystem.h>
+
+#include <numeric>
+#include <algorithm>
+
+#include <fmt/format.h>
+#include <fmt/ostream.h>
+#include <tcb/span.hpp>
+#include <fmt/ostream.h>
```
martinfalk:
```suggestion
```

src/core/common/inviwoapplication.cpp
```diff
@@ -373,6 +374,9 @@ void InviwoApplication::setApplicationUsageMode(UsageMode mode) {
     systemSettings_->applicationUsageMode_.set(mode);
 }
 
+LayerRamResizer* InviwoApplication::getLayerRamResizer() const { return layerRamResizer_; }
+void InviwoApplication::setLayerRamResizer(LayerRamResizer* obj) { layerRamResizer_ = obj; }
```
martinfalk, line 378:
```suggestion

void InviwoApplication::setLayerRamResizer(LayerRamResizer* obj) { layerRamResizer_ = obj; }
```

src/core/io/datawriter.cpp
```diff
@@ -50,7 +53,24 @@ DataWriter& DataWriter::operator=(const DataWriter& that) {
...
+    if (filesystem::fileExists(path) && overwrite == Overwrite::No)
+        throw DataWriterException(IVW_CONTEXT_CUSTOM("DataWriter"),
+                                  "Output file: {} already exists", path);
+}
+
+std::ofstream DataWriter::open(std::string_view path, std::ios_base::openmode mode) const {
+    checkOverwrite(path);
+    auto f = filesystem::ofstream(path, mode);
+    if (!f) {
+        throw FileException(IVW_CONTEXT, "could not open file '{}'", path);
```
martinfalk:
```suggestion
        throw FileException(IVW_CONTEXT, "Could not open file '{}'", path);
```

src/core/util/filesystem.cpp
```diff
@@ -93,35 +93,35 @@ std::string_view removeTrailingSlash(std::string_view path) {
...
 #endif
 }
 
-std::fstream fstream(const std::string& filename, std::ios_base::openmode mode) {
+std::fstream fstream(std::string_view filename, std::ios_base::openmode mode) {
 #if defined(_WIN32)
     return std::fstream(util::toWstring(filename), mode);
 #else
-    return std::fstream(filename, mode);
+    return std::fstream(SafeCStr{filename}.c_str(), mode);
```
martinfalk, line 108:
```suggestion
    return std::fstream(SafeCStr<255>{filename}.c_str(), mode);
```

src/core/util/filesystem.cpp
```diff
@@ -93,35 +93,35 @@ std::string_view removeTrailingSlash(std::string_view path) {
...
 #endif
 }
 
-std::ifstream ifstream(const std::string& filename, std::ios_base::openmode mode) {
+std::ifstream ifstream(std::string_view filename, std::ios_base::openmode mode) {
 #if defined(_WIN32)
     return std::ifstream(util::toWstring(filename), mode);
 #else
-    return std::ifstream(filename, mode);
+    return std::ifstream(SafeCStr{filename}.c_str(), mode);
```
martinfalk, line 116:
```suggestion
    return std::ifstream(SafeCStr<255>{filename}.c_str(), mode);
```

src/core/util/filesystem.cpp
```diff
@@ -93,35 +93,35 @@ std::string_view removeTrailingSlash(std::string_view path) {
...
 #endif
 }
 
-std::ofstream ofstream(const std::string& filename, std::ios_base::openmode mode) {
+std::ofstream ofstream(std::string_view filename, std::ios_base::openmode mode) {
 #if defined(_WIN32)
     return std::ofstream(util::toWstring(filename), mode);
 #else
-    return std::ofstream(filename, mode);
+    return std::ofstream(SafeCStr{filename}.c_str(), mode);
```
martinfalk, line 124:
```suggestion
    return std::ofstream(SafeCStr<255>{filename}.c_str(), mode);
```

src/qt/editor/inviwomainwindow.cpp
```diff
@@ -385,7 +386,10 @@ void InviwoMainWindow::saveCanvases(std::string path, std::string fileName) {
         app_->processFront();
     }
 
-    util::saveAllCanvases(app_->getProcessorNetwork(), path, fileName);
+    util::exportAllFiles(
+        *app_->getProcessorNetwork(), path, fileName,
```
martinfalk, line 390: Would it be reasonable to make this available via the file menu as well? Like "Export workspace" or similar.

tools/ivwpy/regression/reporttest.py
```diff
@@ -55,30 +53,42 @@ def __init__(self, key, differenceTolerance=0.0):
...
+class ReportTextTest(ReportTest):
+    def __init__(self, key):
+        self.key = key
+        self.message = {}
+
+    def test(self, report):
+        tests = report[self.key]["tests"]
+        for test in tests:
+            if test['diff'] > 0:
+                self.message[test['txt']] = f"{test['txt']} differs from the reference" \
```
martinfalk, line 89: why `txt` ? Were there not also csv files?

tools/ivwpy/regression/reporttest.py
```diff
@@ -98,28 +108,32 @@ def test(self, report):
...
+            ReportTest(['images', 'missing_refs'], lambda x: len(x) == 0,
+                       "Missing reference image"),
+            ReportTest(['images', 'missing_imgs'], lambda x: len(x) == 0,
+                       "Missing test image"),
+            ReportImageTest('images', settings.imageDifferenceTolerance),
+            ReportTest(['txts', 'missing_refs'], lambda x: len(x) == 0,
+                       "Missing reference text file"),
+            ReportTest(['txts', 'missing_txts'], lambda x: len(x) == 0,
+                       "Missing test text file"),
+            ReportTextTest('txts'),
```
martinfalk, line 127: csv files counting as `txts` files?

martinfalk (changes requested):
So many changes.
Maybe worthwhile putting some of those into the changelog.

modules/cimg/src/cimgutils.cpp
petersteneteg: That is fine it will go to the heap then...

modules/dataframe/src/io/csvreader.cpp
petersteneteg: at least on our systems long == int...

modules/dataframe/src/io/csvwriter.cpp
petersteneteg: no

modules/dataframe/src/processors/csvsource.cpp
petersteneteg: CSVReader::defaultNumberOfExampleRows think it might still be 50, but it will read more automatically

include/inviwo/core/common/inviwoapplication.h
```diff
@@ -490,6 +495,7 @@ class IVW_CORE_API InviwoApplication : public Singleton<InviwoApplication> {
     WorkspaceManager::SerializationHandle presetsSerializationHandle_;
     WorkspaceManager::DeserializationHandle presetsDeserializationHandle_;
     std::unique_ptr<TimerThread> timerThread_;
+    LayerRamResizer* layerRamResizer_;
```
ResearchDaniel, line 510: We should describe the approach for setting the layerRamResizer. That one should check that the currently set resizer is the owned one and if so set it to null upon deletion.

include/inviwo/core/datastructures/image/layerram.h
```diff
@@ -184,4 +184,10 @@ auto LayerRAM::dispatch(Callable&& callable, Args&&... args) const -> Result {
                                                     std::forward<Args>(args)...);
 }
 
+class IVW_CORE_API LayerRamResizer {
+public:
+    virtual ~LayerRamResizer() = default;
+    virtual bool resize(const LayerRAM& src, LayerRAM& dst) const = 0;
```
ResearchDaniel, line 198: Should state the state of dst after resizing. 
If I am not mistaken the idea is that the contents of src is interpolated to fit into dst.

include/inviwo/core/io/datareaderexception.h
```diff
@@ -32,14 +32,14 @@
 #include <inviwo/core/common/inviwocoredefine.h>
 #include <inviwo/core/util/exception.h>
 
-#include <string>
+#include <string_view>
```
ResearchDaniel, line 35: These includes could be removed?

include/inviwo/core/io/datawriter.h
```diff
@@ -52,13 +55,17 @@ class IVW_CORE_API DataWriter {
     const std::vector<FileExtension>& getExtensions() const;
     void addExtension(FileExtension ext);
 
-    bool getOverwrite() const;
-    void setOverwrite(bool val);
+    Overwrite getOverwrite() const;
+    void setOverwrite(Overwrite val);
+
+    static void checkOverwrite(std::string_view path, Overwrite overwrite);
```
ResearchDaniel, line 65: What does checkOverwrite do? 
This needs to be clarified. 
After going through the code I suggest either adding documentation (it reads reasonably well in the code), or changing the name to something in line with:
```
    static void throwErrorOnOverwriteFail(std::string_view path, Overwrite overwrite);
```

The same goes for checkExists, which is perhaps more obvious. Better to be excplicit about the error throwing in my opinion.

include/inviwo/core/io/datawriterexception.h
```diff
@@ -37,9 +37,7 @@ namespace inviwo {
 
 class IVW_CORE_API DataWriterException : public Exception {
 public:
-    DataWriterException(const std::string& message = "",
-                        ExceptionContext context = ExceptionContext());
-    virtual ~DataWriterException() noexcept = default;
+    using Exception::Exception;
```
ResearchDaniel, line 40: Can remove the string include above now

include/inviwo/core/io/datawriterutil.h
```diff
@@ -0,0 +1,79 @@
...
+    } else {
+        throw DataWriterException(
+            fmt::format("Could not find a writer for {} of the specified extension {}", path,
+                        extension.toString()),
+            IVW_CONTEXT_CUSTOM("datawriterutil"));
+    }
+}
+
+template <typename T>
+std::optional<std::string> saveData(const T& data, std::string_view path, std::string_view name,
```
ResearchDaniel, line 80: path is not entirely true and inconsistent with function above
```suggestion
std::optional<std::string> saveData(const T& data, std::string_view dir, std::string_view name,
```

include/inviwo/core/processors/exporter.h
```diff
@@ -0,0 +1,71 @@
...
+#include <string>
+
+namespace inviwo {
+
+class ProcessorNetwork;
+
+/**
+ * \brief A base class for a Processor that might export a file. For example a CanvasProcessor
+ */
+class IVW_CORE_API Exporter {
```
ResearchDaniel, line 47: The name is a bit broad for its purpose. Could potentially clash with something else in the inviwo namespace.

Perhaps put it in the util namespace along with the function below?

include/inviwo/core/processors/exporter.h
ResearchDaniel: returns nullopt if overwrite is set to No and a file exist?

ResearchDaniel: Could also add info about if exceptions are thrown or captured by the function

include/inviwo/core/util/exception.h
```diff
@@ -73,39 +80,32 @@ std::basic_ostream<Elem, Traits>& operator<<(std::basic_ostream<Elem, Traits>& s
 
 class IVW_CORE_API RangeException : public Exception {
 public:
-    RangeException(std::string_view message = "", ExceptionContext context = ExceptionContext());
-    virtual ~RangeException() noexcept = default;
+    using Exception::Exception;
```
ResearchDaniel, line 83: That's neat!

modules/nifti/src/niftireader.cpp
petersteneteg: thats fine, 120 is only for the stack

modules/plotting/src/utils/axisutils.cpp
petersteneteg: Ahh, did some very annoying Nan debugging there...

modules/png/src/pngreader.cpp
petersteneteg: since there is a cleanup2 below. you can not have them use the same OnScopeExit, that will delete them in thw wrong order. We have UB here before.

modules/pvm/src/pvmvolumereader.cpp
petersteneteg: no need

modules/pvm/src/pvmvolumewriter.cpp
petersteneteg: no need

modules/volume/include/inviwo/volume/processors/volumeregionstatistics.h
petersteneteg: [] means inclusive, will have to come back to this at some later point though.

modules/volume/src/processors/volumeregionstatistics.cpp
petersteneteg: ?

tools/ivwpy/regression/reporttest.py
petersteneteg: txt mean any text based file, xml, csv, txt, anythnig that you can diff with a textual diff more or less

modules/dataframe/src/io/csvwriter.cpp
petersteneteg: Think I will stick with the dynamic cast, there is no real perf aspect here

src/qt/editor/inviwomainwindow.cpp
petersteneteg: you can create a regression tests... that will save them...

petersteneteg: and the workspace file...

modules/base/src/io/ivfvolumewriter.cpp
```diff
@@ -46,21 +47,16 @@ IvfVolumeWriter& IvfVolumeWriter::operator=(const IvfVolumeWriter& that) = defau
...
+    const auto rawPath = filesystem::replaceFileExtension(filePath, "raw");
 
-    if (filesystem::fileExists(filePath) && !overwrite)
-        throw DataWriterException("Output file: " + filePath + " already exists",
-                                  IVW_CONTEXT_CUSTOM("util::writeIvfVolume"));
-
-    if (filesystem::fileExists(rawPath) && !overwrite)
-        throw DataWriterException("Output file: " + rawPath + " already exists",
-                                  IVW_CONTEXT_CUSTOM("util::writeIvfVolume"));
+    DataWriter::checkOverwrite(filePath, overwrite);
```
ResearchDaniel, line 58: Copy-paste error
```suggestion
    DataWriter::checkExists(filePath, overwrite);
```

modules/basegl/glsl/raycasting/raycaster-template.frag
```diff
@@ -60,19 +59,27 @@ uniform float samplingRate = 2.0;
...
-    vec4 result = vec4(0.0);  // The accumulated color along the ray;
-    vec4 picking = vec4(0.0); // The picking color of the ray
-    float rayDepth = -1.0;    // The ray depth value (0 to ray length, -1 mean "no" depth
-    float depth = 1.0;        // The image depth
+    vec4 result = vec4(0.0);   // The accumulated color along the ray;
+    vec4 picking = vec4(0.0);  // The picking color of the ray
+    float rayDepth = -1.0;     // The ray depth value (0 to ray length, -1 mean "no" depth.
+                               // The same space as the rayPosition. Usually used to track
+                               // the depth of the "first" hit in DVR
+    float depth = 1.0;         // The image depth, form far (0.0) to near (1.0).
```
ResearchDaniel:
```suggestion
    float depth = 1.0;         // The image depth, form far (0.0) to near (1.0).
                               // gl_FragDepth is set to rayDepth/rayLength if rayDepth != 1, otherwise depth.
```

modules/basegl/include/modules/basegl/shadercomponents/sampletransformcomponent.h
```diff
@@ -52,6 +53,7 @@ class IVW_MODULE_BASEGL_API SampleTransformComponent : public ShaderComponent {
     virtual std::vector<Segment> getSegments() override;
 
 private:
+    CompositeProperty sampleTransform_;
```
ResearchDaniel, line 56: it is not really the sample (value) that is transformed? It is the position?
Might want to consider changing this to rayPosTransform/RayPosTransformComponent

modules/basegl/src/processors/raycasting/atlasvolumeraycaster.cpp
```diff
@@ -36,11 +36,11 @@ namespace inviwo {
 
 // The Class Identifier has to be globally unique. Use a reverse DNS naming scheme
 const ProcessorInfo AtlasVolumeRaycaster::processorInfo_{
-    "org.inviwo.AtlasVolumeRaycaster",          // Class identifier
-    "Atlas Volume Raycaster",                   // Display name
-    "Volume Rendering",                         // Category
-    CodeState::Experimental,                    // Code state
-    Tag::GL | Tag{"Volume"} | Tag{"Raycaster"}  // Tags
+    "org.inviwo.AtlasVolumeRaycaster",                         // Class identifier
+    "Atlas Volume Raycaster",                                  // Display name
```
ResearchDaniel, line 40: Seems like Atlas is used in some places and Region is used in others?
We should make this consistent

modules/dataframe/include/inviwo/dataframe/datastructures/column.h
```diff
@@ -313,6 +313,8 @@ class IVW_MODULE_DATAFRAME_API CategoricalColumn : public TemplateColumn<std::ui
 
     virtual void add(std::string_view value) override;
 
+    std::function<void(std::string_view)> addMany();
```
ResearchDaniel: This deserves an explanation.

modules/dataframe/include/inviwo/dataframe/io/csvreader.h
```diff
@@ -60,27 +47,48 @@ class IVW_MODULE_DATAFRAME_API CSVDataReaderException : public DataReaderExcepti
...
 
+    CSVReader& setNumberOfExampleRows(size_t rows);
+    size_t getNumberOfExamplesRows() const;
+
+    CSVReader& setLocale(std::string_view loc);
+    const std::string& getLocale() const;
+
+    enum class EmptyField { Throw, DefaultConstruct, NanOrZero };
+
+    CSVReader& setHandleEmptyFields(EmptyField emptyField);
```
ResearchDaniel, line 106: Would be good to mention what it defaults to since it cannot be specified in the constructor

modules/dataframe/include/inviwo/dataframe/io/csvreader.h
```diff
@@ -60,27 +47,48 @@ class IVW_MODULE_DATAFRAME_API CSVDataReaderException : public DataReaderExcepti
...
+    CSVReader& setEnableDoublePrecision(bool doubleprec);
     bool hasDoublePrecision() const;
 
+    CSVReader& setNumberOfExampleRows(size_t rows);
+    size_t getNumberOfExamplesRows() const;
+
+    CSVReader& setLocale(std::string_view loc);
+    const std::string& getLocale() const;
+
+    enum class EmptyField { Throw, DefaultConstruct, NanOrZero };
```
ResearchDaniel: NanOrZero? I think that these options should be separated

modules/dataframe/include/inviwo/dataframe/io/csvreader.h
```diff
@@ -111,10 +119,40 @@ class IVW_MODULE_DATAFRAME_API CSVReader : public DataReaderType<DataFrame> {
...
+    static constexpr bool defaultUnitInHeaders = true;
+    static constexpr std::string_view defaultUnitRegexp = R"((.*)\[(.*)\])";
+    static constexpr bool defaultDoublePrecision = false;
+    static constexpr size_t defaultNumberOfExampleRows = 50;
+    static constexpr std::string_view defaultLocale = "C";
+    static constexpr EmptyField defaultEmptyField = EmptyField::NanOrZero;
+
 private:
+    struct CellType {
+        size_t integer;
```
ResearchDaniel, line 185: nInteger
nReal
nString

modules/dataframe/include/inviwo/dataframe/io/csvreader.h
```diff
@@ -111,10 +119,40 @@ class IVW_MODULE_DATAFRAME_API CSVReader : public DataReaderType<DataFrame> {
...
+        DataFrame& df, const std::vector<CellType>& types,
+        const std::vector<std::string>& headers) const;
+
     std::string delimiters_;
+    bool stripQuotes_;
     bool firstRowHeader_;
+    bool unitsInHeaders_;
+    std::string unitRegexp_;
     bool doublePrecision_;
+    size_t exampleRows_;
```
ResearchDaniel, line 204: nExampleRows_

modules/dataframe/src/io/csvreader.cpp
```diff
@@ -33,262 +33,519 @@
...
 
-    if (stream.bad() || stream.fail()) {
-        throw CSVDataReaderException("Input stream in a bad state", IVW_CONTEXT);
-    }
+template <typename Func>
+size_t parse(std::string_view str, std::string_view delimiters, std::optional<size_t> expectedParts,
+             std::optional<size_t> refLine, Func&& func) {
+
+    std::string lookFor{delimiters};
+    lookFor.push_back('"');
```
ResearchDaniel: Suggest to include an `empty` varable and use it in the code below to make it more explicit
```suggestion
    const auto empty = '"';
    lookFor.push_back(empty); 
```

modules/dataframe/src/io/csvreader.cpp
```diff
@@ -33,262 +33,519 @@
...
+                            ++types[index].real;
+                        } else {
+                            ++types[index].string;
+                        }
+                    });
+        if (i > sampledRows) {
+            if (std::any_of(types.begin(), types.end(), [](const CellType& type) {
+                    return type.integer == 0 && type.real == 0 && type.string == 0;
+                })) {
+                sampledRows *= 2;
```
ResearchDaniel, line 381: TImes 2?
Should be
`sampledRows = i`?

modules/dataframe/src/properties/columnmetadataproperty.cpp
```diff
@@ -55,23 +56,27 @@ ColumnMetaDataProperty::ColumnMetaDataProperty(std::string_view identifier,
         .visibilityDependsOn(*this, [](const auto& p) { return !p.getReadOnly(); })
         .setCurrentStateAsDefault();
 
-    addProperties(header_, range_, unit_, drop_);
+    type_.setReadOnly(true);
```
ResearchDaniel, line 59: Future PR: I think that it would be valid to change the type sometimes. E.g., to make something that was interpreted as integers to categorical. At least if the plotting behaves different for categorical compared to numerical columns

modules/dataframe/tests/unittests/csvreader-test.cpp
```diff
@@ -190,16 +190,16 @@ TEST(CSVdata, delimiterBeforeLineBreakWithHeader) {
     ASSERT_EQ(2, dataframe->getNumberOfRows()) << "row count does not match (empty line)";
 }
 
-TEST(CSVdata, ignoreEmptyLine) {
-    // test whether rows with only delimiters, i.e. ",,,,," are ignored
+TEST(CSVdata, allEmptyElementLine) {
+    // test whether rows with only delimiters, i.e. ",,,,," are keept
```
ResearchDaniel:
```suggestion
    // test whether rows with only delimiters, i.e. ",,,,," are kept
```

modules/opengl/include/modules/opengl/openglexception.h
```diff
@@ -36,15 +36,13 @@ namespace inviwo {
...
 class IVW_MODULE_OPENGL_API OpenGLException : public Exception {
 public:
-    OpenGLException(const std::string& message = "", ExceptionContext context = ExceptionContext());
-    virtual ~OpenGLException() noexcept = default;
+    using Exception::Exception;
 };
 
 class IVW_MODULE_OPENGL_API OpenGLInitException : public ModuleInitException {
 public:
     OpenGLInitException(const std::string& message = "",
```
ResearchDaniel, line 44: Cannot use?
`using ModuleInitException;`

modules/volume/include/inviwo/volume/processors/volumeregionstatistics.h
```diff
@@ -0,0 +1,86 @@
...
+ *                      Index, defaults to World.
+ *
+ */
+class IVW_MODULE_VOLUME_API VolumeRegionStatistics : public PoolProcessor {
```
ResearchDaniel, line 68: We mix atlas and region... We should use one of them

modules/volume/src/processors/volumeregionstatistics.cpp
```diff
@@ -0,0 +1,402 @@
...
+            for (size_t c = 0; c < channels; ++c) {
+                const auto region = regionMapper(getRegion(p) - 1, c);
+                const auto value = getValue(p, c);
+                const auto dpos = dvec3{index2data * dvec4{pos, 1.0}};
+                stats[region].add(dpos, value);
+            }
+        });
+
+        for (auto&& [i, stat] : util::enumerate(stats)) {
+            const auto [region, c] = regionMapper(i);
```
ResearchDaniel, line 353: I think that iterating over regions would be quite useful in other places. If we would have an Atlas/Region volume it would be nice to expose such functionality there. Such a class could also ensure that the indices are consecutive and provide mappings to the original indices if need be

ResearchDaniel (changes requested):
Nice!
Please separate such contributions into multiple PRs in the future.

include/inviwo/core/common/inviwoapplication.h
petersteneteg: the modules usually own everything the app/factories only hold references. So I do the same here. the module is responsible for unsetting the resize when it is removed. 
If a module is unloaded one can not use the resizer even if it was owned by the app.

include/inviwo/core/io/datareader.h
petersteneteg: I don't see why that would be a problem

include/inviwo/core/common/inviwoapplication.h
ResearchDaniel: I changed my comment as I came to the same conclusion. Not sure if your comment is for the old one or the one I updated

include/inviwo/core/io/datareader.h
ResearchDaniel: I don’t think it will be, more of a “keep as closed as possible” principle. It was like that before the change

modules/dataframe/include/inviwo/dataframe/io/csvreader.h
petersteneteg: They already are, DefaultConstruct will generally be 0 for all types, NaNOrZero, sets NaNs for floating points and 0 for others. (for categorical ones the empty sting is not handled as a special case but just a an other value)

modules/dataframe/src/io/csvreader.cpp
petersteneteg: No, I increase the number sampled rows until we find non-empty ones.

modules/dataframe/include/inviwo/dataframe/io/csvreader.h
ResearchDaniel: Ah, I think that from a user perspective we could use Zero, or EmptyOrZero, instead of DefaultConstruct then. While DefaultConstruct might make sense from a pure programmin perspective I think the other alternatives are more informative.
```suggestion
    enum class EmptyField { Throw, 
                            EmptyOrZero, ///< "" for strings/categorical and 0 for numerical values
                            NanOrZero    ///< "" for strings/categorical, NaN for floating point, and 0 for integer values
                                            };
```

include/inviwo/core/io/datareader.h
petersteneteg: Don't think “keep as closed as possible” makes any sense unless there is an invariant that you want to protect. And for the extensions_ I cant't think of any.

petersteneteg:
I wanted to make smaller PRs but don't know how to do that in a effect manner. it is very hard to break these kinds of refactorings apart into smaller pieces that will still compile without a ton of trail and error. And even if you do you end up with a list of several PRs that all depend on each other so they have to be merged order, which menas you have to merge the first, push /rebase the next one, wait for CI /reviews, and then merge, then push the next... which ends up being really slow. If you have any suggestions for how to do that more efficiently that would be appreciated

include/inviwo/core/common/inviwoapplication.h
petersteneteg: added some comment with the getLayerRamResizer and setLayerRamResizer functions

include/inviwo/core/io/datareaderexception.h
petersteneteg: yes, different PR...

include/inviwo/core/io/datawriter.h
petersteneteg: I added some doc string

include/inviwo/core/io/datawriterexception.h
petersteneteg: different PR

include/inviwo/core/io/datawriterutil.h
petersteneteg: changed the other one to filepath

petersteneteg: and added comments

include/inviwo/core/processors/exporter.h
petersteneteg: well we control what is in the Inviwo namespace so I don't worry so much about that.

modules/base/src/io/ivfvolumewriter.cpp
petersteneteg: should have checked the raw file

modules/basegl/include/modules/basegl/shadercomponents/sampletransformcomponent.h
petersteneteg: In this context I don't think of sample as equivalent with value, the value is the result of the sample, you take a sample to get a value, and here I transform how that sample is taken.

modules/basegl/src/processors/raycasting/atlasvolumeraycaster.cpp
petersteneteg: yes, maybe a different PR. Not sure what it the bast names yet... Atlas/Region/Segment/... all have plus and minuses.

modules/dataframe/include/inviwo/dataframe/datastructures/column.h
petersteneteg: mmm

modules/dataframe/include/inviwo/dataframe/io/csvreader.h
petersteneteg: mm

modules/dataframe/include/inviwo/dataframe/io/csvreader.h
petersteneteg: I renamed the struct instead

modules/dataframe/include/inviwo/dataframe/io/csvreader.h
petersteneteg: seems excessive

modules/dataframe/src/io/csvreader.cpp
petersteneteg: it not empty, its a quote, added a comment.

modules/opengl/include/modules/opengl/openglexception.h
petersteneteg: out of scope

modules/volume/src/processors/volumeregionstatistics.cpp
petersteneteg: Im sure that could be nice, but the regionMapper here is just a regular IndexMapper2D.

modules/volume/include/inviwo/volume/processors/volumeregionstatistics.h
petersteneteg: maybe I will rename the atlas renderer at some point.

modules/dataframe/include/inviwo/dataframe/datastructures/column.h
```diff
@@ -313,6 +315,24 @@ class IVW_MODULE_DATAFRAME_API CategoricalColumn : public TemplateColumn<std::ui
...
+     * "outside". This function returns a callable object `AddMany` that enables this optimization
+     * for CategoricalColumns. For example:
+     * ```{.cpp}
+     * CategoricalColumn col{...};
+     * auto adder = col.addMany();
+     * for (const auto& str : some_strings) {
+     *     adder(str);
+     * }
+     * ```
+     */
```
martinfalk:
```suggestion
     * For CategoricalColumn you cannot do the trick of pulling out the data container
     * `auto& data = templateColumn->getEditableRAMRepresentation()->getDataContainer();`
     * like for regular TemplateColumns to efficiently add many elements.
     * This is because adding element to the CategoricalColumn first has to find the string in the
     * lookup table and then insert the corresponding index into the data container. And we can't
     * do that from the "outside". This function returns a callable object `AddMany` that enables
     * this optimization for CategoricalColumns.
     * For example:
     * ```{.cpp}
     * CategoricalColumn col{...};
     * auto adder = col.addMany();
     * for (const auto& str : some_strings) {
     *     adder(str);
     * }
     * ```
     */
```

include/inviwo/core/datastructures/image/layerram.h
ResearchDaniel: I think that this is still quite unclear. 
The layers do not change size, so there is no resize going on.
Come to think of it I think that it might be resolved if we call it LayerRamResampler.

include/inviwo/core/datastructures/image/layerram.h
```diff
@@ -184,4 +184,15 @@ auto LayerRAM::dispatch(Callable&& callable, Args&&... args) const -> Result {
...
 
+class IVW_CORE_API LayerRamResizer {
+public:
+    virtual ~LayerRamResizer() = default;
+
+    /**
+     * Resize and copy src to dst doing some kind of content scaling
+     * @see CIMGLayerRamResizer
+     */
+    virtual bool resize(const LayerRAM& src, LayerRAM& dst) const = 0;
```
ResearchDaniel, line 198:
```suggestion
     /**
     * Resample the contents of src to dst while taking their potentially different sizes into account.
     * @see CIMGLayerRamResizer
     */
    virtual bool resample(const LayerRAM& src, LayerRAM& dst) const = 0;
```

ResearchDaniel (dismissed):
I would change the LayerRamResizer to LayerRamResampler
That would make its intent much clearer

ResearchDaniel:
> I wanted to make smaller PRs but don't know how to do that in a effect manner. it is very hard to break these kinds of refactorings apart into smaller pieces that will still compile without a ton of trail and error...

I understand that this evolved over time. My browser even freezes when github parses this many files though. I think that most string_view changes could be separated into a different PR, which would probably reduce the number of files by 2.

modules/dataframe/src/io/csvreader.cpp
```diff
@@ -33,262 +33,519 @@
...
+#include <charconv>
+#include <cstdlib>
+#include <cerrno>
+#include <clocale>
+#include <functional>
+#include <type_traits>
+#include <regex>
 
-namespace inviwo {
+#include <inviwo/core/util/clock.h>
```
martinfalk: remove?

include/inviwo/core/datastructures/image/layerram.h
martinfalk: // copy the data from \p src to \p dst. This might involve up or down scaling if the dimensions do not match.
// The dimensions of both \p src and \p dst will not change.

modules/volume/include/inviwo/volume/processors/volumeregionstatistics.h
martinfalk: just wanted to make sure that is what you meant. Not that 0 is some reserved value.

martinfalk (changes requested):
Some minor things

ResearchDaniel (approved):
Would still rename the resizer to resampler :)

martinfalk (approved):

PR #1240, opened 2021-12-17
iwyu

petersteneteg:

**Always**

**New features**

**Updated feature / Fixes** 

- Operating system
- Compiler/IDE
- QT/CMake/Python versions


ResearchDaniel (dismissed):
No idea how to use it but the results look ok.
Could you add a ReadMe.md file in the folder where you explain that it is IWYU stands for include what you use and how to use it?

tools/iwyu/glmimp.py
```diff
@@ -0,0 +1,57 @@
...
+from pathlib import Path
+
+
+def begins_with(alist, prefix):
+    if len(prefix) > len(alist):
+        return False
+    return alist[:len(prefix)] == prefix
+
+
+glm_base = "/Users/peter/Work/Inviwo/inviwo/ext/glm/glm"
```
JochenJankowai: Is this a generated file? Otherwise, this path should probably be relative.

JochenJankowai:
I am not too familiar with Jenkins' capabilities but isn't this something that could run there?

tools/iwyu/glmimp.py
petersteneteg: Yes, but you only have to do it ones kind of, or of glm does some large changes...

petersteneteg:
Yes, I would like to run it on Jenkins or GitHub Actions at some point... But at the moment it changes quite a lot of stuff so the interaction of it has to be done with a bit of care to avoid breaking way to much code...

modules/basegl/src/processors/multichannelraycaster.cpp
```diff
@@ -32,9 +32,11 @@
 #include <modules/opengl/volume/volumegl.h>
 #include <modules/opengl/image/layergl.h>
 #include <modules/opengl/texture/textureunit.h>
+
```
martinfalk, line 31: empty lines?

martinfalk (dismissed):
what about qt6_x header files?

Issue #1237, opened 2021-12-16
Evaluating a processor based on subsets of brushing types


ResearchDaniel:
With the really nice new brushing and linking features I have come across an issue with processors that only require a subset of the brushing features. For example only filtering. 
Each hover event will invalidate the processor and, while it is possible to check for modifications and skip computations in that processor it will still cause connected processors to evaluate (even though no new data is set on the outport). 

What would be the recommended action to avoid subsequent processing here? 
Is setting a new isReady_ function the way to go? I would assume that it messes with other things.

@martinfalk @petersteneteg

petersteneteg:
This should not have anything to do with isReady, the key here is in overloading Inport::invalidate. But for that to work we probably need to refactor the B&L a bit. I can put this on my todo...

ResearchDaniel:
Adding a BrushingModification invalidation filter to constructor and/or a settter/getter for it should be fairly straightforward perhaps

ResearchDaniel:
Seems to work well. Will make a PR

petersteneteg:
There are some things related to the propagation in the B&L manager that needs to change for it to work properly I think.

martinfalk:
Implemented in #1241

PR #1241, opened 2021-12-20
BrushingAndLinking: Enable processor invalidation for a subset of bru…

ResearchDaniel:
…shing actions.

Closes issue #1237

### DIFF
--- a/modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
+++ b/modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
@@ -92,7 +92,8 @@ struct IVW_MODULE_BRUSHINGANDLINKING_API BrushingTargetsInvalidationLevel {
  */
 class IVW_MODULE_BRUSHINGANDLINKING_API BrushingAndLinkingManager : public Serializable {
 public:
-    static inline const std::vector<BrushingTarget> AnyBrushingTarget = {}; ///< Helper for invalidation level initialization
+    static inline const std::vector<BrushingTarget> AnyBrushingTarget =
+        {};  ///< Helper for invalidation level initialization
     /**
      * @code
      *  // Only invalidate processor on row filtering and column selection.
--- a/modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
+++ b/modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
@@ -53,15 +53,16 @@ class Deserializer;
 /**
  * Manages brushing and linking events for filtering, selecting, and highlighting. When initialized
  * with a BrushingAndLinking inport, changes are propagated using this port if connected.
- * A Processor can choose to only handle a certain type of 
+ * A Processor can choose to only handle a certain type of
  */
 class IVW_MODULE_BRUSHINGANDLINKING_API BrushingAndLinkingManager : public Serializable {
 public:
     /**
      * @param inport owner of the manager.
      * @param invalidateOn for the modifications that should invalidate the processor
-     * (Processor::process will not be called). 
-     * @param invalidationLevel that will be propagated to the port and thereby upwards in the network.
+     * (Processor::process will not be called).
+     * @param invalidationLevel that will be propagated to the port and thereby upwards in the
+     * network.
      */
     BrushingAndLinkingManager(
         BrushingAndLinkingInport* inport,
@@ -70,8 +71,9 @@ public:
     /**
      * @param outport owner of the manager.
      * @param invalidateOn for the modifications that should invalidate the processor
-     * (Processor::process will not be called). 
-     * @param invalidationLevel that will be propagated to the port and thereby upwards in the network.
+     * (Processor::process will not be called).
+     * @param invalidationLevel that will be propagated to the port and thereby upwards in the
+     * network.
      */
     BrushingAndLinkingManager(
         BrushingAndLinkingOutport* outport,
@@ -289,7 +291,7 @@ private:
     std::unordered_set<BrushingAndLinkingManager*> children_;
     InvalidationLevel invalidationLevel_;
     BrushingModifications
-        invalidateOn_;  ///< Will cause owners to only invalidate processor for these modifications. 
+        invalidateOn_;  ///< Will cause owners to only invalidate processor for these modifications.
 
     std::unordered_map<BrushingTarget, BrushingModifications> modifications_;
 
--- a/modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
+++ b/modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
@@ -92,6 +92,7 @@ struct IVW_MODULE_BRUSHINGANDLINKING_API BrushingTargetsInvalidationLevel {
  */
 class IVW_MODULE_BRUSHINGANDLINKING_API BrushingAndLinkingManager : public Serializable {
 public:
+    static inline const std::vector<BrushingTarget> AnyBrushingTarget = {}; ///< Helper for invalidation level initialization
     /**
      * @code
      *  // Only invalidate processor on row filtering and column selection.
@@ -102,10 +103,10 @@ public:
      * InvalidationLevel::InvalidOutput}
      *  }
      *  );
-     *  // Invalidate processor on any type of action for filtering or column selection.
+     *  // Invalidate processor on filtering or selection for any target.
      *  BrushingAndLinkingManager(port,
      *  {
-     *    {BrushingModification::Filtered | BrushingModification::Selected,
+     *    {AnyBrushingTarget, BrushingModification::Filtered | BrushingModification::Selected,
      * InvalidationLevel::InvalidOutput}
      *  }
      *  );
@@ -117,7 +118,7 @@ public:
      */
     BrushingAndLinkingManager(BrushingAndLinkingInport* inport,
                               std::vector<BrushingTargetsInvalidationLevel> invalidationLevels = {
-                                  {BrushingModifications(flags::any),
+                                  {AnyBrushingTarget, BrushingModifications(flags::any),
                                    InvalidationLevel::InvalidOutput}});
     /**
      * @code
@@ -129,10 +130,10 @@ public:
      * InvalidationLevel::InvalidOutput}
      *  }
      *  );
-     *  // Invalidate processor on any type of action for filtering or column selection.
+     *  // Invalidate processor on filtering or selection for any target.
      *  BrushingAndLinkingManager(port,
      *  {
-     *    {BrushingModification::Filtered | BrushingModification::Selected,
+     *    {AnyBrushingTarget, BrushingModification::Filtered | BrushingModification::Selected,
      * InvalidationLevel::InvalidOutput}
      *  }
      *  );
@@ -144,7 +145,7 @@ public:
      */
     BrushingAndLinkingManager(BrushingAndLinkingOutport* outport,
                               std::vector<BrushingTargetsInvalidationLevel> invalidationLevels = {
-                                  {BrushingModifications(flags::any),
+                                  {AnyBrushingTarget, BrushingModifications(flags::any),
                                    InvalidationLevel::InvalidOutput}});
     virtual ~BrushingAndLinkingManager();
 
@@ -343,10 +344,10 @@ public:
      * InvalidationLevel::InvalidOutput}
      *  }
      *  );
-     *  // Invalidate processor on any type of action for filtering or column selection.
+     *  // Invalidate processor on filtering or selection for any target.
      *  setInvalidationLevels(
      *  {
-     *    {BrushingModification::Filtered | BrushingModification::Selected,
+     *    {AnyBrushingTarget, BrushingModification::Filtered | BrushingModification::Selected,
      * InvalidationLevel::InvalidOutput}
      *  }
      *  );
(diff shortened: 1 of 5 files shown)


modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
```diff
@@ -53,13 +53,32 @@ class Deserializer;
 /**
  * Manages brushing and linking events for filtering, selecting, and highlighting. When initialized
  * with a BrushingAndLinking inport, changes are propagated using this port if connected.
+ * A Processor can choose to only handle a certain type of
```
martinfalk: certain type of?

modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
```diff
@@ -231,6 +250,17 @@ class IVW_MODULE_BRUSHINGANDLINKING_API BrushingAndLinkingManager : public Seria
         std::function<void(BrushingAction, BrushingTarget, const BitSet&, std::string_view)>
             callback);
 
+    /**
+     * Returns the types of actions causing the owner (BrusingAndLinkingInport) to invalidate.
```
martinfalk: affects both ports, suggesting "BrushingAndLinking port" instead.

modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
```diff
@@ -231,6 +250,17 @@ class IVW_MODULE_BRUSHINGANDLINKING_API BrushingAndLinkingManager : public Seria
...
+     */
+    BrushingModifications getInvalidateOn() const;
+
+    /**
+     * Set the types of brushing actions that should invalidate the owner (BrusingAndLinkingInport).
```
martinfalk: see comment above

modules/brushingandlinking/include/modules/brushingandlinking/ports/brushingandlinkingports.h
```diff
@@ -40,18 +40,29 @@
...
 
 class BrushingAndLinkingOutport;
+
 /**
  * \ingroup ports
  * Enables selection/filtering/highlighting between processors.
- * The port has a BrushingManager independent on if it is connected or not, e.g. it does not need to
- * be connected to a BrushingAndLinkingOutport to be valid.
+ * The port has a BrushingManager independent on if it is connected or not, e.g., it does not need
+ * to be connected to a BrushingAndLinkingOutport to be valid.
```
martinfalk:
```suggestion
 * The port has a BrushingManager independent on whether it is connected or not. Therefore it does
 * not need to be connected to a BrushingAndLinkingOutport to be valid.
```

modules/brushingandlinking/include/modules/brushingandlinking/ports/brushingandlinkingports.h
```diff
@@ -40,18 +40,29 @@
...
 class IVW_MODULE_BRUSHINGANDLINKING_API BrushingAndLinkingInport : public Inport {
 public:
     using type = void;
-
-    BrushingAndLinkingInport(std::string identifier);
+    /**
+     * Inport constructor for BrushingAndLinkingManager.
+     * @param identifier of port.
+     * @param invalidateOn flags for the modifications that should invalidate the processor
+     * (Processor::process will not be called).
```
martinfalk: Processor::process will only be called for those

modules/brushingandlinking/src/brushingandlinkingmanager.cpp
```diff
@@ -394,12 +402,18 @@ int BrushingAndLinkingManager::getActionIndex(BrushingAction action) {
...
     modifications_[target] |= fromAction(action);
 
-    if (std::holds_alternative<BrushingAndLinkingOutport*>(owner_)) {
-        auto outport = std::get<BrushingAndLinkingOutport*>(owner_);
-        outport->getProcessor()->invalidate(invalidationLevel_);
-    } else if (!parent_) {
-        auto inport = std::get<BrushingAndLinkingInport*>(owner_);
-        inport->getProcessor()->invalidate(invalidationLevel_);
+    // Only invalidate the top level in the connected brushing manager network
+    if (!parent_) {  // && (getModifiedActions() & invalidateOn_)) {
```
martinfalk: check whether commented part is needed, otherwise remove it.

modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
ResearchDaniel: Right now it only applies to inports, but it might make sense to change it.

modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
ResearchDaniel: fixed

modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
```diff
@@ -53,13 +53,35 @@ class Deserializer;
 /**
  * Manages brushing and linking events for filtering, selecting, and highlighting. When initialized
  * with a BrushingAndLinking inport, changes are propagated using this port if connected.
+ *
+ * Use setInvalidateOn flags if you only want Processor::process to be called for a subset of
```
petersteneteg: Any reason for not considering targets? I would get that it is also as common to only care about row or columns for example.

modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
```diff
@@ -53,13 +53,35 @@ class Deserializer;
...
+    /**
+     * @param inport owner of the manager.
+     * @param invalidateOn for the modifications that should invalidate the processor
+     * (Processor::process will be called for those).
+     * @param invalidationLevel that will be propagated to the port and thereby upwards in the
+     * network.
+     */
+    BrushingAndLinkingManager(
+        BrushingAndLinkingInport* inport,
+        BrushingModifications invalidateOn = BrushingModifications(flags::any),
```
petersteneteg: invalidateOn and the invalidationLevel (it should be invalidationLevel not validationLevel) could be combined into one map ` {{{target1,action1}->invalidationLevel1}, {target2,action2}->invalidationLevel2}, ...}`,
maybe one could also support some shortcuts, 
```
{action}->invalidationLevel1} == {all_targets,actions}->invalidationLevel}
{target}->invalidationLevel1} == {target, all_actions}->invalidationLevel}
{invalidationLevel} == {all_targets, all_actions}->invalidationLevel}
```
And if then the order in the list matters, one could for example do:
```
{
    InvalidationLevel::Valid, 
    {BrushingTarget::Row, BrusingModification::Selected} -> InvalidationLevel::InvalidOutput,
    {BrushingTarget::Column, BrusingModification::Filtered} -> InvalidationLevel::InvalidResources
} 
```

modules/brushingandlinking/include/modules/brushingandlinking/ports/brushingandlinkingports.h
```diff
@@ -40,18 +40,29 @@
 namespace inviwo {
 
 class BrushingAndLinkingOutport;
+
 /**
  * \ingroup ports
  * Enables selection/filtering/highlighting between processors.
- * The port has a BrushingManager independent on if it is connected or not, e.g. it does not need to
- * be connected to a BrushingAndLinkingOutport to be valid.
+ * The port has a BrushingManager independent on whether it is connected or not. Therefore, it does
```
petersteneteg:
```suggestion
 * The inport has it's own BrushingManager therefore, it does
```

modules/brushingandlinking/include/modules/brushingandlinking/ports/brushingandlinkingports.h
```diff
@@ -40,18 +40,29 @@
...
+ * not need to be connected to a BrushingAndLinkingOutport to be valid.
+ *
+ * Use setInvalidateOn flags if you only want Processor::process to be called for a subset of
+ * brushing actions. Use getModifiedActions if you want to know which BrushingModifications caused a
```
petersteneteg:
```suggestion
 * Use setInvalidateOn flags if you only want the processor to be invalidated for a subset of
 * brushing actions. Use getModifiedActions if you want to know which BrushingModifications caused a
```

modules/brushingandlinking/src/brushingandlinkingmanager.cpp
```diff
@@ -394,12 +402,18 @@ int BrushingAndLinkingManager::getActionIndex(BrushingAction action) {
...
-        outport->getProcessor()->invalidate(invalidationLevel_);
-    } else if (!parent_) {
-        auto inport = std::get<BrushingAndLinkingInport*>(owner_);
-        inport->getProcessor()->invalidate(invalidationLevel_);
+    // Only invalidate the top level in the connected brushing manager network
+    if (!parent_) {
+        if (std::holds_alternative<BrushingAndLinkingOutport*>(owner_)) {
+            auto outport = std::get<BrushingAndLinkingOutport*>(owner_);
+            // Processor need to be invalidated for the network evaluation to be notified.
+            outport->getProcessor()->invalidate(invalidationLevel_);
```
petersteneteg: Any reason for nog doing? 
```suggestion
            outport->invalidate(invalidationLevel_);
```

modules/brushingandlinking/src/brushingandlinkingmanager.cpp
```diff
@@ -394,12 +402,18 @@ int BrushingAndLinkingManager::getActionIndex(BrushingAction action) {
...
+        } else if (std::holds_alternative<BrushingAndLinkingInport*>(owner_)) {
+            // Nothing is connected, invalidate the processor itself
+            auto inport = std::get<BrushingAndLinkingInport*>(owner_);
+
+            inport->getProcessor()->invalidate(invalidationLevel_);
```
petersteneteg:
```suggestion
            inport->invalidate(invalidationLevel_);
```
Should be the same, and is cleaner I think. I not the same, explain why.

modules/brushingandlinking/src/ports/brushingandlinkingports.cpp
```diff
@@ -176,6 +185,13 @@ void BrushingAndLinkingInport::setChanged(bool changed, const Outport* source) {
     Inport::setChanged(changed, source);
 }
 
+void BrushingAndLinkingInport::invalidate(InvalidationLevel invalidationLevel) {
+    if (invalidationLevel == InvalidationLevel::Valid ||
```
petersteneteg: why this check `invalidationLevel == InvalidationLevel::Valid` ?

petersteneteg (changes requested):

modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
ResearchDaniel: Good point. I guess that would be a valid filter and perhaps not as uncommon as I originally thought.

modules/brushingandlinking/src/brushingandlinkingmanager.cpp
ResearchDaniel: See comment the line above:
Processor need to be invalidated for the network evaluation to be notified.

modules/brushingandlinking/src/ports/brushingandlinkingports.cpp
ResearchDaniel: Assume that it must be possible to set the port to Valid. Otherwise I think that other logic might become broken

modules/brushingandlinking/src/brushingandlinkingmanager.cpp
petersteneteg: Then you need too specify why that processor needs to invalidate and not the one connected to the outport

modules/brushingandlinking/src/ports/brushingandlinkingports.cpp
petersteneteg: I believe we use setValid to make it valid

modules/brushingandlinking/src/brushingandlinkingmanager.cpp
ResearchDaniel: I just added a comment to make it a bit clearer than before :) Agree that we should clarify even more though
I think that @martinfalk needed `notifyObserversInvalidationBegin(this);` to be called for some use case.

modules/brushingandlinking/src/brushingandlinkingmanager.cpp
ResearchDaniel: Yes, here I think it should be the same

modules/brushingandlinking/src/brushingandlinkingmanager.cpp
petersteneteg: The network should get evaluated either way so which call you make has nothing to do with the network being evaluated. The only question is if the processor owning the outport needs to also be invalidated for some reason, or if it is enough that the down stream processors are evaluated.

ResearchDaniel: Yes, I think that was also one of the desired behaviours.

modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
ResearchDaniel: Fixed in latest commit

petersteneteg (approved):

PR #1242, opened 2021-12-31
Add getHeaders() back to data frame

PetiteViking:


petersteneteg:
Hi

We removed it since I believe it does not belong in the DataFrame Api. I rather see that you do getColumn()->getHeader() or loop over all the columns and ask for the header as you do in the implementation. The reasoning being that it bloats the interface and of DataFrame and is generally inefficient since you will have to copy a lot of strings, which if you just loop over the columns instead you don't have to do (getHeader() gives you a const string&) Also if you add getHeaders to DataFrame, what about the other getters from Column, which ones should also be added to DataFrame, getRanges? getUnits?. Seems better to me to just keep those in Column, but I can probably be convinced to change my mind if you provide a convincing argument :)

JochenJankowai:
As far as I am concerned, the question becomes how often a feature is used. You are right if the getters are not used very frequently. In that case it makes sense to write the loop where its needed. If the getters would be used in many places, writing all that boilerplate everywhere instead will bloat the code that way.
That being said, the fact that you could apparently remove the method without breaking any code other than mine would suggest that the getter does indeed not need to be part of the interface :)

ResearchDaniel:
It did also break my code :) 
I do not have a strong opinion here. Pandas dataframe exposes column headers in their data frame object. I do think that the headers might be used much more frequently than others, but that it is also not too inconvient to go via the columns themselves.

PR #1243, opened 2022-01-06
Mac fixes

ResearchDaniel:
- Fix CEF loading of .proj language files
- Remove warnings about missing plist variables
- CRoaring: Ensure that lib files end up where all other lib files are

Tested on MacOs 12.1 (Monterey) with XCode 13.2.1, and qt 6 installed via brew


cmake/installutils.cmake
```diff
@@ -274,9 +274,15 @@ macro(ivw_qt_add_to_install ivw_comp)
                                             Qt${QT_VERSION_MAJOR}::${qt_module} MODULE_PLUGIN_TYPES)
                         if(plugin_types)
                             foreach(plugin_type IN ITEMS ${plugin_types})
-                                set(plugins_dir "${QT_DIR}/../../../plugins/${plugin_type}")
+                                if(APPLE)
```
petersteneteg, line 277: what about QT_DEFAULT_MAJOR_VERSION above, Don't think that one existed in Qt5

petersteneteg (dismissed):

cmake/installutils.cmake
ResearchDaniel: True, I have changed it to QT_VERSION_MAJOR, which should exist

petersteneteg (dismissed):

ResearchDaniel:
Regression test had issues with the fonts being different. Visually inspecting the regression test images it looked like the new ones were more correct. It is unclear if the differences were due to system updates or if the .pak files where found now and thereby also some font loading was different. Therefore updated the reference images.

petersteneteg (approved):

PR #1245, opened 2022-01-21
Feature/brushingandlinking

martinfalk:
This should fix several issues with the Brushing and Linking.

* fixed propagation of the modification state, now forwarded via port invalidation (otherwise `manager_.getInvalidationLevel()` will not reflect the new state and return `Valid`) 
* to avoid unnecessary invalidations changes are only propagated upward if the indices have changed 
* when clearing indices, all managers should be cleared by propagating to parent and children

### DIFF
--- a/include/inviwo/core/datastructures/bitset.h
+++ b/include/inviwo/core/datastructures/bitset.h
@@ -118,13 +118,13 @@ public:
     BitSetIterator end() const;
 
     /**
-     * return the number of elements the bitset holds
+     * Return the number of elements the bitset holds
      *
      * \see size()
      */
     uint32_t cardinality() const;
     /**
-     * return the number of elements the bitset holds
+     * Return the number of elements the bitset holds
      *
      * \see cardinality()
      */
@@ -135,16 +135,16 @@ public:
     void clear();
 
     /**
-     * check whether this bitset is a subset of \p b
+     * Check whether this bitset is a subset of \p b
      */
     bool isSubsetOf(const BitSet& b) const;
     /**
-     * check whether this bitset is a strict subset of \p b
+     * Check whether this bitset is a strict subset of \p b
      */
     bool isStrictSubsetOf(const BitSet& b) const;
 
     /**
-     * replace the bitset with the contents of \p b, returns true if modified
+     * Replace the bitset with the contents of \p b, returns true if modified
      *
      * @return true if the bitset was modified that is \p this and \p b were different
      */
@@ -180,7 +180,7 @@ public:
     bool addChecked(uint32_t v);
 
     /**
-     * adds all values from the open range [min,max)
+     * Adds all values from the open range [min,max)
      *
      * @param min   lower bound
      * @param max   upper bound (not included)
@@ -188,7 +188,7 @@ public:
     void addRange(uint32_t min, uint32_t max);
 
     /**
-     * adds all values from the closed range [min,max]
+     * Adds all values from the closed range [min,max]
      *
      * @param min   lower bound
      * @param max   upper bound (included)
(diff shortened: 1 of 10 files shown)


include/inviwo/core/datastructures/bitset.h
```diff
@@ -143,6 +143,13 @@ class IVW_CORE_API BitSet : public Serializable {
      */
     bool isStrictSubsetOf(const BitSet& b) const;
 
+    /**
+     * update the bitset with the contents of \p b, if \p b is different
```
ResearchDaniel: Suggestion:
Sets the value of \p b and returns true if modified.

modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
```diff
@@ -357,10 +357,16 @@ class IVW_MODULE_BRUSHINGANDLINKING_API BrushingAndLinkingManager : public Seria
...
 
     /**
-     * propagates the modified state to all child managers and resets the state. Should only be
-     * called by the brushing and linking ports _after_ the process() function has been called.
+     * propagates the modified state to all child managers. Needs to be called by the brushing and
+     * linking ports _before_ the invalidation level is queried.
+     * @see getInvalidationLevel
      */
-    void markAsValid();
+    void propagateModifications();
```
ResearchDaniel, line 364: Would these ever be called by anything but the ports? 
Maybe we could make these private friend functions to the ports.

ResearchDaniel (dismissed):
Did not try it, but looks good from what I can see

modules/brushingandlinking/include/modules/brushingandlinking/brushingandlinkingmanager.h
martinfalk: If you, for whatever reason, have a manager between the ports, these functions need to be called as well in the correct order. Otherwise changes are not propagated downward.

include/inviwo/core/datastructures/bitset.h
```diff
@@ -143,6 +143,13 @@ class IVW_CORE_API BitSet : public Serializable {
      */
     bool isStrictSubsetOf(const BitSet& b) const;
 
+    /**
+     * replace the bitset with the contents of \p b, returns true if modified
```
petersteneteg: I think comments should start with a capital letter. And I would skip the empty row below.

modules/brushingandlinking/include/modules/brushingandlinking/datastructures/indexlist.h
```diff
@@ -43,10 +43,17 @@ class IVW_MODULE_BRUSHINGANDLINKING_API IndexList : public Serializable {
...
     virtual ~IndexList() = default;
 
+    bool empty() const;
     size_t size() const;
     void clear();
 
-    void set(std::string_view src, const BitSet& indices);
+    /**
+     * update the indexlist with source \p src and \p indices, if \p indices are different
+     *
```
petersteneteg, line 52: Same as above

modules/brushingandlinking/src/datastructures/indexlist.cpp
```diff
@@ -50,13 +52,20 @@ const BitSet& IndexList::getIndices() const {
     return indices_;
 }
 
-void IndexList::set(std::string_view src, const BitSet& indices) {
+bool IndexList::set(std::string_view src, const BitSet& indices) {
+    if ((indicesBySource_.count(std::string(src)) > 0)) {
+        if (indicesBySource_[std::string(src)] == indices) return false;
```
petersteneteg: here you are constructing the string twice and doing the same lookup in the map twice? 
```suggestion
    if (auto it = indicesBySource_.find(std::string(src)): it != indicesBySource_.end()) {
        if (it->second == indices) return false;
```

petersteneteg: and again below...

petersteneteg (changes requested):

ResearchDaniel (dismissed):

petersteneteg (dismissed):

petersteneteg (dismissed):

petersteneteg (dismissed):

petersteneteg (approved):

PR #1246, opened 2022-01-28
Volume: Region mapping for Atlas volumes

RasmusHogslatt:


modules/volume/src/processors/volumeregionmap.cpp
```diff
@@ -0,0 +1,116 @@
...
+        });
+};
+}
+
+void VolumeRegionMap::process() {
+    // Get columns
+    auto csv = dataFrame_.getData();
+    auto oldIdx = csv->getColumn(from_.getSelectedValue());
+    auto newIdx = csv->getColumn(to_.getSelectedValue());
+    auto nrows = csv->getNumberOfRows();
```
sigsid: Is there a risk that oldIdx and newIdx are not of the same length? Or should we assume that?

modules/volume/include/inviwo/volume/processors/volumeregionmap.h
```diff
@@ -0,0 +1,76 @@
...
+#include <inviwo/dataframe/datastructures/dataframe.h>
+#include <inviwo/core/processors/poolprocessor.h>
+#include <optional>
+#include <inviwo/dataframe/properties/columnoptionproperty.h>
+
+namespace inviwo {
+
+/** \docpage{org.inviwo.VolumeRegionMap, Volume Region Map}
+ * ![](org.inviwo.VolumeRegionMap.png?classIdentifier=org.inviwo.VolumeRegionMap)
+ * Explanation of how to use the processor.
```
sigsid: Nice processor! Looking forward to be able to use it :) 
Don't forget to add a description here

modules/volume/include/inviwo/volume/processors/volumeregionmap.h
```diff
@@ -0,0 +1,76 @@
...
+ * ### Inports
+ *   * __inputVolume__ Input volume
+ *
+ * ### Outports
+ *   * __outputVolume__ Output volume
+ *
+ * ### Properties
+ *
+ */
+class IVW_MODULE_VOLUME_API VolumeRegionMap : public Processor {
```
ResearchDaniel: Consider changing the name as it may not have anything to do with regions. I guess region is meant to imply that it is is integer values? Might make sense still depending on what other Atlas/Region processors are called
`VolumeAtlasMapper`

modules/volume/src/processors/volumeregionmap.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+    // Store in vectors
+    std::vector<uint32_t> sourceIndices(nrows);
```
ResearchDaniel: A reason why not a map with key = source is used?

modules/volume/src/processors/volumeregionmap.cpp
```diff
@@ -0,0 +1,116 @@
...
+    std::vector<uint32_t> destinationIndices(nrows);
+
+    copyColumn(*oldIdx, sourceIndices,
+               [](const auto& src, auto& dst) { dst = static_cast<uint32_t>(src); });
```
ResearchDaniel: Should uint32_t be assumed?
Possible to use the type of the inport data?

modules/volume/src/processors/volumeregionmap.cpp
```diff
@@ -0,0 +1,116 @@
...
+    auto newVolume = std::shared_ptr<Volume>(inVolume->clone());
+    auto volRep = newVolume->getEditableRepresentation<VolumeRAM>();
+
+    volRep->dispatch<void, dispatching::filter::Scalars>([&](auto volram) {
+        using ValueType = util::PrecisionValueType<decltype(volram)>;
+        ValueType* dataPtr = volram->getDataTyped();
+
+        const auto& dim = volram->getDimensions();
+
+        std::transform(dataPtr, dataPtr + dim.x * dim.y * dim.z, dataPtr, [&](const ValueType& v) {
```
ResearchDaniel: You can also use glm::compMul(dim) if you want. This is perhaps more readable though :)
`std::transform(dataPtr, dataPtr + glm::compMul(dims)...`

modules/volume/src/processors/volumeregionmap.cpp
```diff
@@ -0,0 +1,116 @@
...
+            for (size_t i = 0; i < sourceIndices.size(); ++i) {
```
ResearchDaniel: If you use a map you could replace this loop with:
```
if (auto it = map.find(v); it != map.end() {
  return *it;
}
```

modules/volume/include/inviwo/volume/processors/volumeregionmap.h
```diff
@@ -0,0 +1,76 @@
...
+#include <optional>
+#include <inviwo/dataframe/properties/columnoptionproperty.h>
+
+namespace inviwo {
+
+/** \docpage{org.inviwo.VolumeRegionMap, Volume Region Map}
+ * ![](org.inviwo.VolumeRegionMap.png?classIdentifier=org.inviwo.VolumeRegionMap)
+ * Explanation of how to use the processor.
+ * Indices are remapped to values provided in CSV file. Requires two columns. 1st column is old
+ * indices, second contains values to be mapped to.
```
ResearchDaniel: What happens if the value is not found? 
Should an error be produced?

modules/volume/src/processors/volumeregionmap.cpp
```diff
@@ -0,0 +1,116 @@
...
+                if (static_cast<uint32_t>(v) == sourceIndices[i]) {
+                    return static_cast<ValueType>(destinationIndices[i]);
+                }
+            }
+            return ValueType{0};
```
ResearchDaniel: Perhaps this should be a user-specifed value that defaults to 0?

modules/volume/data/regionmap.csv
```diff
@@ -0,0 +1,29 @@
+﻿pre;new
```
ResearchDaniel, line 1: A suggestion is to refactor the code in the processor into a function, add a unit test for that function and move this file to the unit test. 
I would consider removing the workspace in that case.

ResearchDaniel (changes requested):
Nice work!
Since this is the first commit I have tried to add as detailed comments and suggestions as I could.

modules/volume/include/inviwo/volume/processors/volumeregionmap.h
petersteneteg: We use region for a couple of other ones, so I think it makes an equal amount if sense.

modules/volume/data/regionmap.csv
petersteneteg: We also want a regression test with the processor, and the file is needed for that.

modules/volume/src/processors/volumeregionmap.cpp
petersteneteg: there might also be other even more efficient approaches in some cases.

modules/volume/data/regionmap.csv
ResearchDaniel: Don’t think a regression test is necessary if a unit test is added (much less overhead with a unit test). Would place the file where the test is

petersteneteg: I also want the processor to be tested.

CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ Here we document changes that affect the public API or changes that needs to be
 The welcome widget now has an icon grid view with previews of the first canvas in each network. 
 The workspace and processor search fields gained new features. One can for example search for all networks with a `Volume Raycaster` processor that was created by `Martin` like this: `processors:\"Volume Raycaster\" author:Martin"`. The tooltip for each search field has more details. 
 
+## 2022-01-31 Volume Region Map
```
ResearchDaniel:
```suggestion
## 2022-01-31 New processor: Volume Region Map
Maps each unique integer value in a Volume to another integer value. The value mapping must be provided by two columns in a DataFrame

```

modules/volume/include/inviwo/volume/processors/volumeregionmap.h
```diff
@@ -0,0 +1,78 @@
...
+#include <inviwo/dataframe/datastructures/dataframe.h>
+#include <inviwo/core/processors/poolprocessor.h>
+#include <optional>
+#include <inviwo/dataframe/properties/columnoptionproperty.h>
+
+namespace inviwo {
+
+/** \docpage{org.inviwo.VolumeRegionMap, Volume Region Map}
+ * ![](org.inviwo.VolumeRegionMap.png?classIdentifier=org.inviwo.VolumeRegionMap)
+ * Indices are remapped to values provided in CSV file. Requires two columns. 1st column is old
```
sigsid: I think you can remove the part about 'provided in CSV file' since if I understand it correctly the information is provided in the two columns and not from a file

RasmusHogslatt: True, will do so.

modules/volume/include/inviwo/volume/processors/volumeregionmap.h
RasmusHogslatt: The current implmentation handles non-continuous cases. It adds each pair to an unordered map. If there happens to be  voxel values not stored in the source indices provided, these voxel values are currently set to 0. Any thoughts? Could also return the same value to keep it or maybe implement an option for those values with a "change to zero/display anyway" button.

modules/volume/src/processors/volumeregionmap.cpp
```diff
@@ -0,0 +1,223 @@
...
+        }
+    });
+}
+
+/// <summary>
+/// Tests if a index vector contains sequential numbers without gaps.
+/// </summary>
+/// <param name="src">Index vecotr</param>
+/// <returns>True if sequential. False if not sequential.</returns>
+bool isSortedSequence(const std::vector<unsigned int>& src) {
```
ResearchDaniel: Equivalent to https://en.cppreference.com/w/cpp/algorithm/is_sorted ?

modules/volume/src/processors/volumeregionmap.cpp
```diff
@@ -0,0 +1,223 @@
...
+
+    // Volume
+    auto inVolume = inport_.getData();
+    auto newVolume = std::shared_ptr<Volume>(inVolume->clone());
+    remap(newVolume, sourceIndices, destinationIndices, missingValue);
+
+    outport_.setData(newVolume);
+}
+
+void VolumeRegionMap::remap(std::shared_ptr<Volume>& volume, std::vector<unsigned int> src,
```
ResearchDaniel: Consider moving this function to a folder within the module: algorithm/volume (see for example https://github.com/inviwo/inviwo/tree/master/modules/base/include/modules/base/algorithm/volume ).

modules/volume/src/processors/volumeregionmap.cpp
```diff
@@ -0,0 +1,223 @@
...
+    for (int i = 1; i < src.size(); ++i) {
+        if (src[i] <= prev) {
+            return false;
+        }
+        prev = src[i];
+    }
+    return true;
+}
+
+size_t binarySearch(const std::vector<unsigned int>& src, const unsigned int value) {
```
ResearchDaniel: Equivalent to std::lower_bound or std::upper_bound?
https://en.cppreference.com/w/cpp/algorithm/lower_bound

modules/volume/src/processors/volumeregionmap.cpp
ResearchDaniel: Would be less error prone if a map is used instead of two separate vectors. Or perhaps there is a particular reason for using two vectors as input?

ResearchDaniel (changes requested):
Looks like it is almost in a ready state. There are some functions that may be equivalent to std functions.

modules/volume/src/processors/volumeregionmap.cpp
RasmusHogslatt: Just updated to lower bound.

modules/volume/src/processors/volumeregionmap.cpp
RasmusHogslatt: As of current version, this will yield an exception.

modules/volume/src/processors/volumeregionmap.cpp
RasmusHogslatt: The input type is int as of now. I discussed this with Martin and we figured it makes the most sense if you would like to set something to a negative value. You can now specify if voxel values that are not in the source/destination file should be set to a given value in the GUI or keep the current value.

modules/volume/src/processors/volumeregionmap.cpp
RasmusHogslatt: Nice, made this change now.

modules/volume/src/processors/volumeregionmap.cpp
RasmusHogslatt: There is now an option to keep values that are missing, or set a specific value. Currently, it will default to 0 unless changed in GUI.

modules/volume/src/processors/volumeregionmap.cpp
RasmusHogslatt: This has now been changed.

modules/volume/src/algorithm/volumemap.cpp
```diff
@@ -43,6 +43,12 @@ void remap(std::shared_ptr<Volume>& volume, std::vector<int> src, std::vector<in
                         src.size(), dst.size());
     }
 
+    auto uniqueIterator = std::unique(src.begin(), src.end());
```
petersteneteg: std::unique requires that the input range is sorted, is that the case?

RasmusHogslatt: Did not know that. I just pushed a change where I create a sorted copy of src, and use std::unique for the copy. Of course, it requires an extra sort, but maybe that's not too bad, considering it only has to run once when changing CSV file?

Or should it be assumed that src has no duplicates and not throw any exceptions otherwise? If it has to be sorted, it might make more sense to just use a set or map rather than vector anyway?

martinfalk: Could use `std::is_sorted()` (see [docs](https://en.cppreference.com/w/cpp/algorithm/is_sorted)) or even `std::is_sorted_until()`.

CHANGELOG.md
ResearchDaniel: Ping on this one

modules/volume/include/inviwo/volume/algorithm/volumemap.h
```diff
@@ -0,0 +1,42 @@
...
+
+#include <inviwo/volume/volumemoduledefine.h>
+
+#include <inviwo/core/datastructures/volume/volume.h>
+#include <vector>
+#include <inviwo/core/datastructures/volume/volumeramprecision.h>
+
+namespace inviwo {
+
+void IVW_MODULE_VOLUME_API remap(std::shared_ptr<Volume>& volume, std::vector<int> src,
```
ResearchDaniel: Add documentation about preconditions for src and dst as well as a brief description of what the function does.
Could be a future PR: make it a template function

modules/volume/src/algorithm/volumemap.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+    if (src.size() == 0 || src.size() != dst.size()) {
+        throw Exception(IVW_CONTEXT_CUSTOM("Remap"), "Invalid dataframe size (src = {}, dst = {})",
+                        src.size(), dst.size());
+    }
+
+    // Create sorted copy of src and check if it contains duplicates
+    std::unordered_set<int> set(src.begin(), src.end());
+    if (src.size() != set.size()) {
+        throw Exception(IVW_CONTEXT_CUSTOM("Remap"),
```
ResearchDaniel, line 53: An option could be to simply remove duplicates.

modules/volume/include/inviwo/volume/processors/volumeregionmap.h
```diff
@@ -0,0 +1,81 @@
...
+    virtual const ProcessorInfo getProcessorInfo() const override;
+    static const ProcessorInfo processorInfo_;
+
+private:
+    DataFrameInport dataFrame_;
+    VolumeInport inport_;
+    VolumeOutport outport_;
+    ColumnOptionProperty from_;
+    ColumnOptionProperty to_;
+    BoolProperty defaultMissingValue_;
```
ResearchDaniel:
```suggestion
    BoolProperty useMissingValue_;
```
or 
```suggestion
    BoolProperty mapMissingValue_;
```

modules/volume/include/inviwo/volume/processors/volumeregionmap.h
ResearchDaniel: Still, it should either be VolumeRegionMapper or MapVolumeRegion in a similar fashion as we have VolumeRaycaster

ResearchDaniel (changes requested):
Nice, some minor things that might need to be adjusted

modules/volume/include/inviwo/volume/processors/volumeregionmap.h
martinfalk: Went for `VolumeRegionMapper`.

modules/volume/src/algorithm/volumemap.cpp
martinfalk: Then it would be unclear what happens, that is which index is reused. This means that there need to be at least a log warning which implies the need for logging (and showing the log). An exception is much clearer.

modules/volume/src/processors/volumeregionmap.cpp
martinfalk: The two vectors originate from two separate DataFrame columns. Using a map with the corresponding preconditions would only offload that functionality to the caller.

modules/volume/src/algorithm/volumemap.cpp
```diff
@@ -28,16 +28,19 @@
...
 #include <algorithm>
 #include <unordered_map>
 #include <unordered_set>
-#include <inviwo/core/util/exception.h>
 
 namespace inviwo {
 
-void remap(std::shared_ptr<Volume>& volume, std::vector<int> src, std::vector<int> dst,
-           int missingValue, bool useMissingValue) {
+void remap(std::shared_ptr<Volume>& volume, const std::vector<int>& src,
```
ResearchDaniel: Should not be shared ptr unless the shared functionality is actually needed.

ResearchDaniel (approved):
Would also remove the shared ptr from the arg as it should be the responsibility of the called to make sure that it is alive during the calls

modules/volume/src/algorithm/volumemap.cpp
martinfalk: Good point.

ResearchDaniel (dismissed):

modules/volume/src/algorithm/volumemap.cpp
RasmusHogslatt: Thanks, will review this.

ResearchDaniel (approved):

PR #1248, opened 2022-02-07
Various: missing includes

petersteneteg:


modules/basegl/src/processors/multichannelraycaster.cpp
```diff
@@ -32,9 +32,11 @@
 #include <modules/opengl/volume/volumegl.h>
 #include <modules/opengl/image/layergl.h>
 #include <modules/opengl/texture/textureunit.h>
+
```
martinfalk, line 35: why the empty lines?

modules/fontrendering/src/textrenderer.cpp
```diff
@@ -40,6 +40,7 @@
 #include <modules/opengl/openglutils.h>
 #include <modules/opengl/shader/shaderutils.h>
 #include <modules/opengl/texture/textureutils.h>
+#include <modules/opengl/texture/texture2d.h>                        // for Texture2D
```
martinfalk: remove comment

martinfalk (dismissed):

modules/basegl/src/processors/multichannelraycaster.cpp
petersteneteg: not sure about that

modules/fontrendering/src/textrenderer.cpp
petersteneteg: We might add these comment for every include in the future, or not, we should discuss that.

ResearchDaniel (dismissed):

ResearchDaniel (approved):

martinfalk (approved):

PR #1249, opened 2022-02-08
Correct spelling of QToolBar

Dapid:
See the reference for [Qt5](https://doc.qt.io/qt-5/qtoolbar.html) and  [Qt6](https://doc-snapshots.qt.io/qt6-dev/qtoolbar.html).


petersteneteg (approved):

petersteneteg:
Thanks for noticing and fixing!!! 
Capitalization issues can be hard to notice on system that uses a case insensitive file system ...

PR #1251, opened 2022-02-10
Qt,BaseGL,BrushingAndLinking,DataFrameQt,Volume: warn fixes

martinfalk:


modules/fontrendering/src/processors/textoverlaygl.cpp
```diff
@@ -46,6 +46,7 @@
 #include <locale>
 
 #include <fmt/format.h>
+#include <fmt/args.h>
```
martinfalk, line 51: <fmt/args.h> no longer exists
https://fmt.dev/latest/api.html

cmake/compileoptions.cmake
```diff
@@ -49,7 +49,12 @@ if(CMAKE_GENERATOR STREQUAL "Xcode")
 endif()
 
 function(ivw_define_standard_properties)
-    foreach(target ${ARGN})
+    set(options "QT")
```
ResearchDaniel, line 52:
```suggestion
    set(options "HIDE_QT_WARNINGS")
```

cmake/globalmacros.cmake
```diff
@@ -472,7 +472,7 @@ endfunction()
 # This it called from the inviwo module CMakeLists.txt 
 # that is included from ivw_register_modules. 
 function(ivw_create_module)
-    set(options "NO_PCH")
+    set(options "NO_PCH" "QT")
```
ResearchDaniel, line 475:
```suggestion
    set(options "NO_PCH" "HIDE_QT_WARNINGS")
```

ResearchDaniel (dismissed):

cmake/globalmacros.cmake
petersteneteg: My thinking was that we might use the option for more cases than just the hiding that specific warning.

petersteneteg (approved):

PR #1252, opened 2022-02-10
CSV comment rows support

EmiliaStahlbom:

**Always**

**New features**

**Updated feature / Fixes** 

- Operating system
- Compiler/IDE
- QT/CMake/Python versions

### DIFF
--- a/modules/dataframe/include/inviwo/dataframe/io/csvreader.h
+++ b/modules/dataframe/include/inviwo/dataframe/io/csvreader.h
@@ -93,6 +93,10 @@ public:
     CSVReader& setRowComment(std::string_view comment);
     const std::string& getRowComment() const;
 
+    /** @see CSVReader::defaultKeepOnly */
+    CSVReader& setKeepOnly(std::string_view only);
+    const std::string& getKeepOnly() const;
+
     /** @see CSVReader::defaultLocale */
     CSVReader& setLocale(std::string_view loc);
     const std::string& getLocale() const;
@@ -181,6 +185,8 @@ public:
     static constexpr size_t defaultNumberOfExampleRows = 50;
     /** @see CSVReader::setRowComment */
     static constexpr std::string_view defaultRowComment = "";
+    /** @see CSVReader::setKeepOnly */
+    static constexpr std::string_view defaultKeepOnly = "";
     /** @see CSVReader::setLocale */
     static constexpr std::string_view defaultLocale = "C";
     /** @see CSVReader::setHandleEmptyFields */
@@ -209,6 +215,7 @@ private:
     bool doublePrecision_;
     size_t exampleRows_;
     std::string rowComment_;
+    std::string keepOnly_;
     std::string locale_;
     EmptyField emptyField_;
 };
--- a/modules/dataframe/include/inviwo/dataframe/io/csvreader.h
+++ b/modules/dataframe/include/inviwo/dataframe/io/csvreader.h
@@ -89,17 +89,17 @@ public:
     CSVReader& setNumberOfExampleRows(size_t rows);
     size_t getNumberOfExamplesRows() const;
 
-    /** 
-    * sets the string that indicates that a row should be commented/removed
-    * @see CSVReader::defaultRowComment 
-    */
+    /**
+     * sets the string that indicates that a row should be commented/removed
+     * @see CSVReader::defaultRowComment
+     */
     CSVReader& setRowComment(std::string_view comment);
     const std::string& getRowComment() const;
 
-    /** 
-    * sets the string that indicates that a row should be kept
-    * @see CSVReader::defaultKeepOnly 
-    */
+    /**
+     * sets the string that indicates that a row should be kept
+     * @see CSVReader::defaultKeepOnly
+     */
     CSVReader& setKeepOnly(std::string_view only);
     const std::string& getKeepOnly() const;
 
--- a/modules/dataframe/include/inviwo/dataframe/io/csvreader.h
+++ b/modules/dataframe/include/inviwo/dataframe/io/csvreader.h
@@ -89,11 +89,17 @@ public:
     CSVReader& setNumberOfExampleRows(size_t rows);
     size_t getNumberOfExamplesRows() const;
 
-    /** @see CSVReader::defaultRowComment */
+    /** 
+    * sets the string that indicates that a row should be commented/removed
+    * @see CSVReader::defaultRowComment 
+    */
     CSVReader& setRowComment(std::string_view comment);
     const std::string& getRowComment() const;
 
-    /** @see CSVReader::defaultKeepOnly */
+    /** 
+    * sets the string that indicates that a row should be kept
+    * @see CSVReader::defaultKeepOnly 
+    */
     CSVReader& setKeepOnly(std::string_view only);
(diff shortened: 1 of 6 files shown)


modules/dataframe/include/inviwo/dataframe/io/csvreader.h
```diff
@@ -89,6 +89,14 @@ class IVW_MODULE_DATAFRAME_API CSVReader : public DataReaderType<DataFrame> {
     CSVReader& setNumberOfExampleRows(size_t rows);
     size_t getNumberOfExamplesRows() const;
 
+    /** @see CSVReader::defaultRowComment */
```
martinfalk: this comment doesn't explain what it is doing/what the purpose is. The member only refers back to this function.

modules/dataframe/include/inviwo/dataframe/io/csvreader.h
```diff
@@ -89,6 +89,14 @@ class IVW_MODULE_DATAFRAME_API CSVReader : public DataReaderType<DataFrame> {
...
+    CSVReader& setRowComment(std::string_view comment);
+    const std::string& getRowComment() const;
+
+    /** @see CSVReader::defaultKeepOnly */
```
martinfalk: this comment doesn't explain what it is doing/what the purpose is. The member only refers back to this function.

modules/dataframe/src/io/csvreader.cpp
```diff
@@ -510,6 +524,24 @@ std::shared_ptr<DataFrame> CSVReader::readData(std::istream& stream) const {
         std::remove_if(rows.begin(), rows.end(), [](const auto& row) { return row.first.empty(); }),
         rows.end());
 
+    // Ignore comment lines
+    if (!rowComment_.empty()) {
+        rows.erase(std::remove_if(rows.begin(), rows.end(),
```
martinfalk, line 530: It should be pointed out that the comment _must_ start in the very first column (no leading spaces).

martinfalk: Also mention that comments are removed _before_ keep lines is applied.

martinfalk (changes requested):

modules/dataframe/include/inviwo/dataframe/io/csvreader.h
petersteneteg: no it refers to  defaultRowComment not setRowComment.

martinfalk:
```c++
    /** @see CSVReader::setRowComment */
    static constexpr std::string_view defaultRowComment = "";
```

martinfalk (dismissed):

modules/dataframe/src/processors/csvsource.cpp
```diff
@@ -59,6 +59,8 @@ CSVSource::CSVSource(const std::string& file)
     , stripQuotes_("stripQuotes", "Strip surrounding quotes", CSVReader::defaultStripQuotes)
     , doublePrecision_("doublePrecision", "Double Precision", CSVReader::defaultDoublePrecision)
     , exampleRows_("exampleRows", "Example Rows", CSVReader::defaultNumberOfExampleRows, 0, 10000)
+    , rowComment_("rowComment", "Comment marker", CSVReader::defaultRowComment)
```
ResearchDaniel, line 62: I think that it might be more explicit for a user with:
"Exclude lines starting with"
"Only include lines starting with"

Perhaps too long?
We could place them in a group property with a title "Row filters", which creates a hiearchy in the GUI. Then they could simply be "Exclude", "Include".

modules/dataframe/src/io/csvreader.cpp
```diff
@@ -510,6 +524,25 @@ std::shared_ptr<DataFrame> CSVReader::readData(std::istream& stream) const {
         std::remove_if(rows.begin(), rows.end(), [](const auto& row) { return row.first.empty(); }),
         rows.end());
 
+    // Ignore comment lines 
+    // The row must start with the comment character. Applied before keep only
+    if (!rowComment_.empty()) {
+        rows.erase(std::remove_if(rows.begin(), rows.end(),
+                                  [this](const auto& row) {
+                                      return row.first.substr(0, rowComment_.size()) == rowComment_;
```
ResearchDaniel, line 532: I think that it would make more sense if the rows where never added in the first place. Can't the include/exclude filters be moved to line 519?

ResearchDaniel (changes requested):
I think that the include/exclude filters should be moved to the row parsing on line 519.

Also, this is up for discussion, but I think that it would be more explicit to name the functionality something like "includeRowFilter", "excludeRowFilter"

modules/dataframe/include/inviwo/dataframe/io/csvreader.h
```diff
@@ -89,6 +89,20 @@ class IVW_MODULE_DATAFRAME_API CSVReader : public DataReaderType<DataFrame> {
     CSVReader& setNumberOfExampleRows(size_t rows);
     size_t getNumberOfExamplesRows() const;
 
+    /** 
+    * sets the string that indicates that a row should be commented/removed
```
martinfalk: from [ResearchDaniel](https://github.com/ResearchDaniel)
Suggestion:
Set characters indicating a line comment, for example "//". Lines starting with \p comment will be ignored during parsing. By default it is set to empty, meaning that no lines will be treated as comments.

modules/dataframe/include/inviwo/dataframe/io/csvreader.h
```diff
@@ -89,6 +89,20 @@ class IVW_MODULE_DATAFRAME_API CSVReader : public DataReaderType<DataFrame> {
...
+    * @see CSVReader::defaultRowComment 
+    */
+    CSVReader& setRowComment(std::string_view comment);
+    const std::string& getRowComment() const;
+
+    /** 
+    * sets the string that indicates that a row should be kept
```
martinfalk: from  [ResearchDaniel](https://github.com/ResearchDaniel)
The functionality is a bit ambigous. For example: Does the row need to start with "only"? We might want to have a consistent naming for the comment/keep only. An alternative naming for these functions could be:
**includeRows(...), excludeRows(...)**
Suggest something in the lines of:
```
/*
 * Only parse lines starting with the specified characters. The \p startingWith character(s) will not be considered as values for the line.
 */
CSVReader& includeRows(std::string_view startingWith);
```

modules/dataframe/src/io/csvreader.cpp
petersteneteg: yes, that should work, both this and the empty line check could be there. probably we want to call out to a
` bool filter(std::string_view line)` function there. 
Then the `filter` function can decide what to include/exclude. 

One might have to consider if the Header row is affected by all this somehow?
We might want to apply some filter after the header row is removed.

EmiliaStahlbom: Sounds reasonable to me.

Re:header: We could include a firstRowHeader-check by line 519 to see if it should be kept. Otherwise i guess the filtering should apply to the first row as well?

martinfalk:
Rebased into #1259 and added more filtering.

PR #1253, opened 2022-02-14
Feature/qt cosmetic

martinfalk:
Some cosmetic changes, mostly for the welcome widget.


ResearchDaniel (dismissed):

modules/animationqt/src/animationqtmodule.cpp
```diff
@@ -151,45 +152,47 @@ AnimationQtModule::AnimationQtModule(InviwoApplication* app)
...
-                    editor_->loadState();
-                    // Release pointer if destroyed by Qt before module is destroyed
-                    QObject::connect(editor_.get(), &QObject::destroyed,
-                                     [this](QObject*) { editor_.release(); });
-                    editor_->setVisible(true);
-                } else {
-                    editor_->setVisible(!editor_->isVisible());
-                }
+            QObject::connect(action, &QAction::triggered, [this, win, app, menu, action]() {
+                util::OnScopeExit onexit([action]() { delete action; });
```
petersteneteg: Is it ok to remove the action and disconnect from within the call itself? Think you need to add some comment about what is going on here. I know that you are doing, but it is probably quite hard to figure out I you aren't that familiar with qt. I would also be even better if the "call once" aspects here that are duplicated in 2 places were not. I.e make the "call once" a somewhat generic algorithm here.

petersteneteg (changes requested):

petersteneteg (approved):

PR #1254, opened 2022-02-14
Added support for generation python docx using sphinx

petersteneteg:
image


cmake/pybind11.cmake
```diff
@@ -31,7 +31,17 @@
 
 # Needed for the function pybind11_add_module and dependency pybind11::module 
 
-set(_allPyBindWrappers "" CACHE INTERNAL  "")
+define_property(
+    GLOBAL PROPERTY IVW_PYMODULE_LIST
+    BRIEF_DOCS "List of inviwo python module targets"
```
ResearchDaniel, line 36: Not that it matters really, but it should be Inviwo Python (capital letters)

docs/CMakeLists.txt
```diff
@@ -28,406 +28,51 @@
...
+cmake_dependent_option(IVW_DOXYGEN_OPEN_HTML_AFTER_BUILD
+                       "Open the generated doxygen HTML when build is done"
+                       OFF
+                       "IVW_DOXYGEN_PROJECT"
+                       OFF)
+
+if(IVW_DOXYGEN_PROJECT)
+    # sets: DOXYGEN_FOUND, DOXYGEN_EXECUTABLE, DOXYGEN_DOT_FOUND, DOXYGEN_DOT_EXECUTABLE
+    find_package(Doxygen REQUIRED)
+    find_package(Python3 COMPONENTS Interpreter REQUIRED)
```
ResearchDaniel, line 43: Can sphinx be added here?

ResearchDaniel (dismissed):

docs/CMakeLists.txt
petersteneteg: at the moment I use python3 -m sphinx ... to call it. There is also a script sphinx-build that one can use. But it seems safer to use the module directly. Maybe it would be possible to add some check to see it the sphinx module is installed. And you will only run into that issue when building the target. On the other hand you can then also just pip install it and don't have to rerun cmake...

ResearchDaniel: Is it possible to do add it like we do with numpy:
find_package(Python3 COMPONENTS Interpreter Development **NumPy** REQUIRED)

petersteneteg: That would have been nice, But I think numpy is a special case https://cmake.org/cmake/help/latest/module/FindPython3.html

ResearchDaniel (approved):

PR #1259, opened 2022-02-18
Feature/csv filtering

martinfalk:
* added some generic filtering for CSV files to filter rows and columns
    - removing/including rows starting with strings
    - line ranges
    - string, int, float, and double comparisons
    - int, float, and double ranges
* most of the filters should be applicable to DataFrames as well (future PR)

### DIFF
--- a/include/inviwo/core/io/tempfilehandle.h
+++ b/include/inviwo/core/io/tempfilehandle.h
@@ -45,7 +45,7 @@ namespace util {
 class IVW_CORE_API TempFileHandle {
 public:
     explicit TempFileHandle(const std::string& prefix = "", const std::string& suffix = "",
-                            const char* mode = "rw");
+                            const char* mode = "wb+");
 
     TempFileHandle(const TempFileHandle&) = delete;
     TempFileHandle& operator=(const TempFileHandle&) = delete;
--- a/modules/dataframe/include/inviwo/dataframe/io/csvreader.h
+++ b/modules/dataframe/include/inviwo/dataframe/io/csvreader.h
@@ -93,6 +93,10 @@ public:
     CSVReader& setRowComment(std::string_view comment);
     const std::string& getRowComment() const;
 
+    /** @see CSVReader::defaultKeepOnly */
+    CSVReader& setKeepOnly(std::string_view only);
+    const std::string& getKeepOnly() const;
+
     /** @see CSVReader::defaultLocale */
     CSVReader& setLocale(std::string_view loc);
     const std::string& getLocale() const;
@@ -181,6 +185,8 @@ public:
     static constexpr size_t defaultNumberOfExampleRows = 50;
     /** @see CSVReader::setRowComment */
     static constexpr std::string_view defaultRowComment = "";
+    /** @see CSVReader::setKeepOnly */
+    static constexpr std::string_view defaultKeepOnly = "";
     /** @see CSVReader::setLocale */
     static constexpr std::string_view defaultLocale = "C";
     /** @see CSVReader::setHandleEmptyFields */
@@ -209,6 +215,7 @@ private:
     bool doublePrecision_;
     size_t exampleRows_;
     std::string rowComment_;
+    std::string keepOnly_;
     std::string locale_;
     EmptyField emptyField_;
 };
--- a/modules/dataframe/include/inviwo/dataframe/io/csvreader.h
+++ b/modules/dataframe/include/inviwo/dataframe/io/csvreader.h
@@ -152,6 +152,7 @@ public:
      * * NumberOfExampleRows (size_t)
      * * Locale (string)
      * * HandleEmptyFields (EmptyField)
+     * * Filters (csvfilters::Filters)
      */
     virtual bool setOption(std::string_view key, std::any value) override;
 
@@ -166,6 +167,7 @@ public:
      * * NumberOfExampleRows (size_t)
      * * Locale (string)
      * * HandleEmptyFields (EmptyField)
+     * * Filters (csvfilters::Filters)
      */
     virtual std::any getOption(std::string_view key) override;
 
--- a/modules/dataframe/include/inviwo/dataframe/io/csvreader.h
+++ b/modules/dataframe/include/inviwo/dataframe/io/csvreader.h
@@ -89,11 +89,17 @@ public:
     CSVReader& setNumberOfExampleRows(size_t rows);
     size_t getNumberOfExamplesRows() const;
 
-    /** @see CSVReader::defaultRowComment */
+    /** 
+    * sets the string that indicates that a row should be commented/removed
+    * @see CSVReader::defaultRowComment 
+    */
     CSVReader& setRowComment(std::string_view comment);
     const std::string& getRowComment() const;
 
-    /** @see CSVReader::defaultKeepOnly */
+    /** 
+    * sets the string that indicates that a row should be kept
+    * @see CSVReader::defaultKeepOnly 
+    */
     CSVReader& setKeepOnly(std::string_view only);
     const std::string& getKeepOnly() const;
 
--- a/modules/dataframe/include/inviwo/dataframe/io/csvreader.h
+++ b/modules/dataframe/include/inviwo/dataframe/io/csvreader.h
@@ -34,6 +34,7 @@
 #include <inviwo/core/io/datareader.h>
 #include <inviwo/core/io/datareaderexception.h>
 #include <inviwo/dataframe/datastructures/dataframe.h>
+#include <inviwo/dataframe/util/filters.h>
 
 namespace inviwo {
 
@@ -41,9 +42,9 @@ namespace inviwo {
  * \class CSVReader
  * \ingroup dataio
  *
- * \brief A reader for comma separated value (CSV) files with customizable delimiters.
+ * \brief A reader for comma separated value (CSV) files with customizable delimiters and filters.
  * The default delimiter is ',' and headers are included. Floating point values are stored as
- * float32.
+ * float32 unless double precision is enabled.
  */
 class IVW_MODULE_DATAFRAME_API CSVReader : public DataReaderType<DataFrame> {
 public:
@@ -78,7 +79,7 @@ public:
     const std::string& getUnitRegexp() const;
 
     /**
-     * sets the precision for columns containing floating point values. If @p useDoublePrecision is
+     * Sets the precision for columns containing floating point values. If @p useDoublePrecision is
      * true, values are stored as double (64 bits), otherwise float (32 bits) is used.
      * @see CSVReader::defaultDoublePrecision
      */
@@ -89,23 +90,16 @@ public:
     CSVReader& setNumberOfExampleRows(size_t rows);
     size_t getNumberOfExamplesRows() const;
 
-    /** 
-    * sets the string that indicates that a row should be commented/removed
-    * @see CSVReader::defaultRowComment 
-    */
-    CSVReader& setRowComment(std::string_view comment);
-    const std::string& getRowComment() const;
-
-    /** 
-    * sets the string that indicates that a row should be kept
-    * @see CSVReader::defaultKeepOnly 
-    */
-    CSVReader& setKeepOnly(std::string_view only);
-    const std::string& getKeepOnly() const;
-
     /** @see CSVReader::defaultLocale */
     CSVReader& setLocale(std::string_view loc);
     const std::string& getLocale() const;
+
+    /**
+     * Sets row and column filters.
+     * @see Filters
+     */
+    CSVReader& setFilters(const csvfilters::Filters& filters);
+    const csvfilters::Filters& getFilters() const;
 
     /**
      * How to handle missing / empty data
@@ -189,10 +183,6 @@ public:
     static constexpr bool defaultDoublePrecision = false;
     /** @see CSVReader::setNumberOfExampleRows */
     static constexpr size_t defaultNumberOfExampleRows = 50;
-    /** @see CSVReader::setRowComment */
-    static constexpr std::string_view defaultRowComment = "";
-    /** @see CSVReader::setKeepOnly */
-    static constexpr std::string_view defaultKeepOnly = "";
     /** @see CSVReader::setLocale */
     static constexpr std::string_view defaultLocale = "C";
     /** @see CSVReader::setHandleEmptyFields */
@@ -213,6 +203,8 @@ private:
         DataFrame& df, const std::vector<TypeCounts>& types,
         const std::vector<std::string>& headers) const;
 
+    bool skipRow(std::string_view row, size_t lineNumber, bool filterOnHeader) const;
+
     std::string delimiters_;
     bool stripQuotes_;
     bool firstRowHeader_;
@@ -220,10 +212,9 @@ private:
     std::string unitRegexp_;
     bool doublePrecision_;
     size_t exampleRows_;
-    std::string rowComment_;
-    std::string keepOnly_;
     std::string locale_;
     EmptyField emptyField_;
+    csvfilters::Filters filters_;
 };
 
 }  // namespace inviwo
--- a/modules/dataframe/include/inviwo/dataframe/io/csvreader.h
+++ b/modules/dataframe/include/inviwo/dataframe/io/csvreader.h
@@ -89,6 +89,10 @@ public:
     CSVReader& setNumberOfExampleRows(size_t rows);
     size_t getNumberOfExamplesRows() const;
 
+    /** @see CSVReader::defaultRowComment */
+    CSVReader& setRowComment(std::string_view comment);
+    const std::string& getRowComment() const;
+
     /** @see CSVReader::defaultLocale */
     CSVReader& setLocale(std::string_view loc);
     const std::string& getLocale() const;
@@ -175,6 +179,8 @@ public:
     static constexpr bool defaultDoublePrecision = false;
     /** @see CSVReader::setNumberOfExampleRows */
     static constexpr size_t defaultNumberOfExampleRows = 50;
+    /** @see CSVReader::setRowComment */
+    static constexpr std::string_view defaultRowComment = "";
     /** @see CSVReader::setLocale */
     static constexpr std::string_view defaultLocale = "C";
     /** @see CSVReader::setHandleEmptyFields */
@@ -202,6 +208,7 @@ private:
     std::string unitRegexp_;
     bool doublePrecision_;
     size_t exampleRows_;
+    std::string rowComment_;
     std::string locale_;
     EmptyField emptyField_;
 };
--- a/modules/dataframe/include/inviwo/dataframe/processors/csvsource.h
+++ b/modules/dataframe/include/inviwo/dataframe/processors/csvsource.h
@@ -84,7 +84,6 @@ private:
     BoolProperty doublePrecision_;
     IntSizeTProperty exampleRows_;
     StringProperty rowComment_;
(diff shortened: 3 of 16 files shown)


modules/dataframe/src/processors/csvsource.cpp
```diff
@@ -133,4 +156,92 @@ void CSVSource::deserialize(Deserializer& d) {
...
+    };
+
+    csvfilters::Filters filters;
+    parse(includeFilters_, filters.includeRows, filters.includeItems);
+    parse(excludeFilters_, filters.excludeRows, filters.excludeItems);
+
+    if (!rowComment_.get().empty()) {
+        filters.excludeRows.push_back(csvfilters::rowBegin(rowComment_, true));
+    }
+    if (!keepOnly_.get().empty()) {
```
ResearchDaniel: I would guess that this use case is rather uncommon.
Having “keep top N”/“keep range”, as a special quick access filter would in my opinion be more useful.

modules/dataframe/src/properties/filterlistproperty.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+    {
+        auto rowBegin = std::make_unique<BoolCompositeProperty>("rowBegin", "Row Begin", true);
+        rowBegin->addProperty(onHeaderProp());
+        rowBegin->addProperty(std::make_unique<StringProperty>("match", "Matching String", ""));
+
+        addPrefab(std::move(rowBegin));
+    }
+    {
+        auto lineRange = std::make_unique<BoolCompositeProperty>("lineRange", "Line Range", true);
```
ResearchDaniel: We use “row” for most other cases. We should make it consistent

modules/dataframe/src/properties/filterlistproperty.cpp
```diff
@@ -0,0 +1,207 @@
...
+            std::numeric_limits<float>::max(), 0.1f, 0.0f, InvalidationLevel::InvalidOutput,
+            PropertySemantics::Text));
+
+        addPrefab(std::move(floatRange));
+    }
+    {
+        auto doubleRange =
+            std::make_unique<BoolCompositeProperty>("doubleRangeItem", "Ddouble Range", true);
+        doubleRange->addProperty(columnProp());
+        doubleRange->addProperty(std::make_unique<DoubleMinMaxProperty>(
```
ResearchDaniel, line 206: Would it be possible to only have a Value Range filter of double type and use that for all?

I think that it would be better if a user did not have to care about the value type of the column.

ResearchDaniel: Same reasoning goes for the “match” case

ResearchDaniel (changes requested):
Nice work!
Dealing with the headers adds more complexity. I would have excluded it to simplify the code and usage, but now you have already included it :)

- it feels like some implementation details have sneaked into the end result. I think that we should not expose int, float, double properties to the user. Just use double.
- can we use names for the columns instead of integer? I understand that this might involve quite some work so it is more of a suggestion. The order might not be known beforehand if a user supplies a file. We have default column names if row headers are missing if I am not mistaken (or could assume numbers in that case)

martinfalk:
> * it feels like some implementation details have sneaked into the end result. I think that we should not expose int, float, double properties to the user. Just use double.

Not so sure since the types are not yet known during/before parsing. Assuming you expect integer values, why would you then use double comparisons with a potential epsilon? 

> * can we use names for the columns instead of integer? I understand that this might involve quite some work so it is more of a suggestion. The order might not be known beforehand if a user supplies a file. We have default column names if row headers are missing if I am not mistaken (or could assume numbers in that case)

Column names do not yet exist at that point when parsing the CSV file. The properties will be slightly different for the DataFrame filtering where one can translate existing column names to column indices.

modules/dataframe/src/processors/csvsource.cpp
martinfalk: I left it in there as it is the opposite of comment rows.

modules/dataframe/src/properties/filterlistproperty.cpp
martinfalk: Because here it actually refers to the line numbers in the file and not the index of the data items (rows).

modules/dataframe/src/properties/filterlistproperty.cpp
martinfalk: From a Filter perspective, the type of filter just determines the type a string is converted to before the comparison is applied.

include/inviwo/core/io/tempfilehandle.h
```diff
@@ -45,7 +45,7 @@ namespace util {
 class IVW_CORE_API TempFileHandle {
 public:
     explicit TempFileHandle(const std::string& prefix = "", const std::string& suffix = "",
-                            const char* mode = "rw");
+                            const char* mode = "wb+");
```
petersteneteg, line 48: does this work on linux/mac? seems non standard

modules/dataframe/include/inviwo/dataframe/io/csvreader.h
```diff
@@ -93,6 +94,13 @@ class IVW_MODULE_DATAFRAME_API CSVReader : public DataReaderType<DataFrame> {
     CSVReader& setLocale(std::string_view loc);
     const std::string& getLocale() const;
 
+    /**
+     * Sets row and column filters.
+     * @see Filters
+     */
+    CSVReader& setFilters(const csvfilters::Filters& filters);
```
petersteneteg, line 101: Add to options too

modules/dataframe/include/inviwo/dataframe/processors/csvsource.h
```diff
@@ -80,6 +83,10 @@ class IVW_MODULE_DATAFRAME_API CSVSource : public Processor {
     BoolProperty stripQuotes_;
     BoolProperty doublePrecision_;
     IntSizeTProperty exampleRows_;
+    StringProperty rowComment_;
+    StringProperty keepOnly_;
```
petersteneteg: skip this one.

modules/dataframe/include/inviwo/dataframe/properties/filterlistproperty.h
```diff
@@ -0,0 +1,84 @@
...
+ *********************************************************************************/
+#pragma once
+
+#include <inviwo/dataframe/dataframemoduledefine.h>
+#include <inviwo/core/properties/listproperty.h>
+
+namespace inviwo {
+
+/**
+ * @class FilterListProperty
```
petersteneteg: No need to add @class when you are commenting a class. Only needed if the comment is somewhere else.

modules/dataframe/include/inviwo/dataframe/properties/filterlistproperty.h
```diff
@@ -0,0 +1,84 @@
...
+ * @brief A list property with different sub-properties for defining row and item filters.
+ *
+ * This property has a number of prefab objects for adding row and item filter properties
+ * (BoolCompositeProperty). A row filter is applied to the entire row of a dataset. Item filters, in
+ * contrast, are applied to the individual data items, that is columns. The distinction between row
```
petersteneteg: "data item" is not the came as columns.

modules/dataframe/include/inviwo/dataframe/util/filters.h
```diff
@@ -0,0 +1,135 @@
...
+
+namespace filters {
+
+enum class StringComp { Equal, NotEqual, Regex, RegexPartial };
+
+enum class NumberComp { Equal, NotEqual, Less, LessEqual, Greater, GreaterEqual };
+
+}  // namespace filters
+
+namespace csvfilters {
```
petersteneteg, line 51: Unless the namespace also has a comment, it will not show any comments for the stuff inside in doxygen.

modules/dataframe/src/io/csvreader.cpp
```diff
@@ -484,13 +491,92 @@ std::vector<std::function<void(std::string_view, size_t, size_t)>> CSVReader::ad
...
+                                    retval |=
+                                        std::visit(util::overloaded{stringFilter, intFilter,
+                                                                    floatFilter, doubleFilter},
+                                                   f.filter);
+                                }
+                            }
+                        }
+                    });
+        return retval;
+    };
```
petersteneteg:
```suggestion
    const bool cLocale = locale_ == "C";
    auto test = util::overloaded{
        [&](const std::function<bool(std::string_view)>& func) { return func(cell); },
        [&](const std::function<bool(int)>& func) {
            if (auto val = util::toNumber<int>(cell, cLocale)) {
                return func(*val);
            }
            return neutralValue;
        },
        [&](const std::function<bool(float)>& func) {
            if (auto val = util::toNumber<float>(cell, cLocale)) {
                return func(*val);
            }
            return neutralValue;
        },
        [&](const std::function<bool(double)>& func) {
            if (auto val = util::toNumber<double>(cell, cLocale)) {
                return func(*val);
            }
            return neutralValue;
        }};

    auto filterItems = [&](const std::vector<csvfilters::ItemFilter>& filters, bool neutralValue) {
        bool retval = false;
        util::parse(row, delimiters_, std::nullopt, lineNumber,
                    [&](std::string_view cell, size_t colIndex, [[maybe_unused]] size_t part) {
                        for (const auto& f : filters) {
                            if ((!firstRowHeader_ || (f.filterOnHeader == filterOnHeader)) &&
                                f.column == colIndex) {
                                retval |= std::visit(test, f.filter);
                            }
                        }
                    });
        return retval;
    };
```

modules/dataframe/src/io/csvreader.cpp
```diff
@@ -502,14 +588,11 @@ std::shared_ptr<DataFrame> CSVReader::readData(std::istream& stream) const {
     std::vector<std::pair<std::string_view, size_t>> rows;
     util::parse(content, "\n", std::nullopt, std::nullopt,
                 [&](std::string_view line, [[maybe_unused]] size_t index, size_t lineNumber) {
-                    rows.emplace_back(line, lineNumber);
+                    if (!line.empty() && !skipRow(line, lineNumber, true)) {
```
petersteneteg: Think the empty line test might also be a filter. for the cases when it is needed.

modules/dataframe/src/processors/csvsource.cpp
```diff
@@ -133,4 +156,92 @@ void CSVSource::deserialize(Deserializer& d) {
...
+    auto parse = [&](const auto& filterProp, auto& rowFilters, auto& itemFilters) {
+        for (auto prop : filterProp) {
+            if (auto cp = dynamic_cast<BoolCompositeProperty*>(prop)) {
+                if (!cp->isChecked()) continue;
+
+                const auto& identifier = cp->getIdentifier();
+
+                if (startsWith(identifier, "rowBegin")) {
+                    rowFilters.push_back(csvfilters::rowBegin(
+                        std::string_view{detail::getValue<StringProperty>(cp, "match")},
```
petersteneteg: you should not need to add string_view here

petersteneteg (changes requested):
Nice work, some minor comments..

EmiliaStahlbom (dismissed):
Nice work! This is honestly above my current programming skill level, but I looked through it and it looks good to me.

petersteneteg:
@EmiliaStahlbom One of the benefits of doing code review is that is is not just about checking the code in the PR, but it is also a great opportunity for the reviewer to learn about the codebase in many ways :)

EmiliaStahlbom:
@petersteneteg Yeah I'm sure learning a lot! But I feel like I'm missing a lot of the more fine grained parts, which will come with time and experience  :)

include/inviwo/core/io/tempfilehandle.h
martinfalk: Should be fine according to [cppreference.com](https://en.cppreference.com/w/c/io/fopen)

modules/dataframe/include/inviwo/dataframe/util/filters.h
petersteneteg: see https://github.com/inviwo/inviwo/blob/master/docs/namespaces.dox

include/inviwo/core/io/tempfilehandle.h
martinfalk: "rw" is definitely non-standard and `fopen` in MSVC terminates the program when using unsupported modes.

https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/fopen-wfopen?view=msvc-170#generic-text-routine-mappings

modules/dataframe/src/io/csvreader.cpp
```diff
@@ -49,6 +49,8 @@
 #include <type_traits>
 #include <regex>
 
+#pragma optimize("", off)
```
petersteneteg: remove

modules/dataframe/src/io/csvreader.cpp
```diff
@@ -608,11 +615,18 @@ std::shared_ptr<DataFrame> CSVReader::readData(std::istream& stream) const {
...
+        // demangle duplicate column headers similar to Pandas' default behavior
+        // @see https://pandas.pydata.org/docs/reference/api/pandas.read_csv.html
+        std::unordered_map<std::string, int> headerCount;
+        for (auto& header : headers) {
+            if (int count = ++headerCount[header]; count > 1) {
+                header.append(fmt::format(".{}", count - 1));
+            }
+        }
+        if (headerCount.size() != headers.size()) {
+            LogWarn("Duplicate column headers detected. Affected headers are demangled.");
```
petersteneteg: I would rather either document that we always do this. when we parse the header, or we make it an option. Logging is pointless.

modules/dataframe/src/io/csvreader.cpp
```diff
@@ -608,11 +615,18 @@ std::shared_ptr<DataFrame> CSVReader::readData(std::istream& stream) const {
...
-            std::remove_if(rows.begin(), rows.end(),
-                           [&](const auto& row) { return skipRow(row.first, row.second, false); }),
-            rows.end());
+
+        // demangle duplicate column headers similar to Pandas' default behavior
+        // @see https://pandas.pydata.org/docs/reference/api/pandas.read_csv.html
+        std::unordered_map<std::string, int> headerCount;
+        for (auto& header : headers) {
+            if (int count = ++headerCount[header]; count > 1) {
+                header.append(fmt::format(".{}", count - 1));
```
petersteneteg: How will this interact with extraction of units?

modules/dataframe/src/io/csvreader.cpp
petersteneteg: generally I think one should be extremely careful with changing any data that the user enters without the user explicitly asking for that change. And unless we are trying to enforce unique column headers I see no point. And if we try to do that, that should mostly likely be done in a different PR after some discussion...

modules/dataframe/src/properties/filterlistproperty.cpp
```diff
@@ -84,7 +94,7 @@ FilterListProperty::FilterListProperty(std::string_view identifier, std::string_
             InvalidationLevel::InvalidOutput, PropertySemantics::Text);
     };
 
-    {
+    if (supportedFilters & FilterType::StringItem) {
```
petersteneteg: you don't have to change all of these, but I think 
`supportedFilters.contains(FilterType::StringItem)`
is much more readable.

petersteneteg (changes requested):

modules/dataframe/src/io/csvreader.cpp
martinfalk: very good point.

modules/dataframe/src/io/csvreader.cpp
ResearchDaniel: We could add an option "Require unique headers" to make it more explicit

petersteneteg: unless we enforce that a data frame has unique header names, I don't see the point

modules/dataframe/src/properties/filterlistproperty.cpp
martinfalk: Added a setting to the ListProperty for enabling subsets of properties.

modules/dataframe/src/io/csvreader.cpp
martinfalk: Removed duplicate header handling for now.

modules/dataframe/src/io/csvreader.cpp
```diff
@@ -586,6 +582,7 @@ std::shared_ptr<DataFrame> CSVReader::readData(std::istream& stream) const {
     }
 
     std::string content{std::istreambuf_iterator<char>(stream), std::istreambuf_iterator<char>()};
+    content = rtrim(content);
```
petersteneteg: ahh don't use that function. you should return a string_view not a copy of the content.

modules/dataframe/src/io/csvreader.cpp
```diff
@@ -484,32 +495,106 @@ std::vector<std::function<void(std::string_view, size_t, size_t)>> CSVReader::ad
...
+        trimmed = trimmed.substr(0, pos + 1);
+    }
 
     std::vector<std::pair<std::string_view, size_t>> rows;
-    util::parse(content, "\n", std::nullopt, std::nullopt,
+    util::parse(trimmed, "\n", std::nullopt, std::nullopt,
                 [&](std::string_view line, [[maybe_unused]] size_t index, size_t lineNumber) {
-                    rows.emplace_back(line, lineNumber);
+                    if (!skipRow(line, lineNumber, true)) {
+                        rows.emplace_back(line, lineNumber);
```
ResearchDaniel, line 594: Can't we do the second parsing here instead of below? Would avoid intermediate append and two skipRow calls.
Two skipRow calls should not be necessary?
```
util::parse(line, delimiters_, headers.size(), lineNumber,
                        [&, l = lineNumber](std::string_view cell, size_t index,
                                            [[maybe_unused]] size_t part) {
                            appenders[index](cell, l, index + 1);
                        });
```

modules/dataframe/src/processors/csvsource.cpp
```diff
@@ -133,4 +156,92 @@ void CSVSource::deserialize(Deserializer& d) {
...
+                    itemFilters.push_back(csvfilters::doubleMatch(
+                        detail::getValue<IntProperty>(cp, "column"),
+                        detail::getValue<TemplateOptionProperty<filters::NumberComp>>(cp, "comp"),
+                        detail::getValue<DoubleProperty>(cp, "value"),
+                        detail::getValue<DoubleProperty>(cp, "epsilon")));
+                } else if (startsWith(identifier, "intRangeItem")) {
+                    const auto& range = detail::getValue<IntMinMaxProperty>(cp, "range");
+                    itemFilters.push_back(csvfilters::intRange(
+                        detail::getValue<IntProperty>(cp, "column"), range.x, range.y));
+                } else if (startsWith(identifier, "floatRangeItem")) {
```
ResearchDaniel, line 224: Would also be possible to use float/double comparison depending on if the doubleprecision checkbox is enabled. Then we would get rid of the floatXXX filters

ResearchDaniel (approved):

modules/dataframe/src/io/csvreader.cpp
petersteneteg: you also have to find the types... so you need to parse at least the first part twice. not matter what... 
and the row vector only stores string views so it is very cheap.

petersteneteg (approved):

PR #1260, opened 2022-02-21
Core/Various: Update to Ports/Properties/Factories to use string_views

petersteneteg:
Properties and Ports should take identifiers and display names as string_views.

### DIFF
--- a/include/inviwo/core/ports/portfactoryobject.h
+++ b/include/inviwo/core/ports/portfactoryobject.h
@@ -43,11 +43,11 @@ namespace inviwo {
 
 class IVW_CORE_API InportFactoryObject {
 public:
-    InportFactoryObject(const std::string& className);
+    InportFactoryObject(std::string_view className);
     virtual ~InportFactoryObject() = default;
 
     virtual std::unique_ptr<Inport> create() = 0;
-    virtual std::unique_ptr<Inport> create(const std::string& identifier) = 0;
+    virtual std::unique_ptr<Inport> create(std::string_view identifier) = 0;
     const std::string& getClassIdentifier() const;
 
 protected:
@@ -65,18 +65,18 @@ public:
 
     virtual std::unique_ptr<Inport> create() override { return std::make_unique<T>(className_); }
 
-    virtual std::unique_ptr<Inport> create(const std::string& identifier) override {
+    virtual std::unique_ptr<Inport> create(std::string_view identifier) override {
         return std::make_unique<T>(identifier);
     }
 };
 
 class IVW_CORE_API OutportFactoryObject {
 public:
-    OutportFactoryObject(const std::string& className);
+    OutportFactoryObject(std::string_view className);
     virtual ~OutportFactoryObject() = default;
 
     virtual std::unique_ptr<Outport> create() = 0;
-    virtual std::unique_ptr<Outport> create(const std::string& identifier) = 0;
+    virtual std::unique_ptr<Outport> create(std::string_view identifier) = 0;
     const std::string& getClassIdentifier() const;
 
 protected:
@@ -94,8 +94,12 @@ public:
 
     virtual std::unique_ptr<Outport> create() override { return std::make_unique<T>(className_); }
 
-    virtual std::unique_ptr<Outport> create(const std::string& identifier) override {
-        return std::make_unique<T>(identifier);
+    virtual std::unique_ptr<Outport> create(std::string_view identifier) override {
+        if constexpr (std::is_constructible_v<T, std::string_view>) {
+            return std::make_unique<T>(identifier);
+        } else {
+            return std::make_unique<T>(std::string(identifier));
+        }
     }
 };
 
--- a/include/inviwo/core/util/factory.h
+++ b/include/inviwo/core/util/factory.h
@@ -36,8 +36,9 @@
 #include <inviwo/core/util/logcentral.h>
 
 #include <string>
+#include <string_view>
 #include <memory>
-#include <unordered_map>
+#include <map>
 
 namespace inviwo {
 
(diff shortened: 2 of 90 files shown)


include/inviwo/core/ports/portfactoryobject.h
```diff
@@ -94,8 +94,12 @@ class OutportFactoryObjectTemplate : public OutportFactoryObject {
 
     virtual std::unique_ptr<Outport> create() override { return std::make_unique<T>(className_); }
 
-    virtual std::unique_ptr<Outport> create(const std::string& identifier) override {
-        return std::make_unique<T>(identifier);
+    virtual std::unique_ptr<Outport> create(std::string_view identifier) override {
+        if constexpr (std::is_constructible_v<T, std::string_view>) {
+            return std::make_unique<T>(identifier);
+        } else {
+            return std::make_unique<T>(std::string(identifier));
```
ResearchDaniel, line 101: Too harsh to force the ports to take a string_view and produce a nice interpretable compile error instead?

include/inviwo/core/util/factory.h
```diff
@@ -64,55 +65,100 @@ class FactoryObservable : public Observable<FactoryObserver<T>> {
...
+    std::map<Key, M*, std::less<>> map_;
+};
+
+template <typename K>
+struct FactoryKeyType {
+    using type = std::remove_cv_t<std::remove_reference_t<K>>;
+};
+template <>
+struct FactoryKeyType<std::string_view> {
+    using type = std::string;
```
ResearchDaniel, line 151: This deserves a comment. Why not `using type = string_view`?

include/inviwo/core/util/factory.h
```diff
@@ -64,55 +65,100 @@ class FactoryObservable : public Observable<FactoryObserver<T>> {
...
+    bool hasKey(K key) const;
+    std::vector<Key> getKeys() const;
+
+protected:
+    std::map<Key, M*, std::less<>> map_;
+};
+
+template <typename K>
+struct FactoryKeyType {
+    using type = std::remove_cv_t<std::remove_reference_t<K>>;
```
ResearchDaniel, line 143: Would be nice with a comment on why these remove_reference are needed and what is going on.

include/inviwo/core/util/factory.h
```diff
@@ -64,55 +65,100 @@ class FactoryObservable : public Observable<FactoryObserver<T>> {
...
  * M Models a object with a function create(K key, Args...) that can create objects of type T with
  * constructor T(Args...)
  * M would usually be a "factory object" type
  * K Models a key used to look up T
  */
-template <typename T, typename M, typename K = const std::string&, typename... Args>
-class StandardFactory : public Factory<T, K, Args...>, public FactoryObservable<M> {
+template <typename T, typename M, typename K = std::string_view, typename... Args>
+class StandardFactory : public Factory<T, K, Args...>,
+                        public FactoryRegister<M, factoryKeyType<K>, K> {
```
ResearchDaniel: Not sure it there is a benefit of doing the using declaration for FactoryKeyType<K>::type 
Less readable.

Later, we use StandardFactory like this:
`StandardFactory<T, M, string_view,...>`
then 
```
FactoryRegister::hasKey(string_view)
std::vector<string> FactoryRegister::getKeys()
```

Would it even work with `StandardFactory<T, M, string,...>`? If not, are all of these templates even needed?

ResearchDaniel: Would rename K to Key while at it :)

include/inviwo/core/util/factory.h
```diff
@@ -64,55 +65,100 @@ class FactoryObservable : public Observable<FactoryObserver<T>> {
...
+template <typename M, typename Key, typename K>
+class FactoryRegister : public FactoryObservable<M> {
+public:
+    // The factory will not assume ownership over obj, although is assumes that obj will be
+    // valid for the lifetime of the factory
+    virtual bool registerObject(M* obj);
+    virtual bool unRegisterObject(M* obj);
+
+    bool hasKey(K key) const;
+    std::vector<Key> getKeys() const;
```
ResearchDaniel, line 104: Why would K and Key would need to be different?

ResearchDaniel (dismissed):
Perhaps outside of this PR, but do we have factories that do not use string as key?
We could perhaps simplify to use string and then most of the Factory code-complexity would go away? 

also, do you have a script to do these changes? In that case it would be nice to put that in the changelog

include/inviwo/core/ports/portfactoryobject.h
petersteneteg: I have done that locally, but wanted to avoid braking code...

include/inviwo/core/util/factory.h
petersteneteg: added

include/inviwo/core/util/factory.h
petersteneteg: `factoryKeyType<K> -> typename FactoryKeyType<K>::type`
which almost twice as long...
Not sure what you mean by the other comments. Yes the templates are needed?

petersteneteg:
FactoryBase is only used with string_view so I removed that template. Might be able to remove the Key arg from the standard factory. But that would also be a breaking change. And no there is no script. FYI this is also a much simplified version of the factory code compared to what I had previously...

include/inviwo/core/util/factory.h
ResearchDaniel: Ah, I missed the typename part

ResearchDaniel: Nice commit getting rid of the hasKey type. Can we now also assume
`vector<string> getKeys()`
:)

ResearchDaniel (dismissed):
Might want to double check the overloaded hasKey. The ones with template arguments should not be needed? Just need to make sure that the one with string_view is implemented

include/inviwo/core/util/factory.h
petersteneteg: the FactoryRegister base class this has a `std::vector<Key> getKeys()`

petersteneteg: factoryKeyType<K> is still there.

ResearchDaniel (approved):

PR #1264, opened 2022-03-01
Feature/dataframe filter

martinfalk:
Basic filtering functionality for `DataFrame` and `Column`.


modules/dataframe/include/inviwo/dataframe/util/dataframeutil.h
```diff
@@ -152,6 +153,8 @@ IVW_MODULE_DATAFRAME_API std::shared_ptr<DataFrame> combineDataFrames(
  */
 template <typename Pred>
 std::vector<std::uint32_t> filteredRows(std::shared_ptr<const Column> col, Pred pred);
+IVW_MODULE_DATAFRAME_API std::vector<std::uint32_t> filteredRows(const Column& col,
```
petersteneteg: The filters here are ItemFilters? and which mean they have a column index? would it not be better to match on the whole data frame then? Or is the index just ignored? Would expect an other overload for a data frame in that case?

petersteneteg: it is very easy to get confused by here since filter is so ambiguous, could we saw selectRows instead? or are we doing the opposite?

petersteneteg: maybe we should use predicate instad of filter to? maybe `rowIndexIf` or `selectRowIf` or `selectRowIfNot`

modules/dataframe/src/util/dataframeutil.cpp
```diff
@@ -524,6 +524,64 @@ std::shared_ptr<DataFrame> combineDataFrames(std::vector<std::shared_ptr<DataFra
...
+#include <warn/push>
+#include <warn/ignore/conversion>
+std::vector<std::uint32_t> filteredRows(const Column& col, dataframefilters::Filters filters) {
+
+    if (col.getColumnType() == ColumnType::Categorical) {
+        const auto& catCol = dynamic_cast<const CategoricalColumn&>(col);
+        std::vector<std::uint32_t> rows;
+        for (auto&& [row, value] : util::enumerate<std::uint32_t>(catCol.getValues())) {
+            auto test = util::overloaded{
+                [v = value](const std::function<bool(std::string_view)>& func) { return func(v); },
```
petersteneteg, line 540: since value is a string here, this copy is quite expensive, is there any reason not to do &value or &, nothing should go out of scope here right?

modules/dataframe/src/util/dataframeutil.cpp
```diff
@@ -524,6 +524,64 @@ std::shared_ptr<DataFrame> combineDataFrames(std::vector<std::shared_ptr<DataFra
...
+                [](const auto&) { return false; }};
```
petersteneteg: Might consider adding the different signatures explicitly here and below, to make sure you get and error if the variant types change...

petersteneteg: also should this not return true, when in combination with the any_of below, I would expect a int filter in a categorical column to fail. not be ignored in the any_of?

modules/dataframe/src/util/dataframeutil.cpp
```diff
@@ -524,6 +524,64 @@ std::shared_ptr<DataFrame> combineDataFrames(std::vector<std::shared_ptr<DataFra
...
+            if (std::any_of(filters.include.begin(), filters.include.end(),
```
petersteneteg: no exclude filters here?

modules/dataframe/src/util/dataframeutil.cpp
```diff
@@ -524,6 +524,64 @@ std::shared_ptr<DataFrame> combineDataFrames(std::vector<std::shared_ptr<DataFra
...
 }
 
+#include <warn/push>
+#include <warn/ignore/conversion>
+std::vector<std::uint32_t> filteredRows(const Column& col, dataframefilters::Filters filters) {
+
+    if (col.getColumnType() == ColumnType::Categorical) {
+        const auto& catCol = dynamic_cast<const CategoricalColumn&>(col);
+        std::vector<std::uint32_t> rows;
+        for (auto&& [row, value] : util::enumerate<std::uint32_t>(catCol.getValues())) {
```
petersteneteg: humm getValues returns a new vector of strings that is very expensive :/ need a better way to iterate.

petersteneteg: should add a transformiterator (https://github.com/inviwo/inviwo/blob/master/include/inviwo/core/util/transformiterator.h) to CategoricalColumn that maps from index to string... that could return refs to the lookup table, so no copies should be needed at all...

modules/dataframe/src/util/dataframeutil.cpp
```diff
@@ -524,6 +524,64 @@ std::shared_ptr<DataFrame> combineDataFrames(std::vector<std::shared_ptr<DataFra
...
+            ->getRepresentation<BufferRAM>()
+            ->dispatch<std::vector<std::uint32_t>, dispatching::filter::Scalars>(
+                [filters](auto typedBuf) {
+                    using ValueType = util::PrecisionValueType<decltype(typedBuf)>;
+                    std::vector<std::uint32_t> rows;
+                    for (auto&& [row, value] :
+                         util::enumerate<std::uint32_t>(typedBuf->getDataContainer())) {
+                        auto test = util::overloaded{
+                            [v = value](const std::function<bool(int)>& func) {
+                                if constexpr (std::is_integral_v<ValueType>) {
```
petersteneteg, line 560: Could this test not be outside the loop? should the non matching case not return true? skip the const auto& case as above.

modules/dataframe/src/util/dataframeutil.cpp
```diff
@@ -524,6 +524,64 @@ std::shared_ptr<DataFrame> combineDataFrames(std::vector<std::shared_ptr<DataFra
...
+            }
+        }
+        return rows;
+    } else {
+        return col.getBuffer()
+            ->getRepresentation<BufferRAM>()
+            ->dispatch<std::vector<std::uint32_t>, dispatching::filter::Scalars>(
+                [filters](auto typedBuf) {
+                    using ValueType = util::PrecisionValueType<decltype(typedBuf)>;
+                    std::vector<std::uint32_t> rows;
```
petersteneteg, line 555: wonder how to best adopt this to multiple columns? Might want to pass the rows and loop over both the rows and the column rows in "lockstep", somehow. If a row is already discarded from one column we don't need to test that row in other columns?

petersteneteg (changes requested):

modules/dataframe/include/inviwo/dataframe/util/dataframeutil.h
martinfalk: In case of filtering only a single column, the column index is ignored. For an entire DataFrame, the filters will be matched to the corresponding columns using the index.
`selectedRows` would be better in this case, I think.

modules/dataframe/src/util/dataframeutil.cpp
martinfalk: value is part of a structured binding not a variable that can be captured. 
> If a lambda-expression [...] captures a structured binding (explicitly or implicitly), the program is ill-formed. 
http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/p0588r1.html

modules/dataframe/src/util/dataframeutil.cpp
martinfalk: This is in case the filter type does not match. `true` here is equivalent to keeping the row (which is what we want here).

modules/dataframe/src/util/dataframeutil.cpp
martinfalk: work in progress.

modules/dataframe/src/util/dataframeutil.cpp
martinfalk: Either that or filtering each column on its own in combination with a bitset.

modules/dataframe/include/inviwo/dataframe/datastructures/column.h
```diff
@@ -358,10 +362,19 @@ class IVW_MODULE_DATAFRAME_API CategoricalColumn : public TemplateColumn<std::ui
...
      *
-     * @return all categorical values stored in column
+     * @return all categorical values stored in the column
+     * @see values
      */
     std::vector<std::string> getValues() const;
 
+    /**
+     * returns column contents as iterator range of std::string_view
+     *
```
ResearchDaniel, line 399:
```suggestion
     * Example:
     * \code{.cpp}
     * CategoricalColumn col("example", {"first", "second"});
     * for (auto v : col.values()) {
     *     std::cout << "value: " << v;
     * }   
     * for (auto&& [row, v] : util::enumerate<std::uint32_t>(col.values())) {
     *     std::cout << util::format("Row: {}, value: {}", row, v);
     * }
     * \endcode
```

modules/dataframe/src/util/dataframeutil.cpp
martinfalk: with the transform iterator, `value` is now a `const std::string&`

modules/dataframe/include/inviwo/dataframe/datastructures/column.h
ResearchDaniel: Suggest to add an example since returning auto might not be intuitive to most people.

Could we not get the same results by adding begin/end iterators to the class? That would be more intuitive

modules/dataframe/include/inviwo/dataframe/datastructures/column.h
```diff
@@ -358,10 +362,19 @@ class IVW_MODULE_DATAFRAME_API CategoricalColumn : public TemplateColumn<std::ui
...
+     * @return iterator range over all categorical values stored in the column
```
petersteneteg, line 400: I think this comment is ambiguous might just as well refer to just the values in the lookup tabla

modules/dataframe/include/inviwo/dataframe/util/filters.h
```diff
@@ -43,12 +43,75 @@ enum class StringComp { Equal, NotEqual, Regex, RegexPartial };
...
+     * @p filter is called, that is @p std::string_view, @p int, @p float, or @p double.
+     * @see FilterFunc
+     */
+    FilterFunc filter;
+    int column;  //!< zero-based column index
+    bool filterOnHeader;
+};
+
+/// create an item filter matching strings with @p match based on @p op
+IVW_MODULE_DATAFRAME_API ItemFilter stringMatch(int column, filters::StringComp op,
```
ResearchDaniel, line 71: One option here would be to name these the same and let overloading do the work? 

```
match(int column, filters::StringComp op, std::string_view match);
match(int column, filters::NumberComp op, std::int64_t value);
match(int column, filters::NumberComp op, double value, double epsilon = std::numeric_limits<double>::epsilon());
```

modules/dataframe/include/inviwo/dataframe/properties/filterlistproperty.h
```diff
@@ -69,14 +69,11 @@ using FilterTypes = flags::flags<FilterType>;
  * Item filters own an IntProperty @column defining the target column. Identifiers begin with:
  *  - @p "stringItem":      StringProperty @match for matching strings and a
  *                          TemplateOptionProperty<filters::StringComp> @comp for the comparison
- *  - @p "intItem":         IntProperty @value, TemplateOptionProperty<filters::NumberComp> @comp
+ *  - @p "intItem":         Int64Property @value, TemplateOptionProperty<filters::NumberComp> @comp
```
petersteneteg: there is no doxygen command @value or @comp or @range. Also the quote around "intItem" is not really needed when you also add @p

modules/dataframe/include/inviwo/dataframe/util/dataframeutil.h
petersteneteg: why past tense? that makes no sense.

modules/dataframe/src/util/dataframeutil.cpp
```diff
@@ -253,12 +253,13 @@ std::vector<std::vector<std::uint32_t>> getMatchingRows(
             auto catCol2 = dynamic_cast<const CategoricalColumn*>(rightCol.get());
             IVW_ASSERT(catCol2, "right column is not categorical");
 
-            auto valuesLeft = catCol1->getValues();
-            auto valuesRight = catCol2->getValues();
+            auto valuesLeftIt = catCol1->values().begin();
```
ResearchDaniel: I think that `catCol1->begin()` would make more sense here. See previous comment about CategoricalColumn::values()

modules/dataframe/include/inviwo/dataframe/datastructures/column.h
ResearchDaniel: Also, is it possible to change the values?

ResearchDaniel (dismissed):
Nice!
Would it be possible to add begin/end iterators to CategoricalColumn instead of using values?
Then iterating would work in the same way as other containers.

modules/dataframe/src/util/dataframeutil.cpp
petersteneteg: yes, but when you assign it to `v` it will deduce to a `std::string...

petersteneteg: you need &v = value to avoid the copy.

modules/dataframe/src/util/dataframeutil.cpp
petersteneteg: Think it would make sense to take advantage of the fact that rows is sorted... probably want to use two row vectors and pingponging when doing multiple columns...

petersteneteg (changes requested):

modules/dataframe/include/inviwo/dataframe/datastructures/column.h
martinfalk: there are already begin/end iterators defined by `TemplateColumn`. The use of TransformIterator complicates things quite a bit with respect to the return type (hence auto). Having custom begin/end iterators might require a member function for the TransformIterator instead of a lambda expression.

martinfalk:
> Would it be possible to add begin/end iterators to CategoricalColumn instead of using values?
> Then iterating would work in the same way as other containers.

Not feasible due to `TemplateColumn::begin()/end()`. Unless we want to override that.

modules/dataframe/src/util/dataframeutil.cpp
martinfalk: <s>for now, `catCol1->begin()` returns a `std::uint32_t`, that is the internal representation.</s>

modules/dataframe/include/inviwo/dataframe/properties/filterlistproperty.h
martinfalk: the quote is on purpose to indicate the name of the identifier. @value, @comp, @range are supposed to be `@p "<name>"` as well.

modules/dataframe/include/inviwo/dataframe/util/dataframeutil.h
JochenJankowai: It's not past tense, it's an adjective.

modules/dataframe/include/inviwo/dataframe/datastructures/column.h
petersteneteg: you can add the explicit return types with the lambda if you want to... it is just a little bit of typing to get right...

petersteneteg: In principle we should have replaced the begin/end from the template column, but since we use inheritance to reuse code, that is not very difficult. I would have removed the template col inheritance. But that will probably break some code now...

modules/dataframe/include/inviwo/dataframe/properties/filterlistproperty.h
petersteneteg: Yes I understand that it is on purpose, I just don't think it is needed. and @c would be more appropriate. @p should refer to a function parameter.

petersteneteg:
> > Would it be possible to add begin/end iterators to CategoricalColumn instead of using values?
> > Then iterating would work in the same way as other containers.
> 
> Not feasible due to `TemplateColumn::begin()/end()`. Unless we want to override that.

it is not possible to override since the return types will not match. Need to remove the TemplateColumn inheritance, which would be good, but will probably break some other code. Might give it a try though...

modules/dataframe/include/inviwo/dataframe/properties/filterlistproperty.h
martinfalk: > `\p`:
> Equivalent to [\c](https://www.doxygen.nl/manual/commands.html#cmdc)
https://www.doxygen.nl/manual/commands.html#cmdp

modules/dataframe/include/inviwo/dataframe/util/dataframeutil.h
petersteneteg: yes, true, but a function should be a verb...

modules/dataframe/include/inviwo/dataframe/properties/filterlistproperty.h
petersteneteg: Yes, they are equivalent, but they have different semantic meaning..

modules/dataframe/include/inviwo/dataframe/datastructures/column.h
```diff
@@ -78,7 +78,8 @@ std::basic_ostream<Elem, Traits>& operator<<(std::basic_ostream<Elem, Traits>& s
 }
 
 /**
- * @brief pure interface for representing a data column, i.e. a Buffer with a name
+ * @brief Pure interface for representing a data column, i.e. a Buffer with a name
```
petersteneteg: Don't think buffer should be mentioned. that seems like an implementation detail. name should be header, and might also add unit and range.

modules/dataframe/include/inviwo/dataframe/datastructures/column.h
```diff
@@ -296,22 +281,41 @@ class IVW_MODULE_DATAFRAME_API CategoricalColumn : public TemplateColumn<std::ui
...
-     * \see TemplateColumn
+     * Set the value of row @p idx to @p str.
+     */
+    void set(size_t idx, std::string_view str);
+    /**
+     * Set the value of row @p idx by using the numerical @p id of a category.
+     * @throws RangeException if @p id does not correspond to a registered category
+     * @see addCategory
      */
-    virtual std::shared_ptr<DataPointBase> get(size_t idx, bool getStringsAsStrings) const override;
```
petersteneteg, line 311: So happy that one is gone, that parameter name was so strange

modules/dataframe/include/inviwo/dataframe/datastructures/column.h
```diff
@@ -296,22 +281,41 @@ class IVW_MODULE_DATAFRAME_API CategoricalColumn : public TemplateColumn<std::ui
...
+
+    ///@{
+    /**
+     * Return the numerical id of the category in row @idx.
+     */
+    std::uint32_t getId(size_t idx) const;
+    virtual double getAsDouble(size_t idx) const override;
+    ///@}
+
+    virtual std::string getAsString(size_t idx) const override;
```
petersteneteg: group with get?

modules/dataframe/include/inviwo/dataframe/datastructures/column.h
```diff
@@ -356,19 +360,49 @@ class IVW_MODULE_DATAFRAME_API CategoricalColumn : public TemplateColumn<std::ui
...
 
     /**
-     * \brief add a category \p cat. It will not be added if the category already exists.
+     * \brief Add a category \p cat. It will not be added if the category already exists.
      *
      * @return index of the category
      */
     std::uint32_t addCategory(std::string_view cat);
 
+    auto begin() const;
```
petersteneteg: should probably add some docs here to... does this return the string now, if not, why not? if yes, do we need values? or is values of all of the distinct values?

modules/dataframe/include/inviwo/dataframe/datastructures/column.h
petersteneteg: I wound call this `unique` or `categories` or something like that.

petersteneteg: Im still very confused to what returns what here. The names/comments needs to be much clearar.

modules/dataframe/include/inviwo/dataframe/datastructures/column.h
petersteneteg: I definitiv think this should return strings btw. the ints are just an implementation detail. even if we use that in some plotting.

modules/dataframe/include/inviwo/dataframe/datastructures/column.h
```diff
@@ -356,19 +360,49 @@ class IVW_MODULE_DATAFRAME_API CategoricalColumn : public TemplateColumn<std::ui
...
+    auto end() const;
+
+    virtual std::shared_ptr<BufferBase> getBuffer() override;
```
petersteneteg, line 428: what is this a buffer of?

modules/dataframe/include/inviwo/dataframe/datastructures/column.h
```diff
@@ -388,6 +422,14 @@ class IVW_MODULE_DATAFRAME_API CategoricalColumn : public TemplateColumn<std::ui
 
     virtual std::uint32_t addOrGetID(std::string_view str);
 
+    auto transform() const {
+        return [this](const std::uint32_t& idx) -> const std::string& { return lookUpTable_[idx]; };
+    }
+
```
petersteneteg: see https://godbolt.org/z/qsGha9frx

modules/dataframe/include/inviwo/dataframe/util/dataframeutil.h
```diff
@@ -149,19 +150,44 @@ IVW_MODULE_DATAFRAME_API std::shared_ptr<DataFrame> combineDataFrames(
  * @param col   column containing data for filtering
  * @param pred  predicate to check values from \p col
  * @return list of row indices where rows fulfill the predicate
+ * @see selectedRows(const DataFrame&, dataframefilters::Filters)
  */
 template <typename Pred>
-std::vector<std::uint32_t> filteredRows(std::shared_ptr<const Column> col, Pred pred);
+std::vector<std::uint32_t> selectedRows(std::shared_ptr<const Column> col, Pred pred);
```
petersteneteg: still think we should use the verb from

modules/dataframe/src/datastructures/column.cpp
```diff
@@ -72,33 +119,68 @@ CategoricalColumn* CategoricalColumn::clone(const std::vector<std::uint32_t>& ro
...
+
+Unit CategoricalColumn::getUnit() const { return unit_; }
+
+void CategoricalColumn::setUnit(Unit unit) { unit_ = unit; }
+
+void CategoricalColumn::setCustomRange(std::optional<dvec2> range) { range_ = range; }
+
+std::optional<dvec2> CategoricalColumn::getCustomRange() const { return range_; }
+
+dvec2 CategoricalColumn::getDataRange() const {
```
petersteneteg, line 134: can or should the dataRange for a categorical column ever be different from [0, map_.size())?

modules/dataframe/src/datastructures/column.cpp
```diff
@@ -72,33 +119,68 @@ CategoricalColumn* CategoricalColumn::clone(const std::vector<std::uint32_t>& ro
...
 }
 
+const std::string& CategoricalColumn::get(size_t idx) const { return lookUpTable_[getId(idx)]; }
+
+std::uint32_t CategoricalColumn::getId(size_t idx) const {
+    return buffer_->getRAMRepresentation()->get(idx);
+}
+
+double CategoricalColumn::getAsDouble(size_t idx) const {
+    return util::glm_convert<double>(getId(idx));
```
petersteneteg: this should be the same as just the cast...

modules/dataframe/src/util/dataframeutil.cpp
```diff
@@ -200,8 +203,8 @@ std::vector<std::vector<std::uint32_t>> getMatchingRows(std::shared_ptr<const Co
         auto catCol2 = dynamic_cast<const CategoricalColumn*>(rightCol.get());
         IVW_ASSERT(catCol2, "right column is not categorical");
 
-        auto valuesRight = catCol2->getValues();
-        for (auto&& [i, key] : util::enumerate(catCol1->getValues())) {
+        auto valuesRight = catCol2->values();
+        for (auto&& [i, key] : util::enumerate(catCol1->values())) {
```
petersteneteg: this is an O(n^2) implementation, should be possible to to in nlogn at least... are we using this on any non trivial datasets?

modules/dataframe/src/util/dataframeutil.cpp
petersteneteg: still missing &

petersteneteg (changes requested):

modules/dataframe/include/inviwo/dataframe/datastructures/column.h
martinfalk: It returns `const std::string&` now. Will replace `auto` with the proper type.

modules/dataframe/include/inviwo/dataframe/datastructures/column.h
martinfalk: begin/end iterators are now returning `const std::string&`. Changing values is not supported as this would also affect the internal representation and possibly the lookup tables.

modules/dataframe/include/inviwo/dataframe/util/filters.h
martinfalk: Don't see the reason for that. When processing the filter list property, one needs to go through each individual case anyways.

modules/dataframe/src/datastructures/column.cpp
martinfalk: Should not, but can due to being able to register categories and edit the buffer.

modules/dataframe/include/inviwo/dataframe/datastructures/column.h
ResearchDaniel: The `values()` is not needed anymore?

martinfalk: in principle not. There are however use cases since `values()` returns an iterator range of `begin()` and `end()`

modules/dataframe/src/datastructures/column.cpp
petersteneteg: is this something that we should treat as an invariant? can there be unused labels in the map? are you allowed to change the buffer directly? to we validate that It any change to the buffer still points to a valid label?

martinfalk: Yes, there might be unused labels caused by either registering a new category or when filtering the column and removing a subset. I would argue against compacting those as it might interfere with color mapping for example.
Editing the buffer gives you a bit more freedom than using `addMany()`. There is however no validation at the moment. Might add one which throws if there is a problem (or alternatively needs some heuristics to fix those by introducing dummy strings).

modules/dataframe/src/processors/dataframefilter.cpp
```diff
@@ -0,0 +1,177 @@
...
+    } else {
+        throw Exception(IVW_CONTEXT_CUSTOM("DataFrameFilter::createFilters"),
+                        "Invalid filter property '{}', missing sub-property '{}'",
+                        parent->getIdentifier(), identifier);
+    }
+}
+
+}  // namespace detail
+
+dataframefilters::Filters DataFrameFilter::createFilters() const {
```
petersteneteg, line 131: Nice! this function looks very clean

petersteneteg (dismissed):
Nice! think that turned out quite nicely

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 Here we document changes that affect the public API or changes that needs to be communicated to other developers. 
 
+## 2022-03-09 DataFrame filtering and CSV filtering
+DataFrames can now be filtered with the `DataFrame Filter` processor or when loading a CSV file with the `CSV Source` processor. The `DataFrame Filter` also supports brushing & linking filtering. Filters exist for matching values (`int64` and `double` as well as categorical values), ranges, and commented lines.
+
```
ResearchDaniel, line 5: Could add breaking changes and how to resolve them here.

ResearchDaniel (dismissed):

CHANGELOG.md
martinfalk: Will add the changes in `CategoricalColumn`.

modules/dataframe/src/util/dataframeutil.cpp
martinfalk: Potentially. Could use the row set/BitSet of the previous column to avoid unnecessary tests. Though then there is the trade-off between running a test vs. checking the row set for existence.

ResearchDaniel (dismissed):

petersteneteg (dismissed):

petersteneteg (dismissed):

petersteneteg (approved):

PR #1265, opened 2022-03-04
Feature/pybind update

petersteneteg:


resources/stylesheets/inviwo.qss
```diff
@@ -82,21 +82,11 @@ QToolBar
 QToolBar {
     spacing: 0.1875em;
     padding: 0.0625em;
-    border: none;
-    border-bottom: 0.125em solid #2b2b2d;
-    padding-right: 0.5em;
     background-color: #323235;
-}
-QToolBar::handle {
```
martinfalk, line 90: why were those removed?
image

martinfalk (approved):

CMakePresets.json
```diff
@@ -67,18 +68,16 @@
 
     {
       "name": "msvc-dev",
-      "displayName": "MSVC Development",
-      "description": "MSVC Development configuration",
+      "displayName": "MSVC 2022",
+      "description": "MSVC 2022 Development configuration",
```
ResearchDaniel, line 72:
```suggestion
      "description": "Visual Studio 2022 setup suitable for development",
```

CMakePresets.json
```diff
@@ -67,18 +68,16 @@
...
-      "inherits" : ["modules", "develop"]
+      "inherits" : ["develop"]
     },
     {
       "name": "msvc-dev-vcpkg",
-      "displayName": "MSVC Development VCPKG",
-      "description": "MSVC Development VCPKG configuration",
-      "binaryDir": "${sourceParentDir}/builds/msvc-dev-vcpkg",
+      "displayName": "MSVC 2022 Using vcpkg",
+      "description": "MSVC 2022 Development using vcpkg configuration",
```
ResearchDaniel, line 80:
```suggestion
      "description": "Visual Studio 2022 development using vcpkg package manager",
```

CMakePresets.json
ResearchDaniel: I would prefer if the "development" config could be stripped down a bit to provide the most essential components by excluding the minimal apps. We might want to separate "Core development" and "Module development", where "Module development" would be more light-weight.

If we could define what "development" means that would also be nice to have in the description. This might not be the main purpose of this PR though.

ResearchDaniel (approved):

resources/stylesheets/inviwo.qss
petersteneteg: I made them non movable, and non floatable. then that bar is not visible...

CMakePresets.json
petersteneteg: I can try, incorporate this in a future PR

PR #1267, opened 2022-03-15
Core: default values short, unsigned short, closes #1158

### LINKED ISSUE #1158: Inviwo types do not include `short`
Is there a reason for not including `short` as a basic type/default values in Inviwo?

When using an `enum class MyType : short` in combination with a `TemplateOptionProperty` this will not compile due to the missing default values for short.

martinfalk:


include/inviwo/core/properties/optionproperty.h
```diff
@@ -108,6 +108,27 @@ class OptionPropertyOption : public Serializable {
     bool operator!=(const OptionPropertyOption<T>& rhs) const;
 };
 
+/**
+ * OptionProperty with a custom type @p T.
```
petersteneteg, line 112: Might mention the requirement for non enum types too. I might also start with that all this is only needed for dynamic properties... which is quite rare...

include/inviwo/core/util/defaultvalues.h
```diff
@@ -74,6 +74,21 @@ struct InviwoDefaults<unsigned char> {
     }
 };
 template <>
+struct InviwoDefaults<short> {
+    static constexpr auto get() {
+        return InviwoDefaultData{StaticString{"Int16"}, uvec2(1, 1), short{0},
```
petersteneteg: I think "Short" and "UShort" is more inline with the other names? (If I would have done it from scratch I would have named them all "IntX")

petersteneteg (changes requested):

ResearchDaniel (dismissed):
Looks good to me. I think that Int16 is fine
Should do a using EnumOptionProperty = TemplateOptionProperty ?

ResearchDaniel (approved):

petersteneteg (approved):

PR #1273, opened 2022-03-17
Feature/imagelayoutfix

### LINKED ISSUE #1153: Image Layout + Port Inspector causes incomplete FBO
The `Image Layout` processor in combination with the port inspector causes OpenGL errors like incomplete FBO attachments on unused image inputs. This also subsequently breaks the currently shown layout.

**Cause:** `Image Layout` creates resize events with size `(0, 0)` for unused inputs. Zero-size Images are not handled well in Inviwo.

**Scenario:**
1. create three background processors
2. connect them to a `Image Layout` with a canvas
3. change the layout to horizontal Split
4. hover the connection of the third, unused image 

This will cause an incomplete FBO attachment in the inspector due to an image size of `(0, 0)` and a re-evaluation of the network breaking the layout processor.

martinfalk:
Fixes #1153

Changes proposed in this PR:
 * ImageCache returns empty Image for dimensions `(0,0)`
 * isReady of the `ImageLayout` processor only considers inports which are used in the layout

### DIFF
--- a/modules/basegl/src/processors/imageprocessing/imagelayoutgl.cpp
+++ b/modules/basegl/src/processors/imageprocessing/imagelayoutgl.cpp
@@ -88,15 +88,11 @@ ImageLayoutGL::ImageLayoutGL()
 
     addPort(multiinport_);
     multiinport_.setIsReadyUpdater([this]() {
-        // Ports with zero dimensions will not be active,
-        // so disregard them when considering ready status
+        const auto& outports = multiinport_.getConnectedOutports();
+        size_t minNum = std::min(outports.size(), viewManager_.size());
         return multiinport_.isConnected() &&
-               util::all_of(multiinport_.getConnectedOutports(), [](Outport* p) {
-                   auto ip = static_cast<ImageOutport*>(p);
-                   return (ip->hasData() && glm::any(glm::equal(ip->getDimensions(), size2_t(0))))
-                              ? true
-                              : p->isReady();
-               });
+               util::all_of(outports.begin(), outports.begin() + minNum,
+                            [](Outport* p) { return p->isReady(); });
     });
     // Ensure that viewports are up-to-date
     // before isConnectionActive is called
@@ -481,6 +477,8 @@ void ImageLayoutGL::updateViewports(ivec2 dim, bool force) {
 
     currentDim_ = dim;
     currentLayout_ = layout_.get();
+
+    isReady_.update();
 }
 
 }  // namespace inviwo
--- a/src/core/util/imagecache.cpp
+++ b/src/core/util/imagecache.cpp
@@ -45,7 +45,7 @@ void ImageCache::setMaster(std::shared_ptr<const Image> master) {
 }
 
 std::shared_ptr<const Image> ImageCache::getImage(const size2_t dimensions) const {
-    if (!master_) return std::shared_ptr<const Image>();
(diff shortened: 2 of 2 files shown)


petersteneteg (approved):

modules/basegl/src/processors/imageprocessing/imagelayoutgl.cpp
```diff
@@ -88,15 +88,11 @@ ImageLayoutGL::ImageLayoutGL()
 
     addPort(multiinport_);
     multiinport_.setIsReadyUpdater([this]() {
-        // Ports with zero dimensions will not be active,
-        // so disregard them when considering ready status
+        const auto& outports = multiinport_.getConnectedOutports();
+        size_t minNum = std::min(outports.size(), viewManager_.size());
```
ResearchDaniel, line 92: Should an empty connected port be replaced by a port with data?

That would be inline with "empty views" produced when outports.size() < viewManager_.size()?

src/core/util/imagecache.cpp
```diff
@@ -45,7 +45,7 @@ void ImageCache::setMaster(std::shared_ptr<const Image> master) {
 }
 
 std::shared_ptr<const Image> ImageCache::getImage(const size2_t dimensions) const {
-    if (!master_) return std::shared_ptr<const Image>();
+    if (!master_ || (glm::compMul(dimensions) == 0)) return std::shared_ptr<const Image>();
```
ResearchDaniel, line 48: I think that this is more clear:
```suggestion
    if (!master_ || (glm::compMul(dimensions) == 0)) return nullptr;
```
Should be possible?

ResearchDaniel (changes requested):
Some question that need to be sorted out

modules/basegl/src/processors/imageprocessing/imagelayoutgl.cpp
petersteneteg: From that perspective, why not just have it be always ready in that case?

src/core/util/imagecache.cpp
martinfalk: In principle there is a syntactic difference between a nullptr (no data) and an empty Image (dimensions (0,0)).

modules/basegl/src/processors/imageprocessing/imagelayoutgl.cpp
martinfalk: The reason for the empty port is that the ViewManager inside the `Image Layout` processor sends out resize events of `(0,0)` to all port connections that are not used in the current layout. So if by chance three ports are connected and the layout is changed to horizontal split as describe in #1153, the third connection becomes "unused" and hence an image of size (0,0) is requested.

ResearchDaniel: Could be more intuitive for the user it !ready inputs were indicated in the viewport (like the disconnected canvas), but a different PR

ResearchDaniel (approved):

modules/basegl/src/processors/imageprocessing/imagelayoutgl.cpp
martinfalk: > Could be more intuitive for the user it !ready inputs were indicated in the viewport (like the disconnected canvas), but a different PR

Agreed. There should be a more fine-grained differentiation besides "red", "green", and "yellow". Probably requires a larger reworking of how processors function.

PR #1275, opened 2022-03-18
Feature/listproperty fixes

martinfalk:
Changes proposed in this PR:
 * fixed FilterListProperty serialization
 * ListProperty unfolds when adding a property in the UI

> 

### DIFF
--- a/modules/dataframe/src/properties/filterlistproperty.cpp
+++ b/modules/dataframe/src/properties/filterlistproperty.cpp
@@ -60,25 +60,28 @@ FilterListProperty::FilterListProperty(std::string_view identifier, std::string_
         p->setVisible(enable);
         return p;
     };
+    auto createBoolComposite = [](std::string_view identifier, std::string_view displayName) {
+        auto p = std::make_unique<BoolCompositeProperty>(identifier, displayName, true);
+        p->getBoolProperty()->setSerializationMode(PropertySerializationMode::All);
+        return p;
+    };
 
     if (supportedFilters.contains(FilterType::Rows)) {
         {
-            auto emptyLines =
-                std::make_unique<BoolCompositeProperty>("emptyLines", "Empty Lines", true);
+            auto emptyLines = createBoolComposite("emptyLines", "Empty Lines");
             emptyLines->addProperty(onHeaderProp());
 
             addPrefab(std::move(emptyLines));
         }
         {
-            auto rowBegin = std::make_unique<BoolCompositeProperty>("rowBegin", "Row Begin", true);
+            auto rowBegin = createBoolComposite("rowBegin", "Row Begin");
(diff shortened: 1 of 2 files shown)


modules/dataframe/src/properties/filterlistproperty.cpp
```diff
@@ -60,25 +60,28 @@ FilterListProperty::FilterListProperty(std::string_view identifier, std::string_
         p->setVisible(enable);
         return p;
     };
+    auto createBoolComposite = [](std::string_view identifier, std::string_view displayName) {
+        auto p = std::make_unique<BoolCompositeProperty>(identifier, displayName, true);
+        p->getBoolProperty()->setSerializationMode(PropertySerializationMode::All);
```
petersteneteg: This should be p->setSerializationMode(PropertySerializationMode::All);

petersteneteg (changes requested):

modules/dataframe/src/properties/filterlistproperty.cpp
martinfalk: Why? Then the bool property (owned by the composite) would not be serialized.

petersteneteg: yes it will. you only need to set it on the property that is created dynamically. The bool child property is not dynamic ans is added in the constructor as usual...

petersteneteg: Might even be good if the ListProperty made sure all the Prefabs, has PropertySerializationMode::All

petersteneteg: I suppose this part is a little bit complicated ;)

martinfalk: > yes it will. you only need to set it on the property that is created dynamically. The bool child property is not dynamic and is added in the constructor as usual...

But the constructor of the BoolComposite initializes the bool property by default with `false`, the `FilterListProperty` however sets the bool property to `true`.

petersteneteg: That is not relevant for the serialization mode.

modules/dataframe/src/properties/filterlistproperty.cpp
```diff
@@ -60,119 +82,113 @@ FilterListProperty::FilterListProperty(std::string_view identifier, std::string_
         p->setVisible(enable);
         return p;
     };
+    auto createBoolComposite = [](std::string_view identifier, std::string_view displayName) {
```
petersteneteg, line 90: We should definitely add a few comment here...

petersteneteg (changes requested):

petersteneteg (approved):

PR #1276, opened 2022-03-18
Fix for volume combiner

PetiteViking:
Volume combiner was not setting the result to the outport.


JochenJankowai:
Already fixed in #1266

Issue #1274, opened 2022-03-18
Potential overflow in MinMaxProperty


martinfalk:
The `MinMaxProperty` suffers from numeric overflow in some instances like when limiting the separation or clamping new values.

https://github.com/inviwo/inviwo/blob/f35d449c0882995d962edfa7511c79f4dc0a1e62/include/inviwo/core/properties/minmaxproperty.h#L459-L463

PR #1278, opened 2022-03-23
Feature/scatterplot

### LINKED ISSUE #1212: Scatterplot allows selection of filtered items
The current implementation of the `BoxSelectionInteractionHandler` does not consider brushing events which filter a selection. Thus, it is possible to select invisible points and make them appear.

martinfalk:
Fixes #1212, #1274

Changes proposed in this PR:
 * refactored scatter plot using partitioning of data points in line with the parallel coordinates plotting
 * extended BitSet to convert from `const std::vector<bool>&`
 * extended Buffer and BufferGL interface (`empty()` and `unbind()`), `BufferObjectArray` optionally warns about already attached buffers (was only enabled in `IVW_DEBUG`)
 * updated requirements in Readme 

This has been tested on: Windows/Visual Studio 2022

> 


include/inviwo/core/util/zip.h
```diff
@@ -469,28 +469,30 @@ struct sequence {
...
+            val_ -= static_cast<value_type>(rhs) * inc_;
             clamp();
             return *this;
         }
 
         difference_type operator-(const iterator& rhs) const { return ((val_ - rhs.val_) / inc_); }
         iterator operator+(difference_type rhs) const {
             auto i = *this;
-            return i += rhs;
+            return i += static_cast<value_type>(rhs);
```
petersteneteg: this can't be right += is defined for difference_type not value_type
same issue below...

modules/opengl/src/buffer/bufferobjectarray.cpp
```diff
@@ -67,13 +72,17 @@ BufferObjectArray::~BufferObjectArray() { glDeleteVertexArrays(1, &id_); }
 GLuint BufferObjectArray::getId() const { return id_; }
 
 void BufferObjectArray::clear() {
+    if (id_ == 0) {
+        return;
+    }
+
+    glBindVertexArray(id_);
```
petersteneteg: humm moving this here kind of changes the preconditions... any reason for this? are this used in many places?

modules/opengl/src/buffer/bufferobjectarray.cpp
```diff
@@ -53,9 +58,9 @@ BufferObjectArray& BufferObjectArray::operator=(const BufferObjectArray& that) {
     if (this != &that) {
         if (id_ != 0) {
             // avoid creating the vertex array if is not already initialized
-            bind();
             clear();
```
petersteneteg: now this looks wrong, you unbind without a bind... If clear binds should it not also unbind then? this seems messy.

modules/opengl/include/modules/opengl/buffer/bufferobjectarray.h
```diff
@@ -41,6 +41,21 @@ class IVW_MODULE_OPENGL_API BufferObjectArray {
 public:
     using BindingType = BufferObject::BindingType;
 
+    class IVW_MODULE_OPENGL_API Warn {
+        bool value_ = true;
+        Warn() = default;
+        Warn(std::nullptr_t) : value_{false} {}
```
petersteneteg: Why not an explicit constructor of bool? why nullptr?

modules/plottinggl/include/modules/plottinggl/plotters/scatterplotgl.h
```diff
@@ -94,45 +97,56 @@ class IVW_MODULE_PLOTTINGGL_API ScatterPlotGL : public InteractionHandler {
         FloatProperty minRadius_;
         TransferFunctionProperty tf_;
         FloatVec4Property color_;
-        FloatVec4Property hoverColor_;
-        FloatVec4Property selectionColor_;
+
+        BoolCompositeProperty showHighlighted_;
```
petersteneteg: thought you made property for this?

modules/plottinggl/src/plotters/scatterplotgl.cpp
```diff
@@ -585,6 +390,114 @@ void ScatterPlotGL::invokeEvent(Event* event) {
...
+    attachAttrib(points_.color, 2);
+    attachAttrib(points_.radius, 3);
+
+    if (picking_.isEnabled() && !points_.pickIds.empty()) {
+        attachAttrib(&points_.pickIds, 4, BufferObject::BindingType::Native);
+    } else {
+        points_.boa.detachBufferObject(4);
+    }
+    // need to rebind buffer object array to reattach attributes
+    points_.boa.bind();
```
petersteneteg: this seems like a code smell...

petersteneteg (changes requested):

modules/opengl/src/buffer/bufferobjectarray.cpp
martinfalk: It should ensure that the correct buffer object array is bound before disabling any vertex attribute arrays of other BOAs.

modules/plottinggl/include/modules/plottinggl/plotters/scatterplotgl.h
martinfalk: not yet, probably reasonable to do as part of this PR.

modules/opengl/include/modules/opengl/buffer/bufferobjectarray.h
martinfalk: The code is based on `Build` of `Shader`. Will fix.

modules/opengl/src/buffer/bufferobjectarray.cpp
martinfalk: Probably `clear()` should not call `glDisableVertexAttribArray` in the first place. Then there is no need for bind/unbind at all. However, one would then need to call `bind()` afterward to disable the vertex attribs that are still enabled, which is also somewhat strange. And in case the BOA is bound before calling clear(), it stays bound along with all attributes despite having called clear(). 
It all comes down to mixing the state machine of OpenGL (bound BOA) and our internal states.

modules/plottinggl/src/plotters/scatterplotgl.cpp
martinfalk: See comment above. The first bind should not be necessary.

modules/opengl/src/buffer/bufferobjectarray.cpp
petersteneteg: yes, but previously one had to call bind before calling clear. now one should not do that?

martinfalk:
Superseded by #1280

PR #1280, opened 2022-04-01
Feature/scatterplot2

### LINKED ISSUE #1212: Scatterplot allows selection of filtered items

martinfalk:
Fixes #1212, <s>#1274</s>

Changes proposed in this PR:
 * refactored scatter plot using partitioning of data points in line with the parallel coordinates plotting
 * extended BitSet to convert from `const std::vector<bool>&`
 * <s>extended Buffer and BufferGL interface (`empty()` and `unbind()`), `BufferObjectArray` optionally warns about already attached buffers (was only enabled in `IVW_DEBUG`)</s>
 * <s>updated requirements in Readme</s>

This has been tested on: Windows/Visual Studio 2022

> 


include/inviwo/core/properties/selectioncolorproperty.h
```diff
@@ -0,0 +1,65 @@
...
+ * @brief composite property holding parameters for highlighted and selected data points.
+ */
+class IVW_CORE_API SelectionColorProperty : public BoolCompositeProperty {
+public:
+    virtual std::string getClassIdentifier() const override;
+    static const std::string classIdentifier;
+
+    SelectionColorProperty(
+        std::string_view identifier, std::string_view displayName, bool checked = false,
+        vec3 color = vec3(1.0f, 0.906f, 0.612f), float alpha = 0.75f,
```
ResearchDaniel, line 48: What is this color? :) 
What are the defaults based on?

include/inviwo/core/properties/selectioncolorproperty.h
```diff
@@ -0,0 +1,65 @@
...
+
+    virtual SelectionColorProperty* clone() const override;
+
+    vec4 getColor() const;
+    float getMixIntensity() const;
+
+private:
+    FloatVec3Property color_;
+    FloatProperty alpha_;
+    FloatProperty intensity_;
```
ResearchDaniel, line 65: mix or blend?

modules/dataframe/src/processors/dataframefilter.cpp
```diff
@@ -76,6 +76,13 @@ DataFrameFilter::DataFrameFilter()
 
     includeFilters_.setCollapsed(true);
     excludeFilters_.setCollapsed(true);
+
+    // make this processor a sink if brushing and linking filtering is selected
```
ResearchDaniel: Also, maybe explain why instead?
Make sink because ....
```suggestion
    // make this processor a sink if brushing and linking filtering is enabled
```

modules/opengl/src/buffer/bufferobjectarray.cpp
```diff
@@ -44,76 +69,113 @@ size_t BufferObjectArray::maxSize() const {
...
+
+    glGenVertexArrays(1, &id_);
+}
 
 BufferObjectArray::BufferObjectArray(const BufferObjectArray& rhs)
-    : attachedBuffers_(rhs.attachedBuffers_) {}
+    : id_(0u), attachedBuffers_(rhs.attachedBuffers_) {
+
+    creationContext_ = RenderContext::getPtr()->activeContext();
+    IVW_ASSERT(creationContext_, "An OpenGL Context has to be active");
```
ResearchDaniel: Is this just important during debugging?
Do we have a way to deal with this at (release) runtime?

modules/opengl/src/buffer/bufferobjectarray.cpp
```diff
@@ -44,76 +69,113 @@ size_t BufferObjectArray::maxSize() const {
...
 void BufferObjectArray::clear() {
+    LGL_ERROR_CLASS;
     for (GLuint i = 0; i < static_cast<GLuint>(attachedBuffers_.size()); ++i) {
         if (attachedBuffers_[i].second) {
-            glDisableVertexAttribArray(i);
+            glDisableVertexArrayAttrib(id_, i);
             attachedBuffers_[i].second = nullptr;
         }
     }
-    reattach_ = true;
```
ResearchDaniel: This turned out to be a hack?

modules/plottinggl/include/modules/plottinggl/plotters/scatterplotgl.h
```diff
@@ -151,7 +154,29 @@ class IVW_MODULE_PLOTTINGGL_API ScatterPlotGL : public InteractionHandler {
 
     void setSortingOrder(SortingOrder order);
 
+    /**
+     * Set the brushed indices for filtered, selected, and highlighted data points.
+     * Note that the indices are supposed to be zero-based that is idx in [0, numDataPoints).
```
ResearchDaniel:
```suggestion
     * @pre indices must be zero-based, i.e., idx in [0, numDataPoints).
```

modules/plottinggl/include/modules/plottinggl/plotters/scatterplotgl.h
```diff
@@ -151,7 +154,29 @@ class IVW_MODULE_PLOTTINGGL_API ScatterPlotGL : public InteractionHandler {
...
+     * @see setFilteredIndices setSelectedIndices setHighlightedIndices
+     */
+    void setIndices(const BitSet& filtered, const BitSet& selected, const BitSet& highlighted);
+    /**
+     * Set the brushed indices for filtered data points only.
+     * Note that the indices are supposed to be zero-based that is idx in [0, numDataPoints).
```
ResearchDaniel:
```suggestion
     * Set indices that should not be shown.
     * @pre indices must be zero-based, i.e., idx in [0, numDataPoints).
```

modules/plottinggl/include/modules/plottinggl/plotters/scatterplotgl.h
```diff
@@ -151,7 +154,29 @@ class IVW_MODULE_PLOTTINGGL_API ScatterPlotGL : public InteractionHandler {
...
+     * @see setIndices setSelectedIndices setHighlightedIndices
+     */
+    void setFilteredIndices(const BitSet& indices);
+    /**
+     * Set the brushed indices for selected data points only.
```
ResearchDaniel:
```suggestion
     * Set indices of data points to be selected.
```

modules/plottinggl/include/modules/plottinggl/plotters/scatterplotgl.h
```diff
@@ -151,7 +154,29 @@ class IVW_MODULE_PLOTTINGGL_API ScatterPlotGL : public InteractionHandler {
...
+     * Note that the indices are supposed to be zero-based that is idx in [0, numDataPoints).
```
ResearchDaniel:
```suggestion
     * @pre indices must be zero-based, i.e., idx in [0, numDataPoints).
```

modules/plottinggl/include/modules/plottinggl/plotters/scatterplotgl.h
```diff
@@ -151,7 +154,29 @@ class IVW_MODULE_PLOTTINGGL_API ScatterPlotGL : public InteractionHandler {
...
+     * @see setIndices setFilteredIndices setHighlightedIndices
+     */
     void setSelectedIndices(const BitSet& indices);
+    /**
+     * Set the brushed indices for highlighted data points only.
```
ResearchDaniel:
```suggestion
     * Set indices of points to highlight.
```

modules/plottinggl/include/modules/plottinggl/plotters/scatterplotgl.h
```diff
@@ -151,7 +154,29 @@ class IVW_MODULE_PLOTTINGGL_API ScatterPlotGL : public InteractionHandler {
...
+     * Note that the indices are supposed to be zero-based that is idx in [0, numDataPoints).
```
ResearchDaniel:
```suggestion
     * @pre indices must be zero-based, i.e., idx in [0, numDataPoints).
```

modules/plottinggl/src/plotters/scatterplotgl.cpp
```diff
@@ -64,24 +66,24 @@ ScatterPlotGL::Properties::Properties(std::string_view identifier, std::string_v
     , minRadius_("minRadius", "Min Radius", 0.1f, 0, 10, 0.01f)
     , tf_("transferFunction", "Transfer Function",
           TransferFunction({{0.0, vec4(1.0f)}, {1.0, vec4(1.0f)}}))
-    , color_("color", "Color", vec4(1, 0, 0, 1), vec4(0), vec4(1), vec4(0.1f),
-             InvalidationLevel::InvalidOutput, PropertySemantics::Color)
-    , hoverColor_("hoverColor", "Hover Color", vec4(1.0f, 0.906f, 0.612f, 1))
-    , selectionColor_("selectionColor", "Selection Color", vec4(1.0f, 0.769f, 0.247f, 1))
+    , color_("defaultColor", "Color", util::ordinalColor(1.0f, 0.0f, 0.0f, 1.0f))
```
ResearchDaniel, line 69: I know that nothing really has changed, but IMHO red is not an optimal default color for selection.

modules/plottinggl/src/plotters/scatterplotgl.cpp
```diff
@@ -161,299 +161,66 @@ ScatterPlotGL::ScatterPlotGL(Processor* processor)
...
         [this](const std::vector<bool>& selected, bool append) {
-            ensureSelectAndFilterSizes();
-            for (auto&& [ind, elem] : util::enumerate(selected)) {
-                if (append && elem)
-                    selected_[ind] = elem;
-                else {
-                    selected_[ind] = elem;
-                }
+            BitSet b(selected);
+            // suppress selection of filtered indices
```
ResearchDaniel:
```suggestion
            // Should not be possible to select filtered items
```

modules/plottinggl/src/plotters/scatterplotgl.cpp
```diff
@@ -474,86 +241,98 @@ void ScatterPlotGL::setYAxis(const Column* col) {
...
-        xAxis_ = col->getBuffer();
+        points_.xCoord = col->getBuffer();
         minmaxX_ = vec2(col->getRange());
         properties_.xAxis_.setRange(minmaxX_);
     } else {
-        xAxis_ = nullptr;
+        points_.xCoord = nullptr;
     }
-    boxSelectionHandler_.setXAxisData(xAxis_);
+    partitionDirty_ = true;
```
ResearchDaniel, line 252: partition?

modules/plottinggl/src/plotters/scatterplotgl.cpp
```diff
@@ -604,52 +491,96 @@ void ScatterPlotGL::objectPicked(PickingEvent* p) {
...
-            highlightDirty_ = true;
+            highlightChangedCallback_.invoke(BitSet(id));
         } else if (p->getHoverState() == PickingHoverState::Exit) {
-            highlighted_.clear();
-            highlightChangedCallback_.invoke(highlighted_);
+            highlightChangedCallback_.invoke(BitSet());
         }
     }
 
+    const double movedDistance =
```
ResearchDaniel: Could we refactor this "if moved" into some helper function at some point?

modules/plottinggl/src/plotters/scatterplotgl.cpp
```diff
@@ -604,52 +491,96 @@ void ScatterPlotGL::objectPicked(PickingEvent* p) {
...
+    });
+
+    points_.offsets[0] = 0;
+    points_.offsets[1] = static_cast<std::uint32_t>(std::distance(indices.begin(), lastFilteredIt));
+    points_.offsets[1] = static_cast<std::uint32_t>(std::distance(indices.begin(), lastFilteredIt));
+    points_.offsets[2] = static_cast<std::uint32_t>(std::distance(indices.begin(), lastRegularIt));
+    points_.offsets[3] = static_cast<std::uint32_t>(std::distance(indices.begin(), lastSelectedIt));
+    points_.offsets[4] = numDataPoints;
+
+    // sort each partition and prefer sorting buffer over radius for depth sorting
```
ResearchDaniel:
```suggestion
    // Sort each partition's draw order. Prefer sorting buffer over radius for depth sorting
```

src/core/properties/selectioncolorproperty.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+SelectionColorProperty::SelectionColorProperty(std::string_view identifier,
+                                               std::string_view displayName, bool checked,
+                                               vec3 color, float alpha,
+                                               InvalidationLevel invalidationLevel,
+                                               PropertySemantics semantics)
+    : BoolCompositeProperty(identifier, displayName, checked, invalidationLevel, semantics)
+    , color_("color", "Color", util::ordinalColor(color))
+    , alpha_("alpha", "Alpha", alpha)
+    , intensity_("intensity", "Mixing", 0.7f, 0.01f, 1.0f, 0.001f) {
```
ResearchDaniel, line 45: Why not name  the variable and identifier mix_?

include/inviwo/core/properties/selectioncolorproperty.h
```diff
@@ -0,0 +1,65 @@
...
+        vec3 color = vec3(1.0f, 0.906f, 0.612f), float alpha = 0.75f,
+        InvalidationLevel invalidationLevel = InvalidationLevel::InvalidResources,
+        PropertySemantics semantics = PropertySemantics::Default);
+    SelectionColorProperty(const SelectionColorProperty& rhs);
+    virtual ~SelectionColorProperty() = default;
+
+    virtual SelectionColorProperty* clone() const override;
+
+    vec4 getColor() const;
+    float getMixIntensity() const;
```
ResearchDaniel, line 61:
```suggestion
    * Value in [0 1], where 1 means use this color and 0 means use the item's original color
    float getMix() const;
```

include/inviwo/core/properties/selectioncolorproperty.h
ResearchDaniel, line 65:
```suggestion
    FloatProperty mix_;
```

ResearchDaniel (changes requested):

modules/opengl/src/buffer/bufferobjectarray.cpp
petersteneteg: we can't deal with this at runtime. It will generally just fail. With a black canvas or what ever. Not much one can do at that stage.

modules/opengl/src/buffer/bufferobjectarray.cpp
petersteneteg: Think it was a workaround for context issues. We are trying to make sure we have the right context instead...

include/inviwo/core/util/zip.h
```diff
@@ -469,12 +469,12 @@ struct sequence {
         }
 
         iterator& operator+=(difference_type rhs) {
-            val_ += rhs * inc_;
+            val_ += static_cast<value_type>(rhs) * inc_;
```
petersteneteg: hmmm what happens if value_type is uint32_t and one does `it += -1`

include/inviwo/core/util/zip.h
```diff
@@ -469,12 +469,12 @@ struct sequence {
...
             clamp();
             return *this;
         }
         iterator& operator-=(difference_type rhs) {
-            val_ -= rhs * inc_;
+            val_ -= static_cast<value_type>(rhs) * inc_;
```
petersteneteg: same as above

modules/opengl/src/buffer/bufferobjectarray.cpp
```diff
@@ -44,76 +69,113 @@ size_t BufferObjectArray::maxSize() const {
...
+        if (auto bufferObject = attachedBuffers_[location].second) {
+            glEnableVertexAttribArray(location);
+            bufferObject->bindAndSetAttribPointer(location, attachedBuffers_[location].first);
+        }
+    }
+    unbind();
+}
 
 BufferObjectArray& BufferObjectArray::operator=(const BufferObjectArray& that) {
+    LGL_ERROR_CLASS;
```
petersteneteg: remove?

modules/opengl/src/buffer/bufferobjectarray.cpp
```diff
@@ -44,76 +69,113 @@ size_t BufferObjectArray::maxSize() const {
...
         // print warning if a different buffer is already attached to this location
-        if (attachedBuffers_[location].second && bufferObject &&
+        if (warn && attachedBuffers_[location].second && bufferObject &&
             (attachedBuffers_[location].second != bufferObject)) {
-            LogWarn("BufferObjectArray (" << id_ << "): location " << location
-                                          << " is already bound to different buffer object (id "
-                                          << attachedBuffers_[location].second->getId()
-                                          << "). Replacing with new buffer object (id "
-                                          << bufferObject->getId() << ").");
+            LogWarn(fmt::format(
```
petersteneteg: Is there a better way to handle this warning? What is the use case for having it and not? 
Seems like either it is an error, or you don't really case and then there is not need for logging? 
Would it be better to have to functions attachBufferObject and something like attachBufferObjectForce or what ever. I think we should really try and avoid this kind of warnings in the library code, since it does not really compose.

modules/opengl/src/buffer/bufferobjectarray.cpp
```diff
@@ -44,76 +69,113 @@ size_t BufferObjectArray::maxSize() const {
...
+        glEnableVertexAttribArray(location);
+        bufferObject->bindAndSetAttribPointer(location, bindingType);
     } else {
         LogError("Error: VertexAttribArray location exceeds maximum allowed range");
     }
+    LGL_ERROR_CLASS;
+}
+
+void BufferObjectArray::detachBufferObject(GLuint location) {
+    LGL_ERROR_CLASS;
```
petersteneteg: Remove?

modules/plottinggl/glsl/scatterplot.vert
```diff
@@ -50,6 +50,14 @@ uniform vec2 minmaxC;
 uniform vec2 minmaxR;
 uniform vec4 default_color;
 
+struct SecondaryColor {
```
petersteneteg: Should this be in a header?

modules/opengl/src/buffer/bufferobjectarray.cpp
ResearchDaniel: Might be reasonable to let the user know that a restart is required for it to work properly? In a similar way as the fatal errors

modules/plottinggl/src/plotters/scatterplotgl.cpp
```diff
@@ -585,6 +364,114 @@ void ScatterPlotGL::invokeEvent(Event* event) {
...
+    attachAttrib(points_.xCoord, 0);
+    attachAttrib(points_.yCoord, 1);
+    attachAttrib(points_.color, 2);
+    attachAttrib(points_.radius, 3);
+
+    if (picking_.isEnabled() && !points_.pickIds.empty()) {
+        attachAttrib(&points_.pickIds, 4, BufferObject::BindingType::Native);
+    } else {
+        points_.boa.detachBufferObject(4);
+    }
```
petersteneteg:
```suggestion
    const bool picking = picking_.isEnabled() && !points_.pickIds.empty();
    attachAttrib(picking ? &points_.pickIds : nullptr, 4, BufferObject::BindingType::Native);
```

modules/plottinggl/src/plotters/scatterplotgl.cpp
```diff
@@ -604,52 +491,96 @@ void ScatterPlotGL::objectPicked(PickingEvent* p) {
...
-    if (xAxis_->getSize() != selected_.size() || xAxis_->getSize() != filtered_.size()) {
-        selected_.resize(xAxis_->getSize(), false);
-        filtered_.resize(xAxis_->getSize(), false);
-        selectedIndicesGLDirty_ = true;
-        filteringDirty_ = true;
+void ScatterPlotGL::partitionData() {
+    const auto numDataPoints = static_cast<std::uint32_t>(points_.xCoord->getSize());
+    auto seq = util::make_sequence<std::uint32_t>(0, numDataPoints);
+
+    auto& indices = points_.indices.getEditableRAMRepresentation()->getDataContainer();
```
petersteneteg, line 533: Why do you always reset the list at the start? seems like this will have to do a lot more work than needed.

petersteneteg: in most cases on only has to move a few point around, like the hovered one etc.

modules/plottinggl/src/processors/scatterplotmatrixprocessor.cpp
```diff
@@ -125,13 +121,7 @@ ScatterPlotMatrixProcessor::ScatterPlotMatrixProcessor()
...
     });
 
-    correlectionTF_.get().clear();
-    correlectionTF_.get().add(0.0, vec4(1, 0, 0, 1));
-    correlectionTF_.get().add(0.5, vec4(1, 1, 1, 1));
-    correlectionTF_.get().add(1.0, vec4(0, 0, 1, 1));
-    correlectionTF_.setCurrentStateAsDefault();
-
-    for (auto font : font::getAvailableFonts()) {
+    for (auto& font : font::getAvailableFonts()) {
```
petersteneteg: the does font property not handle this? should be in the constructor preferably

modules/plottinggl/src/processors/scatterplotmatrixprocessor.cpp
```diff
@@ -197,52 +182,61 @@ void ScatterPlotMatrixProcessor::process() {
...
-        auto dfSize = dataframe->getNumberOfRows();
 
-        auto iCol = dataframe->getIndexColumn();
-        auto& indexCol = iCol->getTypedBuffer()->getRAMRepresentation()->getDataContainer();
-
-        auto brushedIndicies = brushing_.getFilteredIndices();
-        indicies = std::make_unique<IndexBuffer>();
-        auto& vec = indicies->getEditableRAMRepresentation()->getDataContainer();
-        vec.reserve(dfSize - brushedIndicies.size());
+        auto indexToRowMap = [&]() {
```
petersteneteg: Is this created in every frame? should probably be kept around then

petersteneteg: Might be better to keep the bitset around to, to avoid creating it every time.

src/core/datastructures/bitset.cpp
```diff
@@ -143,6 +145,12 @@ bool BitSet::set(const BitSet& b) {
 
 void BitSet::add(util::span<const uint32_t> span) { addMany(span.size(), span.data()); }
 
+void BitSet::add(const std::vector<bool>& v) {
+    for (std::uint32_t i = 0; i < static_cast<std::uint32_t>(v.size()); ++i) {
```
petersteneteg: Might throw here if v > uint32::max

Did you try the add many version? think that might be quite good here...

petersteneteg (changes requested):

include/inviwo/core/properties/selectioncolorproperty.h
martinfalk: Defaults are based on either one of the ScatterPlot colors or the PCP ;-)

include/inviwo/core/properties/selectioncolorproperty.h
martinfalk: Inspired by PCP shaders, intensity referring to the fact how visible this color will be.

modules/opengl/src/buffer/bufferobjectarray.cpp
martinfalk: In the constructor of BufferObjectArray we definitely need an OpenGL context for creating the VAO (glGen...). A restart in this context will not even fix this. When trying to access/delete a VAO in a different GL context this is _just_ an OpenGL error and a restart will have no effect at all.

These assertions are the cause for the integration tests to fail with glfw as the GLFW window destruction messes with the context and destroys a CanvasGL, which in turn holds a mesh with a VAO/BufferObjectArray that was created in the default context.

The FBOs rely on the same mechanism.

modules/plottinggl/src/plotters/scatterplotgl.cpp
martinfalk: Same idea as in the PCP. Consider the indices in one container. This container is partitioned/divided into multiple blocks (filtered, normal, selected, and highlighted).

modules/plottinggl/src/plotters/scatterplotgl.cpp
martinfalk: Would be sensible, maybe even part of the PickingEvent?

modules/opengl/src/buffer/bufferobjectarray.cpp
martinfalk: attachBufferObject and attachBufferObjectForce is better and more explicit.

modules/dataframe/src/processors/dataframefilter.cpp
martinfalk: This is independent of whether Brushing and Linking is enabled, it refers to the dropdown option of "filtering". The reason being that in this case you can use the processor to filter indices without ever using its output.

modules/plottinggl/src/plotters/scatterplotgl.cpp
martinfalk: It is not the selection color, it refers to the default color for data points.
See https://github.com/inviwo/inviwo/blob/1ca7e90e221208f40af249687e8b5c33d2fc06c4/modules/plottinggl/src/plotters/scatterplotgl.cpp#L67-L68

src/core/properties/selectioncolorproperty.cpp
martinfalk: I'd rather prefer the display name to be `"Intensity"`

src/core/datastructures/bitset.cpp
martinfalk: Wouldn't addMany require a temporary span/vector of indices?

include/inviwo/core/properties/selectioncolorproperty.h
martinfalk: https://github.com/inviwo/inviwo/blob/1ca7e90e221208f40af249687e8b5c33d2fc06c4/modules/plottinggl/src/plotters/scatterplotgl.cpp#L69

modules/dataframe/src/processors/dataframefilter.cpp
martinfalk: Fixed comment

include/inviwo/core/util/zip.h
martinfalk: This is a general problem with unsigned types. Using a negative increment also breaks the sequence for uint32_t.

modules/opengl/src/buffer/bufferobjectarray.cpp
martinfalk: Will be removed after successfully passing integration tests.

include/inviwo/core/util/zip.h
petersteneteg: But it would have worked before you added the cast... I might give a warning, but the result would be right, now the result is just wrong and with no warning...

martinfalk: That is true. Will remove the cast. Any suggestions whether we should handle negative increments for unsigned types and how?

petersteneteg: we don't need to handle negative values for inc_ that will give a warning in the constructor of sequence. Only sequence += difference_type(-1); 
Have not figured out the proper set of cast to do that correctly yet...

modules/plottinggl/src/processors/scatterplotmatrixprocessor.cpp
martinfalk: `FontProperty` is a composite that handles this internally. Here only a plain `OptionPropertyString` is used. Could consider making a specialization for font faces.

martinfalk:
Changes to `BufferObjectArray` (VAO) and related classes addressed in PR #1288.

include/inviwo/core/properties/selectioncolorproperty.h
```diff
@@ -0,0 +1,69 @@
...
+    virtual SelectionColorProperty* clone() const override;
+
+    vec4 getColor() const;
+    /**
+     * Return the blending factor for mixing the selection color with the item's original color.
+     * @return blending factor in [0,1]
+     */
+    float getMixIntensity() const;
+
+private:
```
petersteneteg: why make them private?

modules/plottinggl/include/modules/plottinggl/processors/scatterplotprocessor.h
```diff
@@ -86,6 +86,8 @@ class IVW_MODULE_PLOTTINGGL_API ScatterPlotProcessor : public Processor {
     ScatterPlotGL::HighlightCallbackHandle highlightChangedCallBack_;
     ScatterPlotGL::SelectionCallbackHandle selectionChangedCallBack_;
     ScatterPlotGL::SelectionCallbackHandle filteringChangedCallBack_;
+
+    std::unordered_map<uint32_t, uint32_t> indexToRowMap_;
```
petersteneteg, line 90: This is the one we eventually wanted in the index column?

modules/plottinggl/src/plotters/scatterplotgl.cpp
```diff
@@ -604,52 +490,95 @@ void ScatterPlotGL::objectPicked(PickingEvent* p) {
...
+                            return data[a] > data[b];
+                        };
+                    }
+                }();
+
+                for (size_t i = 0; i < points_.offsets.size() - 1; ++i) {
+                    auto begin = points_.offsets[i];
+                    auto end = points_.offsets[i + 1];
+                    std::sort(indices.begin() + begin, indices.begin() + end, comp);
+                }
```
petersteneteg:
```suggestion
               auto ascending =  [&data](std::uint32_t a, std::uint32_t b) { return data[a] < data[b]; };
               auto descending = [&data](std::uint32_t a, std::uint32_t b) { return data[a] > data[b]; };

               for (size_t i = 0; i < points_.offsets.size() - 1; ++i) {
                    const auto begin = points_.offsets[i];
                    const auto end = points_.offsets[i + 1];
                    if(sortOrder_ == SortingOrder::Ascending) {
                         std::sort(indices.begin() + begin, indices.begin() + end, ascending);
                    } else {
                         std::sort(indices.begin() + begin, indices.begin() + end, descending);
                    }
                }
```

modules/plottinggl/src/processors/scatterplotmatrixprocessor.cpp
```diff
@@ -183,11 +166,24 @@ ScatterPlotMatrixProcessor::ScatterPlotMatrixProcessor()
...
     if (plots_.empty()) {
         createScatterPlots();
         for (auto& p : plots_) {
             p->properties_.set(&scatterPlotproperties_);
         }
+
+        indexToRowMap_ = [&]() {
+            auto iCol = dataFrame_.getData()->getIndexColumn();
+            auto& indexCol = iCol->getTypedBuffer()->getRAMRepresentation()->getDataContainer();
+            std::unordered_map<uint32_t, uint32_t> indexToRow;
```
petersteneteg, line 169: might reserve some indexCol.size in indexToRow

modules/plottinggl/src/processors/scatterplotmatrixprocessor.cpp
```diff
@@ -197,52 +193,48 @@ void ScatterPlotMatrixProcessor::process() {
...
-        auto& vec = indicies->getEditableRAMRepresentation()->getDataContainer();
-        vec.reserve(dfSize - brushedIndicies.size());
-
-        auto seq = util::sequence<uint32_t>(0, static_cast<uint32_t>(dfSize), 1);
-        std::copy_if(seq.begin(), seq.end(), std::back_inserter(vec),
-                     [&](const auto& id) { return !brushing_.isFiltered(indexCol[id]); });
+    if (brushing_.isConnected() && (brushing_.isFilteringModified() || initialSetup)) {
+        auto transformIdsToRows = [&](const BitSet& b) {
+            BitSet rows;
+            for (const auto& id : b) {
```
petersteneteg: probably better to use auto id here to avoid the indirection

petersteneteg: wonder if the compiler is allowed to remove that it self?

petersteneteg (dismissed):

modules/plottinggl/include/modules/plottinggl/processors/scatterplotprocessor.h
martinfalk: Exactly.

modules/plottinggl/src/processors/scatterplotmatrixprocessor.cpp
martinfalk: Added new `FontFaceOptionProperty` as a custom type for the font face.

include/inviwo/core/interaction/events/pickingevent.h
```diff
@@ -166,6 +166,15 @@ class IVW_CORE_API PickingEvent : public Event {
...
     dvec3 getWorldSpaceDeltaAtPressDepth(const Camera& camera) const;
 
+    /**
+     * Returns true if the distance between the pressed position and current position in screen
+     * coordinates is larger than @p deltaDistance or the current picking ID is different to the
+     * pressed one.
+     *
+     * @see getCurrentGlobalPickingId, getPressedGlobalPickingId, getDeltaPressedPosition
+     */
+    bool getMovedSincePressed(double deltaDistance = 3.0) const;
```
ResearchDaniel, line 176: I think that this distance might be different for high res displays and mouse/touch.
@eriksunden was it necessary to tweak this distance?

modules/plottinggl/src/processors/scatterplotmatrixprocessor.cpp
```diff
@@ -64,13 +65,16 @@ ScatterPlotMatrixProcessor::ScatterPlotMatrixProcessor()
...
+    , fontFace_("fontFace", "Font Face", font::FontType::Caption)
+    , fontSize_("fontSize", "Font size", 20, 0, 144, 1, InvalidationLevel::Valid,
+                PropertySemantics("Fontsize"))
+    , fontFaceStats_("fontFaceStats", "Font Face (stats)", font::FontType::Label)
+    , statsFontSize_("statsFontSize", "Font size (stats)", 14, 0, 144, 1, InvalidationLevel::Valid,
+                     PropertySemantics("Fontsize"))
     , showCorrelationValues_("showStatistics", "Show correlation values", true)
     , parameters_("parameters", "Parameters")
-    , correlectionTF_("correlectionTF", "Correlation TF")
+    , correlectionTF_("correlectionTF", "Correlation TF",
```
ResearchDaniel: Typos (correlection -> correlation)

modules/plottinggl/src/processors/scatterplotmatrixprocessor.cpp
```diff
@@ -64,13 +65,16 @@ ScatterPlotMatrixProcessor::ScatterPlotMatrixProcessor()
...
+                      {{{0.0, vec4(1, 0, 0, 1)}, {0.5, vec4(1, 1, 1, 1)}, {1.0, vec4(0, 0, 1, 1)}}})
```
ResearchDaniel: Feature request: exchange this to our colorbrewer diverging RdBu TF

modules/plottinggl/src/processors/scatterplotprocessor.cpp
```diff
@@ -95,29 +95,29 @@ ScatterPlotProcessor::ScatterPlotProcessor()
...
-                brushingPort_.select(selectedIndices);
+                brushingPort_.select(indices);
             } else {
                 invalidate(InvalidationLevel::InvalidOutput);
             }
         });
     filteringChangedCallBack_ =
-        scatterPlot_.addFilteringChangedCallback([this](const std::vector<bool>& filtered) {
+        scatterPlot_.addFilteringChangedCallback([this](const BitSet& filtered) {
             if (brushingPort_.isConnected()) {
```
ResearchDaniel: This check is not necessary anymore?

ResearchDaniel (dismissed):

petersteneteg (dismissed):

modules/plottinggl/src/processors/scatterplotprocessor.cpp
martinfalk: Good catch. This would have prevented selection without a brushing port connection.

petersteneteg (dismissed):

petersteneteg (approved):

PR #1285, opened 2022-04-07
Qt: clang fix, #1284

martinfalk:
Fixes #1284


petersteneteg (approved):
Think this is a clang compiler bug...

PR #1288, opened 2022-04-11
Feature/opengl

martinfalk:
* Fixes GLFW context handling during destruction. 
 * Refactoring of BufferObjectArray (VAO) including rendering context checks


petersteneteg (approved):
NIce! looks like it got quite a bit cleaner!

PR #1290, opened 2022-04-13
Feature/misc fixes

martinfalk:
Fixes:
* potential deadlock in `forEachVoxel`
* serialization of VolumeInformationProperty
* `Background` processor uses `util::ordinalColor` for colors

Changes proposed in this PR:
 * updated javascript libraries for regression report HTML page
 * added detailed commit information to regression report
 * links to previous regression reports now show timestamp
 * added `ThreadPool::isPoolThread()` to check whether current thread is a pool thread
 * hide up/down arrows on spin boxes where increment is zero

image


modules/base/src/processors/volumesource.cpp
```diff
@@ -144,7 +144,7 @@ void VolumeSource::load(bool deserialize) {
 
         basis_.updateForNewEntity(*(*volumes_)[0], deserialize);
         information_.updateForNewVolume(
-            *(*volumes_)[0], deserialize ? util::OverwriteState::No : util::OverwriteState::Yes);
+            *(*volumes_)[0], deserialize ? util::OverwriteState::Yes : util::OverwriteState::No);
```
petersteneteg, line 147: Do we have this error in other sources?

modules/base/src/properties/volumeinformationproperty.cpp
```diff
@@ -89,7 +89,12 @@ VolumeInformationProperty::VolumeInformationProperty(std::string_view identifier
         },
         props(*this));
 
-    util::for_each_in_tuple([&](auto& p) { addProperty(p); }, meta(*this));
+    util::for_each_in_tuple(
+        [&](auto& p) {
+            p.setCurrentStateAsDefault();
```
petersteneteg: Why?

src/core/util/threadpool.cpp
```diff
@@ -75,6 +75,16 @@ size_t ThreadPool::getQueueSize() {
     return tasks.size();
 }
 
+bool ThreadPool::isPoolThread() const {
+    auto id = std::this_thread::get_id();
+    for (auto& worker : workers) {
```
petersteneteg, line 94: This could be a race condition. Before we only every touched the workers vector on the main thread, hence we did not need any look. but this will touch it from an arbitrary thread.

tools/ivwpy/regression/generatereport.py
```diff
@@ -450,10 +450,21 @@ def images(self, imgs, testdir):
...
-            fail = next((x for x in failures if x[0] == "images"), None)
-            return "ok" if fail is None else "fail"
+            failedImgs = next((x for x in failures if x[0] == "images"), None)
+            failed = (failedImgs is not None and img["image"] in failedImgs[1])
+            diffTolerance = safeget(self.report, "config", "image_test", "differenceTolerance", 
+                                    img["image"], failure=0.0)
+
+            if failed:
+                return "fail"
+            elif img["difference"] > diffTolerance * (1.0 - warnPercentage/100.0):
```
petersteneteg: still not sure how useful this is, differenceTolerance is almost always 0...

petersteneteg (dismissed):

modules/base/src/properties/volumeinformationproperty.cpp
martinfalk: So that the property states are not serialized. This affects `dataRange_, valueRange_, valueName_, valueUnit_, axesNames_, axesUnits_`. These will be set when loading the data or having a custom override.

tools/ivwpy/regression/generatereport.py
martinfalk: Agreed. If it is 0 most of the time, then this is hardly useful.

modules/base/src/properties/volumeinformationproperty.cpp
petersteneteg: but will they not already be default? you don't change them after the initialization?

tools/ivwpy/regression/generatereport.py
petersteneteg: The default is 0, and it is only if we manually change the value in the config file for that test that it will be nonzero. And I don't think we do that very often...

petersteneteg: the differences does not work very well since even a tiny visual change can have a huge difference. We would need some kind of visual similarity here for the tolerance to be useful. But on the other hand then we might miss a lot of regressions that only change minor aspects (like text moving halv a pixel), but that we still want to catch...

modules/qtwidgets/include/modules/qtwidgets/properties/valuedragspinbox.h
```diff
@@ -73,6 +73,11 @@ class IVW_MODULE_QTWIDGETS_API ValueDragSpinBox : public QWidget {
     void setMinimum(int min);
     void setPrefix(const QString& prefix);
     void setRange(int minimum, int maximum);
+    /**
+     * Sets the increment of a single step to @p inc.
+     * If @p inc is zero, the spinbox buttons will be hidden.
+     * @see QDoubleSpinBox::setSingleStep
+     */
     void setSingleStep(int val);
```
ResearchDaniel, line 81: This should also be renamed to `setIncrement`?

src/core/util/filelogger.cpp
```diff
@@ -71,22 +76,22 @@ void FileLogger::log(std::string_view logSource, LogLevel logLevel, LogAudience
...
 
         case LogLevel::Error:
-            fileStream_ << "<div class ='error'><span class='level'>Error: </span>";
+            fileStream_ << "<div class='error nowrap'><span class='level'>Error: </span>";
             break;
     }
 
     auto msg = htmlEncode(logMsg);
-    replaceInString(msg, " ", "&nbsp;");
+    //replaceInString(msg, " ", "&nbsp;");
```
ResearchDaniel: remove

petersteneteg: Remove

src/core/util/filelogger.cpp
```diff
@@ -71,22 +76,22 @@ void FileLogger::log(std::string_view logSource, LogLevel logLevel, LogAudience
...
     replaceInString(msg, "\n", "<br/>");
-    replaceInString(msg, "\t", "&nbsp;&nbsp;&nbsp;&nbsp;");
+    replaceInString(msg, "\t", "&emsp;");
```
petersteneteg: Will this always be 4 spaces?

include/inviwo/core/util/threadpool.h
```diff
@@ -143,10 +150,11 @@ auto ThreadPool::enqueue(F&& f, Args&&... args) -> std::future<std::invoke_resul
 
     std::future<return_type> res = task->get_future();
 
+    std::lock_guard<std::mutex> lock(workers_mutex);
```
petersteneteg: If you put the lock here and one submits a job in an empty pool an that job tries to submit something, that will deadlock. Generally it is a bad idea to hold any locks running arbitrary code, as when executing the task here.

src/core/util/threadpool.cpp
```diff
@@ -43,18 +43,26 @@ ThreadPool::ThreadPool(size_t threads, std::function<void()> onThreadStart,
     }
 }
 
+ThreadPool::~ThreadPool() {
+    std::lock_guard<std::mutex> lock(workers_mutex);
```
petersteneteg: This might also deadlock if any of the currently running jobs want to dispatch something before finishing I think

src/core/util/threadpool.cpp
```diff
@@ -68,17 +76,36 @@ size_t ThreadPool::trySetSize(size_t size) {
...
+            return true;
+        }
+    }
+    return false;
+}
+
+void ThreadPool::enqueueRaw(std::function<void()> task) {
+    std::lock_guard<std::mutex> lock(workers_mutex);
+    if (workers.empty()) {
+        task();  // No worker threads, just run the task.
```
petersteneteg, line 105: Se above

ResearchDaniel (dismissed):
Might want to make the setIncrement naming consistent

petersteneteg (changes requested):

modules/qtwidgets/include/modules/qtwidgets/properties/valuedragspinbox.h
martinfalk: This is in line with QDoubleSpinBox encapsulated within the NumberLineEdit.

src/core/util/filelogger.cpp
martinfalk: No, just a wider "regular" space. Typically it corresponds to the width of an 'm'. If we want something wider we should probably use a span with fixed width.

petersteneteg: But then it wound not align properly. it should always be the same as 4 spaces. The logging expects fixed width fonts and alignment with spaces and tabs to work.

include/inviwo/core/util/threadpool.h
```diff
@@ -143,10 +150,11 @@ auto ThreadPool::enqueue(F&& f, Args&&... args) -> std::future<std::invoke_resul
...
     if (workers.empty()) {
         (*task)();  // No worker threads, just run the task.
     } else {
```
petersteneteg, line 156:
```suggestion
    const bool noWorkers = [&](){
        std::lock_guard<std::mutex> lock(workers_mutex);
        return workers.empty();
    }()
    if (noWorkers) {
        (*task)();  // No worker threads, just run the task.
    } else {
```

petersteneteg:
```suggestion
    if (workers.empty()) {
        lock.unlock();
        (*task)();  // No worker threads, just run the task.
    } else {
```

src/core/util/threadpool.cpp
```diff
@@ -43,18 +43,26 @@ ThreadPool::ThreadPool(size_t threads, std::function<void()> onThreadStart,
     }
 }
 
+ThreadPool::~ThreadPool() {
+    std::lock_guard<std::mutex> lock(workers_mutex);
+    for (auto& worker : workers) worker->state = State::Abort;
+    condition.notify_all();
+    workers.clear();  // this will join all threads.
```
petersteneteg:
```suggestion
    std:: unique_lock lock(workers_mutex);
    for (auto& worker : workers) worker->state = State::Abort;
    condition.notify_all();
    auto localCopy = std::move(workers);
    lock.unlock();
    localCopy();  // this will join all threads.
```
This might be a solution.

include/inviwo/core/util/threadpool.h
petersteneteg:
```suggestion
    std::unique_lock<std::mutex> lock(workers_mutex);
```

src/core/util/filelogger.cpp
martinfalk: According to [mozilla](https://developer.mozilla.org/en-US/docs/Web/CSS/tab-size), there is a css property `tab-size` for setting the spaces, for example in `<pre>` .

martinfalk:
decided to break this PR up into several parts.
See #1301, #1303, and #1304.

PR #1295, opened 2022-04-28
Python3: Packaging for Mac/Linux

ResearchDaniel:
Fixes python packaging for mac and linux, hopefully

### DIFF
--- a/.github/workflows/inviwo.yml
+++ b/.github/workflows/inviwo.yml
@@ -178,8 +178,8 @@ jobs:
         include:
           - os: 'windows-latest'
             installer: 'inviwo-installer-win'
-          #- os: 'macos-latest'
-          #  installer: 'inviwo-installer-macos'
+          - os: 'macos-latest'
+            installer: 'inviwo-installer-macos'
         
     runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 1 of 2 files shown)


petersteneteg:
Should also test at least the Mac installer then...
Not sure if this is they way we want to do this. Also we should isolate our python interpreter to avoid environment issues. 
And we need to provide some way to install additional things?

petersteneteg:
@ResearchDaniel Have you tested this locally? The installer that gets created for MacOS has no python.dll file in it from that I can see

ResearchDaniel:
@petersteneteg I don't think that I managed to get this approach entirely working. Relying on vcpkg for Python would probably be better than downloading installers.

petersteneteg:
mmm, vcpkg will automatically download python and and link it, when we run cmake. The issue is that it will not install any packages. so after vcpkg is done installing everything. cmake will fail since we are missing the numpy headers from python...
So then you have to manually find the python installation from vcpkg and install numpy in that one and then rerun cmake which is kind of horrible... would have been nice if one could have had vcpkg also install all python packages in some requirements.txt file next to the vcpkg.json...

petersteneteg:
I wonder if one could create some local vcpkg port that would do that... might be possible with a bit of hacking...

ResearchDaniel:
>  The issue is that it will not install any packages. so after vcpkg is done installing everything. cmake will fail since we are missing the numpy headers from python...

The idea would be to create a virtual environment of the vcpkg python and install packages into that one. 
We would then need to add the vcpkg python to the installer and run a script that installs pip and necessary packages during installation. That is how I understood it should be done when discussing it with the python vcpkg maintainers

petersteneteg:
Mmmm that might be possible...

PR #1296, opened 2022-05-01
gamepad module to steer Inviwo parameters via a controller

weinkauf:
Implements several processors that work together to enable gamepad support in Inviwo. See the provided workspace for how it works.

This has been tested on: Windows 10 / VS 2019 with different controllers. Qt 5.15.2, cmake 3.22.2, python 3.7

> 

### DIFF
--- a/modules/gamepad/CMakeLists.txt
+++ b/modules/gamepad/CMakeLists.txt
@@ -1,0 +1,40 @@
+ivw_module(Gamepad)
+
+#~ find_package(Qt5Gamepad QUIET REQUIRED)
+# https://doc-snapshots.qt.io/qt6-dev/cmake-qt5-and-qt6-compatibility.html#supporting-older-qt-5-versions   
+find_package(QT NAMES Qt6 Qt5 COMPONENTS Gamepad REQUIRED)
+find_package(Qt${QT_VERSION_MAJOR} COMPONENTS Gamepad REQUIRED)
+
+set(HEADER_FILES
+    include/inviwo/gamepad/gamepadmodule.h
+    include/inviwo/gamepad/gamepadmoduledefine.h
+    include/inviwo/gamepad/processors/gamepadbuttonsswitches.h
+    include/inviwo/gamepad/processors/gamepadcamera.h
+    include/inviwo/gamepad/processors/gamepadinput.h
+    include/inviwo/gamepad/processors/gamepadordinals.h
+    include/inviwo/gamepad/properties/gamepadtextproperty.h
+    include/inviwo/gamepad/properties/gamepadvaluesproperty.h
+)
+ivw_group("Header Files" ${HEADER_FILES})
+
+set(SOURCE_FILES
+    src/gamepadmodule.cpp
+    src/processors/gamepadbuttonsswitches.cpp
+    src/processors/gamepadcamera.cpp
+    src/processors/gamepadinput.cpp
+    src/processors/gamepadordinals.cpp
+    src/properties/gamepadtextproperty.cpp
+    src/properties/gamepadvaluesproperty.cpp
+)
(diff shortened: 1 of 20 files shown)


weinkauf:
Where are we with this? Can I help in any way?

ResearchDaniel:
Sorry for not moving this forward. I tried to get an overview of what was done and I had some questions about if it would be possible to solve some of the problems without using property linking and separate processors, e.g., by providing an API that is used by one or more processors. I am unsure if communicating here is the best approach, verbally would probably be faster. @petersteneteg might have a different view.

ResearchDaniel:
As an example: 
>   This processor has no inputs or outputs in the classic sense.
Everything is done through property connections.
  This processor itself will not steer other Inviwo processors,
  but rather one needs to connect specific gamepad-related processors,
  which will then manipulate the camera or some buttons.

We try to avoid "floating processors" that do not have any input/output and have instead tried to expose such things via the module itself.

petersteneteg:
This looks nice! If I only had a gamepad... 
Im currently on vacation and don't really have time to look into this too much until after midsummer.
I initially thought this should be handled using events as keyboard and mouse are done. But I can see that the current approach might be useful for the case of multiple controllers and since a gamepad don't have the same connection to a canvas as maybe a mouse has. We might want to discuss that in one of out dev meetings. 

The timer vs each press might also be discussed, right now I only saw that using "every press" vs the timer "seemed tedious to use" but I could not figure out why, and since this would be different to how we handle mouse and keyboard events I think that should be made more clear.

Otherwise I think the code looks very good! there are some minor stylistic fixes that should be addressed
* All methods names and variables should use camelCase never PascalCase.
* We don't use any type prefixes in variable names, like `bOn` but something like `isOn`. This also included Properties etc.
* We use IVW_ASSERT for assert so we can inject some nice behavior in the IDEs.
* we __never__ use c-style cast like `(float)someStringVariableThatNowIsANumber
* In new code we generally use libfmt for any string formatting
* We also don't include author information in files in the core Inviwo repo.
    Also comment such as:
    > // Method
    > // Properties
* There is also some commented out code left over that should be removed.

weinkauf:
I can give you a demo at some point after your vacation.

I tried already with the style... Jesus! 😂

weinkauf:
Is there an Inviwo online meeting on June 2? Shall I show it there?

martinfalk:
At the moment there is no Qt6 support for Gamepads, only Qt5.
Independent of that we should potentially consider an alternative "gamepad provider" besides Qt like glfw.

weinkauf:
> At the moment there is no Qt6 support for Gamepads, only Qt5. Independent of that we should potentially consider an alternative "gamepad provider" besides Qt like glfw.

Other alternatives, but I never tried them:
gainput: https://gainput.johanneskuhlmann.de
SDL: https://www.libsdl.org

According to this thread https://www.qt.io/blog/qt-6.0-alpha-released , it will be added in a later release of Qt6. You get the same impression from the git repo for Qt gamepad: https://code.qt.io/cgit/qt/qtgamepad.git/

All in all, a gamepad is cool, but not necessary. I almost feel one should rely on Qt here and only have it available when Qt supports it.

ResearchDaniel:
Bummer:
https://www.qt.io/product/qt6/qt-5-15-vs-6-2-feature-comparison

> **Qt Gamepad** There are no plans to port this add-on module because the related customer projects have been discontinued.

martinfalk:
GLFW has support for gamepads. And Inviwo uses GLFW already for unit tests. Might be interesting to see whether we get GLFW to work along with Qt while not using a GLFW OpenGL context.

martinfalk:
Closing this PR for now. Moved to discussions #1333

PR #1298, opened 2022-05-02
QtApplicationBase: Improve generalized usage of InviwoApplicationQt

ResearchDaniel:
Setting the icon in InviwoApplicationQt overrides icons set externally (e.g., using CMake).

This has been tested on: Mac, Windows

### DIFF
--- a/src/qt/applicationbase/inviwoapplicationqt.cpp
+++ b/src/qt/applicationbase/inviwoapplicationqt.cpp
@@ -135,7 +135,9 @@ void FileSystemObserverQt::fileChanged(QString fileName) {
 
 }  // namespace
 
-InviwoApplicationQt::InviwoApplicationQt(int& argc, char** argv, std::string displayName, std::string organizationName, std::string organizationDomain)
+InviwoApplicationQt::InviwoApplicationQt(int& argc, char** argv, std::string displayName,
+                                         std::string organizationName,
+                                         std::string organizationDomain)
     : QApplication(argc, argv)
     , InviwoApplication(argc, argv, displayName)
     , mainWindow_(nullptr)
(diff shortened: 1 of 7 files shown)


src/qt/applicationbase/inviwoapplicationqt.cpp
```diff
@@ -135,19 +135,20 @@ void FileSystemObserverQt::fileChanged(QString fileName) {
...
+                                         std::string organizationName,
+                                         std::string organizationDomain)
     : QApplication(argc, argv)
     , InviwoApplication(argc, argv, displayName)
     , mainWindow_(nullptr)
     , uiLocal_(getCurrentStdLocale())
     , undoTrigger_{[]() {}} {
 
     setAttribute(Qt::AA_NativeWindows);
-    setWindowIcon(QIcon(":/inviwo/inviwo_light.png"));
```
martinfalk, line 146: does the default CMake file set the appropriate window icon?

ResearchDaniel: Yes, we set the following:

```
set(RES_FILES "")
if(WIN32)
    set(RES_FILES ${RES_FILES} "${IVW_RESOURCES_DIR}/inviwo.rc")
elseif(APPLE)
    set_source_files_properties(${IVW_ROOT_DIR}/resources/inviwo/inviwo_light.icns PROPERTIES MACOSX_PACKAGE_LOCATION Resources)
    set(RES_FILES ${RES_FILES} ${IVW_ROOT_DIR}/resources/inviwo/inviwo_light.icns)
endif()
```

I have not tested this on windows and Linux.

ResearchDaniel: Tested on Windows and there it seems like it is necessary to call setWindowIcon. So, I added an option to specify the icon

ResearchDaniel:
Seems like there was an error when mergin. Will fix

PR #1305, opened 2022-05-09
Feature/linux deploy

ResearchDaniel:
Fixes Linux deployment

Changes proposed in this PR:
 * Use LinuxDeployQt for copying Qt dependencies
 * CI: Disable Python on Mac/Linux since we cannot deploy it (might want to do something differ

### DIFF
--- a/.github/workflows/inviwo.yml
+++ b/.github/workflows/inviwo.yml
@@ -99,7 +99,7 @@ jobs:
         -storepasswordincleartext \
         -name "GitHub" \
         -username "inviwo" \
-        -password "${{ secrets.GITHUB_TOKEN }}" \
+        -password "${{ secrets.GITHUB_TOKEN }}" 
         ${{ matrix.mono }} `./vcpkg/vcpkg fetch nuget | tail -n 1` \
         setapikey "${{ secrets.GITHUB_TOKEN }}" \
         -source "https://nuget.pkg.github.com/inviwo/index.json"
--- a/.github/workflows/inviwo.yml
+++ b/.github/workflows/inviwo.yml
@@ -23,6 +23,7 @@ jobs:
             targets: 'ALL_BUILD DOXY-Inviwo DOXY-Python package'
             installer: 'inviwo-installer-win'
             artifact: 'build/inviwo-v*.exe'
+            bootstrap_vcpkg: "./vcpkg/bootstrap-vcpkg.bat"
           - os: 'macos-latest'
             triplet: 'x64-osx'
             mono: 'mono'
@@ -31,6 +32,7 @@ jobs:
             targets: 'all package'
             installer: 'inviwo-installer-macos'
             artifact: 'build/Inviwo-v*.dmg'
+            bootstrap_vcpkg: "./vcpkg/bootstrap-vcpkg.sh"
           - os: 'ubuntu-20.04'
             triplet: 'x64-linux'
             mono: 'mono'
@@ -41,6 +43,7 @@ jobs:
             targets: 'all package'
             installer: 'inviwo-installer-ubuntu'
             artifact: 'build/inviwo-v*.deb'
+            bootstrap_vcpkg: "./vcpkg/bootstrap-vcpkg.sh"
       fail-fast: false
       
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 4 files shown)


apps/inviwo/packaging/packaging.cmake
```diff
@@ -114,6 +114,31 @@ elseif(APPLE)
...
+      "Name=${IVW_PACKAGE_SELECT_APP}"
+      "Exec=${IVW_PACKAGE_SELECT_APP}"
+      "Icon=${IVW_PACKAGE_SELECT_APP}"
+      "Categories=Science;"
+    )
+    install(FILES ${PROJECT_BINARY_DIR}/${IVW_PACKAGE_SELECT_APP}.desktop 
+            DESTINATION share/applications 
+            COMPONENT Application
+    )
+    install(FILES ${IVW_ROOT_DIR}/resources/inviwo/inviwo-logo-512px.png
```
martinfalk, line 127: Any idea whether SVG icons are also supported?

martinfalk (dismissed):

apps/inviwo/packaging/packaging.cmake
ResearchDaniel: I believe so, but the first goal is to make it work.
One can dive into it here:
https://specifications.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html

.github/workflows/inviwo.yml
```diff
@@ -77,8 +82,15 @@ jobs:
       with:
         repository: 'microsoft/vcpkg'
         path: 'vcpkg' # The cmake vcpkg present expects vcpkg to be next to the inviwo source folder
-    
+
+    - name: 'Bootstrap vcpkg'
+      shell: bash
+      run: |
+        ${{ matrix.bootstrap_vcpkg }} 
+        ./vcpkg/vcpkg integrate install
```
petersteneteg: No we do not want to use that, we use manifests and toolchains so we can control the dependencies

.github/workflows/inviwo.yml
```diff
@@ -77,8 +82,15 @@ jobs:
...
+
     # Cache built dependencies for faster subsequent builds
+    # ./vcpkg/bootstrap-vcpkg.bat must have been called before
```
petersteneteg, line 92: Since when?

petersteneteg (changes requested):

.github/workflows/inviwo.yml
ResearchDaniel: Not sure if this is the issue with the Mac/Linux not working. Thought I would give it a try

https://nicedoc.io/Microsoft/vcpkg/docs/users/binarycaching.md#github-packages

petersteneteg: What is not working for Mac/linux? The caching?

ResearchDaniel: Yes, exactly, it fails when trying to upload. But I think that I have fixed it now (see change for nuget setup). Unsure if the bootstrapping before is necessary.

for some reason it does not work for the visualneuro repo. Perhaps the repo  is used in the nuget setup somehow

petersteneteg:
@ResearchDaniel status? should we close this? We should create an AppImage on linux If we want an installer... just taring the install folder will be very very error prone.

ResearchDaniel:
@petersteneteg This PR tries to do that using linuxdeployqt. I have limited possibilities for testing this beyond github actions and it is not on my prio list. 

 Perhaps we can at least extract the nuget binary caching into a separate PR

petersteneteg:
linuxdeployqt does not do any appImage it just copies dependencies... an appimage is a closer to a container https://appimage.org

ResearchDaniel:
> linuxdeployqt does not do any appImage it just copies dependencies... an appimage is a closer to a container https://appimage.org

I think that you can "just" supply an -appimage argument to linuxdeployqt and then it should be more or less it
https://github.com/coryo/amphetype2/blob/2d41de3b0c19ab9286672ff0d6a7c11eadc13d9c/.travis/deploy.sh
From here:
https://github.com/AppImage/AppImageKit/wiki/Creating-AppImages/cc2441518975caca23e9ce2dba6f08a22c678d1e#4-run-linuxdeployqt-on-your-qt-application

PR #1306, opened 2022-05-18
Feature/msvcfix

martinfalk:
Fixes compiler error with MSVC 17.2.0.

The latest Visual Studio compiler does not like capturing the buffer in the second lambda expression.
```c++
return static_cast<ValueType>(std::inner_product(
  indices.begin(), indices.end(), weights.begin(), T{0}, std::plus<>{},
  [&](uint32_t index, float weight) {   // <-- capture causes error C2440
      return static_cast<T>(buffer[index]) * weight;
  }));
```

```
inviwo\modules\base\src\algorithm\mesh\meshclipping.cpp(492,29): error C2440: '<function-style-cast>': cannot convert from 'initializer list' to 'inviwo::meshutil::clipMeshAgainstPlane::<lambda_1>::()::<lambda_1>::()::<lambda_1>'
inviwo\modules\base\src\algorithm\mesh\meshclipping.cpp(531,17): message : No constructor could take the source type, or constructor overload resolution was ambiguous
```

### DIFF
--- a/modules/base/src/algorithm/mesh/meshclipping.cpp
+++ b/modules/base/src/algorithm/mesh/meshclipping.cpp
@@ -185,10 +185,9 @@ std::vector<std::vector<std::uint32_t>> gatherLoops(std::vector<glm::u32vec2>& e
                     edges.erase(it);
                 }
             } else {
-                LogWarnCustom(
-                    "MeshClipping",
-                    "Found edge, that is not connected to any other edge. This could mean, the "
-                    "clipped mesh was not manifold.");
+                LogWarnCustom("MeshClipping",
+                              "Found edge that is not connected to any other edge. This could mean "
+                              "the clipped mesh was not a manifold.");
                 break;
             }
         }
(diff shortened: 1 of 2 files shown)


ResearchDaniel (approved):

modules/base/src/algorithm/mesh/meshclipping.cpp
```diff
@@ -486,10 +492,8 @@ std::shared_ptr<Mesh> clipMeshAgainstPlane(const Mesh& mesh, const Plane& worldS
                                                const std::vector<uint32_t>& indices,
                                                const std::vector<float>& weights) {
                         return static_cast<ValueType>(std::inner_product(
-                            indices.begin(), indices.end(), weights.begin(), T{0}, std::plus<>{},
-                            [&](uint32_t index, float weight) {
-                                return static_cast<T>(buffer[index]) * weight;
-                            }));
+                            indices.begin(), indices.end(), weights.begin(), T{0}, std::plus<T>{},
```
petersteneteg, line 495: did you try removing the `static`from `static const auto mix`?
Should also use [[maybe_unused]] instead of the void cast.

martinfalk: yes, did not fix the problem. The only way around was to not use `buffer` in the lambda. Capturing `&buffer` led to the same problem.

PR #1308, opened 2022-06-02
Python3: Expose data readers and MeshSeqeuence

ResearchDaniel:
Changes proposed in this PR:
 * Expose data readers and MeshSeqeuence to Python
 

modules/python3/bindings/src/pydatareaders.cpp
```diff
@@ -0,0 +1,94 @@
...
+        .def("getExtensions", &DataReader::getExtensions)
+        .def("addExtension", &DataReader::addExtension)
+        .def("setOption", &DataReader::setOption)
+        .def("getOption", &DataReader::getOption);
+
+    py::class_<DataReaderType<Image>, DataReader>(m, "ImageDataReader")
+        .def("readData", [](DataReaderType<Image>* reader, std::string_view filePath) {
+                return reader->readData(filePath);
+            },
+            py::return_value_policy::reference);
```
petersteneteg: explain why the return value should be reference? seems very strange. 
Why no init function? 
We should have a trampoline class here to

modules/python3/bindings/src/pydatareaders.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+    py::class_<DataReaderFactory>(m, "DataReaderFactory")
+        .def("hasKey", [](DataReaderFactory* f, std::string key) { return f->hasKey(key); })
+        .def("hasKey", [](DataReaderFactory* f, FileExtension key) { return f->hasKey(key); })
+        .def("create", [](DataReaderFactory* f, std::string key) { return f->create(key); })
+        .def("create", [](DataReaderFactory* f, FileExtension key) { return f->create(key); })
+        .def("getImageReader", [](DataReaderFactory* f, std::string_view filePathOrExtension) {
```
petersteneteg: This does not look like the cleanest solution, Is there no better approach?

modules/python3/bindings/src/pymesh.cpp
```diff
@@ -296,7 +299,11 @@ void exposeMesh(pybind11::module& m) {
             return self.addVertex(vertex);
         });
 
+    using MeshSequence = std::vector<std::shared_ptr<Mesh>>;
+    py::bind_vector<MeshSequence>(m, "MeshSequence", py::module_local(false));
```
petersteneteg: Why module_local(false)? added comment

modules/python3/bindings/src/pydatareaders.cpp
```diff
@@ -0,0 +1,94 @@
...
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ *********************************************************************************/
+
+#include <inviwopy/pydatareaders.h>
+#include <inviwopy/vectoridentifierwrapper.h>
+#include <inviwopy/pypropertytypehook.h>
```
petersteneteg: This include is not used.

petersteneteg (changes requested):

modules/python3/bindings/src/pymesh.cpp
ResearchDaniel: Code is based on this https://github.com/inviwo/inviwo/commit/d0b1ddb4c73231cba9dafe252ded032d28814d39

I don't think that the `module_local(false)` is necessary as pybind will try to determine it itself , but you might have had a reason for including it? I tested to remove it and it seems to run without issues at least.

https://pybind11.readthedocs.io/en/stable/advanced/cast/stl.html?highlight=module_local#binding-stl-containers
> When binding STL containers pybind11 considers the types of the container’s elements to decide whether the container should be confined to the local module (via the [Module-local class bindings](https://pybind11.readthedocs.io/en/stable/advanced/classes.html#module-local) feature). If the container element types are anything other than already-bound custom types bound without py::module_local() the container binding will have py::module_local() applied.

modules/python3/bindings/src/pydatareaders.cpp
ResearchDaniel: > explain why the return value should be reference? seems very strange.

Fixed. 

> Why no init function?

They are abstract classes.

> We should have a trampoline class here to

Yes, it seems like many of the exposed classes are missing trampolines so it won't be possible to extend those classes in Python. The Port is one such example.
I have added trampolines for the DataReader and DataReaderType.

modules/python3/bindings/src/pydatareaders.cpp
ResearchDaniel: Refactored a bit, but I could still not find a better way of dealing with the template functions.

modules/python3/bindings/src/pyinviwoapplication.cpp
```diff
@@ -32,7 +32,7 @@
 
 #include <inviwo/core/common/inviwoapplication.h>
 #include <inviwo/core/common/inviwomodule.h>
-#include <inviwo/core/network/processornetwork.h>
+#include <inviwo/core/io/datareaderfactory.h>                #include <inviwo/core/network/processornetwork.h>
```
martinfalk: missing newline

modules/python3/bindings/src/pydatareaders.cpp
petersteneteg: Yes... trampolines did not work very well before I did my last major refactoring. We ran into a lot of lifetime issues that made it very hard/impossible to use...

ResearchDaniel:
Superseded by #1326

Issue #1309, opened 2022-06-08
[BUG]: Python doxygen doc generation 


martinfalk:
### Required prerequisites

- [X] Make sure you've read the [documentation](https://inviwo.org). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/inviwo/inviwo/issues) and [Discussions](https://github.com/inviwo/inviwo/discussionss) to verify that this hasn't already been reported. +1 or comment there if it has.
- [X] Consider asking first in [Slack](https://join.slack.com/t/inviwo/shared_invite/enQtNTc2Nzc2NDQwNzIxLTRiMWM1ZWJiYjljZjkwNWE3OTk3MzYxODZlMDUyMzRmZjUzMzBiZjVhNTM3NWUyNzU1MjI4OWJjMzdkODViMzM) or in a [Discussion](https://github.com/inviwo/inviwo/discussionss).

### Problem description

When enabling `IVW_DOXYGEN_PROJECT`, the build fails for `DOXY-Python` as it expects the source files in the build folder to be `.py` and not `.pyd`.

Same failure occurs in github actions VS build.

### Inviwo version

latest master

### Operating system?

Windows

### Compiler / Version

MSVC 2022

### Inviwo log

```text
1>------ Build started: Project: DOXY-Python, Configuration: RelWithDebInfo x64 ------
1>Generate Inviwo python API documentation with sphinx
1>Running Sphinx v5.0.1
1>making output directory... done
1>[autosummary] generating autosummary for: index.rst
1>
1>CUSTOMBUILD : Extension error (sphinx.ext.autosummary):
1>Handler <function process_generate_options at 0x000001F3EEBA08B0> for event 'builder-inited' threw an exception (exception: source is not a .py file: 'build-vs\\bin\\RelWithDebInfo\\inviwopy.cp310-win_amd64.pyd')
1>C:\Program Files\Microsoft Visual Studio\2022\Enterprise\MSBuild\Microsoft\VC\v170\Microsoft.CppCommon.targets(245,5): error MSB8066: Custom build for 'build-vs\CMakeFiles\5c7356f6e482fb847eb26ad3dccf416d\DOXY-Python.rule' exited with code 2.
```

petersteneteg:
sphinx should most definitely work on the pyd file not any py files.

martinfalk:
[...]

PR #1311, opened 2022-06-30
CI: disabled doxygen for win builds, #1309

martinfalk:
Disable doxygen in CI workflow for windows builds due to #1309.

### DIFF
--- a/.github/workflows/inviwo.yml
+++ b/.github/workflows/inviwo.yml
@@ -18,9 +18,11 @@ jobs:
           - os: 'windows-latest'
             triplet: 'x64-windows'
             mono: ''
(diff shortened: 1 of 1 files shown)


petersteneteg (changes requested):
This output is used on the homepage. Should be fixed immediately not disabled

martinfalk:
> This output is used on the homepage. Should be fixed immediately not disabled

I've looked into it for an entire day and was not able to figure out why `sphinx.ext.autosummary` does not accept the pyd files though the main sphinx loads the python modules. Excluding the autosummary "fixes" the build failing. But of course the generated docs are empty.

petersteneteg:
Did you investigate what might have changed? This did work not that long ago?

martinfalk:
Have not had time to track back the commits. Suspect it has to do with a commit regarding pydocs.

PR #1316, opened 2022-07-06
Feature/atlasisoraycaster

RasmusHogslatt:
Summary:
* Adding one processor AtlasBoundary and a AtlasIsosurfaceComponent added to the current AtlasVolumeRaycaster processor.

AtlasBoundary:
Utilizes brushing and linking to convert a segmented volume into a UINT16 volume with voxel values in range [0, 3] and interpolation set to nearest.

AtlasIsosurfaceComponent (Added to AtlasVolumeRaycaster):
Adds option to render the segments produced by AtlasBoundary and toggle lighting on or off.

This has been tested on: Windows 10 / MSVC 143 / VS 2022 / QT 6.2.4 / Python 3.10


modules/basegl/glsl/raycasting/atlasisosurfacerenderer.frag
```diff
@@ -0,0 +1,130 @@
...
+    cDs.y = getNormalizedVoxelChannel(volume, volumeParams, samplePos + volumeParams.textureSpaceGradientSpacing[1]*scale,channel);
+    cDs.z = getNormalizedVoxelChannel(volume, volumeParams, samplePos + volumeParams.textureSpaceGradientSpacing[2]*scale,channel);
+    // Value at f(x-h)
+    cDs.x = cDs.x - getNormalizedVoxelChannel(volume, volumeParams, samplePos - volumeParams.textureSpaceGradientSpacing[0]*scale,channel);
+    cDs.y = cDs.y - getNormalizedVoxelChannel(volume, volumeParams, samplePos - volumeParams.textureSpaceGradientSpacing[1]*scale,channel);
+    cDs.z = cDs.z - getNormalizedVoxelChannel(volume, volumeParams, samplePos - volumeParams.textureSpaceGradientSpacing[2]*scale,channel);
+    // Note that this computation is performed in world space
+    // f' = ( f(x+h)-f(x-h) ) / 2*volumeParams.worldSpaceGradientSpacing
+    return (cDs)/(2.0*volumeParams.worldSpaceGradientSpacing);
+}
```
petersteneteg: Note: you also have to apply the scaling at the last step when dividing by the step size.

modules/basegl/CMakeLists.txt
```diff
@@ -167,6 +169,7 @@ set(SOURCE_FILES
     src/processors/meshrenderprocessorgl.cpp
     src/processors/multichannelraycaster.cpp
     src/processors/pointrenderer.cpp
+    src/processors/raycasting/atlasisosurfacerenderer.cpp
```
petersteneteg: Is the any difference between this one and using the components? Is this just for debugging? I would prefer to only have the component one for maintenance reasons, unless there is a good reason to have. a separate one.

modules/basegl/include/modules/basegl/processors/raycasting/atlasvolumeraycaster.h
```diff
@@ -93,6 +95,7 @@ class IVW_MODULE_BASEGL_API AtlasVolumeRaycaster : public VolumeRaycasterBase {
     PositionIndicatorComponent positionIndicator_;
     SampleTransformComponent sampleTransform_;
     AtlasComponent atlas_;
+    AtlasIsosurfaceComponent atlasIso_;
```
petersteneteg: Question? is this related to isosurfaces? Is this not for showing the segment boundaries? that is probably not an iso surface right? Maybe rename to SegementSurfaceComponent or RegionSurfaceComponent?

modules/basegl/include/modules/basegl/shadercomponents/atlasisosurfacecomponent.h
```diff
@@ -0,0 +1,71 @@
...
+#include <string>
+#include <vector>
+
+namespace inviwo {
+
+/**
+ * Adds an option to render segments in range [0, 3] of type UINT, and render these
+ * with or without lighting. See the Atlas Boundary processor for such data.
+ */
+class IVW_MODULE_BASEGL_API AtlasIsosurfaceComponent : public ShaderComponent {
```
petersteneteg: Again rename...

modules/basegl/include/modules/basegl/shadercomponents/atlasisosurfacecomponent.h
```diff
@@ -0,0 +1,71 @@
...
+    VolumeInport volume_;
+    std::string name_;
+    //std::shared_ptr<const Volume> smoothVolume_;
+
+    BoolCompositeProperty useAtlasBoundary_;
+    BoolProperty applyBoundaryLight_;
+    SelectionColorProperty showHighlighted_;
+    SelectionColorProperty showSelected_;
+    SelectionColorProperty showFiltered_;
+    SamplerObject nearestSampler_;
```
petersteneteg: You only need the one linear one or?

modules/basegl/src/processors/raycasting/atlasisosurfacerenderer.cpp
```diff
@@ -0,0 +1,143 @@
...
+        throw Exception("Couldn't find VolumeGL representation", IVW_CONTEXT);
+    }
+    utilgl::activateAndClearTarget(outport_);
+    shader_.activate();
+
+    TextureUnitContainer units;
+    utilgl::bindAndSetUniforms(shader_, units, volume, "volume");
+    utilgl::bindAndSetUniforms(shader_, units, entryPort_, ImageType::ColorDepthPicking);
+    utilgl::bindAndSetUniforms(shader_, units, exitPort_, ImageType::ColorDepth);
+
```
petersteneteg: You should bind the SamplerObject to a unit here, and bind the altas to the same unit. Maybe that inviwo::SamplerObject should handle that automatically

modules/basegl/include/modules/basegl/shadercomponents/atlasisosurfacecomponent.h
```diff
@@ -0,0 +1,71 @@
...
+class IVW_MODULE_BASEGL_API AtlasIsosurfaceComponent : public ShaderComponent {
+public:
+    AtlasIsosurfaceComponent(std::string_view volume);
+    virtual std::string_view getName() const override;
+    virtual void process(Shader& shader, TextureUnitContainer&) override;
+    virtual std::vector<Property*> getProperties() override;
+    virtual std::vector<Segment> getSegments() override;
+
+private:
+    VolumeInport volume_;
```
petersteneteg: Should have a name like selectionAtlasInport, Or something descriptive...

modules/basegl/src/processors/raycasting/atlasvolumeraycaster.cpp
```diff
@@ -58,17 +58,19 @@ AtlasVolumeRaycaster::AtlasVolumeRaycaster(std::string_view identifier,
     , light_{&camera_.camera}
     , positionIndicator_{}
     , sampleTransform_{}
-    , atlas_{this, "color", &time_} {
+    , atlas_{this, "color", &time_}
+    , atlasIso_{"atlasiso"} {
```
petersteneteg: Not sure if this should be in the regular altas renders, Has to be optional In that case I think.

modules/basegl/src/shadercomponents/atlasisosurfacecomponent.cpp
```diff
@@ -0,0 +1,178 @@
...
+#include <modules/opengl/volume/volumegl.h>
+#include <inviwo/core/ports/volumeport.h>
+
+#include <inviwo/core/datastructures/buffer/bufferram.h>
+
+namespace inviwo {
+
+AtlasIsosurfaceComponent::AtlasIsosurfaceComponent(std::string_view volume)
+    : ShaderComponent{}
+    , volume_{"atlasboundary"}
```
petersteneteg: You should make this port optional

modules/basegl/src/shadercomponents/atlasisosurfacecomponent.cpp
```diff
@@ -0,0 +1,178 @@
...
+
+    shader.setUniform("useAtlasBoundary", useAtlasBoundary_.getBoolProperty()->get());
+    shader.setUniform("applyBoundaryLight", applyBoundaryLight_.get());
+    shader.setUniform("selectedColor", showSelected_.getColor());
+    shader.setUniform("showSelected", showSelected_.getBoolProperty()->get());
+    shader.setUniform("highlightedColor", showHighlighted_.getColor());
+    shader.setUniform("showHighlighted", showHighlighted_.getBoolProperty()->get());
+    shader.setUniform("filteredColor", showFiltered_.getColor());
+    shader.setUniform("showFiltered", showFiltered_.getBoolProperty()->get());
+
```
petersteneteg: You should bind your samplerObject to a texture unit here.

modules/basegl/src/shadercomponents/atlasisosurfacecomponent.cpp
```diff
@@ -0,0 +1,178 @@
...
+            auto in = std::shared_ptr<Volume>(volume_.getData()->clone());
+            in->setInterpolation(InterpolationType::Linear);
+            smoothVolume_ = in;
+        }
+        GLuint id = smoothVolume_->getRepresentation<VolumeGL>()->getTexture()->getID();
+        smoothVolume_->getRepresentation<VolumeGL>()->bindTexture(id);
+    }*/
+}
+
+std::vector<Property*> AtlasIsosurfaceComponent::getProperties() { return {&useAtlasBoundary_}; }
```
petersteneteg: add the other properties

modules/basegl/src/shadercomponents/atlasisosurfacecomponent.cpp
```diff
@@ -0,0 +1,178 @@
...
+)");
+
+constexpr std::string_view loop = util::trim(R"(
+prevBoundaryValue = boundaryValue;
+boundaryValue = uint(atlasSegment*3.0f + 0.5f);
+
+if(!useAtlasBoundary)
+{
+    // Do nothing
+}
```
petersteneteg: We probably don't what this as a runtime option, we should probably recompile the shader when this changes...
@martinfalk are the compiler smart enough to understand that this value will be the same for all voxels and not have to do both paths?

modules/basegl/src/shadercomponents/raycastingcomponent.cpp
```diff
@@ -70,6 +70,7 @@ RaycastingComponent::RaycastingComponent(std::string_view volume, IsoTFProperty&
 std::string_view RaycastingComponent::getName() const { return raycasting_.getIdentifier(); }
 
 void RaycastingComponent::process(Shader& shader, TextureUnitContainer&) {
+    shader.setUniform("gradientOffsetSpacing", raycasting_.gradientSpacingOffset_.get());
```
petersteneteg: I'm not sure we need this setting in the general case. I have never had any reason to use it before.

modules/opengl/include/modules/opengl/texture/samplerobject.h
```diff
@@ -0,0 +1,64 @@
...
+#include <modules/opengl/openglmoduledefine.h>
+#include <modules/opengl/inviwoopengl.h>
+#include <modules/opengl/glformats.h>
+#include <modules/opengl/texture/texture.h>
+
+namespace inviwo {
+
+/**
+ * Sampler objects store parameters that are not specific to textures. Thus, they store filter modes, wrapping mode but not texture width, height, swizzlemask etc. Binding a sampler object to a texture unit takes that textures uniform location in the shader, and multiple different sampler objects can be assigned to the same texture unit without having to copy the texture data.
+ */
```
petersteneteg: Linewrap

modules/opengl/include/modules/opengl/texture/samplerobject.h
```diff
@@ -0,0 +1,64 @@
...
+class IVW_MODULE_OPENGL_API SamplerObject {
+public:
+    SamplerObject();
+    virtual ~SamplerObject();
```
petersteneteg: Should not be virtual

modules/opengl/src/texture/samplerobject.cpp
```diff
@@ -0,0 +1,82 @@
...
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ *********************************************************************************/
+
+#include <modules/opengl/texture/samplerobject.h>
+
+namespace inviwo {
+
+SamplerObject::SamplerObject() : currentlyBoundTextureUnit_{0} { glGenSamplers(1, &id_); }
```
petersteneteg: Should add parameter options to the constructor

modules/opengl/src/texture/samplerobject.cpp
```diff
@@ -0,0 +1,82 @@
...
+    }*/ // Causes "unable to upload renderui.vert"
+}
+
+// SamplerObject::SamplerObject(const Texture& tex) {
+//	SamplerObject();
+//	bind(tex.getID); }
+
+void SamplerObject::bind(GLuint textureUnitNumber) {
+    glBindSampler(textureUnitNumber, id_);
+    currentlyBoundTextureUnit_ = textureUnitNumber;
```
petersteneteg: why is currentlyBoundTextureUnit_ needed? can it be removed?

modules/opengl/include/modules/opengl/texture/samplerobject.h
```diff
@@ -0,0 +1,64 @@
...
+
+    /// <summary>
+    /// Texture unit number is ONLY the number, such as 0, 1,4 or 5. Not GL_TEXTURE0. This is different from how textures are bound.
+    /// </summary>
+    /// <param name="textureUnitNumber"></param>
+    void bind(GLuint textureUnitNumber);
+    void setMinFilterMode(GLenum type);
+    void setMagFilterMode(GLenum type);
+    void setFilterModeAll(GLenum type);
```
petersteneteg: should add overloads for https://github.com/inviwo/inviwo/blob/614c8793c8b1ac9bdd4444a90d97217d9201a3fb/include/inviwo/core/datastructures/image/imagetypes.h#L49-L54

modules/opengl/src/texture/samplerobject.cpp
```diff
@@ -0,0 +1,82 @@
...
+#include <modules/opengl/texture/samplerobject.h>
+
+namespace inviwo {
+
+SamplerObject::SamplerObject() : currentlyBoundTextureUnit_{0} { glGenSamplers(1, &id_); }
+
+SamplerObject::~SamplerObject() {
+    /*if (id_ != 0) {
+        glDeleteSamplers(1, &id_);
+    }*/ // Causes "unable to upload renderui.vert"
```
petersteneteg: If this fails we are doing something wrong which need to be fixed...

modules/volume/src/processors/atlasboundary.cpp
```diff
@@ -0,0 +1,97 @@
...
+#include <inviwo/volume/processors/atlasboundary.h>
+#include <inviwo/volume/algorithm/volumemap.h>
+
+namespace inviwo {
+
+// The Class Identifier has to be globally unique. Use a reverse DNS naming scheme
+const ProcessorInfo AtlasBoundary::processorInfo_{
+    "org.inviwo.atlasboundary",  // Class identifier
+    "Atlas Boundary",            // Display name
+    "Volume Rendering",          // Category
```
petersteneteg: Don't think Rendering is the right category

modules/volume/src/processors/atlasboundary.cpp
```diff
@@ -0,0 +1,97 @@
...
+            sourceIndices.push_back(0);
+            destinationIndices.push_back(0);
+        }
+
+        // Remap using vectors and cloned volume
+        auto volume = volumeInport_.getData();
+        volume_ = std::shared_ptr<Volume>(volume->clone());
+        util::remap(*volume_, sourceIndices, destinationIndices, 0, true);
+        // Set volume properties
+        volume_->setInterpolation(InterpolationType::Nearest);
```
petersteneteg: remove?

modules/volume/src/processors/atlasboundary.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+        sourceIndices.insert(sourceIndices.end(), selected.begin(), selected.end());
+        sourceIndices.insert(sourceIndices.end(), filtered.begin(), filtered.end());
+        sourceIndices.insert(sourceIndices.end(), highlighted.begin(), highlighted.end());
+        destinationIndices.insert(destinationIndices.end(), selectedDestination.begin(),
+                                  selectedDestination.end());
+        destinationIndices.insert(destinationIndices.end(), filteredDestination.begin(),
+                                  filteredDestination.end());
+        destinationIndices.insert(destinationIndices.end(), highlightedDestination.begin(),
+                                  highlightedDestination.end());
```
petersteneteg: Could use std::fill_n tree time here instead of creating 3 temp vectors...

modules/volume/src/processors/atlasboundary.cpp
```diff
@@ -0,0 +1,97 @@
...
+        highlighted -= filtered;
+        selected -= highlighted;
+        selected -= filtered;
+        std::vector<int> selectedDestination(selected.toVector().size(), 1);
+        std::vector<int> filteredDestination(filtered.toVector().size(), 2);
+        std::vector<int> highlightedDestination(highlighted.toVector().size(), 3);
+
+        // Append to B&L to vectors
+        std::vector<int> sourceIndices;
+        std::vector<int> destinationIndices;
```
petersteneteg: You might want to keep this 2 vectors around as members to avoid having to create them each time...

petersteneteg (changes requested):

modules/basegl/CMakeLists.txt
RasmusHogslatt: Yes, I should have removed this one from the pull request. Atlasisosurfacerenderer is purely for debugging purposes.

modules/basegl/include/modules/basegl/processors/raycasting/atlasvolumeraycaster.h
RasmusHogslatt: Yes, that sounds better.

modules/basegl/include/modules/basegl/shadercomponents/atlasisosurfacecomponent.h
RasmusHogslatt: I actually need both unless I got something wrong. Creating only one linear would automatically take the "spot" of the regular texture sampler in the shader. So, as long as a samplerobject exists for a texture unit, that sampler object will take its spot in the shader. Thus, having both as in the current implementation, the textures parameters will be overwritten by nearestSampler, and linearSampler will get the next spot. The shader thus becomes
`uniform sampler3D nearestSampler; // Without samplerobjects, this would have had the texture's parameters
uniform sampler3D linearSampler;`

modules/basegl/src/processors/raycasting/atlasvolumeraycaster.cpp
RasmusHogslatt: I'm a bit confused. Suggestion marks removal of the atlas_, which is not what I have worked on. Could you explain what should be an option here?

modules/basegl/src/shadercomponents/atlasisosurfacecomponent.cpp
RasmusHogslatt: useAtlasBoundary_ is a boolCompositeProperty with the other properties added to it. I tried adding everything here before, but then I got duplicates of all properties. Martin suggested to put all properties in a composite one for clarity.

modules/basegl/src/shadercomponents/raycastingcomponent.cpp
RasmusHogslatt: Okay, do we still want this option for this processor? If so, should I copy gradientCentralDiff and apply scaling to it, or should there be an overloaded version for all of the gradient calculations? Next comment clarified this.

modules/opengl/src/texture/samplerobject.cpp
RasmusHogslatt: Yes, maybe we can have a look at it during next meeting.

modules/opengl/src/texture/samplerobject.cpp
RasmusHogslatt: Not needed as of right now, but I think it might be useful for debugging to figure out which unit you are working on. Mostly for debugging though. Thus it can be removed. Though, every time you bind, you need to specify which unit to bind to. If you have to bind lots of times, it might be nicer to have what unit it's bound to rather than setting it manually in every bind call.

modules/volume/src/processors/atlasboundary.cpp
RasmusHogslatt: True, will check for a more suitable catergory.

modules/basegl/include/modules/basegl/shadercomponents/atlasisosurfacecomponent.h
petersteneteg: I think this is since you don't assign specific texture units for the sampler but just pass in 0 as texture unit. You should assign a dedicated texture unit for the sampler then I should not mess with existing ones. You also need to assign the uniform associated with that sampler to that new texture unit

modules/basegl/src/processors/raycasting/atlasvolumeraycaster.cpp
petersteneteg: No the altas is still there It's just that the "{" at then end is remove and moved to the end of the atlasIso line.

modules/basegl/src/shadercomponents/atlasisosurfacecomponent.cpp
petersteneteg: ahh ok

modules/basegl/src/shadercomponents/raycastingcomponent.cpp
petersteneteg: Guess we can have a specific one for the surface here.

petersteneteg: you can just create a copy  of the volumeParams in the shader and update the textureSpaceGradientSpacing to the appropriate value

modules/opengl/src/texture/samplerobject.cpp
petersteneteg: yes

modules/opengl/src/texture/samplerobject.cpp
petersteneteg: In out current approach we bind all texture unit every time we change shader anyway so I don't think that is a problem.. There is a limited number of texture units so you have to take care of tracking and reusing them

modules/volume/src/processors/atlasboundary.cpp
RasmusHogslatt: Changed to "Volume Operation", same as for voronoi segmentation.

modules/basegl/include/modules/basegl/shadercomponents/segmentsurfacecomponent.h
```diff
@@ -0,0 +1,70 @@
...
+    virtual std::vector<Segment> getSegments() override;
+
+private:
+    std::string name_;
+    BoolCompositeProperty useAtlasBoundary_;
+    BoolProperty applyBoundaryLight_;
+    SelectionColorProperty showHighlighted_;
+    SelectionColorProperty showSelected_;
+    SelectionColorProperty showFiltered_;
+    SamplerObject nearestSampler_;
```
petersteneteg: This one is not used anymore I think and should be removed

modules/basegl/src/shadercomponents/segmentsurfacecomponent.cpp
```diff
@@ -0,0 +1,154 @@
...
+    , linearSampler_{}
+    , textureSpaceGradientSpacingScale_{"gradientScaling", "Scale gradient spacing", 1.0f, 0.0f,
+                                        5.0f}
+    , atlas_{atlas} {
+
+    useAtlasBoundary_.addProperties(applyBoundaryLight_, textureSpaceGradientSpacingScale_,
+                                    showHighlighted_, showFiltered_, showSelected_);
+
+    nearestSampler_.setFilterModeAll(InterpolationType::Nearest);
+    nearestSampler_.setWrapModeAll(Wrapping::Repeat);
```
petersteneteg: Remove

modules/basegl/src/shadercomponents/segmentsurfacecomponent.cpp
```diff
@@ -0,0 +1,154 @@
...
+    linearSampler_.setFilterModeAll(InterpolationType::Linear);
+    linearSampler_.setWrapModeAll(Wrapping::Repeat);
```
petersteneteg: Should set these as constructor arguments above

modules/basegl/src/shadercomponents/segmentsurfacecomponent.cpp
```diff
@@ -0,0 +1,154 @@
...
+    {
+        vec3 atlasGradient = vec3(1);
+        atlasGradient = normalize(gradientCentralDiff(vec4(0),
+            linearAtlas, scaledAtlasParameters, samplePosition, channel));
+        vec3 worldSpacePosition = 
+            (atlasParameters.textureToWorldNormalMatrix * vec4(samplePosition, 1.0)).xyz;
+        boundaryColor.rgb = APPLY_LIGHTING(lighting, boundaryColor.rgb, boundaryColor.rgb,
+            vec3(1.0), worldSpacePosition, -atlasGradient, cameraDir);  
+    }   
+    #endif
```
petersteneteg: The opening "{" should be on the same line, according to our style guide

modules/opengl/include/modules/opengl/texture/samplerobject.h
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * Sampler objects store parameters that are not specific to textures. Thus, they store filter
+ * modes, wrapping mode but not texture width, height, swizzlemask etc. Binding a sampler object to
+ * a texture unit takes a texture's uniform location in the shader, and multiple different sampler
+ * objects can be assigned to the same texture unit without having to copy the texture data.
+ */
+class IVW_MODULE_OPENGL_API SamplerObject {
+public:
+    SamplerObject();
```
petersteneteg: Add constructor overloads for the InterpolationType and Wrapping

modules/opengl/include/modules/opengl/texture/samplerobject.h
```diff
@@ -0,0 +1,79 @@
...
+    void setWrapModeAll(GLenum wrap);
+    void setWrapMode_S(Wrapping wrap);
+    void setWrapMode_T(Wrapping wrap);
+    void setWrapMode_R(Wrapping wrap);
+    void setWrapModeAll(Wrapping wrap);
+    GLint getID() const;
+
+private:
+    GLuint id_;
+    GLuint currentlyBoundTextureUnit_;
```
petersteneteg: We can remove currentlyBoundTextureUnit_ now.

modules/opengl/include/modules/opengl/texture/samplerobject.h
```diff
@@ -0,0 +1,79 @@
...
+class IVW_MODULE_OPENGL_API SamplerObject {
+public:
+    SamplerObject();
+    SamplerObject(const SamplerObject&) = delete;
+    SamplerObject(SamplerObject&&) noexcept;
+    SamplerObject& operator=(const SamplerObject&) = delete;
+    SamplerObject& operator=(SamplerObject&&) noexcept;
+    ~SamplerObject();
+
+    /// <summary>
```
petersteneteg: We don't use this style (html) for comments

modules/opengl/include/modules/opengl/texture/samplerobject.h
```diff
@@ -0,0 +1,79 @@
...
+#include <modules/opengl/inviwoopengl.h>
+#include <inviwo/core/datastructures/image/imagetypes.h>
+
+namespace inviwo {
+
+/**
+ * Sampler objects store parameters that are not specific to textures. Thus, they store filter
+ * modes, wrapping mode but not texture width, height, swizzlemask etc. Binding a sampler object to
+ * a texture unit takes a texture's uniform location in the shader, and multiple different sampler
+ * objects can be assigned to the same texture unit without having to copy the texture data.
```
petersteneteg: A texture unit should only have one or zero sampler object and one texture bound at any time. 
Might att a link to the bind(sampler object function) in texture unit

petersteneteg (changes requested):
Looking nice now, found a few minor issues to address before merging.

modules/basegl/src/shadercomponents/segmentsurfacecomponent.cpp
RasmusHogslatt: Add filter and wrap mode to this `SegmentSurfaceComponent(VolumeInport& atlas, InterpolationType type, Wrapping wrap)` or add these line to initializer list? Assume the former of the two.

PR #1318, opened 2022-07-06
BaseGL: Instance rendering

petersteneteg:
Changes proposed in this PR:
 * Adds a Instance Renderer processor that can render a number of instances of a geometry in a naive way. A dynamic number of ports of input vectors can be provided to drive how the different instances should be rendered. 
 * A vector to buffer processor
 * A point generation processor that can generate points on a grid or randomly. 

This has been tested on: ...

> 


include/inviwo/core/properties/ordinalproperty.h
```diff
@@ -224,6 +224,54 @@ IVW_CORE_API OrdinalPropertyState<vec3> ordinalLight(
     const vec3& pos, float min = -100.0, float max = 100.0,
     InvalidationLevel invalidationLevel = InvalidationLevel::InvalidOutput);
 
+template <typename T, typename U>
+OrdinalPropertyState<T> ordinalSymmetricVector(
```
ResearchDaniel, line 271: Add documentation.
Is it possible to make U a dependent type of T (`T::value_type`)
Why exclude ConstraintBehavior and increment? I guess this has to do something with intent of Vector, but it is a bit unclear. Also the reason for "symmetric" is unclear since the constraints are set to Ignore anyway. `ordinalUnbounded` might be clearer?

include/inviwo/core/properties/ordinalproperty.h
```diff
@@ -224,6 +224,54 @@ IVW_CORE_API OrdinalPropertyState<vec3> ordinalLight(
...
+            ConstraintBehavior::Ignore,
+            decltype(val){minMax},
+            ConstraintBehavior::Ignore,
+            decltype(val){0.1},
+            invalidationLevel,
+            semantics};
+}
+
+template <typename T = size_t, typename U = T>
+OrdinalPropertyState<T> ordinalCount(
```
ResearchDaniel, line 295: Add documentation. Remove U, it is not be possible to have a different type for the bounds anyway.

include/inviwo/core/properties/ordinalproperty.h
```diff
@@ -224,6 +224,54 @@ IVW_CORE_API OrdinalPropertyState<vec3> ordinalLight(
...
+            ConstraintBehavior::Immutable,
+            decltype(val){max},
+            ConstraintBehavior::Ignore,
+            decltype(val){1},
+            invalidationLevel,
+            semantics};
+}
+
+template <typename T = double, typename U = T>
+OrdinalPropertyState<T> ordinalLength(
```
ResearchDaniel, line 319: Remove U, it is not be possible to have a different type for the bounds anyway.
In hindsight it would probably have been better to name these properties 
ColorProperty, LengthProperty, CountProperty, LengthProperty and so on.

modules/base/include/modules/base/algorithm/pointgeneration.h
```diff
@@ -0,0 +1,144 @@
...
+        } else {
+            std::random_device rd;
+            mt.seed(rd());
+        }
+        return mt;
+    };
+
+    std::visit(
+        util::overloaded{
+            [&](const GridPointGeneration& opts) {
```
ResearchDaniel: The approach with using std::visit and overloaded makes the code more complicated and unnecessarily separates the algorithm from its settings.

Any particular reason for not using the strategy pattern with an abstract base class for point generators?

modules/base/include/modules/base/algorithm/pointgeneration.h
```diff
@@ -0,0 +1,144 @@
...
+#include <inviwo/core/util/stdextensions.h>
+#include <inviwo/core/util/volumeramutils.h>
+#include <optional>
+#include <random>
+
+namespace inviwo {
+
+namespace util {
+
+struct GridPointGeneration {
```
ResearchDaniel: These all assume 3D points. We might want to either loosen the restriction using a template type or append 3D:
GridPointGeneration3D

modules/base/include/modules/base/algorithm/pointgeneration.h
```diff
@@ -0,0 +1,144 @@
...
+
+struct SpherePointGeneration {
+    dvec3 center = dvec3{0.0};
+    dvec2 radius = dvec2{0.0, 1.0};
+    size_t nPoints = 25;
+    std::optional<size_t> seed = std::nullopt;
+};
+
+template <typename T, typename OutIt>
+void generatePoints(
```
ResearchDaniel: Could point to example usage here.
Also, this assumes 3D points. `generate3DPoints`

modules/base/include/modules/base/processors/pointgenerationprocessor.h
```diff
@@ -0,0 +1,91 @@
...
+#include <inviwo/core/properties/ordinalproperty.h>
+#include <inviwo/core/properties/boolcompositeproperty.h>
+#include <inviwo/core/ports/imageport.h>
+#include <inviwo/core/util/glmvec.h>
+
+#include <vector>
+
+namespace inviwo {
+
+class IVW_MODULE_BASE_API PointGenerationProcessor : public Processor {
```
ResearchDaniel: Document. 
Do we want to name it `Point3DGenerationProcessor `

modules/base/include/modules/base/processors/pointgenerationprocessor.h
```diff
@@ -0,0 +1,91 @@
...
+    virtual void process() override;
+
+    virtual const ProcessorInfo getProcessorInfo() const override;
+    static const ProcessorInfo processorInfo_;
+
+private:
+    DataOutport<std::vector<vec3>> outport_;
+
+    struct Grid {
+        FloatVec3Property a1;
```
ResearchDaniel, line 58: BasisProperty

modules/base/include/modules/base/processors/pointgenerationprocessor.h
```diff
@@ -0,0 +1,91 @@
...
+        FloatVec3Property a2;
+        FloatVec3Property a3;
+        FloatVec3Property offset;
+        IntSize3Property nPoints;
+        FloatVec3Property jitter;
+        IntSizeTProperty seed;
+    };
+
+    struct RandomBox {
+        FloatVec3Property a1;
```
ResearchDaniel, line 69: BasisProperty

modules/base/include/modules/base/algorithm/pointgeneration.h
```diff
@@ -0,0 +1,144 @@
...
+namespace util {
+
+struct GridPointGeneration {
+    dmat4 bases = dmat4(1.0);
+    size3_t nPoints = {5, 5, 5};
+    std::optional<dvec3> jitter = std::nullopt;
+    std::optional<size_t> seed = std::nullopt;
+};
+
+struct RandomPointGeneration {
```
ResearchDaniel: Possible to specify distribution?
Otherwise, `RandomUniformPointGeneration `

modules/base/include/modules/base/algorithm/pointgeneration.h
```diff
@@ -0,0 +1,144 @@
...
+    dmat4 bases = dmat4(1.0);
+    size_t nPoints = 25;
+    std::optional<size_t> seed = std::nullopt;
+};
+
+struct SpherePointGeneration {
```
ResearchDaniel: Possible to specify distribution?
Otherwise, `SphereUniformPointGeneration `

modules/base/include/modules/base/processors/vectortobuffer.h
```diff
@@ -0,0 +1,105 @@
...
+    VectorToBuffer()
+        : Processor()
+        , inport_{"inport"}
+        , outport_{"outport"}
+        , bufferUsage_{"bufferUsage",
+                       "Buffer Usage",
+                       {BufferUsage::Static, BufferUsage::Dynamic},
+                       0}
+        , bufferTarget_{
+              "bufferTarget", "Buffer Target", {BufferTarget::Data, BufferTarget::Index}, 0} {
```
ResearchDaniel: BufferTarget::Index should only be available for integer types?
Also, would make it the default for integer types.

modules/base/include/modules/base/processors/vectortobuffer.h
```diff
@@ -0,0 +1,105 @@
...
+template <typename T>
+const ProcessorInfo VectorToBuffer<T>::getProcessorInfo() const {
+    return ProcessorTraits<VectorToBuffer<T>>::getProcessorInfo();
+}
+
+template <typename T>
+struct ProcessorTraits<VectorToBuffer<T>> {
+    static ProcessorInfo getProcessorInfo() {
+        return {
+            fmt::format("org.inviwo.{}.ToBuffer",
```
ResearchDaniel, line 101: In case of IndexBuffer, we probably want to name it `VectorToIndexBuffer`

modules/base/src/basemodule.cpp
```diff
@@ -170,6 +173,10 @@ BaseModule::BaseModule(InviwoApplication* app) : InviwoModule(app, "Base") {
     registerProcessor<TrianglesToWireframe>();
     registerProcessor<TransformMesh>();
     registerProcessor<TransformVolume>();
+    registerProcessor<VectorToBuffer<float>>();
```
ResearchDaniel, line 177: Missing `registerProcessor<VectorToBuffer<unsigned int>>();` for index buffers.

modules/basegl/include/modules/basegl/processors/instancerenderer.h
```diff
@@ -0,0 +1,117 @@
...
+    ImageOutport outport_;
+
+    ListProperty ports_;
+    std::vector<detail::DynPortManager> vecPorts_;
+
+    CameraProperty camera_;
+    CameraTrackball trackball_;
+    SimpleLightingProperty lightingProperty_;
+
+    std::array<StringProperty, 5> transforms_;
```
ResearchDaniel, line 109: `vertexShaderTransforms_`

modules/basegl/include/modules/basegl/processors/instancerenderer.h
```diff
@@ -0,0 +1,117 @@
...
+
+#include <modules/opengl/shader/shader.h>
+#include <modules/opengl/shader/shadersegment.h>
+
+#include <vector>
+#include <tuple>
+
+namespace inviwo {
+
+class InstanceRenderer;
```
ResearchDaniel, line 57: The InstanceRenderer could be something we would like to extract to the rendering folder.

modules/basegl/include/modules/basegl/processors/instancerenderer.h
```diff
@@ -0,0 +1,117 @@
...
+    InstanceRenderer* renderer;
+    std::unique_ptr<Inport> port;
+    std::function<size_t()> size;
+    std::function<void(Shader&, size_t)> set;
+    std::function<void(ShaderObject&)> addUniform;
+};
+
+};  // namespace detail
+
+class IVW_MODULE_BASEGL_API InstanceRenderer : public Processor, public PropertyOwnerObserver {
```
ResearchDaniel, line 81: Document

If I understand the code correctly it renders the mesh input once per added port/buffer?
I would expect an instance renderer to render the mesh using `glDrawArrayInstanced `or `glDrawElementsInstanced`
https://learnopengl.com/Advanced-OpenGL/Instancing

modules/basegl/include/modules/basegl/processors/instancerenderer.h
```diff
@@ -0,0 +1,117 @@
...
+    std::vector<detail::DynPortManager> vecPorts_;
+
+    CameraProperty camera_;
+    CameraTrackball trackball_;
+    SimpleLightingProperty lightingProperty_;
+
+    std::array<StringProperty, 5> transforms_;
+    StringProperty setupVert_;
+
+    std::shared_ptr<StringShaderResource> vert_;
```
ResearchDaniel, line 112: Also geometry shader?

modules/basegl/src/processors/instancerenderer.cpp
```diff
@@ -0,0 +1,394 @@
...
+        std::swap(that.addUniform, addUniform);
+    }
+    return *this;
+}
+DynPortManager::~DynPortManager() {
+    if (port) renderer->removePort(port.get());
+}
+
+template <typename T>
+inline constexpr std::string_view glslTypeName() {
```
ResearchDaniel: Could be extracted to openglutils.h

modules/basegl/src/processors/instancerenderer.cpp
```diff
@@ -0,0 +1,394 @@
...
+            fmt::format("uniform {0} {1} = {0}(0);", detail::glslTypeName<T>(), u->get())});
+    };
+
+    return DynPortManager{theRenderer, std::move(port), std::move(size), std::move(set),
+                          std::move(addUniform)};
+}
+
+}  // namespace detail
+
+InstanceRenderer::InstanceRenderer()
```
ResearchDaniel, line 216: Most of this would also be valid for a general renderer processor, where one can create custom shaders?

modules/basegl/src/processors/instancerenderer.cpp
```diff
@@ -0,0 +1,394 @@
...
+uniform LightParameters lighting;
+uniform CameraParameters camera;
+
+in vec4 color;
+in vec3 normal;
+in vec3 texCoord;
+in vec4 worldPosition;
+flat in vec4 picking;
+ 
+#pragma IVW_SHADER_SEGMENT_PLACEHOLDER_SETUP_FRAG
```
ResearchDaniel: Unused?

modules/basegl/src/processors/instancerenderer.cpp
```diff
@@ -0,0 +1,394 @@
...
+out vec4 color;
+out vec3 normal;
+out vec3 texCoord;
+out vec4 worldPosition;
+flat out vec4 picking;
+ 
+#pragma IVW_SHADER_SEGMENT_PLACEHOLDER_SETUP_VERT
+ 
+void main() {
+#pragma IVW_SHADER_SEGMENT_PLACEHOLDER_TRANSFORMS
```
ResearchDaniel, line 119: `IVW_SHADER_SEGMENT_PLACEHOLDER_TRANSFORMS_VERT`

modules/basegl/src/processors/instancerenderer.cpp
```diff
@@ -0,0 +1,394 @@
...
+ 
+#pragma IVW_SHADER_SEGMENT_PLACEHOLDER_SETUP_FRAG
+ 
+void main() {
+    // Prevent invisible fragments from blocking other objects (e.g., depth/picking)
+    if (color.a == 0) { discard; }
+
+    vec4 fragColor = color;
+    vec3 toCameraDir = camera.position - worldPosition.xyz;
+
```
ResearchDaniel, line 143: Should there be possibilities to change the fragment shader?

modules/basegl/src/processors/instancerenderer.cpp
```diff
@@ -0,0 +1,394 @@
...
+    auto minIt = std::min_element(vecPorts_.begin(), vecPorts_.end(),
+                                  [](const auto& a, const auto& b) { return a.size() < b.size(); });
+    const size_t min = minIt->size();
+
+    for (size_t i = 0; i < min; ++i) {
+        std::for_each(vecPorts_.begin(), vecPorts_.end(),
+                      [&](auto& port) { port.set(shader_, i); });
+
+        drawer.draw();
```
ResearchDaniel, line 373: No glDrawArrayInstanced?

ResearchDaniel (changes requested):

include/inviwo/core/properties/ordinalproperty.h
petersteneteg: My thinking was that the symmetric part was for the fact the the min == -max
Having T and U being independent make it a bit easier since it will allow some more conversions...

include/inviwo/core/properties/ordinalproperty.h
petersteneteg: this way it it is possible to initiate the bound with an int even if T is a size_t if both was T you have to write `OrdinalCount(size3_t{5,5,5}, size_t{100})` instead of just `OrdinalCount(size3_t{5,5,5}, 100)` which I found more intuitive. I can change it if you like, but when using it I would bet that everyone will use the later version and get an error which is a bit cryptic, and then have to change to the first version, this way the second version just works.

include/inviwo/core/properties/ordinalproperty.h
petersteneteg: Se above about U, But they are not properties, just helper functions to configure properties for various use cases. Im not sure we want to create a huge amount if different properties for all these cases...

include/inviwo/core/properties/ordinalproperty.h
ResearchDaniel: > Having T and U being independent make it a bit easier since it will allow some more conversions...

Are you sure? The OrdinalProperty does not differentiate between the types. 
Perhaps there should also be an origin to do the symmetry arround, defaulting to 0

include/inviwo/core/properties/ordinalproperty.h
ResearchDaniel: I see, I would prefer avoiding cryptic error messages :)

include/inviwo/core/properties/ordinalproperty.h
ResearchDaniel: Ah, true

include/inviwo/core/properties/ordinalproperty.h
ResearchDaniel: Perhaps you could add the explanation above as a comment

modules/base/include/modules/base/algorithm/pointgeneration.h
petersteneteg: I very much disagree that this code is more complicated then using a strategy pattern and an abstract base class. For me this is a much cleaner and simpler solution.

modules/base/include/modules/base/algorithm/pointgeneration.h
petersteneteg: I don't think we have that much needed for the other cases. When we need that we can do that...

modules/basegl/include/modules/basegl/processors/instancerenderer.h
petersteneteg: basically it does something like
for(auto uniforms : zip(all ports)){
     set uniforms
    draw
}

the ports ports are for example a std::vector<vec3> port with positions for example. then the we would draw one instance of the geometry at each position. Given how you have configured the transforms.

modules/basegl/include/modules/basegl/processors/instancerenderer.h
petersteneteg: Not so far...

modules/basegl/include/modules/basegl/processors/instancerenderer.h
petersteneteg: Could probably use glDrawArrayInstanced or something, But have not had the time to do that. For the use case I wanted I don't need that optimization... but you are welcome to implement it.

modules/basegl/src/processors/instancerenderer.cpp
petersteneteg: yes

modules/basegl/src/processors/instancerenderer.cpp
petersteneteg: yes

modules/basegl/src/processors/instancerenderer.cpp
petersteneteg: maybe?

modules/basegl/src/processors/instancerenderer.cpp
petersteneteg: Did not need that so far, though about it but did not see a need so far...

modules/basegl/src/processors/instancerenderer.cpp
petersteneteg: should be possible to change into that I think, have not had the time/need for it so far...

modules/base/include/modules/base/algorithm/pointgeneration.h
martinfalk: I disagree. At least "true" 2D points might be an option without a z coordinate.

modules/basegl/src/processors/instancerenderer.cpp
martinfalk: Might go either way. This way it supports generic meshes. And since the uniforms have to be set before each draw call anyway...

include/inviwo/core/properties/ordinalproperty.h
petersteneteg: Point of using these convenience function is that they should be fast and easy to use, i.e. not more options then needed... otherwise one can just use the regular constructor with all the settings... 

Yes, I got exactly that compiler error as I said above... point here is that the min/max argument is just a scalar not a vector... i.e. I assume the min.x == min.y == min.z == min... to make it as easy to use as possible...

modules/base/include/modules/base/algorithm/pointgeneration.h
petersteneteg: renamed

modules/base/include/modules/base/algorithm/pointgeneration.h
petersteneteg: This can be used for any type, if needed. The OutIt can take any dimensions,..

modules/base/include/modules/base/processors/pointgenerationprocessor.h
petersteneteg: forgot about that one

modules/base/include/modules/base/algorithm/pointgeneration.h
petersteneteg: We could add that setting if the need arises

modules/base/include/modules/base/algorithm/pointgeneration.h
petersteneteg: We could add that setting if the need arises

modules/base/include/modules/base/processors/pointgenerationprocessor.h
petersteneteg: but the basis property has a large set of other stuff... it does not really fit that well

modules/base/include/modules/base/processors/vectortobuffer.h
petersteneteg: That is not possible, since not all integer ones will be IndexBuffers and the setting is at runtime.

modules/base/include/modules/base/processors/pointgenerationprocessor.h
petersteneteg: see above

modules/basegl/include/modules/basegl/processors/instancerenderer.h
petersteneteg: maybe, it is fairly processor connected since I use dynamic ports etc, so Im not sure if it makes sense.

modules/base/include/modules/base/processors/pointgenerationprocessor.h
petersteneteg: In the next PR....

modules/base/include/modules/base/processors/pointgenerationprocessor.h
ResearchDaniel: Then why not FloatMat3Property

modules/basegl/src/processors/instancerenderer.cpp
ResearchDaniel: Forgot this one?

ResearchDaniel (approved):

petersteneteg:
I removed VERT from the other one instead, since I don't change the fragment.
I use 3 vectors since that is how I think about... for me it is three basis vectors not a transform matrix, but that is of course just my habits... Would have been nice to use the basisProperty, but that comes with a bunch of other options related to input volumes. We most use that in data sources of some volume transforms I think.

PR #1326, opened 2022-08-19
Python3: Expose data readers and MeshSeqeuence

ResearchDaniel:
Rebase of  #1308

### DIFF
--- a/modules/python3/bindings/include/inviwopy/pydatareaders.h
+++ b/modules/python3/bindings/include/inviwopy/pydatareaders.h
@@ -1,0 +1,41 @@
+/*********************************************************************************
+ *
+ * Inviwo - Interactive Visualization Workshop
+ *
+ * Copyright (c) 2017-2021 Inviwo Foundation
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ *********************************************************************************/
+
+#pragma once
+
+#include <warn/push>
+#include <warn/ignore/shadow>
+#include <pybind11/pybind11.h>
+#include <warn/pop>
+
+namespace inviwo {
+
+void exposeDataReaders(pybind11::module& m);
+
+}  // namespace inviwo
--- a/modules/python3/bindings/include/inviwopy/pydatareaders.h
+++ b/modules/python3/bindings/include/inviwopy/pydatareaders.h
@@ -2,7 +2,7 @@
  *
  * Inviwo - Interactive Visualization Workshop
  *
- * Copyright (c) 2017-2021 Inviwo Foundation
+ * Copyright (c) 2022 Inviwo Foundation
  * All rights reserved.
  *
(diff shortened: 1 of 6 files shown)


modules/python3/bindings/src/pydatareaders.cpp
```diff
@@ -0,0 +1,167 @@
...
+#include <inviwo/core/io/datareader.h>
+#include <inviwo/core/io/datareaderfactory.h>
+#include <inviwo/core/datastructures/geometry/mesh.h>
+#include <inviwo/core/datastructures/image/image.h>
+#include <inviwo/core/datastructures/volume/volume.h>
+
+namespace inviwo {
+
+// Allow overriding virtual functions from Python
+class DataReaderTrampoline : public DataReader, public pybind11::trampoline_self_life_support {
```
petersteneteg: is there any usecase for this one?

petersteneteg (approved):

modules/python3/bindings/src/pydatareaders.cpp
ResearchDaniel: To be able to create new types of readers in Python.
Have not tested it myself though

petersteneteg: I don't think that would work

ResearchDaniel: https://github.com/inviwo/inviwo/pull/1308#discussion_r889686791
:)

petersteneteg: haha nice :) I I will work for the typed ones... but for the base class I don't think there is any reason one would extend that from python and call from C++ since there will be no possibility to actually invoke the data loading...

petersteneteg:
This needs an update with rebase... and a copyright update to 2022 for the new files...

modules/python3/bindings/src/pydatareaders.cpp
```diff
@@ -0,0 +1,167 @@
+/*********************************************************************************
+ *
+ * Inviwo - Interactive Visualization Workshop
+ *
+ * Copyright (c) 2017-2021 Inviwo Foundation
```
petersteneteg: new file from 2017?

modules/python3/bindings/include/inviwopy/pydatareaders.h
```diff
@@ -0,0 +1,41 @@
+/*********************************************************************************
+ *
+ * Inviwo - Interactive Visualization Workshop
+ *
+ * Copyright (c) 2017-2021 Inviwo Foundation
```
petersteneteg: 2017?

petersteneteg (dismissed):

petersteneteg (dismissed):

modules/python3/bindings/src/pydatareaders.cpp
ResearchDaniel: Ok, will remove the base class. Just need to verify that it does not break anything

ResearchDaniel: Worked

modules/python3/bindings/src/pydatareaders.cpp
```diff
@@ -117,26 +98,18 @@ void exposeFactoryReaderType(pybind11::class_<DataReaderFactory>& r, std::string
...
     namespace py = pybind11;
 
-    py::class_<DataReader, DataReaderTrampoline>(m, "DataReader")
-        .def("clone", &DataReader::clone)
-        .def_property_readonly("extensions", &DataReader::getExtensions,
-                               py::return_value_policy::reference_internal)
-        .def("addExtension", &DataReader::addExtension)
-        .def("setOption", &DataReader::setOption)
-        .def("getOption", &DataReader::getOption);
-
```
petersteneteg: Why was the base class removed?

petersteneteg (changes requested):

modules/python3/bindings/src/pydatareaders.cpp
ResearchDaniel: I was assuming that it will not be used as the base class functionality are exposed for DataReaderType anyway. Perhaps there is such as use case though?

petersteneteg: How is the base class exposed in the DataReaderType? You just removed exactly that?

petersteneteg: Otherwise those functions will not be exposed in the typed readers

ResearchDaniel:
@petersteneteg I think that this PR should be possible to merge as soon as the checks passes?

petersteneteg:
yes think so

petersteneteg:
No right, you had removed the DataReader part. Could you add that back?

modules/python3/bindings/src/pydatareaders.cpp
petersteneteg: This part should still be in there? if not pls explain, it is only the DataReaderTrampoline that should be removed.

petersteneteg (changes requested):

ResearchDaniel:
@petersteneteg I think that [the ](https://github.com/inviwo/inviwo/pull/1326#discussion_r984536607) has been addressed in a newer commit

petersteneteg:
then you have to explain where I see no mention of `py::class_<DataReader>(m, "DataReader")` anywhere in the PR

ResearchDaniel:
> then you have to explain where I see no mention of py::class_<DataReader>(m, "DataReader") anywhere in the PR

Missed that. I now expose the DataReader without possibility to define a new one in Python

modules/python3/bindings/src/pydatareaders.cpp
```diff
@@ -97,7 +97,15 @@ void exposeFactoryReaderType(pybind11::class_<DataReaderFactory>& r, std::string
...
+    py::class_<DataReader>(m, "DataReader")
+        .def("clone", &DataReader::clone)
+        .def_property_readonly("extensions", &DataReader::getExtensions,
+                               py::return_value_policy::reference_internal)
+        .def("addExtension", &DataReader::addExtension)
+        .def("setOption", &DataReader::setOption)
+        .def("getOption", &DataReader::getOption);
+        
     // https://pybind11.readthedocs.io/en/stable/advanced/classes.html#binding-classes-with-template-parameters
     py::class_<DataReaderType<Image>, DataReaderTypeTrampoline<Image>>(m, "ImageDataReader")
```
petersteneteg: this one should "derive" from DataReader

petersteneteg (changes requested):

modules/python3/bindings/src/pydatareaders.cpp
ResearchDaniel: Fixed

petersteneteg (approved):

PR #1328, opened 2022-08-24
flags: Fix build errors on VS 2022 17.3.1

ResearchDaniel:
Fixes build errors on VS 2022 17.3.1

### DIFF
--- a/ext/flags/include/flags/flags.h
+++ b/ext/flags/include/flags/flags.h
@@ -85,7 +85,9 @@ public:
     flags(enum_type e, Args... args) noexcept : flags{e, args...} {}
 
     template <class FwIter>
-    flags(FwIter b, FwIter e, typename convertible<decltype(*b)>::type = nullptr) noexcept(
+    flags(FwIter b, FwIter e,
+          typename convertible<decltype(*std::declval<FwIter>())>::type =
+              nullptr) noexcept(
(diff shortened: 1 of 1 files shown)


ext/flags/include/flags/flags.h
```diff
@@ -85,7 +85,9 @@ struct flags {
     flags(enum_type e, Args... args) noexcept : flags{e, args...} {}
 
     template <class FwIter>
-    flags(FwIter b, FwIter e, typename convertible<decltype(*b)>::type = nullptr) noexcept(
+    flags(FwIter b, FwIter e,
+          typename convertible<decltype(*std::declval<FwIter>())>::type =
+              nullptr) noexcept(
```
petersteneteg, line 90:
```
    typename <class FwIter,   typename convertible<decltype(*std::declval<FwIter>())>::type =
              nullptr>
    flags(FwIter b, FwIter e) noexcept(
```

I would use that more canonical version... It also works in 17.3.1

petersteneteg (approved):
The problem is that the SML code will fail after this fix, and that was much harder to fix, at least in a nice way...But im working on it.... 
here is a small example 
https://godbolt.org/z/9PPEa476s

ResearchDaniel:
I can confirm that SML do not compile with this fix

PR #1334, opened 2022-09-02
Feature/qt welcomeappend

martinfalk:
Fixes
* issue when aborting saving the old workspace, it just gets replaced and all changes are lost. Now opening the new workspace is aborted
* items in the gridview in the WelcomeWidget can be deselected

Changes proposed in this PR:
* "append workspace" from File menu is now integrated in WelcomeWidget
* added utility function to query modifiers (append, open with path)
* current workspace is not automatically closed when showing the welcome widget, this enables appending a workspace
* the state of processor and dock widgets is saved before showing the WelcomeWidget and restored when hiding it
* <kbd>Shift</kbd> is now used for opening example and test workspaces with full path since <kbd>Alt/Modifier</kbd> did not work well with the menu on Windows

image

This has been tested on: Windows, MSVS 2022 17.3.2

### DIFF
--- a/include/inviwo/qt/editor/inviwomainwindow.h
+++ b/include/inviwo/qt/editor/inviwomainwindow.h
@@ -252,7 +252,6 @@ private:
     HelpWidget* helpWidget_;
     WelcomeWidget* welcomeWidget_ =
         nullptr;  ///< Use delayed initialization as it can be expensive.
-    std::vector<QDockWidget*> welcomeHidden_;
     AnnotationsWidget* annotationsWidget_ = nullptr;
     InviwoAboutWindow* inviwoAboutWindow_ = nullptr;
 
@@ -294,10 +293,10 @@ private:
         /**
          * show previously hidden processor and dock widgets
          */
-        void restore();
+        void show();
 
         std::vector<Processor*> processors;
-        std::vector<InviwoDockWidget*> dockwidgets;
+        std::vector<QDockWidget*> dockwidgets;
     };
     VisibleWidgets visibleWidgetState_;  //!< holds all processor and dock widgets that were visible
                                          //!< before showing the welcome widget
--- a/include/inviwo/qt/editor/inviwomainwindow.h
+++ b/include/inviwo/qt/editor/inviwomainwindow.h
@@ -141,8 +141,19 @@ public:
      */
     bool appendWorkspace();
 
-    void saveWorkspace();
-    void saveWorkspaceAs();
+    /**
+     * saves the current workspace. If the workspace does not have a name yet, a file dialog will be
+     * shown.
+     * @return true if the workspace was saved, otherwise false.
+     * @see saveWorkspaceAs
+     */
+    bool saveWorkspace();
+    /**
+     * saves the current workspace using a file dialog
+     * @return true if the workspace was saved, otherwise false.
+     * @see saveWorkspaceAs
+     */
+    bool saveWorkspaceAs();
 
     /*
      * Save the current workspace into a new workspace file,
@@ -191,7 +202,12 @@ private:
      * @see askToSaveWorkspaceChanges
      */
     bool openWorkspace(QString workspaceFileName, bool isExample);
-    void saveWorkspace(QString workspaceFileName);
+
+    /**
+     * saves the current workspace to the given \p workspaceFileName.
+     * @return true if the workspace was saved, otherwise false.
+     */
+    bool saveWorkspace(QString workspaceFileName);
     void appendWorkspace(const QString& workspaceFileName);
 
     std::optional<QString> askForWorkspaceToOpen();
--- a/include/inviwo/qt/editor/inviwomainwindow.h
+++ b/include/inviwo/qt/editor/inviwomainwindow.h
@@ -61,6 +61,7 @@ class HelpWidget;
 class WelcomeWidget;
 class AnnotationsWidget;
 class InviwoApplicationQt;
+class InviwoDockWidget;
 class InviwoEditMenu;
 class InviwoAboutWindow;
 class ResourceManagerDockWidget;
@@ -68,6 +69,7 @@ class FileAssociations;
 class ToolsMenu;
 class TextLabelOverlay;
 class MenuKeyboardEventFilter;
+class Processor;
 
 class IVW_QTEDITOR_API InviwoMainWindow : public QMainWindow, public NetworkEditorObserver {
 public:
@@ -283,6 +285,22 @@ private:
     TCLAP::ValueArg<std::string> updateWorkspacesInPath_;
 
     UndoManager undoManager_;
+
+    struct VisibleWidgets {
+        /**
+         * store all visible processor and dock widgets before hiding them
+         */
+        void hide(InviwoMainWindow* win);
+        /**
+         * show previously hidden processor and dock widgets
+         */
+        void restore();
+
+        std::vector<Processor*> processors;
+        std::vector<InviwoDockWidget*> dockwidgets;
+    };
+    VisibleWidgets visibleWidgetState_;  //!< holds all processor and dock widgets that were visible
+                                         //!< before showing the welcome widget
 };
 
 }  // namespace inviwo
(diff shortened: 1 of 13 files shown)


include/inviwo/qt/editor/inviwomainwindow.h
```diff
@@ -267,6 +285,22 @@ class IVW_QTEDITOR_API InviwoMainWindow : public QMainWindow, public NetworkEdit
...
+
+    struct VisibleWidgets {
+        /**
+         * store all visible processor and dock widgets before hiding them
+         */
+        void hide(InviwoMainWindow* win);
+        /**
+         * show previously hidden processor and dock widgets
+         */
+        void restore();
```
ResearchDaniel:
```suggestion
        void show();
```

include/inviwo/qt/editor/welcomewidget.h
```diff
@@ -67,14 +68,18 @@ class IVW_QTEDITOR_API WelcomeWidget : public QSplitter {
     void updateRecentWorkspaces(const QStringList& list);
     void enableRestoreButton(bool hasRestoreWorkspace);
 
+    void updateLoadIcon(bool append, bool openWithPath);
```
ResearchDaniel: This could be made clearer as it is in the public interface. Perhaps leave the logic to the caller and instead allow the icon itself to be specified?
Also, perhaps `setLoadWorkspaceIcon`?

src/qt/editor/inviwomainwindow.cpp
```diff
@@ -184,6 +186,28 @@ class MenuKeyboardEventFilter : public QObject {
...
+                                          [](const auto p) { return p->hasProcessorWidget(); });
+    std::sort(widgetProcessors.begin(), widgetProcessors.end(),
+              [](auto a, auto b) { return iCaseLess(a->getDisplayName(), b->getDisplayName()); });
+    return widgetProcessors;
+}
+
+auto getFloatingDockWidgets(InviwoMainWindow* win) {
+    auto dockWidgets = util::copy_if(win->findChildren<InviwoDockWidget*>(),
+                                     [](const auto p) { return p->isFloating(); });
+    std::sort(dockWidgets.begin(), dockWidgets.end(), [](auto a, auto b) {
```
ResearchDaniel, line 201: Why are these sorted?

src/qt/editor/inviwomainwindow.cpp
```diff
@@ -118,6 +119,7 @@ QString addAppendSuffix(const QString& str, bool append) {
 class MenuKeyboardEventFilter : public QObject {
 public:
     MenuKeyboardEventFilter(QObject* parent) : QObject(parent) {}
+    virtual ~MenuKeyboardEventFilter() = default;
```
petersteneteg: Why add this?

ResearchDaniel (dismissed):
Looks good, but have some suggested changes

src/qt/editor/inviwomainwindow.cpp
```diff
@@ -1337,28 +1346,26 @@ void InviwoMainWindow::saveWorkspaceAsCopy() {
...
         showWelcomeScreen();
     }
 }
 
 void InviwoMainWindow::showWelcomeScreen() {
-    if (!newWorkspace()) return;
-
     setUpdatesEnabled(false);
+
+    visibleWidgetState_.hide(this);
```
petersteneteg, line 1376: Seems there are overlap between this one and welcomehidden_ why are both needed? Think this should be unified

src/qt/editor/inviwomainwindow.cpp
```diff
@@ -1384,6 +1391,9 @@ void InviwoMainWindow::hideWelcomeScreen() {
         return;
     }
     setUpdatesEnabled(false);
+
+    visibleWidgetState_.restore();
```
petersteneteg: Se above

src/qt/editor/inviwomainwindow.cpp
```diff
@@ -1414,14 +1424,23 @@ WelcomeWidget* inviwo::InviwoMainWindow::getWelcomeWidget() {
 
         connect(welcomeWidget_, &WelcomeWidget::loadWorkspace, this,
                 [this](const QString& filename, bool isExample) {
-                    if (askToSaveWorkspaceChanges()) {
+                    if (QApplication::keyboardModifiers() ==
+                        util::getKeyboardModifier(ModifierAction::AppendWorkspace)) {
```
petersteneteg: This function is backwards it should take the QModifierKeys as argument and return the ModifierAction

src/qt/editor/inviwomainwindow.cpp
```diff
@@ -1414,14 +1424,23 @@ WelcomeWidget* inviwo::InviwoMainWindow::getWelcomeWidget() {
 
         connect(welcomeWidget_, &WelcomeWidget::loadWorkspace, this,
                 [this](const QString& filename, bool isExample) {
-                    if (askToSaveWorkspaceChanges()) {
+                    if (QApplication::keyboardModifiers() ==
```
petersteneteg: I think it is better for the welcome widget to have two buttons and two signals, one for loading and one for appending

src/qt/editor/welcomewidget.cpp
```diff
@@ -217,11 +219,37 @@ class WorkspaceFilter : public QSortFilterProxyModel {
     SearchDSL<QModelIndex> dsl_;
 };
 
+class KeyboardEventFilter : public QObject {
```
petersteneteg: Why not just have two buttons

src/qt/editor/workspacegridview.cpp
```diff
@@ -305,6 +306,35 @@ class ChunkProxyModel : public QAbstractProxyModel {
...
+public:
+    SelectionModel(QAbstractItemModel* model, QObject* parent)
+        : QItemSelectionModel(model, parent) {}
+    virtual ~SelectionModel() = default;
+
+    virtual void select(const QModelIndex& index,
+                        QItemSelectionModel::SelectionFlags command) override {
+        if (command.testFlag(QItemSelectionModel::Select) && isSelected(index)) {
+            // ignore de-selection of already selected items, the initial selection is not using
+            // this function
```
petersteneteg, line 319: Why that makes no sense, what problem are you fixing?

src/qt/editor/workspacegridview.cpp
```diff
@@ -318,6 +348,10 @@ WorkspaceGridView::WorkspaceGridView(QAbstractItemModel* theModel, QWidget* pare
     setItemDelegate(new SectionDelegate(itemSize_, this));
     setIndentation(0);
 
+    // use custom selection model to prevent unselecting an already selected item
```
petersteneteg: Why should I not be able to unselect?!?

petersteneteg (changes requested):

src/qt/editor/inviwomainwindow.cpp
petersteneteg: They were initially only used for listing all the windows in the window menu

src/qt/editor/inviwomainwindow.cpp
martinfalk: My reasoning was to leave that decision to the main window. But OK.

src/qt/editor/workspacegridview.cpp
martinfalk: Currently, you select a workspace with the first mouse click. The second mouse click clears the current selection (as per Qt docs). I find this a rather strange behavior.
Put differently: why should clicking on the same item deselect it?

src/qt/editor/inviwomainwindow.cpp
martinfalk: In combination with adding another button, probably more sensible.

martinfalk:
<s>There is at least one instance where hiding the widgets might backfire:
When the welcome widget is shown with a modified workspace and Inviwo is closed. Then saving the current workspace will serialize the current hidden states.</s>

Addressed in f87537e

src/qt/editor/welcomewidget.cpp
martinfalk: Will do.

src/qt/editor/workspacegridview.cpp
petersteneteg: How else to unselect? when we have the drop icon in the center, if you want that back you have to be able to unselect... (With the append functionality we might even consider multiple selection)

petersteneteg (approved):

PR #1336, opened 2022-09-06
Feature/no usage mode

petersteneteg:
Changes proposed in this PR:
 * Removes the old never functional Usage Mode
 * Adds a new dialog for configuring which properties that should be shown in a compositeprocessor
image


modules/base/src/algorithm/image/imagecontour.cpp
```diff
@@ -28,12 +28,106 @@
...
+namespace {
+struct ImageContourDispatcher {
+    using type = std::shared_ptr<Mesh>;
+    template <typename Result, typename T>
+    std::shared_ptr<Mesh> operator()(const LayerRepresentation* in, size_t channel, double isoValue,
+                                     vec4 color);
+};
+
+template <typename Result, class DataType>
+std::shared_ptr<Mesh> ImageContourDispatcher::operator()(const LayerRepresentation* in,
```
ResearchDaniel, line 47: Is it not necessary to have the template code in the header?

include/inviwo/core/network/processornetwork.h
```diff
@@ -186,6 +188,11 @@ class IVW_CORE_API ProcessorNetwork : public Serializable,
     template <typename C>
     void forEachProcessor(C callback);
 
+    auto processorView() const {
```
ResearchDaniel: Any reason why we should not use:
processorRange
connectionRange
linkRange

modules/base/src/processors/imagecontourprocessor.cpp
```diff
@@ -37,25 +39,24 @@ const ProcessorInfo ImageContourProcessor::processorInfo_{
     "Image Contour",                     // Display name
     "Image Processing",                  // Category
     CodeState::Experimental,             // Code state
-    Tags::None,                          // Tags
-};
+    Tags::CPU,                           // Tags
+    R"(Does marching squares on the image to extract a contour mesh.)"_unindentHelp};
```
ResearchDaniel:
```suggestion
    R"(Extracts a contour line for a specific value in the image using marching squares. The output contour is provided as a line mesh.)"_unindentHelp};
```

modules/base/src/processors/imagecontourprocessor.cpp
```diff
@@ -37,25 +39,24 @@ const ProcessorInfo ImageContourProcessor::processorInfo_{
...
 ImageContourProcessor::ImageContourProcessor()
     : Processor()
-    , image_("image", true)
-    , mesh_("mesh")
-    , channel_("channel", "Channel", 0, 0, 4)
-    , isoValue_("iso", "ISO Value", 0.5, 0, 1)
-    , color_("color", "Color", vec4(1.0)) {
+    , image_("image", "Input image"_help, OutportDeterminesSize::Yes)
+    , mesh_("mesh", "Contour mesh"_help)
+    , channel_("channel", "Channel", "The image channel to use compare the iso value to"_help, 0,
```
ResearchDaniel:
```suggestion
    , channel_("channel", "Channel", "The image channel used to extract the contour"_help, 0,
```

modules/base/src/processors/imagecontourprocessor.cpp
```diff
@@ -37,25 +39,24 @@ const ProcessorInfo ImageContourProcessor::processorInfo_{
...
+               {0, ConstraintBehavior::Immutable}, {4, ConstraintBehavior::Editable})
+    , isoValue_("iso", "ISO Value", "The contour iso value"_help, 0.5,
+                {0, ConstraintBehavior::Ignore}, {1, ConstraintBehavior::Ignore})
+    , color_("color", "Color",
+             util::ordinalColor(vec4(1.0)).set("The color of the resulting mesh"_help)) {
```
ResearchDaniel:
```suggestion
             util::ordinalColor(vec4(1.0)).set("The contour color"_help)) {
```

src/core/processors/compositeprocessorutils.cpp
```diff
@@ -283,7 +285,7 @@ void util::expandCompositeProcessorIntoNetwork(CompositeProcessor& composite) {
             network.addLink(l);
         }
 
-        network.removeProcessor(&composite);
+        auto tmp = network.removeProcessor(&composite);
```
ResearchDaniel: Does tmp have something to do with deleting the composite?
Either name it accordingly or add a comment in that case.

src/qt/editor/subpropertyselectiondialog.cpp
```diff
@@ -0,0 +1,687 @@
...
+            return Qt::ItemIsEnabled;
+        }
+    }
+
+    virtual QVariant data(const QModelIndex& index, int role) const override {
+        if (role ==  Qt::ForegroundRole) {
+            return {};
+        }
+    
+        if (index.column() == 3) {
```
ResearchDaniel: These magic numbers could perhaps be named?

src/qt/editor/subpropertyselectiondialog.cpp
```diff
@@ -0,0 +1,687 @@
...
+
+        QSizePolicy sp = tree->sizePolicy();
+        sp.setHorizontalStretch(3);
+        sp.setVerticalStretch(3);
+        sp.setVerticalPolicy(QSizePolicy::Expanding);
+        sp.setHorizontalPolicy(QSizePolicy::MinimumExpanding);
+        tree->setSizePolicy(sp);
+    };
+
+    layout->addWidget(new QLabel("Composite Network"), 0, 0);
```
ResearchDaniel:
```suggestion
    layout->addWidget(new QLabel("All Properties"), 0, 0);
```

ResearchDaniel (changes requested):
Very nice! Some minor comments.
Could you also update the changelog and state that we now advice people to use CanvasWithProperties instead of the UsageMode

modules/base/src/algorithm/image/imagecontour.cpp
petersteneteg: Only if you plan on using it in some other translation unit. this is only used further down in this translation unit, then there is no need to have it in the header....

include/inviwo/core/network/processornetwork.h
petersteneteg: no particular reason, can change

src/core/processors/compositeprocessorutils.cpp
petersteneteg: don't think is was needed with the holder up top

ResearchDaniel (dismissed):

ResearchDaniel (approved):

PR #1337, opened 2022-09-08
Core,OpenGL,PlottingGL: shader uniform for SelectionColorProperty

martinfalk:
Changes proposed in this PR:
* GLSL shader uniform for SelectionColorProperty
* GLSL utility function to apply  a selection color

### DIFF
--- a/modules/opengl/glsl/utils/selectioncolor.glsl
+++ b/modules/opengl/glsl/utils/selectioncolor.glsl
@@ -2,7 +2,7 @@
  *
  * Inviwo - Interactive Visualization Workshop
  *
- * Copyright (c) 2016-2022 Inviwo Foundation
+ * Copyright (c) 2022 Inviwo Foundation
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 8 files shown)


petersteneteg (dismissed):

modules/opengl/glsl/utils/selectioncolor.glsl
```diff
@@ -0,0 +1,50 @@
...
+};
+
+/**
+ * Blend @p srcColor with @p brushingColor using separate interpolants for color and alpha.
+ * @param srcColor
+ * @param brushingColor
+ * @return color resulting from mixing @p srcColor with @p brushingColor
+ */
+vec4 applySelectionColor(in vec4 srcColor, in SelectionColor brushingColor) {
+    return mix(srcColor, brushingColor.color, vec4(vec3(brushingColor.colorMixIn), brushingColor.alphaMixIn));
```
ResearchDaniel: The visible state is never used. There should be a check here?

modules/opengl/glsl/utils/selectioncolor.glsl
```diff
@@ -0,0 +1,50 @@
...
+    vec4 color;
+    float colorMixIn;
+    float alphaMixIn;
+    bool visible;
+};
+
+/**
+ * Blend @p srcColor with @p brushingColor using separate interpolants for color and alpha.
+ * @param srcColor
+ * @param brushingColor
```
ResearchDaniel: Would call it selectionColor

ResearchDaniel (changes requested):

modules/opengl/glsl/utils/selectioncolor.glsl
martinfalk: The visible state is used in the molvis shaders to reject stuff in shaders. Both scatterplot and parallel coords handle this on the C++ side.

modules/opengl/glsl/utils/selectioncolor.glsl
martinfalk: minor oversight. Might have missed renaming those.

modules/opengl/glsl/utils/selectioncolor.glsl
ResearchDaniel: I see. The code becomes a bit diluted though as the struct is defined just above and then the bool unused.

ResearchDaniel (approved):

petersteneteg (approved):

modules/opengl/glsl/utils/selectioncolor.glsl
martinfalk: With respect to the helper function, yes. But not regarding the SelectionColorProperty. I also rather have a `discard()` directly in the shader instead of some utility function.

PR #1340, opened 2022-09-14
Fix/tiff in-source build

ResearchDaniel:
Fixes 
* Hide tiff release project under ext/tiff

Changes proposed in this PR:
* Allow building in the source directory
 * Ignore folders build and out (Visual Studio uses out per default)

This has been tested on: Windows, Visual studio


ext/tiff/CMakeLists.txt
```diff
@@ -207,7 +207,8 @@ message(STATUS "")
 #[===[ Inviwo modification ]===]
 ivw_suppress_compiler_warnings(tiff)
 ivw_default_install_targets(tiff)
-ivw_folder(tiff ext)
+ivw_folder(tiff ext/tiff)
+ivw_folder(release ext/tiff)
```
petersteneteg: better to remove the include(Release) higher up instead...

petersteneteg (dismissed):

ext/tiff/CMakeLists.txt
ResearchDaniel: Yes and no. The release writes the version in the header. Not sure if any code relies on that

petersteneteg: Should not be needed there is already a generated file in the in the source tree. So Im fairly sure we will not need this.

petersteneteg: Also I really dislike having a generic target name like that around if it not inviwo related...

petersteneteg (dismissed):

petersteneteg:
I was expecting that doing a rebase and force push, would discard my review, but it seems it was smart enough not to do that somehow, nice!

ResearchDaniel:
> I was expecting that doing a rebase and force push, would discard my review, but it seems it was smart enough not to do that somehow, nice!

Cool, but I added .vs folder to ignore so you have to review again anyway :D

petersteneteg (approved):

PR #1343, opened 2022-09-16
Feature/changelog drawer

ResearchDaniel:
Fixes: More space for workspace previews by making it possible to collapse the changelog. The changelog is collapsed by default.

Changes proposed in this PR:
image
image


petersteneteg:
Why hide the changelog by default?

ResearchDaniel:
> Why hide the changelog by default?

My reasoning is that it is not too relevant for a person starting to use Inviwo and hiding it makes it easier to focus on the main parts (loading a workspace or creating a new one).

petersteneteg:
My thinking is that the more people that see that the better.... having it hidden mean even less people will ever look at it.

martinfalk:
Reworked `WelcomeWidget` and the collapsible splitter slightly.
* central widget got replaced with a stacked widget that allows easy switching between the drag&drop hint and the workspace details
* the splitter handle of the changelog now has a label and arrow indicator
* can be collapsed with dragging, clicking the arrow, or a double click
* got rid of the mini "toolbar" since it introduced artifacts.
* restored previous split ratios

image

**Collapsed changelog**
image


image

ResearchDaniel:
Still think that it should be collapsed by default and make the grid view occupy a majority of the space

petersteneteg:
> Still think that it should be collapsed by default and make the grid view occupy a majority of the space

I agree with Martin, The center view should be the larger view, anything else is just too visually unbalanced. 
Also thing thing the changes should be visible by default.

src/qt/editor/CMakeLists.txt
```diff
@@ -10,7 +10,9 @@ set(MOC_FILES
     ${IVW_INCLUDE_DIR}/inviwo/qt/editor/networkeditor.h
     ${IVW_INCLUDE_DIR}/inviwo/qt/editor/processordraghelper.h
     ${IVW_INCLUDE_DIR}/inviwo/qt/editor/processormimedata.h
+    ${IVW_INCLUDE_DIR}/inviwo/qt/editor/verticallabel.h
     ${IVW_INCLUDE_DIR}/inviwo/qt/editor/subpropertyselectiondialog.h
+    ${IVW_INCLUDE_DIR}/inviwo/qt/editor/verticallabel.h
```
petersteneteg, line 14: duplicate?

src/qt/editor/CMakeLists.txt
```diff
@@ -182,7 +185,9 @@ if(IVW_CFG_BUILD_CHANGELOG)
         OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/CopyIfFound.cmake"
         CONTENT 
             "if (EXISTS \"${buildFile}\")
+
```
petersteneteg: why?

src/qt/editor/welcomewidget.cpp
```diff
@@ -538,36 +565,18 @@ WelcomeWidget::WelcomeWidget(InviwoApplication* app, QWidget* parent)
...
-    auto getScreen = []() {
-#if QT_VERSION < QT_VERSION_CHECK(5, 14, 0)
-        return QGuiApplication::screenAt(utilqt::getApplicationMainWindow()->pos());
-#else
-        return utilqt::getApplicationMainWindow()->screen();
-#endif
-    };
-
-    if (getScreen()->size().width() < 1200) {
-        setSizes(QList<int>({1, 0}));
```
petersteneteg, line 551: is this still not a good idea?

petersteneteg (dismissed):

src/qt/editor/welcomewidget.cpp
ResearchDaniel: Only if we display it by default

petersteneteg: which I still think we should

petersteneteg (dismissed):

petersteneteg (approved):

PR #1346, opened 2022-09-23
XxxRAMPrecision headers moved into XxxRAM headers

petersteneteg:
Changes proposed in this PR:
 * Moved VolumeRAMPrecision to VolumeRAM, LayerRAMPrecision to LayerRAM, and BufferRAMPrecision to BufferRAM

This make life a bit easier with respect to the dispatching... since XxxRAM has the dispatch function but to call it you have to include XxxRAMPrecision which is very error prone. And I don't think we gain much by keeping them in separate headers.


petersteneteg:
will remove the XxxRAMPrecision includes at a later stage.

martinfalk (dismissed):
Maybe worthwhile a changelog entry.

martinfalk (approved):

PR #1350, opened 2022-09-30
Feature/iwyu part 2

petersteneteg:
Changes proposed in this PR:
 * Include cleanup for a number of modules


martinfalk (approved):

modules/glfw/src/filewatcher.cpp
```diff
@@ -29,21 +29,11 @@
...
-#include <atomic>
-#include <chrono>
-#include <algorithm>
+#include <inviwo/core/util/assertion.h>      // for IVW_ASSERT
+#include <inviwo/core/util/logcentral.h>     // for LogCentral, LogWarn
+#include <inviwo/core/util/stdextensions.h>  // for erase_remove
+
+#include <algorithm>  // for find
 
 #ifdef WIN32
```
martinfalk, line 38: Build fails on Windows due to missing includes
mutex, chrono, thread, ...

martinfalk (changes requested):

modules/glfw/src/filewatcher.cpp
petersteneteg: fixed

martinfalk (approved):

PR #1356, opened 2022-10-14
Feature/bugfixes

petersteneteg:
various minor bug fixes


martinfalk (dismissed):

modules/base/include/modules/base/processors/dataexport.h
```diff
@@ -74,10 +74,10 @@ class DataExport : public Processor {
 template <typename DataType, typename PortType>
 DataExport<DataType, PortType>::DataExport()
     : Processor()
-    , port_("data")
-    , file_("file", "File name", "", "mesh")
-    , export_("export", "Export")
-    , overwrite_("overwrite", "Overwrite", false) {
+    , port_("data", "The data to export"_help)
+    , file_("file", "File name", "The file to save data to"_help, "", "mesh")
```
ResearchDaniel:
```suggestion
    , file_("file", "File name", "The path to save data to"_help, "", "mesh")
```

ResearchDaniel: Why "mesh" in the end?

modules/base/include/modules/base/processors/vectorelementselectorprocessor.h
```diff
@@ -66,8 +66,8 @@ class VectorElementSelectorProcessor : public Processor {
 template <typename T, typename OutportType>
 VectorElementSelectorProcessor<T, OutportType>::VectorElementSelectorProcessor()
     : Processor()
-    , inport_("inport")
-    , outport_("outport")
+    , inport_("inport", "Sequence of data to select from"_help)
+    , outport_("outport", "The select item"_help)
```
ResearchDaniel:
```suggestion
    , outport_("outport", "The selected item"_help)
```

modules/plotting/src/interaction/boxselectioninteractionhandler.cpp
```diff
@@ -83,14 +83,18 @@ void BoxSelectionInteractionHandler::invokeEvent(Event* event) {
                 // Click in empty space
                 switch (dragRectSettings_.getMode()) {
                     case BoxSelectionSettingsInterface::Mode::Selection: {
-                        // selection changed
-                        std::vector<bool> selected(xAxis_->getSize(), false);
-                        selectionChangedCallback_.invoke(selected, append);
+                        if (xAxis_) {
```
ResearchDaniel, line 86: Why do we only use xAxis_ and not also yAxis_?

tools/meta/templates/processor.cpp
```diff
@@ -42,7 +42,7 @@ const ProcessorInfo {{ file/name }}::processorInfo_{
 
 const ProcessorInfo {{ file/name }}::getProcessorInfo() const { return processorInfo_; }
 
-{{ file/name }}::{{file / name}}()
+{{ file/name }}::{{ file/name}}()
```
ResearchDaniel: Should it start with a space?

```suggestion
{{ file/name }}::{{file/name}}()
```

vcpkg.json
```diff
@@ -32,6 +34,12 @@
         "tinyxml2",
         "utfcpp",
         "zlib"
+    ],
+    "overrides" : [
+        {
+            "$reason" : "We get very off link errors with 1.7.0. 1.7.0 changed how the export header is created, maybe something broke upstream",
```
ResearchDaniel:
```suggestion
            "$reason" : "We get linker errors with 1.7.0. 1.7.0 changed how the export header is created, maybe something broke upstream",
```

ResearchDaniel (dismissed):
Nice, some suggestions and questions

modules/base/include/modules/base/processors/dataexport.h
petersteneteg: that would be the "contentType" I did not change that, copied in the doc string. 
the content type should depend on with kid of data we want to export. It is probably left over since before this was a template. Now is should be something like ContentType<DataType>::value if we had such a construct...

modules/plotting/src/interaction/boxselectioninteractionhandler.cpp
petersteneteg: I have don't know, It just crashed if the xAxis was nullptr...

tools/meta/templates/processor.cpp
petersteneteg: generally we have a space before and after

modules/dataframe/tests/benchmarks/join.cpp
```diff
@@ -65,7 +64,10 @@ std::vector<int> left_ = [&]() {
     return left;
 }();
 
-std::vector<int> right_ = [&]() {
+const std::vector<int> right_ = []() {
+    std::random_device rd;
+    std::mt19937 gen(rd());
```
martinfalk, line 69: doesn't this now give exactly the same sequence as for left_?

petersteneteg: No, one does a shuffle the other uses distrib?

martinfalk (approved):

PR #1357, opened 2022-10-27
Feature/various bug fixes

martinfalk:
Fixes
* when updating a `CompositeProperty`, that is a `CollapsibleGroupBoxWidgetQt`, the internal widget lost its parent and the group box started floating
* when replacing buffers of a `Mesh`, the updating the `MeshGL` representation fails since the buffer addresses are different.
* incorrect lower limits for double properties in `FilterListProperty`
* serialization fix in `BrushingAndLinkingManager`

Changes proposed in this PR:
 * `MeshGL::update()` enforces the buffer attachment without checking the buffer address


ResearchDaniel (dismissed):

include/inviwo/core/datastructures/geometry/mesh.h
```diff
@@ -125,16 +125,17 @@ class IVW_CORE_API Mesh : public DataGroup<Mesh, MeshRepresentation>,
...
+     * @return the removed buffer
      */
     std::pair<BufferInfo, std::shared_ptr<BufferBase>> replaceBuffer(
         size_t idx, BufferInfo info, std::shared_ptr<BufferBase> att);
 
     /**
-     * Replaces buffer at index with new buffer
-     * Does nothing if index out of range.
-     * @param old   Old buffer to replace, if the old buffer is not found the new one is appended.
+     * Replaces the old buffer \p old with the new buffer \p att
```
petersteneteg: att is not the best parameter name...

modules/dataframe/src/properties/filterlistproperty.cpp
```diff
@@ -181,7 +181,7 @@ FilterListProperty::FilterListProperty(std::string_view identifier, std::string_
                 {"greater", "greater (>)", filters::NumberComp::Greater},
                 {"greaterEqual", "Greater Equal (>=)", filters::NumberComp::GreaterEqual}});
         detail::addProperty<DoubleProperty>(*doubleItem, "value", "Value", 0.0,
-                                            std::numeric_limits<double>::lowest(),
+                                            -std::numeric_limits<double>::max(),
```
petersteneteg: Why? That should be the same?

modules/opengl/src/geometry/meshgl.cpp
```diff
@@ -95,13 +95,13 @@ void MeshGL::update(bool editable) {
 
     bufferArray_.bind();
     Mesh* owner = this->getOwner();
-    // update all buffers except index buffers, i.e. position, color, normals, etc.)
+    // update regular buffers except index buffers
```
petersteneteg: What is a regular buffer?

modules/dataframe/src/properties/filterlistproperty.cpp
petersteneteg: Might need an explanation in a comment, otherwise I would change it back when I see it

petersteneteg (dismissed):

modules/dataframe/src/properties/filterlistproperty.cpp
martinfalk: > Returns the lowest finite value representable by the numeric type T, that is, a finite value x such that there is no other finite value y where y < x. This is different from [std::numeric_limits](http://en.cppreference.com/w/cpp/types/numeric_limits)<T>::min() for floating-point types.

modules/opengl/src/geometry/meshgl.cpp
martinfalk: everything beside an index buffer, might change it back to "all".

modules/dataframe/src/properties/filterlistproperty.cpp
martinfalk: You are correct. Might have been mislead by an outdated serialized state.

include/inviwo/core/datastructures/geometry/mesh.h
martinfalk: renamed all occurrences of `att` to `buffer`

modules/brushingandlinking/src/brushingandlinkingmanager.cpp
```diff
@@ -413,7 +413,7 @@ void BrushingAndLinkingManager::serialize(Serializer& s) const {
             s.serialize(toString(action), map, "selection", {},
                         [](BrushingTarget t) { return t.getString(); }, {});
         } else if (std::holds_alternative<IndexListTargets>(targetmap)) {
-            auto& map = std::get<BitSetTargets>(targetmap);
+            auto& map = std::get<IndexListTargets>(targetmap);
```
petersteneteg: Using get_if (https://en.cppreference.com/w/cpp/utility/variant/get_if) here would avoid writing the type twice or maybe using a visitor...

petersteneteg (dismissed):

martinfalk:
added new styles to `ColorScaleLegend` with a split background (checkerboard and opaque TF) and opaque TF only.
image

modules/plottinggl/glsl/legend.frag
```diff
@@ -60,11 +66,30 @@ vec4 over(vec4 colorB, vec4 colorA) {
...
+    src.rgb *= src.a;
+    return src;    
+}
+
 void main() {
     vec2 texCoord = gl_FragCoord.xy - viewport.xy;
     vec2 outputDim = viewport.zw - vec2(2 * borderWidth);
     vec2 centeredPos = (texCoord - viewport.zw * 0.5);
-    vec2 normPos = texCoord / outputDim;
+    vec2 normPos = (texCoord - vec2(borderWidth)) / outputDim;
```
ResearchDaniel, line 92: In case the intention is to sample pixel-centered:
(texCoord + 0.5 - borderWidth) / outputDim

ResearchDaniel (dismissed):

modules/plottinggl/glsl/legend.frag
martinfalk: The reason here is to have normalized coordinates [0,1] within the border corresponding to the given pixel size of the legend. Pixel-centered sampling would be a bonus but requires outputDim-1 for normalization, doesn't it.

martinfalk:
More fixes, added import/export for XML-based TFs (sciviscolor.org/ParaView format). Number of recent files is now adjustable in system settings.

modules/plottinggl/src/processors/volumeaxis.cpp
```diff
@@ -317,15 +333,19 @@ void VolumeAxis::process() {
     const auto volume = inport_.getData();
     const dmat4 m = volume->getCoordinateTransformer().getDataToModelMatrix();
     const dmat3 nm = glm::transpose(glm::inverse(m));
-    const double offset = axisOffset_.get() / 100;
+    // the mean length of the three basis vectors is used for a relative axis offset (%)
+    const double offset =
+        axisOffset_.get() / 100 *
+        (glm::length(dvec3(m[0])) + glm::length(dvec3(m[1])) + glm::length2(dvec3(m[2]))) / 3.0;
```
petersteneteg: glm::length2 ?

modules/plottinggl/src/processors/volumeaxis.cpp
```diff
@@ -317,15 +333,19 @@ void VolumeAxis::process() {
     const auto volume = inport_.getData();
     const dmat4 m = volume->getCoordinateTransformer().getDataToModelMatrix();
```
petersteneteg, line 334: Why not just use dataToWorld here? or why would that be wrong?

src/core/datastructures/transferfunction.cpp
```diff
@@ -139,43 +141,235 @@ vec4 TransferFunction::sample(double v) const { return interpolateColor(v); }
 vec4 TransferFunction::sample(float v) const { return interpolateColor(v); }
 
 std::vector<FileExtension> TransferFunction::getSupportedExtensions() const {
-    return {{"itf", "Inviwo Transfer Function"}, {"png", "Transfer Function Image"}};
+    return {{"itf", "Inviwo Transfer Function"},
```
petersteneteg: This is nice!, but it should really be using the DataReader/WriterFactory I think.

src/core/datastructures/transferfunction.cpp
```diff
@@ -139,43 +141,235 @@ vec4 TransferFunction::sample(double v) const { return interpolateColor(v); }
...
+
+    for (size_t i = 0; i < size; ++i) {
+        dptr[i] = static_cast<glm::u8vec4>(glm::clamp(sptr[i] * 255.0f, vec4(0.0f), vec4(255.0f)));
+    }
+
+    auto factory = InviwoApplication::getPtr()->getDataWriterFactory();
+    auto writer = factory->getWriterForTypeAndExtension<Layer>(ext);
+    if (!writer) {
+        writer = factory->getWriterForTypeAndExtension<Layer>(ext);
+    }
```
petersteneteg: ?

src/core/datastructures/transferfunction.cpp
```diff
@@ -139,43 +141,235 @@ vec4 TransferFunction::sample(double v) const { return interpolateColor(v); }
...
-        if (invalidData_) calcTransferValues();
+    const auto size = glm::compMul(dataRepr_->getDimensions());
+    const auto sptr = dataRepr_->getDataTyped();
+    const auto dptr = uint8DataRepr->getDataTyped();
+
+    for (size_t i = 0; i < size; ++i) {
+        dptr[i] = static_cast<glm::u8vec4>(glm::clamp(sptr[i] * 255.0f, vec4(0.0f), vec4(255.0f)));
+    }
+
+    auto factory = InviwoApplication::getPtr()->getDataWriterFactory();
```
petersteneteg: you can use factoryutil.h here

src/core/datastructures/transferfunction.cpp
```diff
@@ -139,43 +141,235 @@ vec4 TransferFunction::sample(double v) const { return interpolateColor(v); }
...
+        writer = factory->getWriterForTypeAndExtension<Layer>(ext);
+    }
+    if (!writer) {
+        throw DataWriterException("Data writer not found for requested format", IVW_CONTEXT);
+    }
+    writer->setOverwrite(Overwrite::Yes);
+    writer->writeData(unit8Data.get(), filename);
+}
+
+void TransferFunction::loadFromLayer(std::string_view filename, const FileExtension& ext) {
```
petersteneteg: We could register data readers specifically for image file to tf

petersteneteg (changes requested):
Some minor suggestions, let me know if you want me to move the tf things to the factories...

modules/plottinggl/src/processors/volumeaxis.cpp
martinfalk: The world transform get's applied during rendering of the axis (line 348).
```c++
axisRenderers_[axisIdx].render(&camera_.get(), volume->getWorldMatrix(), dims, start, stop, tickDir);
```

src/core/datastructures/transferfunction.cpp
martinfalk: good catch. The second one should request a writer for the file's actual `extension`.

modules/plottinggl/src/processors/volumeaxis.cpp
petersteneteg: Yes but Why?

include/inviwo/core/io/datareaderfactory.h
```diff
@@ -118,14 +118,22 @@ std::vector<FileExtension> DataReaderFactory::getExtensionsForType() const {
...
+    for (auto& [ext, reader] : map_) {
+        if (util::iCaseEndsWith(path, ext.extension_)) {
+            if (auto r = dynamic_cast<DataReaderType<T>*>(reader)) {
+                candidates.emplace_back(ext.extension_.size(), r);
             }
         }
     }
-    return std::unique_ptr<DataReaderType<T>>{};
+
+    if (auto it = std::max_element(candidates.begin(), candidates.end(),
```
martinfalk, line 131: Might want to copy the removed comment from above.
```suggestion
    // Sort extensions by length to first compare long extensions. Avoids selecting extension xyz in
    // case xyzw exist, see getReaderForTypeAndExtension
    if (auto it = std::max_element(candidates.begin(), candidates.end(),
```

include/inviwo/core/io/datawriter.h
```diff
@@ -71,6 +72,25 @@ class IVW_CORE_API DataWriter {
...
      */
     void checkOverwrite(std::string_view path) const;
+    
+    /**
+     * @brief Set writer specific options
+     * See the documentation of the specific writer about which options that are available
+     * @param key the option to set
+     * @param value the new value for the option
+     * @return true of the option was recognized and set, otherwise false
+     */
```
martinfalk:
```suggestion
     * @brief Set writer specific options
     * See the documentation of the specific writer for the available options
     * @param key the option to set
     * @param value the new value for the option
     * @return true if the option was recognized and set, otherwise false
     */
```

include/inviwo/core/io/datawriter.h
```diff
@@ -71,6 +72,25 @@ class IVW_CORE_API DataWriter {
...
+    virtual bool setOption([[maybe_unused]] std::string_view key, [[maybe_unused]] std::any value) {
+        return false;
+    }
+
+    /**
+     * @brief Query the value of an writer specific option
+     * @param key the option to query
+     * @return an std::any with the requested option of an empty std::any of the option was not
+     * found
+     */
```
martinfalk:
```suggestion
    /**
     * @brief Query the value of a writer specific option
     * @param key the option to query
     * @return an std::any with the requested option or an empty std::any if the option was not
     * found
     */
```

include/inviwo/core/util/factory.h
```diff
@@ -243,7 +243,7 @@ bool FactoryRegister<M, Key, K>::unRegisterObject(M* obj) {
     size_t removed =
         util::map_erase_remove_if(map_, [obj](const auto& elem) { return elem.second == obj; });
     if (removed > 0) {
-        this->notifyObserversOnRegister(obj);
+        this->notifyObserversOnUnRegister(obj);
```
martinfalk, line 246: maybe should be called `notifyObserversOnUnregister()`

petersteneteg: not part of this pr, I just change the typo of the missing "Un"...

petersteneteg (dismissed):

petersteneteg (approved):

PR #1361, opened 2022-11-16
Feature/plotting fixes

martinfalk:
Fixes
 * `VolumeAxis` did not consider world transformations of the volume
 * `enumToStr` conversion for Image port-related enums

Changes proposed in this PR:
 * refactored `VolumeSlice` processor into `VolumeSliceExtractor`
     + output format is either source format, `uint8`, or `float32`. 
     + optionally, a TF can be applied (implicitly sets output format to `glm::vec<4, output format>`
 * LayerInformationProperty and ImagePlotProcessor consider layer basis and offset.

This has been tested on: VS 2022, 17.4.0

image


modules/base/src/processors/volumesliceextractor.cpp
```diff
@@ -0,0 +1,421 @@
...
+        case CartesianCoordinateAxis::X:
+            return vec2(offset.z, offset.y);
+        case CartesianCoordinateAxis::Y:
+            return vec2(offset.x, offset.z);
+        case CartesianCoordinateAxis::Z:
+            return vec2(offset.x, offset.y);
+    }
+}
+
+SwizzleMask getSwizzleMask(int channels) {
```
petersteneteg: What about the source swizzle mask? what about wrapping?

petersteneteg (dismissed):

petersteneteg (dismissed):

modules/basegl/src/processors/imageprocessing/columnrowlayout.cpp
```diff
@@ -341,6 +341,7 @@ void ColumnLayout::updateViewports(ivec2 dim) {
     for (auto i = 0; i < std::max(1, numViews); ++i) {
         viewManager_.push_back(ivec4(positions[i], 0, positions[i + 1] - positions[i], dim.y));
     }
+    notifyObserversActiveConnectionsChange(this);
```
ResearchDaniel, line 344: Should this not be placed here?
```
    inport_.onConnect([this]() { updateSplitters(true); });
    inport_.onDisconnect([this]() { updateSplitters(false); });
```

martinfalk: `updateViews()` is also called in `updateSliders()` since the active state of port connections changes depending on the viewports/viewport manager.

petersteneteg (approved):

PR #1366, opened 2022-12-13
Feature/volumeaxis offset

martinfalk:
Changes proposed in this PR:
 * relative offsets for labels and captions in `VolumeAxis`
 * modules sorted by name in Tools menu
 * added help for axis related properties
 * typo/warn fix

Relative scaling of axis labels and captions as well as tick lengths ensures reasonable axes when the volume basis changes. In the newly added regression test (see image), the basis of the volume is scaled by 1, 10000, and 0,00001 while using the same axis offsets.

image

This has been tested on: Windows/MSVC 2022


modules/plotting/src/properties/axisproperty.cpp
```diff
@@ -67,34 +68,53 @@ class PlotTextSettings;
...
+    , color_{"color", "Color",
+             util::ordinalColor(vec4{0.0f, 0.0f, 0.0f, 1.0f}).set("Color of the axis"_help)}
+    , width_{"width", "Width", util::ordinalLength(2.5f, 20.0f).set("Line width of the axis"_help)}
     , useDataRange_{"useDataRange", "Use Data Range", true}
-    , range_("range", "Axis Range", 0.0, 100.0, -1.0e6, 1.0e6, 0.01, 0.0,
-             InvalidationLevel::InvalidOutput, PropertySemantics::Text)
-    , flipped_{"flipped", "Swap Label Position", false}
+    , range_{"range",
+             "Axis Range",
+             "Current range of the axis"_help,
```
ResearchDaniel:
```suggestion
             "Value range of the axis"_help,
```

modules/plotting/src/properties/categoricalaxisproperty.cpp
```diff
@@ -64,32 +64,34 @@ CategoricalAxisProperty::CategoricalAxisProperty(
...
 
-    color_.setSemantics(PropertySemantics::Color);
-
-    addProperty(visible_);
-    addProperty(color_);
-    addProperty(width_);
-    addProperty(flipped_);
-    addProperty(orientation_);
-    addProperty(placement_);
+    scalingFactor_.setVisible(1.0f);
```
ResearchDaniel:
```suggestion
    scalingFactor_.setVisible(false);
```

modules/plottinggl/src/processors/volumeaxis.cpp
```diff
@@ -161,18 +162,29 @@ const ProcessorInfo VolumeAxis::processorInfo_{
     "Plotting",               // Category
     CodeState::Stable,        // Code state
     "GL, Plotting",           // Tags
-};
+    "Renders an x, y, and z axis next to the input volume."_help};
```
ResearchDaniel:
```suggestion
    "Renders x, y, and z axes next to the input volume."_help};
```

modules/plottinggl/src/processors/volumeaxis.cpp
```diff
@@ -385,6 +402,36 @@ void VolumeAxis::process() {
...
+                        mode = offsetScaling_.get()]() {
+            switch (mode) {
+                case OffsetScaling::MinExtent:
+                    return glm::compMin(l) / 100.0f;
+                case OffsetScaling::MaxExtent:
+                    return glm::compMax(l) / 100.0f;
+                case OffsetScaling::MeanExtent:
+                    return glm::compAdd(l) / (3.0f * 100.0f);
+                case OffsetScaling::Diagonal:
+                    return glm::compAdd(l) / 100.0f;
```
ResearchDaniel: Should be sqrt(3)*100?

modules/plotting/include/modules/plotting/datastructures/axissettings.h
```diff
@@ -57,6 +57,8 @@ class IVW_MODULE_PLOTTING_API AxisSettings {
     virtual bool getFlipped() const = 0;
     virtual vec4 getColor() const = 0;
     virtual float getWidth() const = 0;
+    // scaling factor affecting tick lengths and offsets of axis caption and labels
+    virtual float getScalingFactor() const = 0;
```
ResearchDaniel, line 60: Factor is redundant. Could name it scaling, transform, or offsetScaling. OffsetScaling seem to be used in later.
```suggestion
    virtual float getOffsetScaling() const = 0;
```

modules/plottinggl/src/processors/volumeaxis.cpp
```diff
@@ -161,18 +162,29 @@ const ProcessorInfo VolumeAxis::processorInfo_{
...
     : Processor()
-    , inport_{"volume"}
-    , imageInport_{"imageInport"}
-    , outport_{"outport"}
-    , axisOffset_{"axisOffset", "Axis Offset (%)", 0.1f, 0.0f, 10.0f}
+    , inport_{"volume", "Input volume"_help}
+    , imageInport_{"imageInport", "Background image (optional)"_help}
+    , outport_{"outport",
+               "Output image containing the rendered volume axes and the optional input image"_help}
+    , offsetScaling_{"offsetScaling",
```
ResearchDaniel, line 175: What would be the reason for choosing one over the other here? If all options are necessary we might want to provide some guidelines in the help.

ResearchDaniel (dismissed):
Looks good overall, but have some questions and suggestions. The axes labels in the regression test seem to overlap the tick text. Is it possible avoid this? If not, is it possible to change the workspace so that it does not?

modules/plotting/include/modules/plotting/datastructures/axissettings.h
petersteneteg: Since it affects more than the offset, according to the comment, I think getScalingFactor or getScaleFactor is appropriate

petersteneteg (dismissed):

modules/plottinggl/src/processors/volumeaxis.cpp
martinfalk: Assuming you have a rather uniform basis, then it doesn't matter. However when loading volumes with rather different dimensions these options have different effects. Think of narrow vs. large volumes in a time series.

modules/plottinggl/src/processors/volumeaxis.cpp
martinfalk: Should read `glm::length(m[0]+m[1]+m[2])` instead.

modules/plottinggl/src/processors/volumeaxis.cpp
martinfalk: If you load volumes which grow in one dimension, you probably want min to be consistent.
 If the smallest side is shrinking you don't want that one.

martinfalk: Added the following help
```c++
                     R"(Offset scaling affects tick lengths and offsets of axis captions and labels.
                        + *None* No scaling, offsets and lengths are given in world coordinates.
                        + *Min* Relative scaling based on the shortest extent of the volume. Useful 
                                when visualizing a growing volume.
                        + *Max* Relative scaling based on the longest extent of the volume. Useful 
                                when visualizing a shrinking volume.
                        + *Mean* Relative scaling basd on the mean volume extent.
                        + *Diagonal* Relative scaling based on the diagonal of the volume.
                      )"_unindentHelp,
```
Might consider to either use `mean` or `diagonal`, if we want to reduce the number of options.

ResearchDaniel: Would it be possible to choose this automatically based on these guidelines with manual override?

petersteneteg: I don't think it is worth the effort to try and automate that, just use mean or diagonal as default.

modules/plotting/src/properties/axisproperty.cpp
```diff
@@ -52,10 +52,12 @@
 #include <algorithm>  // for transform
 #include <cstddef>    // for size_t
 #include <iterator>   // for back_insert_iterator
+#include <limits>
 
 #include <fmt/core.h>    // for basic_string_view
 #include <fmt/printf.h>  // for sprintf
 #include <glm/vec2.hpp>  // for vec<>::(anonymous)
+#include <glm/gtx/vec_swizzle.hpp>
```
petersteneteg, line 60: unused?

petersteneteg:
modules/plottinggl/tests/regression/axiscombinationtest/VolumeAxis2.png
seems to have gone for the worse I think, The units are lost? and the labels are 2 close to the axes, I the new one is the new defaults, they should be further away, especially in the 2d plot

modules/plotting/src/properties/axisproperty.cpp
martinfalk: Used in lines 283/284
```c++
        labelSettings_.font_.anchorPos_.set(glm::yx(labelSettings_.font_.anchorPos_.get()));
        captionSettings_.font_.anchorPos_.set(glm::yx(captionSettings_.font_.anchorPos_.get()));
```

martinfalk:
Revised the internal state of the `AxisProperty` and `ColorScaleLegend`. Presets for axis placement and label alignment are available as buttons.

Additional changes:
* switched blending in PCP to pre-multiplied to be consistent with all other plots, text rendering, and the default blending in the background processor
* label and caption rotation in the AxisProperty are now handled identically
* added functionality to `2D Mesh Renderer` processor to use the Image of the outport for determining the view frustum. This allows rendering geometry directly in screen coordinates.

image

**2D Mesh Renderer drawing axis bounding boxes using screen coordinates**

modules/plotting/include/modules/plotting/datastructures/majorticksettings.h
```diff
@@ -50,8 +50,17 @@ class IVW_MODULE_PLOTTING_API MajorTickSettings {
     virtual double getTickDelta() const = 0;
     virtual bool getRangeBasedTicks() const = 0;
 };
+
 IVW_MODULE_PLOTTING_API bool operator==(const MajorTickSettings& a, const MajorTickSettings& b);
 IVW_MODULE_PLOTTING_API bool operator!=(const MajorTickSettings& a, const MajorTickSettings& b);
+
+namespace utilplot {
```
petersteneteg: Do we use utilplot anywhere else, otherwise I would just use util, or if you really want a different namespace I would just go with plot

modules/plotting/include/modules/plotting/datastructures/plottextsettings.h
```diff
@@ -56,6 +59,13 @@ class IVW_MODULE_PLOTTING_API PlotTextSettings {
 IVW_MODULE_PLOTTING_API bool operator==(const PlotTextSettings& a, const PlotTextSettings& b);
 IVW_MODULE_PLOTTING_API bool operator!=(const PlotTextSettings& a, const PlotTextSettings& b);
 
+namespace utilplot {
+/**
+ * flip inside and outside direction of placement \p p
+ */
+IVW_MODULE_PLOTTING_API LabelPlacement flip(LabelPlacement p);
```
petersteneteg, line 65: I think Inviwo::plot::untilplot::flip is a bit overkill

modules/plotting/include/modules/plotting/properties/tickproperty.h
```diff
@@ -57,6 +57,9 @@ class IVW_MODULE_PLOTTING_API MajorTickProperty : public MajorTickSettings,
     virtual std::string getClassIdentifier() const override;
     static const std::string classIdentifier;
 
+    MajorTickProperty(std::string_view identifier, std::string_view displayName, Document help,
```
petersteneteg: Im not sure If we need to pass in individual help text into each Composite property here. Maybe it is better just to hardcode the help text in the Constructor of the property...

modules/plotting/src/properties/axisproperty.cpp
```diff
@@ -112,56 +132,88 @@ AxisProperty::AxisProperty(std::string_view identifier, std::string_view display
...
-    adjustAlignment();
     updateLabels();
-
-    setCurrentStateAsDefault();
 }
 
+AxisProperty::AxisProperty(std::string_view identifier, std::string_view displayName,
+                           Orientation orientation, bool includeOrientationProperty,
+                           InvalidationLevel invalidationLevel, PropertySemantics semantics)
+    : AxisProperty{identifier,        displayName, {}, orientation, includeOrientationProperty,
```
petersteneteg: If you are updating the help added a default help text here would be good...

modules/plotting/src/properties/axisproperty.cpp
```diff
@@ -112,56 +132,88 @@ AxisProperty::AxisProperty(std::string_view identifier, std::string_view display
...
+    , alignment_{rhs.alignment_} {
 
-    addProperties(color_, width_, useDataRange_, range_, flipped_, orientation_, placement_);
+    addProperties(color_, width_, useDataRange_, range_, alignment_, scalingFactor_, mirrored_,
+                  captionSettings_, labelSettings_, majorTicks_, minorTicks_);
 
-    range_.setReadOnly(useDataRange_.get());
-    useDataRange_.onChange([this]() { range_.setReadOnly(useDataRange_.get()); });
+    // insert orientation property only if rhs owns one, too
+    if (rhs.getPropertyByIdentifier(rhs.orientation_.getIdentifier())) {
```
petersteneteg: might wrap the orientation one in an std::optional?

modules/plotting/src/properties/axisstyleproperty.cpp
```diff
@@ -55,18 +55,27 @@ AxisStyleProperty::AxisStyleProperty(std::string_view identifier, std::string_vi
                                      InvalidationLevel invalidationLevel,
                                      PropertySemantics semantics)
     : CompositeProperty(identifier, displayName, invalidationLevel, semantics)
```
petersteneteg: Might add a help text here to...

modules/plotting/src/properties/plottextproperty.cpp
```diff
@@ -46,47 +46,59 @@ const std::string PlotTextProperty::classIdentifier = "org.inviwo.PlotTextProper
...
+
+    addProperties(title_, placement_, color_, offset_, position_, rotation_, font_);
 
     font_.anchorPos_.set(vec2(0.0f, 0.0f));
 }
 
+PlotTextProperty::PlotTextProperty(std::string_view identifier, std::string_view displayName,
+                                   bool checked, InvalidationLevel invalidationLevel,
+                                   PropertySemantics semantics)
+    : PlotTextProperty(identifier, displayName, {}, checked, invalidationLevel, semantics) {}
```
petersteneteg: might add help text here

modules/plotting/src/properties/tickproperty.cpp
```diff
@@ -167,17 +170,17 @@ TickProperty::TickProperty(std::string_view identifier, std::string_view display
     : CompositeProperty(identifier, displayName, invalidationLevel, semantics)
     , majorTicks_("majorTicks", "Major Ticks")
     , minorTicks_("minorTicks", "Minor Ticks") {
-    addProperty(majorTicks_);
-    addProperty(minorTicks_);
+
+    addProperties(majorTicks_, minorTicks_);
```
petersteneteg: Could be just get rid of this property? Would be nice to not have to unfold that level since it is only 2 nodes in it?

petersteneteg (dismissed):
Nice Work! looking forward to trying it out!

modules/plotting/src/properties/tickproperty.cpp
martinfalk: Removed this property since it was used nowhere.

petersteneteg (dismissed):

petersteneteg (approved):

martinfalk:
@ResearchDaniel 
> Looks good overall, but have some questions and suggestions. The axes labels in the regression test seem to overlap the tick text. Is it possible avoid this? If not, is it possible to change the workspace so that it does not?

Resolved all questions and addressed the overlap in the regression test.

image

PR #1368, opened 2022-12-16
Textured iso renderer

petersteneteg:
Changes proposed in this PR:
 * Adds a TexturedIsoSurfaceRenderer. based on the ShaderComponent system. Basically it renders a iso surface using raycasting and then colors it using a different texture volume and TF

bild


modules/basegl/include/modules/basegl/shadercomponents/isocomponent.h
```diff
@@ -0,0 +1,64 @@
...
+#include <string>       // for string
+#include <string_view>  // for string_view
+#include <vector>       // for vector
+
+namespace inviwo {
+
+/**
+ * Adds a IsoValueProperty, and binds it to uniforms in the shader.
+ */
+class IsoComponent : public ShaderComponent {
```
ResearchDaniel, line 50: Not crucial, but this would make it consistent with naming of IsoValueProperty
```suggestion
class IsoValueComponent : public ShaderComponent {
```

modules/basegl/include/modules/basegl/shadercomponents/tfcomponent.h
```diff
@@ -0,0 +1,73 @@
...
+#include <vector>       // for vector
+
+#include <fmt/core.h>  // for format
+
+namespace inviwo {
+
+/**
+ * Adds a TransferFunctionProperty, and binds it to uniforms in the shader.
+ */
+class TFComponent : public ShaderComponent {
```
ResearchDaniel, line 57: Would also make this consistent with other naming, e.g., TransferFunctionProperty 
```suggestion
class TransferFunctionComponent : public ShaderComponent {
```

modules/basegl/src/processors/embeddedvolumeslice.cpp
```diff
@@ -204,6 +201,8 @@ void EmbeddedVolumeSlice::process() {
         utilgl::bindAndSetUniforms(shader_, units, backgroundPort_, ImageType::ColorDepthPicking);
     }
 
+    embeddedMesh_.setModelMatrix(inport_.getData()->getModelMatrix());
```
ResearchDaniel, line 204: Our general guideline nowadays is to avoid doing computations on `property.onChange`. Instead, check modifications in the process function and do the processing there.
So, in that case it would be better to check 
```
if (planePosition_.modified() || planeNormal_.modified()) {
    planeSettingsChanged(); 
}
```

martinfalk (dismissed):

modules/basegl/src/shadercomponents/volumecomponent.cpp
```diff
@@ -60,10 +64,11 @@ namespace {
 constexpr std::string_view uniforms = util::trim(R"(
 uniform VolumeParameters {0}Parameters;
 uniform sampler3D {0};
+uniform vec4 {0}Zero;
 )");
 
 constexpr std::string_view voxelFirst = util::trim(R"(
-vec4 {0}VoxelPrev = vec4(0);
+vec4 {0}VoxelPrev = vec4({0}Zero);
```
ResearchDaniel: Suggest to remove the uniform and use the formula instead
```suggestion
// Initialize previous value with 0 
vec4 {0}VoxelPrev = (0.0 + {0}Parameters.formatOffset) * (1.0 - {0}Parameters.formatScaling);
```

modules/basegl/include/modules/basegl/processors/raycasting/texturedisosurfacerenderer.h
```diff
@@ -0,0 +1,107 @@
...
+
+    IsoValueProperty* iso;
+    TransferFunctionProperty* tf;
+
+    OptionPropertyInt isoChannel;
+    OptionPropertyInt colorChannel;
+    FloatProperty samplingRate;
+};
+
+class IVW_MODULE_BASEGL_API TexturedIsoSurfaceRenderer : public VolumeRaycasterBase {
```
ResearchDaniel: Would be good to describe the functionality of this one. One has to dig quite deep into the code to understand it.

ResearchDaniel (changes requested):
If I understand it correctly from the code:
The iso-surface  color at location x is determined by mapping the value of a secondary volume through a TF?

One could say that the secondary volume contains a 1D texture coordinate.

I would have expected a "TexturedIsoSurfaceRenderer" to extract 2D texture coordinates from the surface and then map those coordinates through a 2D texture. 

Fair enough that we assume 1D texture coordinates using a TF, but perhaps we could find an alternative name that is not so strongly associated with surface texturing?

The zero volume initialization should be changed to not use a uniform in my opininon.

modules/basegl/src/processors/embeddedvolumeslice.cpp
petersteneteg: maybe, not really related to this pr though

modules/basegl/include/modules/basegl/processors/raycasting/texturedisosurfacerenderer.h
petersteneteg: we put that into the processorInfo for processors.

modules/basegl/src/shadercomponents/volumecomponent.cpp
petersteneteg: How sure are you that that is correct?

petersteneteg:
how would you extract 2D texture coordinates from an iso surface? 

what do you mean by 
> One could say that the secondary volume contains a 1D texture coordinate.
is a 3D texture used for texturing the surface

modules/basegl/include/modules/basegl/shadercomponents/tfcomponent.h
martinfalk: The reason for `TransferFunction` is legacy since we did not want to break too many things when introducing the IsoTFProperty.

modules/basegl/src/shadercomponents/volumecomponent.cpp
martinfalk: see https://github.com/inviwo/inviwo/blob/3bbed809efa238fcb1bd29356585e7b295e7d4dd/modules/opengl/glsl/utils/sampler3d.glsl#L43-L54

ResearchDaniel: Would probably be good to wrap in a function to make it more clear, voxelDataToNormalized or something

modules/basegl/include/modules/basegl/processors/raycasting/texturedisosurfacerenderer.h
ResearchDaniel: Ok, would be nice with a reference to that. Would make it easier when looking at the code.

modules/basegl/src/processors/raycasting/texturedisosurfacerenderer.cpp
```diff
@@ -0,0 +1,280 @@
...
+
+// The Class Identifier has to be globally unique. Use a reverse DNS naming scheme
+const ProcessorInfo TexturedIsoSurfaceRenderer::processorInfo_{
+    "org.inviwo.TexturedIsoSurfaceRenderer",  // Class identifier
+    "Textured Iso Surface Renderer",          // Display name
+    "Volume Rendering",                       // Category
+    CodeState::Stable,                        // Code state
+    Tags::GL,                                 // Tags
+    R"(Render a iso surface defined by the Volume data.
+       The iso surface is textured by the second color
```
ResearchDaniel: The second volume is not a color volume? 
If that were so there would be no need for a transfer function (we could just use the color volume).

The iso surface retrieves its color using the second volume’s value mapped through the transfer function. This makes it possible to visualize the other volume’s value at the the iso surface location

modules/basegl/src/shadercomponents/volumecomponent.cpp
petersteneteg: But all this does not say if it is the same or not!

petersteneteg: I know that what I have is correct.... so I don't se the point of spending all this time on a pointless change

modules/basegl/src/processors/raycasting/texturedisosurfacerenderer.cpp
petersteneteg: No the second volume is used to color the iso surface, it is not a volume of color values.

modules/basegl/src/shadercomponents/volumecomponent.cpp
ResearchDaniel: Yes, they must be the same.

Should it be initialized with zero in the first place? Why not equal to the first sample?

Uniforms can be expensive, at least it was a scarce resource before on GPUs, which could result in slower code. The volume component is a base component so as a precausion I think that it would be better to keep it as slim as possible.
In principle, I don't see a major reason for caching values when they can easily be derived from already uploaded uniforms.

petersteneteg: consider the other crapton of uniforms we set all the time. I would not be to bothered.

I still don't se why the need to be the same. The formatOffset and formatScaling. take into account the scaling done by opengl it self. Which the uniform I set does not. And should not do unless we change other part of the iso rendering. Hence Im not convinced the are the same.

ResearchDaniel: > The formatOffset and formatScaling. take into account the scaling done by opengl it self. Which the uniform I set does not.

I think that they are equivalent. Both should take the "raw" values and transform them to [0 1]. The formatOffset//formatScaling should be adapted to take the internal OpenGL format when doing the transformation. 

> Which the uniform I set does not. And should not do

The iso-surface computation must work in the same value space. Otherwise I don't see how it could be correct. The fact that it does work for you indicate that the computations indeed are the same.

modules/basegl/src/processors/raycasting/texturedisosurfacerenderer.cpp
ResearchDaniel: Ok, then I suggest that we update the text to something in the line of:
Render an iso surface defined by the Volume data, but color it based on the values of a second volume.
The iso surface retrieves its color using the second volume’s value mapped through the transfer function. This makes it possible to visualize the other volume’s value at the the iso surface location.

Then we are good to go

ResearchDaniel (dismissed):

martinfalk (approved):

PR #1372, opened 2022-12-21
Doc: change placeholders<modulePath> to ~modulePath~

petersteneteg:
Fixes help links on windows

### DIFF
--- a/modules/base/src/processors/convexhull2dprocessor.cpp
+++ b/modules/base/src/processors/convexhull2dprocessor.cpp
@@ -77,7 +77,7 @@ const ProcessorInfo ConvexHull2DProcessor::processorInfo_{
     Tags::CPU,                           // Tags
     R"(Computes the convex hull of a 2D mesh.
     
-    Example Workspace: [core/convexhull.inv](file:///<basePath>/data/workspaces/convexhull.inv)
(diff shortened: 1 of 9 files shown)


martinfalk (dismissed):

modules/base/src/processors/convexhull2dprocessor.cpp
```diff
@@ -77,7 +77,7 @@ const ProcessorInfo ConvexHull2DProcessor::processorInfo_{
     Tags::CPU,                           // Tags
     R"(Computes the convex hull of a 2D mesh.
     
-    Example Workspace: [core/convexhull.inv](file:///<basePath>/data/workspaces/convexhull.inv)
+    Example Workspace: [core/convexhull.inv](file:~basePath~/data/workspaces/convexhull.inv)
```
ResearchDaniel, line 80: The // should remain?

ResearchDaniel (changes requested):
Does not ~tilde~ cause a strikethrough?

modules/base/src/processors/convexhull2dprocessor.cpp
petersteneteg: No

petersteneteg:
> Does not ~tilde~ cause a strikethrough?

No not in a link
~ are allowed in links without escaping < are not...

ResearchDaniel (dismissed):

PR #1373, opened 2022-12-21
More apps

petersteneteg:
Changes proposed in this PR:
 * removed the inviwoapplicationqt to decrease system coupling and make it easer it integrate inviwo into other systems. Needed functionality move over to include/inviwo/qt/applicationbase/qtapptools.h. An example of how to set up a app can be seen in apps/minimals/qt/qtminimum.cpp
 * Better handling or runtime module loading
 * Better support for running inviwo from python


CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
...
+Previously the `PropertyListWidget` has shown the properties of all selected processors in the network. 
+This can sometimes be very slow to render for large selections, and in most cases one does not care about the properties when selection many processors.
+But one will usually do some operation on the processor network (more/copy/delete/etc). 
+So to make the behavior better we now only show the properties of the first selected processor if multiple ones are selected.
+But if you hold down `shift` we will fall back to the old behavior of showing all properties.
+
+## 2022-11-25 InviwoApplicationQt removed
+The `InviwoApplicationQt` class has been remove in favor of using a standalone `InviwoApplication` and a `QApplication` to reduce coupling and increase flexibility.
+The functionality that could be found in InviwoApplcationQt can now be access as a set of utility function in `qt/applicationbase/qtapptools.h`
+
```
martinfalk:
```suggestion
## 2022-11-25 The PropertyListWidget defaults to showing one processor
Previously the `PropertyListWidget` has shown the properties of all selected processors in the network. 
This can sometimes be very slow to render for large selections, and in most cases one does not care about the properties when selecting many processors.
But one will usually do some operation on the processor network (move/copy/delete/etc.). 
So to make the behavior better, we now only show the properties of the first selected processor if multiple processors are selected.
By holding down <kbd>shift</kbd> during the selection all properties will be shown.

## 2022-11-25 InviwoApplicationQt removed
The `InviwoApplicationQt` class has been removed in favor of using a standalone `InviwoApplication` and a `QApplication` to reduce the coupling and increase flexibility.
The functionality earlier found in `InviwoApplcationQt` can now be accessed as a set of utility functions in `qt/applicationbase/qtapptools.h`

```

include/inviwo/qt/applicationbase/qtapptools.h
```diff
@@ -0,0 +1,114 @@
...
+ */
+IVW_QTAPPLICATIONBASE_API void configurePoolResizeWait(InviwoApplication& app, QWidget* window);
+
+/**
+ * Sets a new style sheet file in the QApplication
+ */
+IVW_QTAPPLICATIONBASE_API void setStyleSheetFile(std::string_view file);
+
+/**
+ * \brief getCurrentStdLocale
```
martinfalk:
```suggestion
```

modules/python3/src/pythoninterpreter.cpp
```diff
@@ -71,11 +73,55 @@ PythonInterpreter::PythonInterpreter() : embedded_{false}, isInit_(false) {
     if (!Py_IsInitialized()) {
 
         LogInfo("Python version: " + toString(Py_GetVersion()));
-        static wchar_t programName[] = L"PyInviwo";
-        Py_SetProgramName(programName);
 
         try {
             py::initialize_interpreter(false);
+
```
martinfalk: remove commented block?

martinfalk (dismissed):

modules/python3/src/pythoninterpreter.cpp
petersteneteg: mmm

PR #1374, opened 2022-12-21
Feature/animation rendering

petersteneteg:
Changes proposed in this PR:
 * Make the rendering mode in the animation editor usable

### DIFF
--- a/modules/animation/include/modules/animation/datastructures/invalidationtrack.h
+++ b/modules/animation/include/modules/animation/datastructures/invalidationtrack.h
@@ -51,6 +51,9 @@ class ProcessorNetwork;
 
 namespace animation {
 
+/**
+ * @see InvalidationTrack
+ */
 class IVW_MODULE_ANIMATION_API InvalidationKeyframe : public BaseKeyframe {
 public:
     using value_type = void;
@@ -60,10 +63,11 @@ public:
     InvalidationKeyframe& operator=(const InvalidationKeyframe&);
     virtual ~InvalidationKeyframe();
     virtual InvalidationKeyframe* clone() const override;
-
-private:
 };
 
+/**
+ * @see InvalidationTrack
+ */
 class IVW_MODULE_ANIMATION_API InvalidationKeyframeSequence
     : public BaseKeyframeSequence<InvalidationKeyframe> {
 public:
@@ -79,9 +83,13 @@ public:
     virtual void serialize(Serializer& s) const override;
     virtual void deserialize(Deserializer& d) override;
 
-    std::string processorId;
+    std::string path;
 };
 
+/**
+ * A special track to invalidate processors and or properties. Mainly useful for debugging of if a
+ * processor has side effects. Can also be useful as a quick profiling tool.
+ */
 class IVW_MODULE_ANIMATION_API InvalidationTrack : public BaseTrack<InvalidationKeyframeSequence> {
 public:
     InvalidationTrack(ProcessorNetwork* network);
@@ -98,6 +106,6 @@ public:
     ProcessorNetwork* network;
 };
 
-}  // namespace animation
+}  //namespace animation
 
-}  // namespace inviwo
+}  //namespace inviwo
--- a/modules/animation/include/modules/animation/datastructures/invalidationtrack.h
+++ b/modules/animation/include/modules/animation/datastructures/invalidationtrack.h
@@ -1,0 +1,103 @@
+/*********************************************************************************
+ *
+ * Inviwo - Interactive Visualization Workshop
+ *
+ * Copyright (c) 2022 Inviwo Foundation
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ *********************************************************************************/
+#pragma once
+
+#include <modules/animation/animationmoduledefine.h>
+
+#include <inviwo/core/io/serialization/deserializer.h>  // for ContainerWrapper<>::Item
+#include <inviwo/core/util/exception.h>                 // for Exception
+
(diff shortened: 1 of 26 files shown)


modules/animation/src/datastructures/invalidationtrack.cpp
```diff
@@ -0,0 +1,93 @@
...
+InvalidationTrack::~InvalidationTrack() = default;
+
+InvalidationTrack* InvalidationTrack::clone() const { return new InvalidationTrack(*this); }
+
+std::string InvalidationTrack::classIdentifier() {
+    return "org.inviwo.animation.InvalidationTrack";
+}
+std::string InvalidationTrack::getClassIdentifier() const { return classIdentifier(); }
+
+AnimationTimeState InvalidationTrack::operator()(Seconds from, Seconds to,
```
martinfalk: `from` unused?

martinfalk (dismissed):

modules/animation/include/modules/animation/datastructures/invalidationtrack.h
```diff
@@ -0,0 +1,103 @@
...
+
+    virtual InvalidationKeyframeSequence* clone() const override;
+
+    virtual void serialize(Serializer& s) const override;
+    virtual void deserialize(Deserializer& d) override;
+
+    std::string processorId;
+};
+
+class IVW_MODULE_ANIMATION_API InvalidationTrack : public BaseTrack<InvalidationKeyframeSequence> {
```
ResearchDaniel, line 93: I think that the connection to what is invalidated could be more explicit as both properties and processors can be invalidated.
I am also not sure about the use case for this track. If a processor has not been invalidated there will not be any change in its output. Care to elaborate on this by documenting the class?
```suggestion
class IVW_MODULE_ANIMATION_API ProcessorInvalidationTrack : public BaseTrack<InvalidationKeyframeSequence> {
```

ResearchDaniel: If the name is changed I also think that corresponding changes should be made to the KeyFrame and the KeyFrameSequence

modules/animation/src/animationcontroller.cpp
```diff
@@ -135,105 +112,79 @@ AnimationController::AnimationController(Animation& animation, InviwoApplication
...
+                     exts.begin(), exts.end(), std::back_inserter(state.options),
+                     [](const auto& ext) -> OptionPropertyOption<FileExtension> { return ext; });
+                 auto it = std::find_if(exts.begin(), exts.end(),
+                                        [&](auto& e) { return e.extension_ == "png"; });
+                 if (it != exts.end()) {
+                     state.selectedIndex = std::distance(exts.begin(), it);
+                 }
+                 return state;
+             }())
+    , renderFPS("renderFPS", "Frames per Second", 100, 2, 1000000, 1,
```
ResearchDaniel: Ranges need to be adapted because the meaning of this property changed
```suggestion
    , renderFPS("renderFPS", "Frames per Second", 24.0,
                      {1, ConstraintBehavior::Immutable},
                      {60, ConstraintBehavior::Editable}, 1.0
```

modules/animation/src/animationcontroller.cpp
```diff
@@ -87,42 +88,18 @@ class Layer;
...
                      0)
     , playWindow("PlayFirstLastTime", "Window", 0, 10, 0, 1e5, 1, 0.0,
                  InvalidationLevel::InvalidOutput, PropertySemantics::Text)
-    , framesPerSecond(
-          "PlayFramesPerSecond", "Frames per Second",
-          [this]() { return 1.0 / std::abs(deltaTime_.count()); },
-          [this](const double& value) { deltaTime_ = Seconds(1.0 / value); },
-          {0.001, ConstraintBehavior::Immutable}, {1000.0, ConstraintBehavior::Immutable}, 1.0,
-          InvalidationLevel::InvalidOutput, PropertySemantics::Text)
+    , framesPerSecond("PlayFramesPerSecond", "Frames per Second", 25.0,
```
ResearchDaniel: Out of scope of PR, but not sure why it was 25 before. I think that 24, 30, or 60 would make most sense.
```suggestion
    , framesPerSecond("PlayFramesPerSecond", "Frames per Second", 24.0,
```

modules/animation/src/animationcontroller.cpp
```diff
@@ -469,46 +303,88 @@ void AnimationController::tick() {
...
+        return std::max(d, 4);
+    }();
+
+    // Get all active canvases
+    std::vector<std::pair<Exporter*, std::string>> exporters;
+    network->forEachProcessor([&](Processor* p) {
+        if (p->isSink()) {
+            if (auto exporter = dynamic_cast<Exporter*>(p)) {
+                auto base = renderBaseName.get();
+                replaceInString(base, "UPN", p->getIdentifier());
```
ResearchDaniel, line 341: Where does this UPN come from and what is its purpose?

modules/animation/src/animationcontroller.cpp
```diff
@@ -469,46 +303,88 @@ void AnimationController::tick() {
...
+
+    // render frames
+    for (int currentFrame = 0; currentFrame < numFrames; ++currentFrame) {
+        // Evaluate animation
+        Seconds newTime = firstTime + (lastTime - firstTime) / (numFrames - 1) * currentFrame;
+        eval(currentTime_, newTime);
+
+        lock.reset();
+
+        while (app_->getProcessorNetwork()->runningBackgroundJobs() > 0) {
```
ResearchDaniel, line 365: This functionality seems to be a reocurring pattern.
See for example `waitForNetwork` in the python API.
Should we expose this in the application or is it already there through `waitForPool` (should processEvents be included in waitForPool?)? 
If not, what is the difference between these ways of waiting?

modules/animation/src/animationcontroller.cpp
```diff
@@ -469,46 +303,88 @@ void AnimationController::tick() {
...
     }
 
-    // Evaluate animation
-    eval(currentTime_, newTime);
+    using duration_double = std::chrono::duration<double, std::ratio<1>>;
+    auto seconds = std::chrono::duration_cast<duration_double>(
+                       std::chrono::high_resolution_clock::now() - start)
+                       .count();
+    LogInfo(fmt::format("Rendered {} frames in {:.3f} seconds, {:.3f} per frame", numFrames,
+                        seconds, seconds / numFrames));
```
ResearchDaniel, line 387: Does use of int not produce a warning?
```suggestion
                        seconds, seconds / static_cast<double>(numFrames)));
```

ResearchDaniel (changes requested):

modules/animation/src/animationcontroller.cpp
petersteneteg: Think it's an old invention by Rickard. it means "use processor name"
There are probably better ways to do that... but that's how we have done it for a long time...

modules/animation/src/animationcontroller.cpp
petersteneteg: I think double / int should be fine

modules/animation/src/animationcontroller.cpp
```diff
@@ -91,67 +91,65 @@ namespace animation {
 AnimationController::AnimationController(Animation& animation, InviwoApplication* app)
     : playOptions("PlayOptions", "Play Settings")
     , playWindowMode("PlayFirstLastTimeOption", "Time",
-                     {{"FullTimeWindow", "Play full animation", 0},
+                     {
+                         {"FullTimeWindow", "Play full animation",  0},
+                         {"UserTimeWindow", "Selected time window", 1}
```
ResearchDaniel: Not sure how much is given by the context, but it might be better with:
“Play selected”
Or 
“Play selected frames”

modules/animation/src/animationcontroller.cpp
```diff
@@ -425,6 +423,6 @@ Seconds AnimationController::deltaTime() const {
 
 Seconds AnimationController::getCurrentTime() const { return currentTime_; }
 
-}  // namespace animation
```
ResearchDaniel, line 586: Is this a new format setting?
Personally, I think that there should be a space

modules/animationqt/src/sequenceeditor/invalidationsequenceeditor.cpp
```diff
@@ -130,15 +130,15 @@ InvalidationSequenceEditor::InvalidationSequenceEditor(KeyframeSequence& sequenc
     layout->addLayout(subLayout);
     subLayout->setContentsMargins(0, 0, 0, 0);
     subLayout->setSpacing(7);
-    processor_ = new QLineEdit();
-    processor_->setToolTip("Invalidate the specified processor at each network evaluation");
-    subLayout->addWidget(new QLabel("Processor Identifier"), 0, 0);
-    subLayout->addWidget(processor_, 0, 1);
+    path_ = new QLineEdit();
+    path_->setToolTip("Invalidate the specified processor / property at each network evaluation");
```
ResearchDaniel: Suggest to add an example as people might not know what a path is/looks like:
(processorId.propertyId, e.g., Canvas or Canvas.saveLayer)

ResearchDaniel (dismissed):
Space after comment “//“ should not be removed?

modules/animation/src/animationcontroller.cpp
petersteneteg: yes, I have added it back. had some local version update of clang-format which did the wrong thing...

ResearchDaniel (approved):

PR #1378, opened 2023-01-12
Feature/improvments

petersteneteg:
__Changes proposed in this PR:__
* Tools: remove VLD
* Core/Qt: use one widget for all option properties
* Various: better errors in data.h
* Core: Mesh add constexpr
* Animation: threaded image save when randering
* DataFrame: added column constructor for data add lookup
* Qt: editor added function for growing a processor selection
* Base: More volume info


modules/animation/src/animationcontroller.cpp
```diff
@@ -369,9 +378,27 @@ void AnimationController::render() {
...
+            std::visit(util::overloaded{
+                           [&](Exporter* exporter) {
+                               exporter->exportFile(renderLocation.get(), file, {writer.get()},
+                                                    Overwrite::Yes);
+                           },
+                           [&](CanvasProcessor* cp) {
+                               // Hackish: Make sure LayerRAM is the last valid rep, so that
+                               // is the one that will be cloned, This also forces the
+                               // download to happen on the main thread instead of in the back
+                               // ground.
```
martinfalk:
```suggestion
                               // Hackish: Make sure LayerRAM is the last valid rep, so that it
                               // is the one that will be cloned. This also forces the
                               // download to happen on the main thread instead of in the 
                               // background.
```

modules/base/include/modules/base/processors/volumeinformation.h
```diff
@@ -70,16 +70,21 @@ class IVW_MODULE_BASE_API VolumeInformation : public Processor {
     DoubleMinMaxProperty minMaxChannel3_;
     DoubleMinMaxProperty minMaxChannel4_;
 
-    FloatMat4Property worldTransform_;
     FloatMat3Property basis_;
     FloatVec3Property offset_;
+    DoubleVec3Property voxelSize_;
+
+    FloatMat4Property modelMatrix_;
```
martinfalk, line 77: add similar functionality to `MeshInformation`?

modules/basegl/src/shadercomponents/atlascomponent.cpp
```diff
@@ -297,6 +297,13 @@ void AtlasComponent::process(Shader& shader, TextureUnitContainer& cont) {
 
         util::IndexMapper2D im(colors_.getDimensions());
 
+        auto indexCheck = [&](uint32_t i, std::string_view type) {
+            if (i >= nSegments) {
+                throw Exception(IVW_CONTEXT, "Fund {} index {} outside of expected range [0,{})",
```
martinfalk:
```suggestion
                throw Exception(IVW_CONTEXT, "{} index {} outside of expected range [0,{})",
```

martinfalk (dismissed):
some minor comments

martinfalk (approved):

Issue #1379, opened 2023-01-30
[BUG]: Dangling pointer in RepresentationFactoryManager upon InviwoApplication destruction


ResearchDaniel:
### Required prerequisites

- [X] Make sure you've read the [documentation](https://inviwo.org). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/inviwo/inviwo/issues) and [Discussions](https://github.com/inviwo/inviwo/discussionss) to verify that this hasn't already been reported. +1 or comment there if it has.
- [ ] Consider asking first in [Slack](https://join.slack.com/t/inviwo/shared_invite/enQtNTc2Nzc2NDQwNzIxLTRiMWM1ZWJiYjljZjkwNWE3OTk3MzYxODZlMDUyMzRmZjUzMzBiZjVhNTM3NWUyNzU1MjI4OWJjMzdkODViMzM) or in a [Discussion](https://github.com/inviwo/inviwo/discussionss).

### Problem description

`RepresentationFactoryManager` keeps reference to InviwoApplication's representation factory.
If InviwoApplication is destroyed and initialized again RepresentationFactoryManager::representationMetaFactory_ will point to the destroyed instance. 
 
[RepresentationFactoryManager::getRepresentationFactory() ](https://github.com/inviwo/inviwo/blob/a832ca089466d8bf50c221eed7974cbe10708398/include/inviwo/core/datastructures/representationfactorymanager.h#L186)

One possible solution:
Add functionality to reset the static factories and do this in the InviwoApplication destructor.

Any objections to such a solution or suggestions on another one?

petersteneteg:
Don't think we support destroying and recreating InviwoApplication

PR #1381, opened 2023-01-31
Core: Fix potential dangling pointer in RepresentationFactoryManager

ResearchDaniel:
Closes issue #1379


src/core/tests/unittests/without-initialization/init-deinit-tests.cpp
```diff
@@ -0,0 +1,65 @@
...
+#include <inviwo/core/common/inviwoapplication.h>
+#include <inviwo/core/datastructures/buffer/buffer.h>
+#include <inviwo/core/common/inviwomodulefactoryobject.h>
+#include <inviwo/core/common/coremodulesharedlibrary.h>
+
+namespace inviwo {
+
+void forceBufferRepresentationCreation() {
+    LogCentral::init();
+    util::OnScopeExit deleteLogcentral([]() { LogCentral::deleteInstance(); });
```
petersteneteg:
```suggestion
    inviwo::LogCentral logger;
    inviwo::LogCentral::init(&logger);
```

src/core/CMakeLists.txt
```diff
@@ -800,7 +800,10 @@ set(TEST_FILES
     tests/unittests/volumesequenceutils-tests.cpp
     tests/unittests/zip-test.cpp
 )
+# Default inviwo unit tests run after InviwoApplication initialization
 ivw_add_unittest(${TEST_FILES})
+# Tests for initialization/deinitialization of InviwoApplication 
+add_subdirectory(tests/unittests/without-initialization)
```
petersteneteg: I think it would be better to keep this outside of core, and also have it graded by IVW_TEST_INTEGRATION_TESTS
and not IVW_TEST_UNIT_TESTS

hence I would add the test here instead. 
https://github.com/inviwo/inviwo/blob/a832ca089466d8bf50c221eed7974cbe10708398/CMakeLists.txt#L109

that way we might also enable more modules and so on to increase the test surface in the future? 

I would then more the files to /tests/recycle-inviwoapplication/ to something like that.
without-initialization does not seem very descriptive of the test per say...

petersteneteg (changes requested):
Some minor organizational changes, cool that even this seems to work! we should try adding more modules...

src/core/CMakeLists.txt
ResearchDaniel: That is actually where I placed it first, but since the tests only dealt with core I thought it would be better to keep it there. 

I can move it and do renaming, but i will keep core as the only dependency for now. Another PR could in that case extend the testing to check more modules with singletons.

src/core/tests/unittests/without-initialization/init-deinit-tests.cpp
ResearchDaniel: This suggested change will leave the LogCentral with a dangling pointer when `logger `goes out of scope

src/core/CMakeLists.txt
petersteneteg: yes, that was my thinking also

Eventually the goal would be that it would work with all modules.

src/core/tests/unittests/without-initialization/init-deinit-tests.cpp
petersteneteg: Nope, it will be set to nullptr automatically by the Singleton destructor.

petersteneteg (dismissed):
Any special reason for putting it inside integrationtests and not next to it?

src/core/tests/unittests/without-initialization/init-deinit-tests.cpp
ResearchDaniel: Fixed. Also changed in the integration tests main function

ResearchDaniel:
> Any special reason for putting it inside integrationtests and not next to it?

Clarity. As GTest automatically runs the tests, it would be hard to understand which main function that is running for the tests if they would be in the same folder.

The other alternative would be to use one main function and a special init/deinit for the different types of tests

petersteneteg (approved):
I meant in a folder next to the integration-test folder :)

Issue #1391, opened 2023-03-16
[BUG]: Assimp import broken in vcpkg build


martinfalk:
### Required prerequisites

- [X] Make sure you've read the [documentation](https://inviwo.org). Your issue may be addressed there.
- [X] Search the [issue tracker](https://github.com/inviwo/inviwo/issues) and [Discussions](https://github.com/inviwo/inviwo/discussionss) to verify that this hasn't already been reported. +1 or comment there if it has.
- [X] Consider asking first in [Slack](https://join.slack.com/t/inviwo/shared_invite/enQtNTc2Nzc2NDQwNzIxLTRiMWM1ZWJiYjljZjkwNWE3OTk3MzYxODZlMDUyMzRmZjUzMzBiZjVhNTM3NWUyNzU1MjI4OWJjMzdkODViMzM) or in a [Discussion](https://github.com/inviwo/inviwo/discussionss).

### Problem description

Using Assimp via vcpkg instead of submodules breaks the import of non-triangle meshes like quad meshes. The `DrawType` is determined incorrectly which results in point clouds instead of regular meshes. The AssimpReader also assumes the same primitive type in all meshes in one file.

![image](https://user-images.githubusercontent.com/9251300/225662323-ca5d38fa-a0ab-4519-8509-8e113d2dd95b.png)

### Inviwo version

982bd77cb3707f7878edf5273809b35f29e197b8

### Operating system?

Windows

PR #1392, opened 2023-03-16
Feature/assimp fix

### LINKED ISSUE #1390: [BUG]: file names not serialized when using data visualizers
### Required prerequisites

### Problem description

Workspaces where files have been dropped into the Inviwo editor do not correctly serialize the file names. When loading such a workspace, the file names are reset to empty string.

Affects meshes, volumes, data frames.

### Inviwo version

982bd77cb3707f7878edf5273809b35f29e197b8

### Operating system?

Windows

### Compiler / Version

_No response_

### Qt Version

_No response_

### Reproducible example code

```text
+ drop any mesh (e.g. from inviwo/modules/userinterfacegl/data/meshes/) onto Inviwo and select "Mesh renderer"
+ save workspace
+ reload workspace

The filename in Mesh source is empty and no data is loaded.
```

### Inviwo log

_No response_

martinfalk:
Fixes #1390, #1391

Changes proposed in this PR:
 * fixed file import in assimp reader (DrawType set incorrectly for n-gons)
 * assimp reader creates multiple index buffers per mesh based on geometry types
 * serialization fix for data visualizers (file names not serialized)


modules/base/include/modules/base/processors/datasource.h
```diff
@@ -98,13 +99,17 @@ class DataSource : public Processor {
...
      */
-    DataSource(DataReaderFactory* rf = util::getDataReaderFactory(), std::string_view file = "",
+    DataSource(DataReaderFactory* rf = util::getDataReaderFactory(), std::string_view filename = "",
                std::string_view contentType = FileProperty::defaultContentType);
     virtual ~DataSource() = default;
 
     virtual void process() override;
     virtual void deserialize(Deserializer& d) override;
 
+    FileProperty file;
```
ResearchDaniel:
```suggestion
    FileProperty filePath;
```

modules/base/include/modules/base/processors/datasource.h
```diff
@@ -44,24 +44,25 @@
...
     } else {
-        reader.setSelectedValue(file.getSelectedExtension());
+        extReader.setSelectedValue(file.getSelectedExtension());
     }
 }
 
 template <typename... Types>
 void updateFilenameFilters(const DataReaderFactory& rf, FileProperty& file,
-                           OptionProperty<FileExtension>& reader) {
+                           OptionProperty<FileExtension>& extReader) {
```
ResearchDaniel: I don’t understand why the name should be changed.

modules/base/include/modules/base/processors/volumesource.h
```diff
@@ -60,7 +60,7 @@ class IVW_MODULE_BASE_API VolumeSource : public Processor {
     virtual const ProcessorInfo getProcessorInfo() const override;
     static const ProcessorInfo processorInfo_;
 
-    VolumeSource(InviwoApplication* app, const std::string& file = "");
+    VolumeSource(InviwoApplication* app, std::string_view file = {});
```
ResearchDaniel:
```suggestion
    VolumeSource(InviwoApplication* app, std::string_view filePath = {});
```

modules/base/src/processors/volumesource.cpp
```diff
@@ -82,7 +82,7 @@ const ProcessorInfo VolumeSource::processorInfo_{
 
 const ProcessorInfo VolumeSource::getProcessorInfo() const { return processorInfo_; }
 
-VolumeSource::VolumeSource(InviwoApplication* app, const std::string& file)
+VolumeSource::VolumeSource(InviwoApplication* app, std::string_view file)
```
ResearchDaniel:
```suggestion
VolumeSource::VolumeSource(InviwoApplication* app, std::string_view filePath)
```

modules/base/src/processors/volumesource.cpp
```diff
@@ -98,6 +98,11 @@ VolumeSource::VolumeSource(InviwoApplication* app, const std::string& file)
     addProperties(file_, reader_, reload_, information_, basis_, volumeSequence_);
     volumeSequence_.setVisible(false);
 
+    // ensure that the file name is always serialized by setting the default to ""
+    if (!file.empty()) {
```
ResearchDaniel: Does this make sense? Should be
```suggestion
    if (file.empty()) {
```

ResearchDaniel (changes requested):

petersteneteg (dismissed):

modules/base/include/modules/base/processors/datasource.h
petersteneteg: filename conveys no more relevant information information then file. If you want a different name it should convey more meaning. In the case `filePath` would be more descriptive.

ResearchDaniel: Agree, edited the suggestion

modules/base/include/modules/base/processors/datasource.h
martinfalk: Since this here refers to list entries tying an extension to a certain reader.

modules/base/src/processors/volumesource.cpp
martinfalk: no, in case the file name is empty, the default is an empty string "" as well. We need to override the default in case `filePath` is set. This is due to inherent mechanism of property constructors.
If the file name is set for the constructor, it will not be serialized unless changed. See #1390

modules/base/include/modules/base/processors/datasource.h
ResearchDaniel: But it is not an extension reader, which I think that the name implies. Would not `extensions` be more appropriate if you want to change the name?

modules/base/include/modules/base/processors/datasource.h
```diff
@@ -93,18 +94,22 @@ class DataSource : public Processor {
     /**
      * Construct a DataSource
      * @param app An InviwoApplication.
-     * @param file A filename passed into the FileProperty
+     * @param filePath A filename passed into the FileProperty
```
ResearchDaniel:
```suggestion
     * @param filePath A file path passed into the FileProperty
```

modules/base/include/modules/base/processors/datasource.h
```diff
@@ -126,57 +128,63 @@ class DataSource : public Processor {
 };
 
 template <typename DataType, typename PortType>
-DataSource<DataType, PortType>::DataSource(DataReaderFactory* rf, std::string_view file,
+DataSource<DataType, PortType>::DataSource(DataReaderFactory* rf, std::string_view filePath,
                                            std::string_view content)
     : Processor()
+    , filePath{"filename", "File", filePath, content}
```
ResearchDaniel, line 134: To make the Python api consistent with the c++ api (could break code in many places and requires a network update): 
```suggestion
    , filePath{"filePath", "File", filePath, content}
```
Not sure if the naming refactoring of the property is worth the trouble. Should be done in a separate PR

modules/base/src/processors/volumesource.cpp
```diff
@@ -98,6 +98,11 @@ VolumeSource::VolumeSource(InviwoApplication* app, const std::string& file)
     addProperties(file_, reader_, reload_, information_, basis_, volumeSequence_);
     volumeSequence_.setVisible(false);
 
+    // ensure that the file name is always serialized by setting the default to ""
```
ResearchDaniel: I think that a better solution could be to add the option of specifying the default value in the FileProperty constructor. Otherwise:
```suggestion
    // The FileProperty default state is the one provided in the constructor (filePath). The default state is not serialized. So, to ensure that a custom file path is serialied we set the default state to ""
```

ResearchDaniel (changes requested):
I think that a better option would be to add the default value option to the FileProperty constructor.

also, as the file name refactoring is a breaking change I think that it should be done in a separate PR. Here, I think that it should be ensured that the identifiers are consistent with the naming in C++. However, this would break a lot of code using the source processors, so I’m not sure if the refactoring is worth it.

modules/base/src/processors/volumesource.cpp
ResearchDaniel: No need for the if statement then

ResearchDaniel:
Come to think of it. Perhaps the initial default state of the file property should always be “”?

modules/base/include/modules/base/processors/datasource.h
martinfalk: that was by accident. Will be reverted.

modules/base/src/processors/volumesource.cpp
martinfalk: This will affect all our properties. In this case, FileProperty is based on TemplateProperty<std::string>. Will use your suggestion.

ResearchDaniel: No, would work by making the change in the FileProperty constructor?
https://github.com/inviwo/inviwo/blob/master/src/core/properties/fileproperty.cpp#L50

modules/base/src/processors/volumesource.cpp
martinfalk: What about the overhead of `Property::setStateAsDefault()`? That is not necessary if no file name was set.
https://github.com/inviwo/inviwo/blob/903f8b9e1253c4a7b8acb79c2bd234661bccb7e3/include/inviwo/core/properties/property.h#L469

petersteneteg (dismissed):

petersteneteg (approved):

martinfalk:
@ResearchDaniel 
> Come to think of it. Perhaps the initial default state of the file property should always be “”?

Changed in 32f5288

ResearchDaniel (approved):

PR #1399, opened 2023-03-24
Feature/layerpy examples

martinfalk:
Fixes
 * `VolumePy` representation destroyed by Python

Changes proposed in this PR:
 * added `LayerPy` including converters to RAM and GL representations based on `VolumePy`
 * revised python scripts and processors
 * added Python processors demonstrating the use of `Layer`, `LayerPy`, `Volume`, and `VolumePy` as well as `Mesh` and `BasicMesh`


petersteneteg (dismissed):

modules/basegl/src/processors/imageprocessing/imageglprocessor.cpp
```diff
@@ -149,15 +153,22 @@ void ImageGLProcessor::createDefaultImage(const size2_t& dim, ImageInport& inpor
...
+    const auto wrapping = colorLayer->getWrapping();
+    const auto interpolation = colorLayer->getInterpolation();
 
     if (!outport.hasEditableData() || format != outport.getData()->getDataFormat() ||
         dim != outport.getData()->getDimensions() ||
-        swizzleMask != outport.getData()->getColorLayer()->getSwizzleMask()) {
+        swizzleMask != outport.getData()->getColorLayer()->getSwizzleMask() ||
+        wrapping != outport.getData()->getColorLayer()->getWrapping() ||
+        interpolation != outport.getData()->getColorLayer()->getInterpolation()) {
         Image* img = new Image(dim, format);
```
petersteneteg: Wny not just clone the input image? or use Image(const Image& rhs, NoData);

petersteneteg: Also you don't need to create a new image just to change the wrapping or interpolation for example...

petersteneteg: Would also be good to put all this in a well named function...

modules/python3qt/templates/templateprocessor.py
```diff
@@ -21,7 +21,8 @@ def processorInfo():
             displayName="{name}",
             category="Python",
             codeState=ivw.CodeState.Stable,
-            tags=ivw.Tags.PY
+            tags=ivw.Tags.PY,
+            help=ivw.md2doc("")
```
petersteneteg: maybe we should use something like help=ivw.md2doc({name}.__doc__) here...

petersteneteg (dismissed):

petersteneteg (dismissed):

petersteneteg (approved):

PR #1405, opened 2023-04-14
Use standard FindHDF5 cmake module for system-provided HDF5 build

rartino:
This adjusts the modules/hdf5/CMakeLists.txt when using system libraries. This makes it possible to build Inviwo on Ubuntu 22.04 against its apt-provided hdf5 packages.

I'm not sure what "hdf5"-finding module the present cmake setup is supposed to be using (?), but the [FindHDF5](https://cmake.org/cmake/help/latest/module/FindHDF5.html) this PR changes it into seems to have been standard in cmake for quite a while.

### DIFF
--- a/modules/fontrendering/CMakeLists.txt
+++ b/modules/fontrendering/CMakeLists.txt
@@ -50,6 +50,8 @@ ivw_handle_shader_resources(${CMAKE_CURRENT_SOURCE_DIR}/glsl ${SHADER_FILES})
 # Add font directory to pack
 ivw_add_to_module_pack(${CMAKE_CURRENT_SOURCE_DIR}/fonts)
 
+find_package(utf8cpp CONFIG REQUIRED)
+
 option(IVW_USE_EXTERNAL_FREETYPE "Freetype is provided externaly" OFF)
 if(NOT IVW_USE_EXTERNAL_FREETYPE)
     add_subdirectory(ext/freetype)
@@ -60,17 +62,33 @@ if(NOT IVW_USE_EXTERNAL_FREETYPE)
     )
     ivw_default_install_targets(freetype)
     ivw_make_package(freetype freetype)
-else()
-    find_package(freetype CONFIG REQUIRED)
-    ivw_vcpkg_install(freetype MODULE FontRendering)
-endif()
 
-find_package(utf8cpp CONFIG REQUIRED)
-target_link_libraries(inviwo-module-fontrendering 
-    PUBLIC 
-        freetype
-    PRIVATE
-        utf8cpp
-)
+    target_link_libraries(inviwo-module-fontrendering
+        PUBLIC
+            freetype
+        PRIVATE
+            utf8cpp
+    )
+
+else()
+    find_package(freetype CONFIG)
+    if(FREETYPE_FOUND)
+        ivw_vcpkg_install(freetype MODULE FontRendering)
+        target_link_libraries(inviwo-module-fontrendering
+            PUBLIC
+            freetype
+            PRIVATE
+            utf8cpp
+        )
+    else()
+      find_package(Freetype REQUIRED)
+      target_link_libraries(inviwo-module-fontrendering
+        PUBLIC
+            Freetype::Freetype
+        PRIVATE
+            utf8cpp
+        )
+    endif()
+endif()
 
 ivw_make_package(InviwoFontRenderingModule inviwo-module-fontrendering)
--- a/modules/hdf5/CMakeLists.txt
+++ b/modules/hdf5/CMakeLists.txt
@@ -109,7 +109,8 @@ if (NOT IVW_USE_EXTERNAL_HDF5)
     ivw_default_install_targets(${h5-targets})
 
 else()
-    find_package(hdf5 REQUIRED)
+    # Try vcpkg-provided Findhdf5
+    find_package(hdf5 QUIET)
     if(HDF5_FOUND)
         if(BUILD_SHARED_LIBS)
             target_link_libraries(inviwo-module-hdf5
@@ -131,6 +132,7 @@ else()
         ivw_vcpkg_install(hdf5 MODULE HDF5)
         set(version ${HDF5_VERSION_STRING})
     else()
+        # Try cmake standard FindHDF5
         if(NOT BUILD_SHARED_LIBS)
            set(HDF5_USE_STATIC_LIBRARIES ON)
         endif()
--- a/modules/hdf5/CMakeLists.txt
+++ b/modules/hdf5/CMakeLists.txt
@@ -109,24 +109,17 @@ if (NOT IVW_USE_EXTERNAL_HDF5)
     ivw_default_install_targets(${h5-targets})
 
 else()
-    find_package(hdf5 REQUIRED)
-    if(BUILD_SHARED_LIBS)
-        target_link_libraries(inviwo-module-hdf5 
-            PUBLIC 
-                hdf5::hdf5-shared 
-                hdf5::hdf5_hl-shared 
-                hdf5::hdf5_cpp-shared 
-                hdf5::hdf5_hl_cpp-shared
-        )
-    else()
-        target_link_libraries(inviwo-module-hdf5 
-            PUBLIC 
-                hdf5::hdf5-static 
-                hdf5::hdf5_hl-static 
-                hdf5::hdf5_cpp-static 
-                hdf5::hdf5_hl_cpp-static
-        )
+    if(NOT BUILD_SHARED_LIBS)
+        set(HDF5_USE_STATIC_LIBRARIES ON)
     endif()
+    find_package(HDF5 REQUIRED)
+    target_link_libraries(inviwo-module-hdf5
+        PUBLIC
+        ${HDF5_LIBRARIES}
+        ${HDF5_HL_LIBRARIES}
+        ${HDF5_CXX_LIBRARIES}
+        ${HDF5_HL_CXX_LIBRARIES}
+    )
     ivw_vcpkg_install(hdf5 MODULE HDF5)
     set(version ${HDF5_VERSION_STRING})
 endif()
--- a/modules/hdf5/CMakeLists.txt
+++ b/modules/hdf5/CMakeLists.txt
@@ -112,14 +112,15 @@ else()
     if(NOT BUILD_SHARED_LIBS)
         set(HDF5_USE_STATIC_LIBRARIES ON)
     endif()
-    find_package(HDF5 REQUIRED)
+    find_package(HDF5 REQUIRED COMPONENTS C CXX)
     target_link_libraries(inviwo-module-hdf5
         PUBLIC
         ${HDF5_LIBRARIES}
-        ${HDF5_HL_LIBRARIES}
         ${HDF5_CXX_LIBRARIES}
-        ${HDF5_HL_CXX_LIBRARIES}
-    )
+        )
+    target_include_directories(inviwo-module-hdf5
+        PUBLIC
+        ${HDF5_INCLUDE_DIRS})
     ivw_vcpkg_install(hdf5 MODULE HDF5)
     set(version ${HDF5_VERSION_STRING})
 endif()
--- a/modules/hdf5/CMakeLists.txt
+++ b/modules/hdf5/CMakeLists.txt
@@ -109,20 +109,39 @@ if (NOT IVW_USE_EXTERNAL_HDF5)
     ivw_default_install_targets(${h5-targets})
 
 else()
-    if(NOT BUILD_SHARED_LIBS)
-        set(HDF5_USE_STATIC_LIBRARIES ON)
+    find_package(hdf5 REQUIRED)
+    if(HDF5_FOUND)
+        if(BUILD_SHARED_LIBS)
+            target_link_libraries(inviwo-module-hdf5
+                PUBLIC
+                    hdf5::hdf5-shared
+                    hdf5::hdf5_hl-shared
+                    hdf5::hdf5_cpp-shared
+                    hdf5::hdf5_hl_cpp-shared
+            )
+        else()
+            target_link_libraries(inviwo-module-hdf5
+                PUBLIC
+                    hdf5::hdf5-static
+                    hdf5::hdf5_hl-static
+                    hdf5::hdf5_cpp-static
+                    hdf5::hdf5_hl_cpp-static
+            )
+        endif()
+        ivw_vcpkg_install(hdf5 MODULE HDF5)
+        set(version ${HDF5_VERSION_STRING})
+    else()
+        if(NOT BUILD_SHARED_LIBS)
+           set(HDF5_USE_STATIC_LIBRARIES ON)
+        endif()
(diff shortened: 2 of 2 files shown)


petersteneteg:
I need to test this on more systems. find_package(hdf5) will use the hdf5-config.cmake that is provided directly by HDF5. There will probably be some problem with the casing of hdf5 vs HDF5. Since on windows for example HDF5 will probably not match hdf5... 
Also we always try to use the provided targets when linking over using various variables directly.

rartino:
I suppose that is the issue: Linux dev packages don't provide `hdf5-config.cmake`, e.g.:

- https://packages.ubuntu.com/jammy/amd64/libhdf5-dev/filelist
- https://rpmfind.net/linux/RPM/epel/9/x86_64/Packages/h/hdf5-devel-1.12.1-7.el9.1.x86_64.html

Perhaps they used to (?), but I cannot find any trace of it, e.g., going through the Ubuntu versions here:

- https://packages.ubuntu.com/search?suite=jammy&arch=any&mode=exactfilename&searchon=contents&keywords=hdf5-config.cmake

This PR only affects the cmake branch where `IVW_USE_EXTERNAL_HDF5 = ON`, which is when there is no reason to expect a `hdf5-config.cmake` to be present (in difference to a self-compiled HDF5).

I also note that [FindHDF5](https://cmake.org/cmake/help/latest/module/FindHDF5.html) apparently uses `hdf5-config.cmake` if it can find it:
> HDF5_NO_FIND_PACKAGE_CONFIG_FILE *New in version 3.8.* Set true to skip trying to find hdf5-config.cmake.

The casing difference seems intentional. The lowercase version is apparently the one distributed with the HDF5 sources (?), whereas the upper case version points to the "finder" distributed with cmake itself.

petersteneteg:
Yes, unfortunately we do to little testing with dist packages. But I would like to do more.

> I suppose that is the issue: Linux dev packages don't provide `hdf5-config.cmake`, e.g.:
> 
> * https://packages.ubuntu.com/jammy/amd64/libhdf5-dev/filelist
> * https://rpmfind.net/linux/RPM/epel/9/x86_64/Packages/h/hdf5-devel-1.12.1-7.el9.1.x86_64.html
> 
> Perhaps they used to (?), but I cannot find any trace of it, e.g., going through the Ubuntu versions here:
> 
> * https://packages.ubuntu.com/search?suite=jammy&arch=any&mode=exactfilename&searchon=contents&keywords=hdf5-config.cmake

Doing a standard install from HDF5 will add a hdf5-config.cmake I think

> 
> This PR only affects the cmake branch where `IVW_USE_EXTERNAL_HDF5 = ON`, which is when there is no reason to expect a `hdf5-config.cmake` to be present (in difference to a self-compiled HDF5).

We also use vcpkg to provide dependencies externally (we will remove the submodules at some point in favor of using vcpkg) together with IVW_USE_EXTERNAL_HDF5. And that is generally fine. The only issue is that all external providers has to agree on what the find_packages argument is called. For vcpkg that need to be called hdf5 with lowercase.

> 
> I also note that [FindHDF5](https://cmake.org/cmake/help/latest/module/FindHDF5.html) apparently uses `hdf5-config.cmake` if it can find it:
> 
> > HDF5_NO_FIND_PACKAGE_CONFIG_FILE _New in version 3.8._ Set true to skip trying to find hdf5-config.cmake.

Yes, I śaw that but it seems that one is also looking HDF5 not hdf5

but maybe that works anyway, I need to test that on a few systems. 

> 
> The casing difference seems intentional. The lowercase version is apparently the one distributed with the HDF5 sources (?), whereas the upper case version points to the "finder" distributed with cmake itself.

Yes. that could be the reason.

rartino:
> Yes, unfortunately we do to little testing with dist packages. But I would like to do more.

Comparing the three strategies; bundled libraries; externally built libraries (e.g. via src, pip, etc.); and dist packages I think dist packages are the way to go for stability. Especially in the interaction with the system Python it seems easy to get into trouble when mixing in non-dist libraries; and pulling in Qt from somewhere else when the system one would work is a bit nasty also for other reasons.

I saw that you have a Ubuntu 22.04 build in your CI, so I'm trying to adopt those steps to use dist libraries as much as possible. Would you be interested in adding such a build to the CI alongside the one you have now? Continuously testing a dist-oriented Ubuntu build would probably make for a bit more user-friendly install experience on Linux.

>> HDF5_NO_FIND_PACKAGE_CONFIG_FILE New in version 3.8. Set true to skip trying to find hdf5-config.cmake.

> Yes, I śaw that but it seems that one is also looking HDF5 not hdf5

(I guess you are looking here?: https://github.com/Kitware/CMake/blob/master/Modules/FindHDF5.cmake#L510 )
This indeed seems odd. Searching around I am now very confused about where the lowercase hdf5 comes from, because I only see the uppercase being used in find_package. Is it perhaps specifically a decision (bug) of the (apparently rather recent) vcpkg HDF5 support to install the file with the lowercase name?

To support both, perhaps one could do something along the lines of:
```
find_package(hdf5)
if(HDF5_FOUND)
  ...
else()
  find_package(HDF5 REQUIRED COMPONENTS C CXX)
  ...
endif()
```

rartino:
Working on the dist-library build, I'm running into a similar issue with freetype as I did with HDF5, and I found a very helpful stackoverflow discussion. Probably the same thing is going on for both freetype/Freetype and hdf5/HDF5:

https://stackoverflow.com/questions/60531709/add-freetype-dependency-with-cmake-supporting-all-vcpkg-apt-get-brew

rartino:
As a minor followup, I have improved the PR so that it should handle HDF5 provided by either vcpkg or via the build-in cmake FindHDF5. I also added the same thing for Freetype, which has the same issue.

With the version in this PR I have engaged in trial-and-error to use Ubuntu 22.04 dist packages as far as possible and ended up with this recipe:
```
sudo apt install \
     python3 python3-pip git python3-numpy python3-h5py python3-pybind11 python3-scipy python3-regex \
     libglew-dev ninja-build gcc-12 g++-12 \
     build-essential cmake extra-cmake-modules git freeglut3-dev xorg-dev \
     openexr zlib1g zlib1g-dev qt6-base-dev qt6-tools-dev qt6-tools-dev libqt6svg6-dev \
     libjpeg-dev libtiff-dev libtirpc-dev \
     libhdf5-dev libpng-dev libglu1-mesa-dev libxrandr-dev libxinerama-dev libxcursor-dev \
     googletest libgtest-dev libglm-dev nlohmann-json3-dev libfmt-dev libglew-dev pybind11-json-dev \
     libfreetype-dev libassimp-dev cimg-dev libnifti-dev libznz-dev libopenexr-dev libtclap-dev

sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-12 120 --slave /usr/bin/g++ g++ /usr/bin/g++-12 --slave /usr/bin/gcov gcov /usr/bin/gcov-12
sudo update-alternatives --set gcc /usr/bin/gcc-12

qtchooser -install sys-qt6 /usr/lib/qt6/bin/qmake
export QT_SELECT="sys-qt6"
eval `qtchooser --print-env`

mkdir ~/Inviwo/inviwo-build
cd ~/Inviwo/inviwo-build
export CMAKE_PREFIX_PATH="/usr/lib/cmake/nlohmann_json/:/usr/share/ECM/find-modules/"
cmake -G "Unix Makefiles" \
  -DCMAKE_BUILD_TYPE="RelWithDebInfo" \
  -DCMAKE_C_COMPILER="gcc-12" \
  -DCMAKE_CXX_COMPILER="g++-12" \
  -DBUILD_SHARED_LIBS=ON \
  -DIVW_APP_MINIMAL_GLFW=ON \
  -DIVW_APP_MINIMAL_QT=OFF \
  -DIVW_TEST_BENCHMARKS=OFF \
  -DIVW_MODULE_HDF5=ON \
  -DIVW_MODULE_PYTHON3=ON \
  -DIVW_MODULE_PYTHON3QT=ON \
  -DIVW_MODULE_QTWIDGETS=ON \
  -DIVW_PACKAGE_INSTALLER=ON \
  -DIVW_USE_OPENEXR=ON \
  -DIVW_CFG_BUILD_CHANGELOG=ON \
  -DIVW_USE_EXTERNAL_ASSIMP=ON \
  -DIVW_USE_EXTERNAL_EIGEN=ON \
  -DIVW_USE_EXTERNAL_FREETYPE=ON \
  -DIVW_USE_EXTERNAL_GLEW=ON \
  -DIVW_USE_EXTERNAL_HDF5=ON \
  -DIVW_USE_EXTERNAL_IMG=ON \
  -DIVW_USE_EXTERNAL_JSON=ON \
  -DIVW_USE_EXTERNAL_SGCT=ON \
  -DIVW_USE_EXTERNAL_TCLAP=ON \
  -DIVW_USE_EXTERNAL_TIFF=ON \
  -DIVW_USE_EXTERNAL_UTFCPP=ON \
  -DIVW_USE_EXTERNAL_GLM=ON \
  -DIVW_USE_EXTERNAL_PYBIND11=OFF \
  -DIVW_USE_EXTERNAL_PEGTL=OFF \
  -DIVW_USE_EXTERNAL_ROARING=OFF \
  -DIVW_USE_EXTERNAL_GTEST=OFF \
  -DIVW_USE_EXTERNAL_NIFTI=OFF \
  -DIVW_USE_EXTERNAL_FMT=OFF \
  -DIVW_USE_EXTERNAL_INJA=OFF \
  -DINJA_USE_EMBEDDED_JSON=ON \
  -DIVW_USE_EXTERNAL_TINYDIR=OFF \
  -DIVW_USE_EXTERNAL_CIMG=OFF \
  -DIVW_USE_EXTERNAL_OPENEXR=OFF \
  -DIVW_USE_OPENMP=OFF \
  -S ~/Inviwo/inviwo -B ./
make -j4   
```

With the following notes of things which are still OFF:

- **inja**: there is no dist package for https://github.com/pantor/inja. 
- **tinydir**: there is no dist package for https://github.com/cxong/tinydir. 
- **nifti**: gets `NIFTI::znz` libznz path wrong, sets it to `/usr/lib` instead of `/usr/lib/x86_64-linux-gnu/`.
  Patch the dist-provided cmake finder? (And report upstream...)
- **cimg**: there is no standard cmake cimg finder, and not cmake file provided in dist package. Provide own? 
- **gtest**: configures with the dist cmake finder, but then have issues with multiple expected targets. Are the vcpkg and dist Find modules named the same but incompatible?
- **openexr**: possibly same thing as for qtest, ends up with an error about `OpenEXR::OpenEXR` being undefined.
- **fmt**: dist package gives compiler errors: `include/inviwo/core/util/glmfmt.h:40:67: error: expected class-name before ‘{’ token 40 | struct fmt::formatter<glm::vec<L, T, Q>> : fmt::ostream_formatter {};`
- **pybind11**: dist package gives compiler error: `fatal error: pybind11/trampoline_self_life_support.h: No such file or directory`
- `DIVW_APP_MINIMAL_QT`: I'm not sure what setting this to ON does, but it is used in the present CI. Inviwo builds with this, but segfaults on startup.

petersteneteg:
Hi! that is great work! I have been wanting to to that for some time! It is a lot of trial and error getting all that working, I will 
try and get this merged and add an ubuntu GitHub action runner using this, to hopefully keep it working. 
Getting it up and running is one thing, but keeping it working over time is even harder :/ 

Looking att some of the missing dependencies, tinydir will soon be removed, replaced with std::filesystem.

vcpkg tries to follow the upstream target names as closely as possible. But sometime upstream changes names...
The Find packages in cmake are often quite old (at least they used to be), and maintained differently for the upstream sources, which makes keeping them in sync and upto date hard... 

I think ideally cmake should not come with any build in find packages, but what ever package manager that is used should provide suitable package descriptions, but that is in an ideal world...

We use a special branch of pybind11, SMART_HOLDERS, I think it is called, it will probably get merged at some point, but it still some time out. Anyway it is a header only lib to that should not make a big difference. 

cimg is also header only and should be fine as long as all the image libs used a linked from the system. 

I think fmt error you have is because the version is too old. We use 9.+, but Im not 100% sure.  

-DIVW_APP_MINIMAL_QT is just there as on option to build an additional target, a small qt app In that case, can be disabled without any problems.

petersteneteg:
@rartino did you attempt to build the python package PySide6 (the official python qt bindings) with this set of system packages? I could not find a recent version of that on apt, and if one uses pip to get it is come with its own qt libs. It has to be built with the same qt as inviwo for it to work together. I don't know if it is possible to force pip to build it using the system libs?

rartino:
@petersteneteg 
> did you attempt to build the python package PySide6 (the official python qt bindings) with this set of system packages? I could not find a recent version of that on apt, and if one uses pip to get it is come with its own qt libs. It has to be built with the same qt as inviwo for it to work together. I don't know if it is possible to force pip to build it using the system libs?

No, I did not; but I think I see the same thing as you. A trivial 'pip3 install PySide6' does not use the system Qt libraries. Peeking into site-packages it seems PySide6 goes out of its way to bundle its own Qt, maybe to avoid conflicts between itself and the Qt version it runs on. 

However, I also see that PySlide6 seems to correlate their versions with Qt6, so I suspect what one is meant to do is: `pip3 install --user PySide6==6.2.4` to still get bundled, but version-matched, Qt library files. Did you try that? (Or do you have an example PySlide + Inviwo script I can try?).

> Looking att some of the missing dependencies, tinydir will soon be removed, replaced with std::filesystem.

Right, tindydir and inja simply have to remain Inviwo-built for now.

> I think ideally cmake should not come with any build in find packages, but what ever package manager that is used should provide suitable package descriptions, but that is in an ideal world...

IMO the whole design with these "Find"-packages isn't great for backwards and forwards compatibility. It is very easy to end up with build configs that only work for exactly a small interval of cmake versions (which applies regardless of who maintain the Find packages).

> cimg is also header only and should be fine as long as all the image libs used a linked from the system.

So, what is the right way to get that into the build system? Because with `DIVW_USE_EXTERNAL_CIMG=OFF` I assume Inviwo builds and uses its own, right?

> I think fmt error you have is because the version is too old. We use 9.+, but Im not 100% sure.

Probably; which means it needs to be built, as there is no newer dist package.

> -DIVW_APP_MINIMAL_QT is just there as on option to build an additional target, a small qt app In that case, can be disabled without any problems.

Do you understand why, when I build with that ON, running the `Inviwo` binary gives me a segfault? Whereas the exact same build with it OFF gives a working Inviwo?

petersteneteg:
I think all this should be merged in various other branches now. 
I hope that the HDF and freetype find package should work for both vcpkg and our submodules and apt now. 
The number of permutations is very large, when you also add in that the versions of each package can vary across the sources over time.. 
We now have a GHA test for building with as many apt packages as possible,
https://github.com/inviwo/inviwo/blob/7a46274536fb9b2bb3f874555421a57981f5a1c0/.github/workflows/inviwo.yml#L285-L360
thanks @rartino for figuring out all the apt packages to use! Let me know If there are more problems!

petersteneteg:
right we also have saved the apt cmake config in https://github.com/inviwo/inviwo/blob/7a46274536fb9b2bb3f874555421a57981f5a1c0/CMakePresets.json#L74-L106

PR #1406, opened 2023-04-24
Feature/animation fix

### LINKED ISSUE #1395: [BUG]: workspace with animation crashes on load
### Problem description

Workspaces containing animations crash during deserialization, if not all previous modules/processors are present. 

### Inviwo version

c570dac

### Compiler / Version

MSVC 2022

### Qt Version

6.3.1

### Reproducible example code

```text
1. Create workspace and animation with one processor/module that can be temporarily disabled (e.g. OpenGL, BaseGL)
2. Disable the module
3. Recompile Inviwo without the module
4. Load the workspace
```

petersteneteg:
Fixes #1395

Changes proposed in this PR:
 * Fixes an issue the index deserializer where serialization errors where not handled.
 * Fixes lifetime isses in WorkspaceAnimations
 * ...

### DIFF
--- a/include/inviwo/core/io/serialization/deserializer.h
+++ b/include/inviwo/core/io/serialization/deserializer.h
@@ -437,14 +437,18 @@ public:
     template <typename C>
     void operator()(Deserializer& d, C& container) {
         size_t count = 0;
+        T tmp{};
         ContainerWrapper<T> cont(
             itemKey_, [&](std::string_view, size_t ind) -> typename ContainerWrapper<T>::Item {
-                ++count;
                 if (ind < container.size()) {
-                    return {true, container[ind], [&](T&) {}};
+                    return {true, container[ind], [&](T&) { ++count; }};
                 } else {
-                    container.emplace_back(makeNewItem_());
-                    return {true, container.back(), [&](T&) { onNewItem_(container.back()); }};
+                    tmp = makeNewItem_();
+                    return {true, tmp, [&](T& value) {
+                                container.emplace_back(std::move(value));
+                                ++count;
+                                onNewItem_(container.back());
+                            }};
                 }
             });
 
--- a/modules/animation/include/modules/animation/mainanimation.h
+++ b/modules/animation/include/modules/animation/mainanimation.h
@@ -90,7 +90,7 @@ private:
 
     Animation* animation_;
     InviwoApplication* app_;
-    
+
     // create this lazy so we have all the factories ready when we create it
     mutable std::optional<AnimationController> controller_;
 
--- a/modules/animation/include/modules/animation/mainanimation.h
+++ b/modules/animation/include/modules/animation/mainanimation.h
@@ -88,9 +88,10 @@ private:
      */
     void addKeyframeSequenceCallback(Property* property);
 
-    // create this lazy so we have all the factories ready when we create it
     Animation* animation_;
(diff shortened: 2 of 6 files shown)


include/inviwo/core/io/serialization/deserializer.h
```diff
@@ -437,14 +437,18 @@ class IndexedDeserializer {
...
         ContainerWrapper<T> cont(
             itemKey_, [&](std::string_view, size_t ind) -> typename ContainerWrapper<T>::Item {
-                ++count;
                 if (ind < container.size()) {
-                    return {true, container[ind], [&](T&) {}};
+                    return {true, container[ind], [&](T&) { ++count; }};
                 } else {
-                    container.emplace_back(makeNewItem_());
-                    return {true, container.back(), [&](T&) { onNewItem_(container.back()); }};
+                    tmp = makeNewItem_();
```
ResearchDaniel, line 446:
```suggestion
                    tmp = makeNewItem_();
Any particular reason for having tmp outside?
```
```suggestion
                    T tmp = makeNewItem_();
```

modules/animation/include/modules/animation/mainanimation.h
```diff
@@ -88,9 +88,10 @@ class IVW_MODULE_ANIMATION_API MainAnimation {
      */
     void addKeyframeSequenceCallback(Property* property);
 
-    // create this lazy so we have all the factories ready when we create it
     Animation* animation_;
     InviwoApplication* app_;
+    
+    // create this lazy so we have all the factories ready when we create it
```
ResearchDaniel, line 94: I think that this comment applies to animation_ as well

modules/animation/include/modules/animation/workspaceanimations.h
```diff
@@ -96,8 +97,10 @@ class IVW_MODULE_ANIMATION_API WorkspaceAnimations : public AnimationControllerO
     Animation& operator[](size_t i);
     const Animation& operator[](size_t i) const;
 
-    std::vector<Animation>::const_iterator begin() const;
-    std::vector<Animation>::const_iterator end() const;
+
+    using iterator = util::IndirectIterator<std::vector<std::unique_ptr<Animation>>::const_iterator>;
```
ResearchDaniel:
```suggestion
    using const_iterator = util::IndirectIterator<std::vector<std::unique_ptr<Animation>>::const_iterator>;
```

ResearchDaniel (dismissed):
Looks good, but have some suggestions for changes

include/inviwo/core/io/serialization/deserializer.h
petersteneteg: Its lifetime needs to outlive that lambda that it is created within, since the lambda will return a reference to the created object.  I use the same pattern for the other ContainerWrapper cases...

modules/animation/include/modules/animation/mainanimation.h
petersteneteg: we don't create the animation we are given it in the constructor...

modules/animation/include/modules/animation/workspaceanimations.h
petersteneteg: unless we add a non-const version I don't see the need.

ResearchDaniel (approved):

PR #1410, opened 2023-04-27
OpenGL: GLSL illumination for negative normals

martinfalk:
Currently in Inviwo's volume rendering, negative gradients result in regions that are not properly illuminated (first image).

Changes proposed in this PR:
 * ensure that normals/gradients are always pointing toward the camera for illumination in GLSL (second image)


modules/opengl/glsl/utils/shading.glsl
```diff
@@ -85,6 +85,10 @@ vec3 shadeSpecular(LightParameters light_, vec3 materialSpecularColor, vec3 posi
 
 vec3 shadeBlinnPhong(LightParameters light_, vec3 materialAmbientColor, vec3 materialDiffuseColor,
                      vec3 materialSpecularColor, vec3 position, vec3 normal, vec3 toCameraDir) {
+    // important for negative gradients/normals (inside/outside illumination)
+    if (dot(normal, toCameraDir) < 0.0) {
```
ResearchDaniel: Use faceforward
https://registry.khronos.org/OpenGL-Refpages/gl4/html/faceforward.xhtml

ResearchDaniel (dismissed):
It should be optional to apply back face lighting?

I guess the question is also what should be the default. 
I am also surprised that the gradient is not facing the camera on the "front side" of the objects...

petersteneteg (changes requested):
I don't think this approach is valid. We don't want to enable 2-sided lighting, But define the "outside" to be at "0" and the inside to be at either higher densities or lower densities. This will not to that correctly, judging from the regressions failures. For any data set that is only uses [0, +] nothing should change. 

I think it would probably make more sense to multiply the normal with the sign of the value of current sample.
Potentially we need to adjust for the data range scaling to figure out where the value 0 will end up in the normalized 0,1 range.

modules/opengl/glsl/utils/shading.glsl
martinfalk: faceforward for volume gradients?

martinfalk:
Reverted to flip only the volume gradient based on the sign of the density value, that is
```glsl
gradient *= sign(voxel[channel] / (1.0 - volumeParameters.formatScaling) - volumeParameters.formatOffset);
```

petersteneteg (dismissed):

petersteneteg (dismissed):

petersteneteg (approved):

modules/basegl/glsl/volume_gradient.frag
```diff
@@ -34,15 +34,17 @@ uniform VolumeParameters volumeParameters;
...
 uniform int channel = 0;
+uniform vec2 dataRange = vec2(0, 1);
 
 in vec4 texCoord_;
 
 void main() {
 #ifdef ADD_DATA_CHANNEL
-    float v = getNormalizedVoxelChannel(volume, volumeParameters, texCoord_.xyz,channel);
+    float v = getNormalizedVoxelChannel(volume, volumeParameters, texCoord_.xyz, channel) 
+        * (dataRange.y - dataRange.x) + dataRange.x;
```
ResearchDaniel, line 44: This should be value range?

modules/basegl/src/processors/volumeprocessing/volumegradientprocessor.cpp
```diff
@@ -97,14 +97,17 @@ VolumeGradientProcessor::~VolumeGradientProcessor() = default;
...
+    shader_.setUniform("dataRange", vec2(inport_.getData()->dataMap_.dataRange));
 }
 
 void VolumeGradientProcessor::postProcess() {
-    volume_->dataMap_.valueAxis.name = "gradient";
-    volume_->dataMap_.valueAxis.unit =
-        inport_.getData()->dataMap_.valueAxis.unit / inport_.getData()->axes[0].unit;
-    volume_->dataMap_.dataRange = dvec2(-1.0, 1.0);
-    volume_->dataMap_.valueRange = dvec2(-1.0, 1.0);
+    if (!dataInChannel4_) {
```
ResearchDaniel, line 104: This parameters need to be set when dataInChannel4_ is true as well

modules/basegl/src/shadercomponents/volumecomponent.cpp
```diff
@@ -82,6 +82,9 @@ vec3 {0}GradientPrev = vec3(0);
 vec3 {0}Gradient = useSurfaceNormals ? -texture(surfaceNormal, texCoords).xyz :
     normalize(COMPUTE_GRADIENT_FOR_CHANNEL({0}Voxel, {0}, {0}Parameters,
                                            samplePosition, channel));
+if (!useSurfaceNormals) {{
+    {0}Gradient *= sign({0}Voxel[channel] / (1.0 - {0}Parameters.formatScaling) - {0}Parameters.formatOffset);
```
ResearchDaniel, line 86: If the aim is to flip sign if inside an object it would be more appropriate to detect when opacity goes from 0 to > 0 (entering an object) and going from opacity != 0 to opacity == 0 (leaving an object).

modules/opengl/glsl/raycasting.frag
```diff
@@ -147,7 +147,11 @@ vec4 rayTraversal(vec3 entryPoint, vec3 exitPoint, vec2 texCoords, float backgro
             } else {
                 gradient = COMPUTE_GRADIENT_FOR_CHANNEL(voxel, volume, volumeParameters, samplePos, channel);
                 gradient = normalize(gradient);
+
+                // make sure that the gradient always points away from zero
+                gradient *= sign(voxel[channel] / (1.0 - volumeParameters.formatScaling) - volumeParameters.formatOffset);
```
ResearchDaniel, line 152: Same comment:
If the aim is to flip sign if inside an object it would be more appropriate to detect when opacity goes from 0 to > 0 (entering an object) and going from opacity != 0 to opacity == 0 (leaving an object).

ResearchDaniel (changes requested):
Two points:
1. If the aim is to flip sign if inside an object it would be more appropriate to detect when opacity goes from 0 to > 0 (entering an object) and going from opacity != 0 to opacity == 0 (leaving an object). 
2. It would have been nice to separate the gradient computation into another PR. Overall, although a big change, I think that the gradients should be computed based on value range

modules/opengl/glsl/raycasting.frag
petersteneteg: No

modules/basegl/glsl/volume_gradient.frag
martinfalk: No, the intention here is to replicate the original value stored in the input volume without any mapping. For example, `getNormalizedVoxel()` returns a [0,1] range for a uint16 texture with a data range of [0, 65535] independent of ValueRange.

modules/basegl/src/processors/volumeprocessing/volumegradientprocessor.cpp
martinfalk: Since the fourth channel contains now float values in the original input range, we need to reuse the original data and value ranges here.

A slightly better solution would be to have a range mapping for each channel (and stop mixing up data format and actual data ranges).

ResearchDaniel: But that means that the gradients will be scaled incorrectly

petersteneteg: well otherwise the data will be scaled incorrectly...

PR #1419, opened 2023-05-09
Feature/core fixes

petersteneteg:
Fixes #...

### DIFF
--- a/include/inviwo/core/network/networkedge.h
+++ b/include/inviwo/core/network/networkedge.h
@@ -44,14 +44,12 @@ class ProcessorNetwork;
  * Represents a edge in the ProcesorNetwork, either a PortConnection or a PropertyLink, as a
  * pair of paths, i.e. dot separated identifiers.
  */
-struct IVW_CORE_API NetworkEdge : Serializable {
+struct IVW_CORE_API NetworkEdge {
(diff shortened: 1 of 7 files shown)


include/inviwo/core/network/networkedge.h
```diff
@@ -44,15 +44,13 @@ class ProcessorNetwork;
  * Represents a edge in the ProcesorNetwork, either a PortConnection or a PropertyLink, as a
  * pair of paths, i.e. dot separated identifiers.
  */
-struct IVW_CORE_API NetworkEdge : Serializable {
+struct IVW_CORE_API NetworkEdge {
```
ResearchDaniel, line 47: Why? The functions are still there…

ResearchDaniel (changes requested):

include/inviwo/core/network/networkedge.h
petersteneteg: It is enough to have the functions to be serializeable now, there is no need to also inherit the base class. This avoids some vtables and stuff, don't think it matters so much in this case.

martinfalk (approved):

ResearchDaniel (approved):

PR #1424, opened 2023-05-29
BaseGL: volume mask processor

boonta:
Changes proposed in this PR:
 * added a processor for masking a volume


petersteneteg (dismissed):
might be some corner cases to think about, that if mask*volume is outside the data range? Which about different dimensions? Also a regression test is always good.

modules/basegl/glsl/volumemask.frag
```diff
@@ -0,0 +1,45 @@
...
+uniform sampler3D volume;
+uniform sampler3D mask;
+uniform VolumeParameters volumeParameters;
+uniform VolumeParameters maskParameters;
+
+in vec4 texCoord_;
+
+void main() {
+    vec4 voxel = getVoxel(volume, volumeParameters, texCoord_.xyz);
+    float mask = getVoxel(mask, maskParameters, texCoord_.xyz).r;
```
ResearchDaniel: Could there be one mask per channel?

modules/basegl/glsl/volumemask.frag
```diff
@@ -0,0 +1,45 @@
...
+
+    FragData0 = voxel * mask;
```
ResearchDaniel: `mask` might be larger than 1?
Could also be a good idea to have an option for the value to use in non-mask regions (replaceWithValue).
```suggestion
    FragData0 = mix(replaceWithValue, voxel, mask != 0);
```

ResearchDaniel (changes requested):
Looks good, but I think that it would be good to change to a user specified value when the mask is zero. Some volumes might have other meanings for the value "0", which means that for example "-1" might be better.

modules/basegl/glsl/volumemask.frag
petersteneteg: this seems more like a feature request.

modules/basegl/glsl/volumemask.frag
petersteneteg: Also a feature request.

modules/basegl/glsl/volumemask.frag
boonta: Me and Martin were working on how to filter the flow data. There is flow data and a mask as inputs. The VolumeCombiner could not work on this issue. So the feature here is like to request how we can apply a mask on the flow data. You can freely modify it.

modules/basegl/glsl/volumemask.frag
ResearchDaniel: Could agree depending on what is to be expected of the mask volume. The current approach almost assumes a binary mask, but this information is a bit hidden. 
Perhaps enough to be more explicit about what the processor does:
    R"(Mask the first input volume with the second volume through multiplication, i.e., masked = voxel * mask)"_unindentHelp};

modules/basegl/src/processors/volumemasker.cpp
```diff
@@ -33,25 +33,56 @@
...
+data range of the input.
+)"_unindentHelp};
 
 const ProcessorInfo VolumeMasker::getProcessorInfo() const { return processorInfo_; }
 
 VolumeMasker::VolumeMasker()
     : VolumeGLProcessor{"volumemask.frag"}
-    , mask_{"mask", "Volume mask applied to the input volume"_help} {
+    , mask_{"mask", "Volume mask applied to the input volume"_help}
+    , useWorldSpace_{"useWorldSpace", "Use World Coordinates",
```
ResearchDaniel, line 59: May be more consistent with other processors to use an option property

https://github.com/inviwo/inviwo/blob/6c264c17af7e31ba7f3d25a8848c45ea6a6dfee6/modules/base/include/modules/base/processors/transform.h#LL118C1-L118C1

modules/basegl/src/processors/volumemasker.cpp
```diff
@@ -33,25 +33,56 @@
...
+                                                 R"(
+Mask `inputVolume` with the second `mask` volume through multiplication. The output volume inherits all 
+parameters from the input volume including value/data ranges and transformations. Its contents are set to
+```glsl
+output = input * mask.r,
+```
+where the mask volume is assumed to be a single-channel volume.
+
+**Note:** 
+Depending on the data ranges of the input volume and the mask volume, the result may lie outside the
```
ResearchDaniel, line 50: Any reason for not adjusting the range instead?

ResearchDaniel (dismissed):
Nice with the regression test. Note that regression tests might be faster if a slicer is used instead of a raycaster.

modules/basegl/src/processors/volumemasker.cpp
petersteneteg: Think the prime reason is that you would not expect the the masking to change it...

petersteneteg (dismissed):

modules/basegl/src/processors/volumemasker.cpp
ResearchDaniel: That is why multiplication might not be a good idea :)

src/core/io/serialization/serializer.cpp
```diff
@@ -60,7 +60,7 @@ void Serializer::serialize(std::string_view key, const std::filesystem::path& pa
         setAttribute(rootElement_, key, path.string());
     } else {
         auto nodeSwitch = switchToNewNode(key);
-        setAttribute(rootElement_, SerializeConstants::ContentAttribute, path.string());
+        setAttribute(rootElement_, SerializeConstants::ContentAttribute, path.generic_string());
```
petersteneteg, line 63: might fix the one on line 60 also

petersteneteg (changes requested):

modules/basegl/src/processors/volumemasker.cpp
martinfalk: And without (binary) multiplication masking is no longer applicable. Meaning you should then probably not use this processor.

ResearchDaniel: I don’t understand. This is what I had in mind:
`val = val ? mask : 0`

martinfalk: You probably meant `val = mask ? val : 0`. But this is the same as multiplying by 0/1.

boonta: If the mask is binary (0 and 1) that would be the same. Otherwise the mask (with values in a range) could benefit from the multiplication. The question could be what kind of mask is non-binary in this case. I am not familiar much in this domain but probably there might be a mask with non-binary values. any suggestion ?

petersteneteg: An other question is what happens with the interpolation when using the ternary operator... multiplication would handle that much better...

ResearchDaniel: Agree, multiplication is much better if you want smooth masking.
One could put a requirement on the mask volume value range to be in [0 1] to ensure that the processor does what is intended.

martinfalk:
the test installer fails due to `fmt.dll` and `assimp-vc143-mt.dll` not being found.

petersteneteg (approved):

PR #1425, opened 2023-06-01
Feature/locale utf8

martinfalk:
Changes proposed in this PR:
 * use UTF-8 locale as default
 * serialization/deserialization defaults to UTF-8
 * logging a `std::filesystem::path` with `operator<<` uses the wrong code page for some reason. Switched to fmt::format instead.
     ```
     Workspace saved to: "D:/�������+fo.inv"   (before)
     Workspace saved to: "D:/äääööåå+fo.inv" (using fmt::format)
     ```
* fixed file type for custom workspace directories

### DIFF
--- a/modules/qtwidgets/src/editorsettings.cpp
+++ b/modules/qtwidgets/src/editorsettings.cpp
@@ -43,12 +43,16 @@ EditorSettings::EditorSettings(InviwoApplication* app)
     , restoreFrequency("restoreFrequency", "Restore Frequency",
                        "Minutes between new backup files"_help, 10,
                        {1, ConstraintBehavior::Immutable}, {100, ConstraintBehavior::Ignore})
-
-    , workspaceDirectories("workspaceDirectories", "Workspace Directories",
-                           std::make_unique<FileProperty>("directory", "Directory")) {
+    , workspaceDirectories(
+          "workspaceDirectories", "Workspace Directories",
+          std::make_unique<FileProperty>("directory", "Directory", "Workspace directory"_help, "",
+                                         AcceptMode::Open, FileMode::Directory)) {
 
     addProperties(workspaceAuthor, numRecentFiles, numRestoreFiles, restoreFrequency,
                   workspaceDirectories);
+
+    workspaceDirectories.setHelp(
+        "Additional workspace directories listed on the Get Started screen."_help);
 
     load();
 }
--- a/src/qt/applicationbase/qtlocale.cpp
+++ b/src/qt/applicationbase/qtlocale.cpp
@@ -46,6 +46,9 @@ std::locale utilqt::getCurrentStdLocale() {
 #else
         std::string localeName(QLocale::system().name().toStdString());
 #endif
+        if (localeName.size() < 6 || localeName.substr(localeName.size() - 6) != ".UTF-8") {
+            localeName += ".UTF-8";
+        }
 
         // try to use the system locale provided by Qt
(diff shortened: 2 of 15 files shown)


modules/qtwidgets/src/editorsettings.cpp
```diff
@@ -43,13 +43,17 @@ EditorSettings::EditorSettings(InviwoApplication* app)
...
+    , workspaceDirectories(
+          "workspaceDirectories", "Workspace Directories",
+          std::make_unique<FileProperty>("directory", "Directory", "Workspace directory"_help, "",
+                                         AcceptMode::Open, FileMode::Directory)) {
 
     addProperties(workspaceAuthor, numRecentFiles, numRestoreFiles, restoreFrequency,
                   workspaceDirectories);
 
+    workspaceDirectories.setHelp(
+        "Additional workspace directories listed on the Get Started screen."_help);
```
petersteneteg, line 55: why not put this in the constructor?

src/qt/applicationbase/qtlocale.cpp
```diff
@@ -46,6 +46,9 @@ std::locale utilqt::getCurrentStdLocale() {
 #else
         std::string localeName(QLocale::system().name().toStdString());
 #endif
+        if (localeName.size() < 6 || localeName.substr(localeName.size() - 6) != ".UTF-8") {
```
petersteneteg: Is this potentially error prone? maybe split on '.' and check the second part?

petersteneteg: will this be a problem on non-windows systems?

petersteneteg (dismissed):

src/qt/applicationbase/qtlocale.cpp
martinfalk: As far as I can see, the locale names returned by QLocale::name() do not contain .UTF-8, just ll_rr  (that is language and region).

modules/qtwidgets/src/editorsettings.cpp
martinfalk: No appropriate constructor for a single prefab with help.

https://github.com/inviwo/inviwo/blob/64d43ff0306166184d84755c9678fdd08a5b2c24/include/inviwo/core/properties/listproperty.h#L103-L126

src/qt/applicationbase/qtlocale.cpp
petersteneteg: yes... but what happens on linux/Mac when you set that?

modules/qtwidgets/src/editorsettings.cpp
petersteneteg: ahh

src/qt/applicationbase/qtlocale.cpp
petersteneteg: I think it should work.

petersteneteg (dismissed):

petersteneteg (approved):
Does this mean we can remove the configureCodePage function now?

martinfalk:
> Does this mean we can remove the configureCodePage function now?

Unfortunately not. The locale during startup is still "C". The manifest, however, prevents Python to mess up part of the locale.

Locale _after_ the Python module has been loaded
```
LC_COLLATE=Swedish_Sweden.utf8;LC_CTYPE=Swedish_Sweden.1252;LC_MONETARY=Swedish_Sweden.utf8;LC_NUMERIC=Swedish_Sweden.utf8;LC_TIME=Swedish_Sweden.utf8
```

PR #1427, opened 2023-06-08
Feature/property move

### LINKED ISSUE #178: Reorder Properties in UI
Would be nice to be able to reorder properties, i.e. move them up or down. For example in the composite property. At the moment, the order is predetermined by the calls to `addProperty()`.

Reordering properties in a PropertyOwner is not sufficient, since deserialization is affected by the order of `addProperty()`
Suggestion: store some metadata to reorder properties upon deserialization.

petersteneteg:
Fixes #178

Changes proposed in this PR:
 * Added support for moving a property in the propertyowner

### DIFF
--- a/modules/postprocessing/src/processors/depthoffield.cpp
+++ b/modules/postprocessing/src/processors/depthoffield.cpp
@@ -212,8 +212,12 @@ void DepthOfField::process() {
             std::swap(nextOutImg_, prevOutImg_);
         }
         evalCount_++;
-        dispatchFront([this, camera, maxEvalCount, focusDepth]() {
-            moveCamera(camera, maxEvalCount, focusDepth);
+        dispatchFront([this, maxEvalCount, focusDepth]() {
+            if (auto* SkewedCamera = dynamic_cast<SkewedPerspectiveCamera*>(&camera_.get())) {
+                moveCamera(SkewedCamera, maxEvalCount, focusDepth);
+            } else {
+                invalidate(InvalidationLevel::InvalidOutput);
+            }
         });
     } else {
         // Set output
@@ -480,6 +484,8 @@ void DepthOfField::synthesizeLightfield(TextureUnitContainer& cont) {
 
 void DepthOfField::moveCamera(SkewedPerspectiveCamera* camera, int maxEvalCount,
                               double focusDepth) {
+    if (evalCount_ >= maxEvalCount) return;
+
     vec3 lookFrom = ogCamera_->getLookFrom();
     vec3 lookTo = ogCamera_->getLookTo();
     vec3 lookUp = ogCamera_->getLookUp();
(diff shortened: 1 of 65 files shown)


ResearchDaniel (dismissed):

modules/postprocessing/src/processors/depthoffield.cpp
```diff
@@ -212,8 +212,15 @@ void DepthOfField::process() {
             std::swap(nextOutImg_, prevOutImg_);
         }
         evalCount_++;
-        dispatchFront([this, camera, maxEvalCount, focusDepth]() {
-            moveCamera(camera, maxEvalCount, focusDepth);
+        dispatchFront([weakSelf = weak_from_this(), maxEvalCount, focusDepth]() {
```
martinfalk, line 215: this pattern should probably be mentioned in the docs

modules/pvm/src/pvmvolumereader.cpp
```diff
@@ -99,9 +99,14 @@ std::shared_ptr<Volume> PVMVolumeReader::readPVMData(const std::filesystem::path
     unsigned char* parameter = nullptr;
     unsigned char* comment = nullptr;
 
-    unsigned char* pvmdata = readPVMvolume(filePath.string().c_str(), &udim.x, &udim.y, &udim.z,
-                                           &bytesPerVoxel, &spacing.x, &spacing.y, &spacing.z,
-                                           &description, &courtesy, &parameter, &comment);
+    auto file = filePath.string();  // convert to string before changing locale
+
+    auto locale = setlocale(LC_ALL, nullptr);  // save current locale
```
martinfalk, line 104: also applicable to Inviwo dat/raw files?

src/qt/editor/welcomewidget.cpp
```diff
@@ -157,33 +182,62 @@ using Role = WorkspaceTreeModel::Role;
 
 namespace {
 
+bool contains(std::string_view str, std::string_view substr) {
+    auto it = std::search(
+        str.begin(), str.end(),
+        std::boyer_moore_horspool_searcher(
```
martinfalk, line 188: Interesting ;-)

martinfalk (approved):

modules/postprocessing/src/processors/depthoffield.cpp
petersteneteg: Well I would rather that know one did anything like this ;)

modules/pvm/src/pvmvolumereader.cpp
petersteneteg: Has not had any problems with it for our files so far.

src/qt/editor/welcomewidget.cpp
petersteneteg: :) probably a premature optimization... but I might refactor it in the future

PR #1428, opened 2023-06-09
Feature/various

martinfalk:
Changes proposed in this PR:
 * added the value of BoolProperty to its description
 * fixed read-only states of BoolProperty and OptionProperty Qt widgets (once set to read-only, the widget state never changed back to active)
 * fixed the animation in the ray casting `TimeComponent`. If disabled, the same time t is used instead of the current system time, which led to a pseudo animation while interacting.

### DIFF
--- a/modules/base/src/processors/meshclipping.cpp
+++ b/modules/base/src/processors/meshclipping.cpp
@@ -75,48 +75,55 @@
 
 namespace inviwo {
 
-const ProcessorInfo MeshClipping::processorInfo_{
-    "org.inviwo.MeshClipping",  // Class identifier
-    "Mesh Clipping",            // Display name
-    "Mesh Creation",            // Category
-    CodeState::Experimental,    // Code state
-    Tags::CPU,                  // Tags
-};
+const ProcessorInfo MeshClipping::processorInfo_{"org.inviwo.MeshClipping",  // Class identifier
+                                                 "Mesh Clipping",            // Display name
+                                                 "Mesh Creation",            // Category
+                                                 CodeState::Stable,          // Code state
+                                                 Tags::CPU,                  // Tags
+                                                 R"(
+    Remove parts of a mesh that are on the backside of a plane. Replaces removed
+    parts with triangles aligned with the plane. Link the camera property to move
+    the camera along the plane or to align plane with the view direction.
+    Coordinates are specified in world space.
+
+    Supports `SimpleMesh` and `BasicMesh`.
+)"_unindentHelp};
 const ProcessorInfo MeshClipping::getProcessorInfo() const { return processorInfo_; }
 
 MeshClipping::MeshClipping()
     : Processor()
-    , inport_("inputMesh")
-    , outport_("clippedMesh")
-    , clippingPlane_("clippingPlane")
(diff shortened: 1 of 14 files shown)


modules/base/src/processors/meshclipping.cpp
```diff
@@ -149,16 +156,18 @@ void MeshClipping::process() {
      *     triangle strip list.
      *   - Build new mesh from the triangle strip list and return it.
      */
-    auto plane = std::make_shared<Plane>(planePoint_.get(), planeNormal_.get());
+    const auto normal =
```
ResearchDaniel, line 159: Does this make sense for the camera movement? It will mean that the plane will move closer to the camera for ClipSide::Front

modules/basegl/src/shadercomponents/timecomponent.cpp
```diff
@@ -94,9 +94,12 @@ TimeComponent::TimeComponent(std::string_view name,
 std::string_view TimeComponent::getName() const { return name_; }
 
 void TimeComponent::process(Shader& shader, TextureUnitContainer&) {
-    shader.setUniform(name_, std::chrono::duration_cast<std::chrono::duration<float, std::milli>>(
-                                 std::chrono::steady_clock::now().time_since_epoch())
-                                 .count());
+    std::chrono::duration<float, std::milli> t(0.0f);
+    if (enabled_) {
```
ResearchDaniel, line 98: Is is not enough with stopping the timer when it is disabled?

ResearchDaniel (dismissed):
Might be some issues with flipping the normal

modules/basegl/src/shadercomponents/timecomponent.cpp
martinfalk: No, `time_since_epoch()` will always return a different value every time `process()` is called. Hence the impression of an animation while for example moving the camera or triggering a re-evaluation of the corresponding processor.
Also any brushing interaction triggers an auto start of the animation.

modules/base/src/processors/meshclipping.cpp
martinfalk: So? This is already the case for the non-flipped normal if the camera is positioned on the "other" side of the plane.

ResearchDaniel: The original idea behind this functionality was that the user enables clipping and that the clipping plane follows the camera. Perhaps this is not a use case when the normal is flipped? That’s why I wondered if it makes sense that the “follow camera” instead moves closer to the camera.

petersteneteg: What is the use case for the "follow camera" mode?

ResearchDaniel: It is used for touch table interaction when clipping along the view direction.

modules/basegl/src/shadercomponents/timecomponent.cpp
ResearchDaniel: I see. I know that this is not exactly this PR, but it seams to be error prone to have both enabled and running. They seem very similar to me and their interaction is quite intricate (e.g `setRunning(true)` will do nothing when disabled)

petersteneteg (approved):
Nice!

PR #1430, opened 2023-06-28
QtWidgets: filesystem fix for TF previews

martinfalk:
Fixes
* regression of file paths of TF previews in the context menu of the TF properties


petersteneteg (dismissed):
When was this broken? long time ago?

martinfalk:
> When was this broken? long time ago?

Broken by the introduction of std::filesystem. Previously `file` contained the full path, but that was changed.

PR #1431, opened 2023-07-11
Changed auto pointer to Processor pointer in qtminimum to compile on gcc 11.3.0

ylegoc:
Changes proposed in this PR:
 * Change type of pointer in qtminimum.cpp

Code compiled and tested on:
Ubuntu 22.04.1 LTS / gcc 11.3.0 / Qt6

### DIFF
--- a/apps/minimals/qt/qtminimum.cpp
+++ b/apps/minimals/qt/qtminimum.cpp
@@ -125,7 +125,7 @@ int main(int argc, char** argv) {
             inviwoApp.getProcessorNetwork()->addObserver(&obs);
             inviwoApp.getProcessorNetworkEvaluator()->addObserver(&obs);
             inviwoApp.getProcessorNetwork()->forEachProcessor(
-                [&](auto* p) { p->ProcessorObservable::addObserver(&obs); });
+                [&](inviwo::Processor* p) { p->ProcessorObservable::addObserver(&obs); });
         },
         200);
 


petersteneteg:
Hi, that for the PR!

Could you perhaps explain what the GCC compiler error is? The code that you changed is as far as I know correct? And I want to avoid adding complier workarounds unless they are really needed and in that cases well documented (Otherwise someone will just put that auto back, or get really confused why you would not use auto).

ylegoc:
Hi,
Here is the gcc compilation error:
```
qtminimum.cpp:128:35: error: request for member ‘ProcessorObservable’ in non-class type ‘__typeof__ (p->)’
  128 |                 [&](auto* p) { p->ProcessorObservable::addObserver(&obs); });
```

That version 11.3.0 of the gcc compiler is unable to deduce *inviwo::Processor*.
However I also tested the compilation on Ubuntu 23.10 with gcc 12.3.0 and it succeeded to compile the code.

So it is necessary to replace *auto* with *inviwo::Processor* but I would suggest to comment: 
```
// Use explicit inviwo::Processor type rather than auto to avoid a compilation error on some compilers e.g. gcc 11.3.0 
// that are unable to deduce the auto type.
[&](inviwo::Processor* p) { p->ProcessorObservable::addObserver(&obs); });
```

I hope this helps ;)

petersteneteg:
I think the real problem it that we need to qualify ProcessorObservable. i.e. the code should look like this: 

> [&](auto* p) { p->inviwo::ProcessorObservable::addObserver(&obs); });

petersteneteg:
fixed in #1434

PR #1451, opened 2023-10-17
Core: Make ImageInport::hasData() consistent

ResearchDaniel:
Fixes case where ImageInport::hasData() would return true even though ImageInport::getData() returns nullptr

### DIFF
--- a/include/inviwo/core/ports/imageport.h
+++ b/include/inviwo/core/ports/imageport.h
@@ -323,8 +323,7 @@ std::vector<std::shared_ptr<const Image>> BaseImageInport<N>::getVectorData() co
 
     for (auto outport : this->connectedOutports_) {
         auto imgport = static_cast<ImageOutport*>(outport);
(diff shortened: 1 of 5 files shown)


include/inviwo/core/ports/imageport.h
```diff
@@ -389,6 +393,20 @@ Document BaseImageInport<N>::getInfo() const {
     return doc;
 }
 
+template <size_t N>
+bool BaseImageInport<N>::hasData() const {
+    if constexpr (N == 0) {
+        return this->isConnected() && this->begin() != this->end();
```
petersteneteg: What about this case?

ResearchDaniel: I kept this the same as for DataInport. However, I agree that the logic here seems flawed. Should be 
`isConnected() && util::any_of…`

petersteneteg (dismissed):

petersteneteg (approved):

PR #1453, opened 2023-10-26
Feature/various fixes

martinfalk:
Changes proposed in this PR:
 * added Mesh constructor supporting multiple buffers (`Mesh(const BufferVector& buffers, const IndexVector& indices);`)
 * extended `BufferObject` and added convenience function for uploading std::vector<> directly instead of `void*`
 * the data upload in `BufferObject` now also supports a grow policy, that is growing only or resize to fit
 * added a default line color to the `LineSettingsInterface` which is used when a mesh does not have color vertex attributes


ResearchDaniel (dismissed):

modules/opengl/include/modules/opengl/buffer/bufferobject.h
```diff
@@ -62,6 +62,16 @@ class IVW_MODULE_OPENGL_API BufferObject : public Observable<BufferObjectObserve
                               //!< integral types
     };
 
+    /**
+     * Policy for managing the buffer size when uploading new data to the GPU.
+     *
+     * \see upload
+     */
+    enum class GrowPolicy {
```
petersteneteg: GrowPolicy seems a bit misleading, since it might also shrink, when ResizeToFit is set. SizePolicy seems more fitting I think.

modules/opengl/src/buffer/bufferobject.cpp
```diff
@@ -245,8 +249,9 @@ void BufferObject::initialize(const void* data, GLsizeiptr sizeInBytes) {
     forEachObserver([](BufferObjectObserver* o) { o->onAfterBufferInitialization(); });
 }
 
-void BufferObject::upload(const void* data, GLsizeiptr sizeInBytes) {
-    if (sizeInBytes <= getSizeInBytes()) {
+void BufferObject::upload(const void* data, GLsizeiptr sizeInBytes, GrowPolicy policy) {
```
petersteneteg: How do we keep track of the last used size? can I ask for that? Seems we should have both a size and capacity here?

petersteneteg (dismissed):

petersteneteg (approved):

PR #1455, opened 2023-11-09
Feature/instance renderer fix and fit

petersteneteg:
Instance renderer
Fixes crash when disconnection port

uses transform feedback to enable fit to view


modules/basegl/src/processors/instancerenderer.cpp
```diff
@@ -304,7 +340,9 @@ InstanceRenderer::InstanceRenderer()
                  "",
                  InvalidationLevel::InvalidResources,
                  PropertySemantics::Multiline}
-
+    , numberOfvertices_{0}
+    , feedbackBuffer_{1000 * sizeof(vec4), GLFormats::get(DataVec4Float32::id()), GL_STATIC_READ,
```
ResearchDaniel: 1000?

modules/basegl/src/processors/instancerenderer.cpp
```diff
@@ -412,14 +450,67 @@ void InstanceRenderer::initializeResources() {
...
+
+size_t numberOfVertices(const MeshGL& meshGL, size_t index) {
+    auto mi = meshGL.getMeshInfoForIndexBuffer(index);
+    auto nIndices = meshGL.getIndexBuffer(index)->getSize();
+    return numberOfVertices(mi, nIndices);
+}
+
+size_t numberOfVertices(const MeshGL& meshGL, Mesh::MeshInfo defaultMeshInfo) {
+    const std::size_t numIndexBuffers = meshGL.getIndexBufferCount();
+    size_t count = 0;
```
ResearchDaniel: Declare inside if

modules/basegl/src/processors/instancerenderer.cpp
```diff
@@ -269,7 +280,32 @@ InstanceRenderer::InstanceRenderer()
         types can be added.)"_unindentHelp,
              prefabs()}
     , vecPorts_{}
-    , camera_("camera", "Camera", util::boundingBox(inport_))
+    , camera_("camera", "Camera",
+              [this]() -> std::optional<mat4> {
+                  if (numberOfvertices_ > 0) {
+                      rendercontext::activateDefault();
+                      feedbackBuffer_.bindBase(0);
```
ResearchDaniel: I think that it would make more sense, and make the code cleaner, if we could use compute shaders for this. Adding transform feedback to the rendering will impact the performance negatively?

ResearchDaniel (changes requested):
Would it not be better to use compute shaders instead? This could add overhead to the rendering for a feature that is not used every frame?

modules/basegl/src/processors/instancerenderer.cpp
petersteneteg: removed

modules/basegl/src/processors/instancerenderer.cpp
petersteneteg: removed

modules/basegl/src/processors/instancerenderer.cpp
petersteneteg: only calculate it on demand now... transform feedback works on macOS compute shaders does not...

modules/basegl/src/processors/instancerenderer.cpp
```diff
@@ -412,38 +428,150 @@ void InstanceRenderer::initializeResources() {
...
-    const auto& mesh = *inport_.getData();
-    MeshDrawerGL::DrawObject drawer{mesh.getRepresentation<MeshGL>(), mesh.getDefaultMeshInfo()};
-    utilgl::setShaderUniforms(shader_, mesh, "geometry");
+void forEach(const MeshGL& meshGL, Shader& shader, auto& ports, size_t nInstances,
+             auto&& callback) {
+    if (meshGL.empty()) return;
+
+    const std::size_t numIndexBuffers = meshGL.getIndexBufferCount();
+
+    for (size_t i = 0; i < nInstances; ++i) {
```
ResearchDaniel, line 491: Should the instance rendering be in the inner loop, just before the callback? This would reduce the number of buffer switches

modules/opengl/include/modules/opengl/shader/shader.h
```diff
@@ -216,6 +216,8 @@ class IVW_MODULE_OPENGL_API Shader {
     std::shared_ptr<std::function<void()>> onReloadScoped(std::function<void()> callback);
     void removeOnReload(const BaseCallBack* callback);
 
+    void setTransformFeedbackVaryings(std::span<const char*> varyings, GLenum bufferMode);
```
ResearchDaniel, line 225:
```suggestion
/*
 * Specify values to record in transform feedback buffers.
 * @param names Array of names of the varying variables to use for transform feedback.
 * @param bufferMode The mode used to capture the varying variables when transform feedback is active. Must be GL_INTERLEAVED_ATTRIBS or GL_SEPARATE_ATTRIBS.
 */
    void setTransformFeedbackVaryings(std::span<const char*> names, GLenum bufferMode);
```

modules/opengl/src/geometry/meshgl.cpp
```diff
@@ -81,6 +81,7 @@ const Mesh::MeshInfo& MeshGL::getMeshInfoForIndexBuffer(size_t idx) const {
               "MeshGL::getMeshInfoForIndexBuffer(): index out of bounds");
     return indexBuffers_[idx].first;
 }
+const Mesh::MeshInfo& MeshGL::getDefaultMeshInfo() const { return defaultMeshInfo_; }
```
ResearchDaniel, line 84: Why not 
```suggestion
const Mesh::MeshInfo& MeshGL::getDefaultMeshInfo() const { return owner->getDefaultMeshInfo(); }
```

modules/opengl/src/shader/shader.cpp
```diff
@@ -715,4 +715,9 @@ void Shader::setUniform(std::string_view name, const TextureUnit& texUnit) const
     if (location != -1) glUniform1i(location, texUnit.getUnitNumber());
 }
 
+void Shader::setTransformFeedbackVaryings(std::span<const GLchar*> varyings, GLenum bufferMode) {
+    glTransformFeedbackVaryings(program_.id, varyings.size(), varyings.data(), bufferMode);
```
ResearchDaniel:
```suggestion
    glTransformFeedbackVaryings(program_.id, static_cast<GLsizei>(varyings.size()), varyings.data(), bufferMode);
```

modules/basegl/include/modules/basegl/processors/instancerenderer.h
```diff
@@ -92,6 +95,8 @@ class IVW_MODULE_BASEGL_API InstanceRenderer : public Processor, public Property
     virtual void initializeResources() override;
     virtual void process() override;
 
+    std::optional<mat4> process(bool enableBoundingBoxCalc);
```
ResearchDaniel: For clarity I think this could be called render
```suggestion
/* Render all meshes. 
* @return The axis-aligned bounding box of all vertices if enableBoundingBoxCalc, otherwise nullptr.
    std::optional<mat4> render(bool enableBoundingBoxCalc);
```

ResearchDaniel (dismissed):
Some minor suggestions

modules/opengl/src/geometry/meshgl.cpp
petersteneteg: I wanted the representation to be slightly more self reliant. For other data you can for example create a Volume from a VolumeRAM. But we can't to that with meshes at the moment. And this does not fix that, but it's a step in that direction...

modules/basegl/src/processors/instancerenderer.cpp
petersteneteg: But it increases the number of uniform calls. effectively doing it that way results in about a 40% drop in fps

modules/opengl/include/modules/opengl/shader/shader.h
ResearchDaniel: Note: names in comment should be varyings, or change variable name

martinfalk (approved):

PR #1462, opened 2023-11-22
Core: FileProperty refactorings

petersteneteg:
Various cleanup and refactoring of file and multifile property

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -92,6 +92,7 @@ endif()
 
 option(IVW_ENABLE_QT "Enable Qt support in Inviwo" ON)
 if(IVW_ENABLE_QT)
+    set(CMAKE_MODULE_PATH_TMP ${CMAKE_MODULE_PATH})
     find_package(Qt6 COMPONENTS Core)
     if(NOT Qt6Core_FOUND) 
         message(FATAL_ERROR
@@ -99,6 +100,7 @@ if(IVW_ENABLE_QT)
             "Consider adding Qt to the CMAKE_PREFIX_PATH,\n" 
             "i.e. start cmake with the option -DCMAKE_PREFIX_PATH=\"/path/to/qt\"")
     endif()
+    set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH_TMP})
 endif()
 
 include(cmake/globalconfig.cmake)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -92,6 +92,7 @@ endif()
 
 option(IVW_ENABLE_QT "Enable Qt support in Inviwo" ON)
 if(IVW_ENABLE_QT)
+    # Find Qt will add paths to CMAKE_MODULE_PATH which we don't want.
     set(CMAKE_MODULE_PATH_TMP ${CMAKE_MODULE_PATH})
     find_package(Qt6 COMPONENTS Core)
     if(NOT Qt6Core_FOUND) 
--- a/include/inviwo/core/properties/fileproperty.h
+++ b/include/inviwo/core/properties/fileproperty.h
@@ -89,6 +89,15 @@ protected:
     std::unique_ptr<FileDialog> createFileDialog(std::string_view title,
                                                  const std::filesystem::path& file) const;
 
+    /**
+     * A variadic any_of. This will evaluate all the arguments.
+     * But the order is arbitrary.
+     */
+    template <typename Arg, typename... Args>
+    [[nodiscard]] static constexpr bool any_of(Arg&& arg, Args&&... args) {
+        return (arg || ... || args);
+    }
+
     std::vector<FileExtension> nameFilters_;
     ValueWrapper<FileExtension> selectedExtension_;
     ValueWrapper<AcceptMode> acceptMode_;
@@ -119,8 +128,8 @@ public:
      * @param displayName displayName for the property
      * @param help descriptive text
      * @param value the path to the file
-     * @param acceptMode
-     * @param fileMode
+     * @param acceptMode @see AcceptMode
+     * @param fileMode @see FileMode
      * @param contentType
      * @param invalidationLevel
      * @param semantics Can be set to Editor
@@ -186,6 +195,10 @@ public:
     void set(const std::filesystem::path& file, const FileExtension& selectedExtension);
 
     void set(const FileProperty* property);
+    /*
+     * Assigns the first path of the MultiFileProperty to this.
+     * If the MultiFileProperty is empty, this path will be empty.
+     */
     void set(const MultiFileProperty* property);
     virtual void set(const Property* property) override;
 
--- a/include/inviwo/core/properties/fileproperty.h
+++ b/include/inviwo/core/properties/fileproperty.h
@@ -38,6 +38,9 @@
 
 namespace inviwo {
 
+class MultiFileProperty;
+class FileDialog;
+
 class IVW_CORE_API FileRequestable {
 public:
     virtual ~FileRequestable() = default;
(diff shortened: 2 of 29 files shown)


include/inviwo/core/properties/multifileproperty.h
```diff
@@ -77,35 +74,29 @@ class IVW_CORE_API MultiFileProperty : public TemplateProperty<std::vector<std::
...
-    virtual void clearNameFilters();
-    virtual std::vector<FileExtension> getNameFilters();
+    operator const std::vector<std::filesystem::path>&() const { return files_; }
+    const std::vector<std::filesystem::path>& get() const { return files_; }
+    const std::vector<std::filesystem::path>& operator*() const { return files_; };
+    const std::vector<std::filesystem::path>* operator->() const { return &files_.value; }
 
-    virtual void setAcceptMode(AcceptMode mode);
-    AcceptMode getAcceptMode() const;
+    const std::filesystem::path* front() const;
```
martinfalk, line 119: Why pointers?

martinfalk: What about begin(), end()?

petersteneteg: since it might be empty... you can get the whole list and do that ever... this is for connivence in a few places... could remove back, since I don't use that. but it looked nice together

modules/qtwidgets/src/properties/collapsiblegroupboxwidgetqt.cpp
```diff
@@ -776,12 +777,13 @@ void CollapsibleGroupBoxWidgetQt::insertPropertyWidget(PropertyWidgetQt* propert
...
-    if (checkable_) {
-        setFocusPolicy(checkBox_->focusPolicy());
-        setFocusProxy(checkBox_);
-    } else {
-        setFocusPolicy(Qt::NoFocus);
-    }
+    setFocusPolicy(btnCollapse_->focusPolicy());
+    setFocusProxy(btnCollapse_);
+
+//    if (checkable_) {
```
martinfalk: remove if no longer needed

src/core/properties/fileproperty.cpp
```diff
@@ -32,21 +32,159 @@
...
+    fileDialog->setAcceptMode(getAcceptMode());
+    fileDialog->setFileMode(getFileMode());
+    fileDialog->setContentType(getContentType());
+
+    return fileDialog;
+}
+
+void FileBase::serialize(Serializer& s, PropertySerializationMode mode) const {
+    // We don't serialize the filename Filter since they are usually a runtime thing depending
+    // on which readers/writers that a available
```
martinfalk:
```suggestion
    // We don't serialize the filename filter since these are typically depending on readers/writers available at runtime
```

CMakeLists.txt
```diff
@@ -92,13 +92,15 @@ endif()
 
 option(IVW_ENABLE_QT "Enable Qt support in Inviwo" ON)
 if(IVW_ENABLE_QT)
+    set(CMAKE_MODULE_PATH_TMP ${CMAKE_MODULE_PATH})
```
ResearchDaniel, line 96: Add a comment to explain this workaround

include/inviwo/core/properties/fileproperty.h
```diff
@@ -131,14 +179,15 @@ class IVW_CORE_API FileProperty : public Property {
...
-    virtual void set(const std::filesystem::path& file);
+    void set(const std::filesystem::path& file);
     /**
      * Set the file name and the selected extension.
      */
-    virtual void set(const std::filesystem::path& file, const FileExtension& selectedExtension);
+    void set(const std::filesystem::path& file, const FileExtension& selectedExtension);
 
+    void set(const FileProperty* property);
+    void set(const MultiFileProperty* property);
```
ResearchDaniel, line 202: Document which part of the multiFileProperty that will be used

include/inviwo/core/properties/multifileproperty.h
```diff
@@ -41,22 +41,19 @@ namespace inviwo {
...
- * @see TemplateProperty, FileProperty
+ *  A class for a list of file/directory paths
+ * @see FileProperty
  */
-class IVW_CORE_API MultiFileProperty : public TemplateProperty<std::vector<std::filesystem::path>> {
+class IVW_CORE_API MultiFileProperty : public Property, public FileBase {
 public:
     virtual std::string getClassIdentifier() const override;
     static const std::string classIdentifier;
+    static constexpr std::string_view defaultContentType = "default";
```
ResearchDaniel: Remove as it is defined in FileBase?

include/inviwo/core/properties/multifileproperty.h
ResearchDaniel: The implementation goes a bit against the one in std::vector, which does not work for empty vectors. So, this could be a bit misleading. Not sure if calling it something else would be better.  I’m fine with keeping it but, we should document that it returns a null ptr if the list is empty.

include/inviwo/core/properties/multifileproperty.h
```diff
@@ -77,35 +74,29 @@ class IVW_CORE_API MultiFileProperty : public TemplateProperty<std::vector<std::
...
     virtual MultiFileProperty* clone() const override;
     virtual ~MultiFileProperty() = default;
 
     void set(const std::filesystem::path& value);
-    virtual void set(const std::vector<std::filesystem::path>& values) override;
+    void set(const std::vector<std::filesystem::path>& values);
+    void set(const std::vector<std::filesystem::path>& files,
+             const FileExtension& selectedExtension);
+    void set(const MultiFileProperty* property);
+    void set(const FileProperty* property);
```
ResearchDaniel, line 103: Maybe add info about if the list will be replaced

src/core/properties/fileproperty.cpp
```diff
@@ -32,21 +32,159 @@
...
+            if (allIt != nameFilters_.end() && selectedExtension_ != *allIt) {
+                modified |= selectedExtension_.update(*allIt);
+            }
+        }
     }
-};
+    return modified;
+}
 
+void FileBase::setAsDefault() {
```
ResearchDaniel, line 132: No modified return here?

src/core/properties/fileproperty.cpp
```diff
@@ -87,50 +221,40 @@ FileProperty& FileProperty::operator=(const std::filesystem::path& file) {
...
+    if (file_.update(file) || selectedExtension_.update(selectedExtension)) {
+        propertyModified();
+    }
 }
 
 void FileProperty::set(const FileProperty* property) {
     if (!property) return;
 
-    bool modified = file_.update(property->file_);
+    if (file_.update(property->file_) || setFileBase(*property)) {
```
ResearchDaniel: I think that this might cause the second statement not to be evaluated

ISO/IEC C++ Standard 14882 §5.15 Logical OR operator [expr.log.or]

logical-or-expression:
    logical-and-expression
    logical-or-expression || logical-and-expression
The || operator groups left-to-right. The operands are both implicitly converted to bool (clause 4). It returns true if either of its operands is true, and false otherwise. Unlike |, || guarantees left-to-right evaluation; moreover, the second operand is not evaluated if the first operand evaluates to true.

src/core/properties/fileproperty.cpp
```diff
@@ -87,50 +221,40 @@ FileProperty& FileProperty::operator=(const std::filesystem::path& file) {
...
 
-    modified |= update(selectedExtension_, property->selectedExtension_);
-    modified |= update(acceptMode_, property->acceptMode_);
-    modified |= update(fileMode_, property->fileMode_);
-    modified |= update(contentType_, property->contentType_);
+void FileProperty::set(const MultiFileProperty* property) {
+    if (!property) return;
 
-    if (modified) propertyModified();
+    if (file_.update(property->front() ? *property->front() : std::filesystem::path{}) ||
```
ResearchDaniel: Also here

src/core/properties/fileproperty.cpp
```diff
@@ -331,7 +370,9 @@ FileProperty& FileProperty::setDefault(const std::filesystem::path& value) {
     return *this;
 }
 FileProperty& FileProperty::resetToDefaultState() {
-    if (file_.reset()) propertyModified();
+    if (file_.reset() || FileBase::reset()) {
```
ResearchDaniel: Also here

src/core/properties/multifileproperty.cpp
```diff
@@ -62,51 +61,74 @@ MultiFileProperty::MultiFileProperty(std::string_view identifier, std::string_vi
...
-        }
+        modified |= selectedExtension_.update(FileExtension{});
+    }
+
+    if (modified) propertyModified();
+}
+
+void MultiFileProperty::set(const std::vector<std::filesystem::path>& files,
+                            const FileExtension& selectedExtension) {
+    if (files_.update(files) || selectedExtension_.update(selectedExtension)) {
```
ResearchDaniel: Also here

src/core/properties/multifileproperty.cpp
```diff
@@ -62,51 +61,74 @@ MultiFileProperty::MultiFileProperty(std::string_view identifier, std::string_vi
...
+        propertyModified();
+    }
+}
+
+void MultiFileProperty::set(const MultiFileProperty* property) {
+    if (!property) return;
+
+    if (files_.update(property->files_) || setFileBase(*property)) {
```
ResearchDaniel: Also here

src/core/properties/multifileproperty.cpp
```diff
@@ -62,51 +61,74 @@ MultiFileProperty::MultiFileProperty(std::string_view identifier, std::string_vi
...
+        propertyModified();
+    }
+}
+
+void MultiFileProperty::set(const FileProperty* property) {
+    if (!property) return;
+
+    if (files_.update(std::vector{property->get()}) || setFileBase(*property)) {
```
ResearchDaniel: Also here

src/core/properties/multifileproperty.cpp
```diff
@@ -118,195 +140,144 @@ void MultiFileProperty::serialize(Serializer& s) const {
...
+    FileBase::setAsDefault();
+    files_.setAsDefault();
+    return *this;
+}
+MultiFileProperty& MultiFileProperty::setDefault(const std::vector<std::filesystem::path>& value) {
+    files_.defaultValue = value;
+    return *this;
+}
+MultiFileProperty& MultiFileProperty::resetToDefaultState() {
+    if (files_.reset() || FileBase::reset()) {
```
ResearchDaniel: Also here

ResearchDaniel (dismissed):

src/core/properties/fileproperty.cpp
petersteneteg: This only changes the default state it does not change the "value" parts...

src/core/properties/fileproperty.cpp
petersteneteg: ahh that is right, did not consider that, but of course that will short circut, Good catch.

src/core/properties/multifileproperty.cpp
petersteneteg: that was somewhat annoying... :/ liked not having to same the modified state in some variable...

petersteneteg: think I fixed all of them

src/core/util/settings/systemsettings.cpp
```diff
@@ -48,8 +48,11 @@ SystemSettings::SystemSettings(InviwoApplication* app)
     , enablePickingProperty_("enablePicking", "Enable picking", true)
     , enableSoundProperty_("enableSound", "Enable sound", true)
     , logStackTraceProperty_("logStackTraceProperty", "Error stack trace log", false)
-    , moduleSearchPaths_("moduleSearchPaths", "Module Search Paths", {}, AcceptMode::Open,
-                         FileMode::Directory)
+    , moduleSearchPaths_(
+          "moduleSearchPaths", "Module Search Paths",
+          "The system will look for inviwo module lib in these path to load at start up. "
```
martinfalk, line 53:
```suggestion
          "The system will look for Inviwo module libs in these paths to load at start up. "
```

martinfalk (approved):

PR #1466, opened 2023-11-23
QtWidget: property search in the network search dialog

petersteneteg:
QtWidget: better keyboard support in the property list widget, Added support for property search in the networksearch dialog using `p:propertyid`

### DIFF
--- a/modules/qtwidgets/src/properties/stringmultilinepropertywidgetqt.cpp
+++ b/modules/qtwidgets/src/properties/stringmultilinepropertywidgetqt.cpp
@@ -29,7 +29,8 @@
 
 #include <modules/qtwidgets/properties/stringmultilinepropertywidgetqt.h>
 
-#include <inviwo/core/properties/stringproperty.h>          // for StringProperty
+#include <inviwo/core/properties/stringproperty.h>  // for StringProperty
+#include <inviwo/core/properties/multifileproperty.h>
 #include <modules/qtwidgets/editablelabelqt.h>              // for EditableLabelQt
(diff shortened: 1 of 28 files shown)


modules/qtwidgets/src/properties/stringmultilinepropertywidgetqt.cpp
```diff
@@ -104,6 +105,59 @@ void StringMultilinePropertyWidgetQt::updateFromProperty() {
...
+MultiFileStringPropertyWidgetQt::MultiFileStringPropertyWidgetQt(MultiFileProperty* property)
+    : PropertyWidgetQt(property), property_(property) {
+
+    QHBoxLayout* hLayout = new QHBoxLayout;
+    setSpacingAndMargins(hLayout);
+
+    label_ = new EditableLabelQt(this, property_);
+    hLayout->addWidget(label_);
+
+    textEdit_ = new MultilineTextEdit;
```
martinfalk, line 117: The constructor should be called.
```suggestion
    textEdit_ = new MultilineTextEdit();
```

martinfalk (dismissed):

modules/qtwidgets/src/properties/stringmultilinepropertywidgetqt.cpp
petersteneteg: the initializer is optional here https://en.cppreference.com/w/cpp/language/new

petersteneteg: But sure we usually put one in

martinfalk (approved):

PR #1485, opened 2024-01-23
Feature/meshaxis

martinfalk:
Fixes:
* incorrect TetraMesh bounding box in `TetraMeshBoundingBox`

Changes proposed in this PR:
 * new processor `MeshAxis` similar to `VolumeAxis`
 * refactored VolumeAxis processor into commonly used components and properties (`Axis3DProcessorHelper`)

This has been tested on: Windows, MSVC 17.8.5


modules/plottinggl/src/utils/axis3dprocessorhelper.cpp
```diff
@@ -0,0 +1,412 @@
...
+        offset = entity->getOffset();
+
+        worldTrafo = entity->getCoordinateTransformer().getDataToWorldMatrix();
+    }
+
+    std::string xCaption{"x"};
+    std::string yCaption{"y"};
+    std::string zCaption{"z"};
+
+    if (auto volume = dynamic_cast<const Volume*>(entity)) {
```
petersteneteg: you could add an extra argument for a std::function<std::string(size_t axis)> to get the captions, then you don'e have to deal with the volume and dynamic cast it here...

petersteneteg: Or we go all the way and add that to SpatialEntity as we talked about

petersteneteg (dismissed):
might fix the range warning and remove 
include/inviwo/core/datastructures/image/imageramprecision.h
if you want to...

modules/plottinggl/src/utils/axis3dprocessorhelper.cpp
martinfalk: Added virtual function `const Axis* getAxis(size_t) const` to SpatialEntity.

petersteneteg: hmm you still need the dims? 
should cast to StructuredGridEntity in principle... that should also pull in the layer I suppose

modules/tetramesh/src/datastructures/volumetetramesh.cpp
```diff
@@ -171,4 +171,11 @@ dvec2 VolumeTetraMesh::getDataRange() const {
     }
 }
 
+const Axis* VolumeTetraMesh::getAxis(size_t index) const {
+    if (volume_) {
+        return volume_->getAxis(index);
```
petersteneteg, line 176: this would have been difficult the just the axis member in the base...

petersteneteg (dismissed):

modules/plottinggl/src/utils/axis3dprocessorhelper.cpp
martinfalk: > hmm you still need the dims?

Yes, for referring to texels/voxels.
Good point with `StructuredGridEntity`.

petersteneteg (approved):

petersteneteg (approved):

PR #1490, opened 2024-01-31
Feature/misc

martinfalk:
Fixes various issues and bugs.


include/inviwo/core/datastructures/volume/volume.h
```diff
@@ -73,7 +73,15 @@ class IVW_CORE_API Volume : public Data<Volume, VolumeRepresentation>,
...
-    Volume(const Volume&, NoData);
+    /**
+     * Create a volume based on \p rhs without copying any data. If \p defaultFormat is a nullptr,
+     * the format of \p rhs is used.
+     * @param rhs             source volume providing the necessary information like dimensions,
+     *                        swizzle masks, interpolation, spatial transformations, etc.
+     * @param defaultFormat   data format of the new volume. If equal to nullptr, the format of \p
+     *                        rhs is used instead.
+     */
+    Volume(const Volume& rhs, NoData, const DataFormatBase* defaultFormat = nullptr);
```
petersteneteg: Potentially we could have optional arguments for all the parameters,
```suggestion
    Volume(const Volume& rhs, NoData, std::optional<size3_t> dims = {}, const DataFormatBase* defaultFormat = nullptr, std:optional<SwizzleMask> defaultSwissleMaks = {}, std::optional<InterpolationType> interpolation = {}, std::optional<Wrapping3D> wrapping = {});
```

I think it would be nice to have the same order of arguments as in the main constructor above...

modules/base/src/processors/volumesubset.cpp
```diff
@@ -67,26 +68,24 @@ const ProcessorInfo VolumeSubset::processorInfo_{
...
-    , rangeZ_("rangeZ", "Z Slices", 0, 256, 0, 256, 1, 1) {
-    addPort(inport_);
-    addPort(outport_);
-    addProperty(enabled_);
-    addProperty(adjustBasisAndOffset_);
-    addProperty(rangeX_);
-    addProperty(rangeY_);
-    addProperty(rangeZ_);
-    dims_ = size3_t(1, 1, 1);
+    , rangeX_("rangeX", "X Slices", "Range X axis"_help, 0, 256, 0, 256, 1, 1)
```
petersteneteg: might be nice to explain what type of range this is (open, closed, etc)

petersteneteg (dismissed):

include/inviwo/core/datastructures/volume/volume.h
martinfalk: Implemented for Volume and Layer.

petersteneteg (approved):

PR #1492, opened 2024-02-02
ModuleSystem: Fix CMake error when network editor is not enabled

ResearchDaniel:
Fixes CMake error when network editor is not enabled

### DIFF
--- a/src/sys/CMakeLists.txt
+++ b/src/sys/CMakeLists.txt
@@ -50,8 +50,10 @@ ivw_install_helper(TARGET inviwo-module-system
 # Save information for python tools. used for regression
 ivw_mod_dep_to_mod_name(ivw_module_mod_names ${enabled_modules})
 ivw_mod_name_to_class(ivw_module_classes ${ivw_module_mod_names})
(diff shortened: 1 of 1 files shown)


petersteneteg:
Did you test if this still works for a regular build? I don't think the inviwo target exists when this is called, but since it uses generator expressions in the file generation it will look for inviwo at generation time when it does exist.

petersteneteg:
I can refactor it if you want

ResearchDaniel:
> Did you test if this still works for a regular build? I don't think the inviwo target exists when this is called, but since it uses generator expressions in the file generation it will look for inviwo at generation time when it does exist.

I did not. The runners should fail in that case I hope. More appropriate to have this code in the editor? Feel free to refactor.

petersteneteg:
yes this broke regression http://jenkins.inviwo.org:8080/job/inviwo/job/PR-1492/

ResearchDaniel:
Using IVW_APP_INVIWO instead of the target

petersteneteg (approved):

PR #1494, opened 2024-02-05
Feature/layer

martinfalk:
Changes proposed in this PR:
 * added a `LayerGLProcessor` similar to the ones used for `Image` and `Volume`
 * expose Layer ports in Python
 * added a `VolumeChannelCombiner` processor for creating a single volume from multiple channels, which shares the implementation with `LayerCombiner`
 * fixed `VolumeSubSet` processor. Slice values now match the ones of `CubeProxyGeometryProcessor`


modules/basegl/src/processors/layerprocessing/layerglprocessor.cpp
```diff
@@ -0,0 +1,115 @@
...
+void LayerGLProcessor::process() {
+    bool reattach = false;
+
+    if (internalInvalid_) {
+        reattach = true;
+        internalInvalid_ = false;
+        layer_ = std::make_shared<Layer>(*inport_.getData(), noData, std::nullopt, dataFormat_);
+        if (dataFormat_) {
+            layer_->setSwizzleMask(swizzleMask_);
+        }
```
petersteneteg: why if(dataFormat)?

modules/basegl/include/modules/basegl/processors/layerprocessing/layerglprocessor.h
```diff
@@ -0,0 +1,102 @@
...
+
+    LayerInport inport_;
+    LayerOutport outport_;
+    std::shared_ptr<Layer> layer_;
+
+    const DataFormatBase* dataFormat_;
+    // if a custom data format is specified, i.e. dataFormat_ != nullptr, this swizzle mask is used
+    SwizzleMask swizzleMask_;
+    bool internalInvalid_;
+    std::string fragmentShader_;
```
petersteneteg: fragmentShader string?

modules/basegl/include/modules/basegl/processors/layerprocessing/layerglprocessor.h
```diff
@@ -0,0 +1,102 @@
...
+     * overwrite this function in the derived class to be notified of inport onChange events
+     */
+    virtual void afterInportChanged();
+
+    LayerInport inport_;
+    LayerOutport outport_;
+    std::shared_ptr<Layer> layer_;
+
+    const DataFormatBase* dataFormat_;
+    // if a custom data format is specified, i.e. dataFormat_ != nullptr, this swizzle mask is used
```
petersteneteg: but why?

modules/basegl/include/modules/basegl/processors/layerprocessing/layerglprocessor.h
```diff
@@ -0,0 +1,102 @@
...
+     *
+     * overwrite this function in the derived class to be notified of inport onChange events
+     */
+    virtual void afterInportChanged();
+
+    LayerInport inport_;
+    LayerOutport outport_;
+    std::shared_ptr<Layer> layer_;
+
+    const DataFormatBase* dataFormat_;
```
petersteneteg: Pass this in the constructor? default nullptr?

modules/basegl/include/modules/basegl/processors/layerprocessing/layerglprocessor.h
```diff
@@ -0,0 +1,102 @@
...
+     *
+     * overwrite this function in the derived class to perform post-processing
+     */
+    virtual void postProcess();
+
+    /*! \brief this function gets called whenever the inport changes
+     *
+     * overwrite this function in the derived class to be notified of inport onChange events
+     */
+    virtual void afterInportChanged();
```
petersteneteg: Do we need to expose this?

petersteneteg (changes requested):

modules/basegl/include/modules/basegl/processors/layerprocessing/layerglprocessor.h
martinfalk: Yes, the derived processors may have properties depending on the input data.

https://github.com/inviwo/inviwo/blob/c696c112bf966e14dd90001cc70dcb1ba7974e02/modules/basegl/src/processors/imageprocessing/imagemapping.cpp#L88-L105
https://github.com/inviwo/inviwo/blob/c696c112bf966e14dd90001cc70dcb1ba7974e02/modules/basegl/src/processors/volumeprocessing/volumegradientmagnitude.cpp#L86-L100

modules/basegl/include/modules/basegl/processors/layerprocessing/layerglprocessor.h
martinfalk: Some processors may need to change the data format. This is the same mechanism as in `ImageGLProcessor` and `VolumeGLProcessor`.
For example
https://github.com/inviwo/inviwo/blob/c696c112bf966e14dd90001cc70dcb1ba7974e02/modules/basegl/src/processors/imageprocessing/imagemapping.cpp#L88-L105

modules/basegl/include/modules/basegl/processors/layerprocessing/layerglprocessor.h
martinfalk: Good catch. Left-over from https://github.com/inviwo/inviwo/commit/9c67ca9eeabe3621b75d89f3d0e3fcfa017379f8

modules/basegl/src/processors/layerprocessing/layerglprocessor.cpp
martinfalk: So that swizzleing is only applied if there is a custom output format defined. Based on ImageGL and VolumeGL processors.

modules/basegl/include/modules/basegl/processors/layerprocessing/layerglprocessor.h
petersteneteg: Ok. although the first one could just as well run in process from what I see...
For the second case I would rather just have all 4 channels always a give an error if the data does not have the channel specified, but sure.  We should generally try to avoid having properties depend on data as much as possible.

modules/basegl/include/modules/basegl/processors/layerprocessing/layerglprocessor.h
petersteneteg: yes, but why does that change the swizzle?

petersteneteg: ahh now I think I understand, you expect the user to set the dataforamt and swizzle in the afterInportChange function... That is _not_ very clear...

You should have a function that you override like getDataFormat(const Layer& input) const; 
and similar for swizzle, with appropriate defaults. and just remove the format and swizzle members.

petersteneteg (changes requested):
see comments

petersteneteg:
@martinfalk you might want to have a look at my updates

petersteneteg (approved):

PR #1497, opened 2024-02-08
Editor: Move UndoManager to applicationbase to allow other apps to use it

ResearchDaniel:


src/qt/applicationbase/undomanager.cpp
```diff
@@ -97,13 +95,6 @@ class AutoSaver {
                 }
 
                 if (str) {
-                    int numRestoreFiles = 100000;
```
petersteneteg, line 100: The new version will need to restart inviwo to have this take effect... :( 
Better to pass a lambda if you don't what to couple to the editorsettings

src/qt/applicationbase/undomanager.cpp
```diff
@@ -174,23 +165,24 @@ class AutoSaver {
...
 
 {
 
     QApplication::instance()->installEventFilter(this);
 
-    mainWindow_->getInviwoApplication()->getProcessorNetwork()->addObserver(this);
+    app->getProcessorNetwork()->addObserver(this);
 
-    undoAction_ = new QAction(QIcon(":/svgicons/undo.svg"), QAction::tr("&Undo"), mainWindow_);
+    undoAction_ = new QAction(QIcon(":/svgicons/undo.svg"), QAction::tr("&Undo"), mainWindow);
```
petersteneteg: It here a reason to parent them to the mainWindow and not just this?

src/qt/applicationbase/undomanager.cpp
```diff
@@ -174,23 +165,24 @@ class AutoSaver {
     std::thread saver_;
 };
 
-UndoManager::UndoManager(InviwoMainWindow* mainWindow)
-    : mainWindow_(mainWindow)
-    , manager_{mainWindow_->getInviwoApplication()->getWorkspaceManager()}
-    , refPath_{filesystem::findBasePath()}
-    , autoSaver_{std::make_unique<AutoSaver>(mainWindow_->getInviwoApplication())}
+UndoManager::UndoManager(InviwoApplication* app, QObject* mainWindow, int numRestoreFiles,
```
petersteneteg: I would pass just the Network, and the Workspace Manger instead.

petersteneteg (changes requested):

src/qt/applicationbase/undomanager.cpp
ResearchDaniel: No change from before

src/qt/applicationbase/undomanager.cpp
ResearchDaniel: Have not tested. 
Could try and see

src/qt/applicationbase/undomanager.cpp
petersteneteg: Yes this is a change from before since it was pull on every save before

ResearchDaniel: fixed

src/qt/applicationbase/undomanager.cpp
ResearchDaniel: Tested with this and could not see any difference

src/qt/applicationbase/undomanager.cpp
ResearchDaniel: Done

petersteneteg (dismissed):

petersteneteg (approved):

PR #1499, opened 2024-02-28
Feature/error reporting

petersteneteg:
Changes proposed in this PR:
 * The indicator light on processors will now indicate errors for each processor

image

Is ready has a reason
image

source processor error:
image

### DIFF
--- a/include/inviwo/core/processors/processorstatus.h
+++ b/include/inviwo/core/processors/processorstatus.h
@@ -1,0 +1,114 @@
+/*********************************************************************************
+ *
+ * Inviwo - Interactive Visualization Workshop
+ *
+ * Copyright (c) 2024 Inviwo Foundation
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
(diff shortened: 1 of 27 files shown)


include/inviwo/qt/editor/processorgraphicsitem.h
```diff
@@ -119,18 +124,23 @@ class IVW_QTEDITOR_API ProcessorGraphicsItem : public EditorGraphicsItem,
...
-#if IVW_PROFILING
     virtual void onProcessorAboutToProcess(Processor*) override;
     virtual void onProcessorFinishedProcess(Processor*) override;
-#endif
+
+    virtual void onProcessorStartBackgroundWork(Processor*, size_t jobs) override {
+        backgroundJobs_ += jobs;
+    };
+    virtual void onProcessorFinishBackgroundWork(Processor*, size_t jobs) override {
+        backgroundJobs_ -= jobs;
```
martinfalk, line 134: overflow/underflow handling due to size_t?

martinfalk (dismissed):

include/inviwo/core/processors/processorstatus.h
```diff
@@ -0,0 +1,114 @@
...
+#include <inviwo/core/util/sourcecontext.h>
+#include <inviwo/core/util/exception.h>
+
+#include <string_view>
+
+namespace inviwo {
+
+class IVW_CORE_API ProcessorStatus {
+public:
+    enum class Status { Ready, NotReady, Error };
```
ResearchDaniel, line 42: Invalid is not necessary?

modules/vectorfieldvisualizationgl/src/processors/3d/streamparticles.cpp
```diff
@@ -169,6 +182,15 @@ StreamParticles::StreamParticles(InviwoApplication* app)
 }
 
 void StreamParticles::initializeResources() {
+    if (OpenGLCapabilities::getOpenGLVersion() < 430) {
+        isReady_.setUpdate([]() -> ProcessorStatus {
+            static constexpr std::string_view error{
+                "The StreamParticles processor requires OpenGL 4.3"};
+            return {ProcessorStatus::Error, error};
+        });
+        throw Exception(IVW_CONTEXT, "The StreamParticles processor requires OpenGL 4.3");
```
ResearchDaniel, line 191: Could it be that it is always desirable to set isReady with error status when an exception is thrown? In that case, could we manage this behind the scenes? Or make a new type of exception that does it for you?

modules/vectorfieldvisualizationgl/src/vectorfieldvisualizationglmodule.cpp
```diff
@@ -80,10 +78,7 @@ VectorFieldVisualizationGLModule::VectorFieldVisualizationGLModule(InviwoApplica
     registerProcessor<TMIP>();
     registerProcessor<VectorFieldGenerator4D>();
 
-    auto& gl = app->getModuleByType<OpenGLModule>()->getOpenGLCapabilities();
-    if (gl.getOpenGLVersion() >= 430) {
-        registerProcessor<StreamParticles>();
-    }
+    registerProcessor<StreamParticles>();
```
ResearchDaniel, line 81: Nice!

include/inviwo/qt/editor/processorgraphicsitem.h
petersteneteg: As long as nothing else is wrong, that should never happen...

ResearchDaniel (approved):
Nice!
Would probably be nice to have a close button in the top right corner of the info.

modules/vectorfieldvisualizationgl/src/processors/3d/streamparticles.cpp
petersteneteg: Exceptions are also handled automatically

include/inviwo/core/processors/processorstatus.h
petersteneteg: No, we have never had an invalid state for the isReady that is orthogonal

petersteneteg:
> Nice! Would probably be nice to have a close button in the top right corner of the info.

If you unselect the processor it is gone

modules/vectorfieldvisualizationgl/src/processors/3d/streamparticles.cpp
petersteneteg: In a different way using the evaluator error handle

PR #1500, opened 2024-02-28
VolumeSequence

petersteneteg:
Added a class for VolumeSequence instead of using a std::vector of shared_ptr Volume.

Fixes #Make it possible to create a VolumeSequence of shared_ptr<const Volume>

ensures that a const VolumeSequence returns const Volumes. i.e. it propagates const.

### DIFF
--- a/include/inviwo/core/util/transformiterator.h
+++ b/include/inviwo/core/util/transformiterator.h
@@ -54,6 +54,7 @@ struct TransformIterator {
     using difference_type = typename std::iterator_traits<Iter>::difference_type;
     using iterator_category = typename std::iterator_traits<Iter>::iterator_category;
 
+    using base_iterator = Iter;
     using base_value = typename std::iterator_traits<Iter>::value_type;
     using base_reference = typename std::iterator_traits<Iter>::reference;
     using base_pointer = typename std::iterator_traits<Iter>::pointer;
@@ -63,7 +64,9 @@ struct TransformIterator {
     using pointer = std::add_pointer_t<value_type>;
 
     TransformIterator() = default;
-    TransformIterator(Iter iterator) : transform_{}, iterator_(iterator) {}
+    TransformIterator(Iter iterator)
+        requires std::default_initializable<Transform>
+        : transform_{Transform{}}, iterator_(iterator) {}
     TransformIterator(Transform transform, Iter iterator)
         : transform_{std::move(transform)}, iterator_(iterator) {
         static_assert(std::is_same<base_reference, decltype(*std::declval<Iter>())>::value, "");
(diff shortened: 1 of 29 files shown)


include/inviwo/core/util/transformiterator.h
```diff
@@ -91,6 +94,13 @@ struct TransformIterator {
         return *this;
     }
 
+    template <typename Other>
+    TransformIterator(const Other& other)
+        requires(!std::is_same_v<TransformIterator, Other>) &&
```
martinfalk, line 99: Why must Other be different?

modules/base/src/processors/volumesequencesource.cpp
```diff
@@ -92,14 +94,20 @@ std::optional<std::filesystem::path> getFirstFileInFolder(const std::filesystem:
...
+    , inputType_(
+          "inputType", "Input type",
+          "Select the input type, either select a single file to a 4D dataset or use a folder"_help,
+          {{"singlefile", "Single File", InputType::SingleFile},
+           {"folder", "Folder", InputType::Folder}},
+          1)
+    , file_("filename", "Volume file", "If using single file mode, the file to load"_help)
+    , folder_("folder", "Volume folder",
+              "If using folder mode, the folder to look for data sets in"_help)
+    , filter_(
```
martinfalk, line 107: In the future, we should probably use a file pattern property here instead of a plain string.

modules/python3/bindings/src/pyvolume.cpp
```diff
@@ -144,7 +144,101 @@ void exposeVolume(pybind11::module& m) {
...
 
-    py::bind_vector<VolumeSequence>(m, "VolumeSequence", py::module_local(false));
+    py::class_<VolumeSequence>(m, "VolumeSequence", py::module_local(false))
+        .def(py::init<>())
+        .def(py::init<std::span<std::shared_ptr<const Volume>>>(), py::arg("volumes"))
+        .def(py::init<const VolumeSequence&>(), "Copy constructor")
+
+        .def("__getitem__",
+             [](const VolumeSequence& v, std::ptrdiff_t i) -> std::shared_ptr<const Volume> {
+                 if (i < 0 && (i += v.size()) < 0) {
```
martinfalk, line 154: Why the special handling of negative indices? Does this cover the cases like `v[-1]` ?

Use parenthesis around right condition as i is modified?

martinfalk (approved):

modules/python3/bindings/src/pyvolume.cpp
petersteneteg: Yes, this is for [-1] in python

modules/base/src/processors/volumesequencesource.cpp
petersteneteg: if someone cleans up those...

include/inviwo/core/util/transformiterator.h
petersteneteg: if they are the same the copy constructor will be used.

PR #1505, opened 2024-03-01
Feature/layer2

martinfalk:
Changes proposed in this PR:
* shader clean-up getting rid of `_` in uniform names
 * most 2D vector field processors now use Layer instead of Image
    + extended 2D LIC processor with option for image post processing (brightness, contrast, gamma)
 * data on gradient spacing for Layers similar to Volume
 * added more LayerGL processors matching existing ImageGL ones
    + refactored some ImageGL processors 
 * added Layer resampling processor using Cimg scaling


data/workspaces/portinspectors/imageportinspector.inv
petersteneteg, line 1: Guess one could manually remove the previews from the port inspectors?

petersteneteg: and other unneeded stuff...

modules/base/include/modules/base/algorithm/image/layercontour.h
```diff
@@ -40,10 +40,8 @@ namespace inviwo {
 class LayerRepresentation;
 class Mesh;
 
-class IVW_MODULE_BASE_API ImageContour {
-public:
-    static std::shared_ptr<Mesh> apply(const LayerRepresentation* in, size_t channel,
-                                       double isoValue, vec4 color = vec4(1.0));
-};
+IVW_MODULE_BASE_API std::shared_ptr<Mesh> computeLayerContour(const LayerRepresentation* in,
```
petersteneteg, line 43: what about more then one contour? would be nice to get them the same mesh right?

modules/base/include/modules/base/algorithm/randomutils.h
```diff
@@ -91,21 +91,23 @@ std::vector<T> haltonSequence(size_t base, size_t numberOfPoints) {
...
  */
 template <typename T>
-std::shared_ptr<Image> haltonSequence(size2_t dims, size_t numberOfPoints, size_t baseX = 2,
+std::shared_ptr<Layer> haltonSequence(size2_t dims, size_t numberOfPoints, size_t baseX = 2,
                                       size_t baseY = 3) {
 
-    std::shared_ptr<Image> img = std::make_shared<Image>(dims, DataFormat<T>::get());
-    auto ram = static_cast<LayerRAMPrecision<T>*>(
-        img->getColorLayer()->getEditableRepresentation<LayerRAM>());
+    std::shared_ptr<Layer> layer =
```
petersteneteg: create the LayerRAMPrecision<T> first and add that to the layer...

modules/base/include/modules/base/algorithm/randomutils.h
```diff
@@ -222,15 +224,17 @@ template <typename T, typename Rand = std::mt19937,
           typename Dist = typename std::conditional<std::is_integral<T>::value,
                                                     std::uniform_int_distribution<T>,
                                                     std::uniform_real_distribution<T>>::type>
-std::shared_ptr<Image> randomImage(size2_t dims, Rand& randomNumberGenerator = Rand(),
+std::shared_ptr<Layer> randomLayer(size2_t dims, Rand& randomNumberGenerator = Rand(),
                                    Dist& distribution = Dist(0, 1)) {
-    std::shared_ptr<Image> img = std::make_shared<Image>(dims, DataFormat<T>::get());
-    auto ram = static_cast<LayerRAMPrecision<T>*>(
-        img->getColorLayer()->getEditableRepresentation<LayerRAM>());
+    std::shared_ptr<Layer> layer =
```
petersteneteg: create the LayerRAMPrecision first and add that to the layer...

modules/base/include/modules/base/algorithm/randomutils.h
```diff
@@ -261,39 +265,41 @@ std::shared_ptr<Volume> randomVolume(size3_t dims, Rand& randomNumberGenerator =
...
-        samplers.push_back(TemplateImageSampler<float, float>(img1.get()));
-        levels.push_back(img1);
+        auto randomLayer = util::randomLayer<float>(layerSize, randomNumberGenerator, dist);
+        samplers.push_back(TemplateImageSampler<float, float>(randomLayer.get()));
+        levels.push_back(randomLayer);
         currentSize *= 2;
         currentPersistance *= persistence;
     }
 
+    std::shared_ptr<Layer> layer =
```
petersteneteg: create the LayerRAMPrecision first and add that to the layer...

petersteneteg: this might even just work with LayerRAMPrecision<float> no need for layers here?

modules/base/include/modules/base/algorithm/randomutils.h
```diff
@@ -346,11 +350,15 @@ std::shared_ptr<Image> poissonDisk(size2_t dims, size_t poissonDotsAlongX, size_
         return glm::i32vec2(newX, newY);
     };
 
-    auto gridImg = std::make_unique<Image>(gridSize, DataVec2Int32::get());
-    auto grid = gridImg->getColorLayer()->getEditableRepresentation<LayerRAM>();
+    std::shared_ptr<Layer> layer =
```
petersteneteg: create the LayerRAMPrecision first and add that to the layer...

or retur the LayerRAMPrecision<float> instead

modules/base/include/modules/base/processors/layercontour.h
```diff
@@ -24,30 +24,35 @@
...
-    vec2 texCoords = gl_FragCoord.xy * outportParameters_.reciprocalDimensions;
-    vec2 gradient;
-    // compute gradient using central differences
-    gradient.x = (texture(inport_,texCoords + dx )[channel] - texture(inport_,texCoords - dx )[channel]) * 0.5;
-    gradient.y = (texture(inport_,texCoords + dy )[channel] - texture(inport_,texCoords - dy )[channel]) * 0.5;
+private:
+    LayerInport inport_;
+    MeshOutport outport_;
+    OptionPropertyInt channel_;
+    DoubleProperty isoValue_;
```
petersteneteg, line 54: probably want several values. Use the IsoValueCollection

modules/base/src/processors/imagecontourprocessor.cpp
```diff
@@ -69,14 +69,14 @@ const ProcessorInfo ImageContourProcessor::getProcessorInfo() const { return pro
 ImageContourProcessor::ImageContourProcessor()
     : Processor()
     , image_("image", "Input image"_help, OutportDeterminesSize::Yes)
-    , mesh_("mesh", "Contour mesh"_help)
+    , outport_("mesh", "Contour mesh"_help)
     , channel_("channel", "Channel", "The image channel used to extract the contour"_help, 0,
                {0, ConstraintBehavior::Immutable}, {4, ConstraintBehavior::Editable})
-    , isoValue_("iso", "ISO Value", "The contour iso value"_help, 0.5,
+    , isoValue_("iso", "Isovalue", "The isovalue of the contour "_help, 0.5,
```
petersteneteg, line 75: use IsoValueCollection

modules/base/src/processors/noisegenerator2d.cpp
```diff
@@ -154,36 +156,35 @@ NoiseProcessor::NoiseProcessor()
...
 
-void NoiseProcessor::process() {
+void NoiseGenerator2D::process() {
     if (useSameSeed_.get()) {
         mt_.seed(seed_.get());
     }
 
     std::uniform_real_distribution<float> r(range_.get().x, range_.get().y);
-    std::shared_ptr<Image> img;
+    std::shared_ptr<Layer> layer;
```
petersteneteg, line 149: You might add a LayerInformationProperty to this so you can sett some "meta" data if needed

petersteneteg: also maybe the basis property

modules/basegl/glsl/img_binary.frag
```diff
@@ -29,15 +29,17 @@
...
 
-uniform sampler2D inport_;
-uniform ImageParameters outportParameters_;
+#ifndef COMPARE
+#define COMPARE >
+#endif
 
-
-uniform float threshold;
+uniform sampler2D inport;
```
petersteneteg, line 36: NIce!

petersteneteg (changes requested):
Nice! Good to get all those uniforms consistently named

modules/base/include/modules/base/algorithm/image/layercontour.h
martinfalk: separate PR?

petersteneteg: sure

modules/base/src/processors/imagedistancetransform.cpp
```diff
@@ -58,24 +59,38 @@
...
-    "Image Operation",                       // Category
-    CodeState::Stable,                       // Code state
-    Tags::CPU,                               // Tags
+const ProcessorInfo ImageDistanceTransform::processorInfo_{
+    "org.inviwo.ImageDistanceTransform",  // Class identifier
+    "Image Distance Transform",           // Display name
+    "Image Operation",                    // Category
+    CodeState::Stable,                    // Code state
+    Tags::CPU,                            // Tags
+    R"(Computes the distance transform of a layer dataset using a threshold value
```
ResearchDaniel:
```suggestion
    R"(Computes the closest distance to a threshold value for each pixel in a Layer.
```

modules/base/src/processors/imagedistancetransform.cpp
```diff
@@ -58,24 +59,38 @@
...
+    The result is the distance from each pixel to the closest feature. It will only work
```
ResearchDaniel:
```suggestion
    It uses the Saito's algorithm to compute
     the Euclidean distance. Note: Only works correctly for Layers with orthogonal basis.
```

modules/base/src/processors/imagedistancetransform.cpp
```diff
@@ -58,24 +59,38 @@
...
+    correctly for layers with a orthogonal basis. It uses the Saito's algorithm to compute
```
ResearchDaniel:
```suggestion
```

modules/base/src/processors/imagedistancetransform.cpp
```diff
@@ -58,24 +59,38 @@
...
+    the Euclidean distance.)"_unindentHelp,
```
ResearchDaniel:
```suggestion
    )"_unindentHelp,
```

modules/base/src/processors/imagedistancetransform.cpp
```diff
@@ -58,24 +59,38 @@
...
-    , flip_("flip", "Flip", false)
-    , normalize_("normalize", "Use normalized threshold", true)
-    , resultDistScale_("distScale", "Scaling Factor", 1.0f, 0.0f, 1.0e3, 0.05f)
-    , resultSquaredDist_("distSquared", "Squared Distance", false)
+    , imagePort_("inputImage", "Input image"_help)
+    , outport_("outputImage", "Scalar image representing the distance transform (float)"_help,
+               DataVec4UInt8::get(), HandleResizeEvents::No)
+    , threshold_(
+          "threshold", "Threshold",
+          "Pixels with a value  __larger___ then the then the threshold will be considered as features, i.e. have a zero distance."_help,
```
ResearchDaniel:
```suggestion
          "Pixels with a value  __larger___ than the threshold will be considered as features, i.e. have a zero distance."_help,
```

modules/base/src/processors/imagedistancetransform.cpp
```diff
@@ -58,24 +59,38 @@
...
+          0.5, {0.0, ConstraintBehavior::Ignore}, {1.0, ConstraintBehavior::Ignore})
+    , flip_("flip", "Flip",
+            "Consider features as pixels with a values __smaller__ then threshold instead."_help,
```
ResearchDaniel:
```suggestion
            "Consider features as pixels with a values __smaller__ than threshold instead."_help,
```

modules/base/src/processors/imagedistancetransform.cpp
```diff
@@ -58,24 +59,38 @@
...
-    , resultDistScale_("distScale", "Scaling Factor", 1.0f, 0.0f, 1.0e3, 0.05f)
-    , resultSquaredDist_("distSquared", "Squared Distance", false)
+    , imagePort_("inputImage", "Input image"_help)
+    , outport_("outputImage", "Scalar image representing the distance transform (float)"_help,
+               DataVec4UInt8::get(), HandleResizeEvents::No)
+    , threshold_(
+          "threshold", "Threshold",
+          "Pixels with a value  __larger___ then the then the threshold will be considered as features, i.e. have a zero distance."_help,
+          0.5, {0.0, ConstraintBehavior::Ignore}, {1.0, ConstraintBehavior::Ignore})
+    , flip_("flip", "Flip",
```
ResearchDaniel, line 83: Should be three options: zero if larger, zero if smaller, sign (e.g negative if smaller)

modules/base/src/processors/layerdistancetransform.cpp
```diff
@@ -0,0 +1,138 @@
...
+namespace inviwo {
+
+// The Class Identifier has to be globally unique. Use a reverse DNS naming scheme
+const ProcessorInfo LayerDistanceTransform::processorInfo_{
+    "org.inviwo.LayerDistanceTransform",  // Class identifier
+    "Layer Distance Transform",           // Display name
+    "Layer Operation",                    // Category
+    CodeState::Stable,                    // Code state
+    Tags::CPU,                            // Tags
+    R"(Computes the distance transform of a Layer using a threshold value
```
ResearchDaniel: See comments in image transform

modules/basegl/glsl/img_binary.frag
```diff
@@ -29,15 +29,17 @@
...
+uniform float threshold = 0.5;
 
 void main() {
-    vec2 texCoords = gl_FragCoord.xy * outportParameters_.reciprocalDimensions;
-    bool b = texture(inport_, texCoords).r >= threshold;
-    float v = b ? 1.0f : 0.0f;
-    FragData0 = vec4(v,v,v,1.0);
+    vec2 texCoords = gl_FragCoord.xy * outportParameters.reciprocalDimensions;
+    bvec4 b = bvec4(texture(inport, texCoords)[channel] COMPARE threshold);
+    FragData0 = mix(vec4(0), vec4(1), b);
```
ResearchDaniel, line 44: This will change behavior of last channel. Makes sense that it is zero if mask is zero

modules/basegl/glsl/img_normalize.frag
```diff
@@ -24,29 +24,23 @@
...
+uniform ImageParameters outportParameters;
 
-uniform vec4 min_;    
-uniform vec4 max_;    
+uniform vec4 minValue = vec4(0);
+uniform vec4 maxValue = vec4(1);
 
-uniform float typeMax_;
-uniform float typeMin_;
+uniform float minDataType = 0.0;
```
ResearchDaniel, line 38: Passing in scale and offset would be more efficient

modules/basegl/src/processors/imageprocessing/imagebinary.cpp
```diff
@@ -56,12 +48,42 @@ const ProcessorInfo ImageBinary::getProcessorInfo() const { return processorInfo
...
+
+    auto layer = outport_.getEditableData()->getColorLayer();
+    layer->dataMap.dataRange = DataMapper::defaultDataRangeFor(layer->getDataFormat());
+    layer->dataMap.valueRange = dvec2{0.0, 1.0};
+}
+
+void ImageBinary::afterInportChanged() {
+    if (inport_.hasData()) {
+        const DataFormatBase* format = inport_.getData()->getColorLayer()->getDataFormat();
+        dataFormat_ = DataFormatBase::get(format->getNumericType(), 1, format->getPrecision());
```
ResearchDaniel: This should always be 8-bit uint?

modules/basegl/src/processors/imageprocessing/imagegradient.cpp
```diff
@@ -86,12 +80,23 @@ ImageGradient::ImageGradient()
...
+    const auto gradientSpacing{layer->getWorldSpaceGradientSpacing()};
+    shader_.setUniform("worldSpaceGradientSpacing", gradientSpacing);
+    shader_.setUniform("textureSpaceGradientSpacing",
+                       mat2{glm::scale(layer->getCoordinateTransformer().getWorldToTextureMatrix(),
+                                       vec3{gradientSpacing, 1.0f})});
+
     shader_.setUniform("channel", channel_.getSelectedValue());
-    shader_.setUniform("renormalization_", renormalization_.get() ? 1 : 0);
+
+    const double gradientEstimate = glm::compMax(glm::abs(layer->dataMap.dataRange)) /
```
ResearchDaniel, line 95: Should be value range?
```suggestion
    const double gradientEstimate = glm::compMax(glm::abs(layer->dataMap.valueRange)) /
```

modules/basegl/src/processors/layerprocessing/layerbinary.cpp
```diff
@@ -97,9 +97,12 @@ void LayerBinary::preProcess(TextureUnitContainer&, const Layer&, Layer&) {
 }
 
 LayerConfig LayerBinary::outputConfig(const Layer& input) const {
+    auto format = input.getDataFormat();
```
ResearchDaniel, line 77: Should not depend on input? Can always be 8 bit

modules/basegl/src/processors/layerprocessing/layergradient.cpp
```diff
@@ -0,0 +1,75 @@
...
+    shader_.setUniform("worldSpaceGradientSpacing", gradientSpacing);
+    shader_.setUniform("textureSpaceGradientSpacing",
+                       mat2{glm::scale(input.getCoordinateTransformer().getWorldToTextureMatrix(),
+                                       vec3{gradientSpacing, 1.0f})});
+
+    shader_.setUniform("channel", channel_.getSelectedValue());
+}
+
+LayerConfig LayerGradient::outputConfig(const Layer& input) const {
+    const double gradientEstimate = glm::compMax(glm::abs(input.dataMap.dataRange)) /
```
ResearchDaniel, line 67: Why dataRange?

modules/basegl/src/processors/layerprocessing/layermapping.cpp
```diff
@@ -34,20 +34,21 @@
...
 // The Class Identifier has to be globally unique. Use a reverse DNS naming scheme
-const ProcessorInfo LayerMapping::processorInfo_{"org.inviwo.LayerMapping",  // Class identifier
-                                                 "Layer Mapping",            // Display name
-                                                 "Layer Operation",          // Category
-                                                 CodeState::Stable,          // Code state
-                                                 Tags::GL,                   // Tags
-                                                 R"(
-Maps the input image to an output image with the help of a transfer function.
-)"_unindentHelp};
+const ProcessorInfo LayerMapping::processorInfo_{
```
ResearchDaniel: Would love to have called this color mapping…

modules/basegl/src/processors/layerprocessing/layermapping.cpp
```diff
@@ -34,20 +34,21 @@
...
+    "org.inviwo.LayerMapping",  // Class identifier
+    "Layer Mapping",            // Display name
+    "Layer Operation",          // Category
+    CodeState::Stable,          // Code state
+    Tags::GL,                   // Tags
+    R"(Applies a transferfunction to one channel of a Layer. The format of the RGBA output will
+    match the input with a data range of [0, 1].)"_unindentHelp,
```
ResearchDaniel:
```suggestion
    match the input with a value range of [0, 1].)"_unindentHelp,
```

modules/basegl/src/processors/layerprocessing/layermapping.cpp
```diff
@@ -34,20 +34,21 @@
...
+};
 
 const ProcessorInfo LayerMapping::getProcessorInfo() const { return processorInfo_; }
 
 LayerMapping::LayerMapping()
     : LayerGLProcessor{utilgl::findShaderResource("img_mapping.frag")}
-    , channel_{"channel", "Channel", "Selected channel used for mapping"_help,
+    , channel_{"channel", "Channel", "Selected channel used for TF mapping"_help,
```
ResearchDaniel:
```suggestion
    , channel_{"channel", "Channel", "Selected channel used for color mapping"_help,
```

modules/cimg/src/processors/layerresampling.cpp
```diff
@@ -0,0 +1,131 @@
...
+              1}
+    , inputDimensions_{"inputDimensions", "Input Dimensions",
+                       util::ordinalCount(size2_t{0}, size2_t{4096})
+                           .set(PropertySemantics::Text)
+                           .set("Size of the input layer (read-only)"_help)}
+    , outputDimensions_{"outputDimensions", "Output Dimensions",
+                        util::ordinalCount(size2_t{0}, size2_t{4096})
+                            .set("Size of the resampled layer"_help)}
+    , scaling_{"scaling", "Scaling",
+               util::ordinalScale(1.0f, 4.0f)
```
ResearchDaniel, line 79: Not smaller?

modules/vectorfieldvisualizationgl/glsl/vector2dcurl.frag
```diff
@@ -30,24 +30,27 @@
...
-void main(void) {
-	vec2 o = inportParameters.reciprocalDimensions;
-	mat2 J;  
-	J[0] =  texture(inportColor,texCoord_.xy  + vec2(o.x,0)).xy;
-	J[0] -= texture(inportColor,texCoord_.xy  - vec2(o.x,0)).xy;
-	J[1] =  texture(inportColor,texCoord_.xy  + vec2(0,o.y)).xy;
-	J[1] -= texture(inportColor,texCoord_.xy  - vec2(0,o.y)).xy;
-	J[0] *= 0.5;
-	J[1] *= 0.5;
+float partialDiff(in vec2 texcoord, in vec2 gradientTextureSpacing, in float gradientWorldSpacing, in int channel) {
```
ResearchDaniel: Duplicate code

modules/base/include/modules/base/algorithm/randomutils.h
```diff
@@ -50,6 +50,33 @@ namespace inviwo {
 
 namespace util {
 
+namespace detail {
+
+/**
+ * Create a data container of type @p T from a RAM representation @p ramRep and set its data range
```
ResearchDaniel:
```suggestion
 * Create a data container of type @p T from a RAM representation @p ramRep and set its value range
```

modules/base/include/modules/base/algorithm/randomutils.h
```diff
@@ -211,6 +234,31 @@ void randomSequence(T* data, size_t numberOfElements, Rand& randomNumberGenerato
                   [&]() { return distribution(randomNumberGenerator); });
 }
 
+/**
+ * Generate a LayerRAM representation with white noise based using C++ a given random number
```
ResearchDaniel: Rephrase

modules/basegl/include/modules/basegl/processors/layerprocessing/layerglprocessor.h
```diff
@@ -49,10 +49,11 @@ class TextureUnitContainer;
...
  *
  * The LayerGLProcessor provides the basic structure for layer processing on the GPU.
- * Derived shaders have to provide a custom fragment shader which is used during rendering.
- * Optionally, derived classes can overwrite LayerGLProcessor::postProcess() to perform
- * post-processing of the layer data set in the outport. Furthermore, it is possible to
- * be notified of changes in the input layer by overwriting LayerGLProcessor::afterInportChanged().
+ * Derived processors have to provide a custom fragment shader which is used during rendering.
+ * Optionally, derived classes can overwrite LayerGLProcessor::preProcess() and
+ * LayerGLProcessor::postProcess() to perform pre- and post-processing of the Layer directly before
+ * and after rendering. Furthermore, it is possible to be specify the data format of the output
```
ResearchDaniel:
```suggestion
 * and after rendering. Furthermore, it is possible to specify the data format of the output
```

ResearchDaniel (dismissed):

modules/basegl/glsl/img_normalize.frag
martinfalk: But then you have to compute it beforehand. And in this minimal shader, other things cause way more overhead than the mad operation.

modules/basegl/src/processors/imageprocessing/imagebinary.cpp
martinfalk: Any particular reason why?

modules/basegl/src/processors/layerprocessing/layerbinary.cpp
martinfalk: Yes, it should definitely. I do not want layers to change their dataformat.

modules/basegl/src/processors/layerprocessing/layermapping.cpp
martinfalk: No, the value range is not considered at all for normalization before applying the TF. Value Range is just cosmetic.

modules/cimg/src/processors/layerresampling.cpp
martinfalk: ordinalScale initializes the factor with 1.0f and ignored max of 4.0f. The lower limit is 0 + epsilon (fixed).

CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
...
+
+Layer operation processors have been added matching most of the existing image operation processors. Note that the output dimensions of these processors will typically be identical to the input. That is the dimensions are _not_ based on the size of the canvas. Use the `Layer Resampling` processor to upsample or downsample the Layer to the desired dimensions beforehand.
+
+The 2D processors in `VectorFieldVisualizationGL` module work now directly with Layers and no longer with Image. This may require restoring the port connections manually and/or adding `Layer To Image` or `Image To Layer` converters in between. Basic image post-processing functionality was added to the 2D LIC processor to increase brightness and contrast directly.
+
+## 2024-03-01 `Volume::dataMap_` renamed to `Volume::dataMap`
+The DataMapper member in Volume `dataMap_` was renamed to `dataMap` since it is a public class member.
```
petersteneteg, line 13: 👍

modules/base/include/modules/base/algorithm/randomutils.h
```diff
@@ -50,6 +50,33 @@ namespace inviwo {
...
+        typename T::repr;
+        std::is_base_of_v<typename T::repr, TRAMrep>;
+        data.template getRepresentation<TRAMrep>();
+    }
+{
+    auto data = std::make_shared<T>(ramRep);
+    data->dataMap.dataRange = DataMapper::defaultDataRangeFor(data->getDataFormat());
+    data->dataMap.valueRange = dvec2{0.0, 1.0};
+    return data;
+}
```
petersteneteg: Might consider something like this

```
#include <iostream>
#include <type_traits>
#include <memory>

template <template <typename...> class F>
struct conversion_tester {
        template <typename... Ts>
        conversion_tester (const F<Ts...> &);
};

template <class From, template <typename...> class To>
constexpr bool is_instance_of = std::is_convertible_v<From, conversion_tester<To>>;

template< typename T, typename U >
struct Data{
    using repr = U;
};

struct Volume : Data<int, float> {};

template<typename T>
concept DataType = is_instance_of<T, Data>;

template<typename D, typename Rep>
concept RepresentationOf = is_instance_of<D, Data> && std::is_base_of_v<typename D::repr, Rep>;


template <DataType T, RepresentationOf<T> TRAMrep>
std::shared_ptr<T> createFromRAMRepresentation(std::shared_ptr<TRAMrep> ramRep) {
    auto data = std::make_shared<T>(ramRep);
    //data->dataMap.dataRange = DataMapper::defaultDataRangeFor(data->getDataFormat());
    //data->dataMap.valueRange = dvec2{0.0, 1.0};
    return data;
}


int main() {
    std::cout << std::boolalpha << is_instance_of<Volume, Data> << '\n'; 
    std::cout << std::boolalpha << is_instance_of<int, Data> << '\n';
}
```

petersteneteg: Might add the concepts to data.h

petersteneteg: On the other hand you might just put that in the constructor for Volume and layer

usually we make the valueRange = dataRange unless one explicitly sets some other range.

modules/base/include/modules/base/algorithm/randomutils.h
```diff
@@ -91,35 +118,35 @@ std::vector<T> haltonSequence(size_t base, size_t numberOfPoints) {
...
     const util::IndexMapper2D index(dims);
-    auto data = ram->getDataTyped();
+    auto data = layerRam->getDataTyped();
 
     for (auto&& [xi, yi] : util::zip(x, y)) {
         auto coord = dimsf * vec2(xi, yi);
         data[index(coord)] = 1;
     }
-    return img;
+    return detail::createFromRAMRepresentation<Layer>(layerRam);
```
petersteneteg: just use the constructor of Layer

modules/base/include/modules/base/algorithm/randomutils.h
```diff
@@ -91,35 +118,35 @@ std::vector<T> haltonSequence(size_t base, size_t numberOfPoints) {
...
+    });
 
     const auto x = util::haltonSequence<T>(baseX, numberOfPoints);
     const auto y = util::haltonSequence<T>(baseY, numberOfPoints);
 
     const auto dimsf = vec2(dims - size2_t(1));
 
     const util::IndexMapper2D index(dims);
-    auto data = ram->getDataTyped();
+    auto data = layerRam->getDataTyped();
```
petersteneteg: use getView()

modules/base/include/modules/base/algorithm/randomutils.h
```diff
@@ -91,35 +118,35 @@ std::vector<T> haltonSequence(size_t base, size_t numberOfPoints) {
...
-    std::shared_ptr<Image> img = std::make_shared<Image>(dims, DataFormat<T>::get());
-    auto ram = static_cast<LayerRAMPrecision<T>*>(
-        img->getColorLayer()->getEditableRepresentation<LayerRAM>());
+    auto layerRam = std::make_shared<LayerRAMPrecision<T>>(LayerReprConfig{
+        .dimensions = dims,
+        .swizzleMask = swizzlemasks::defaultData(1),
+    });
 
     const auto x = util::haltonSequence<T>(baseX, numberOfPoints);
     const auto y = util::haltonSequence<T>(baseY, numberOfPoints);
```
petersteneteg: We should just add a function for the nth Halton number, and use that instead of creating to temp vectors here.

modules/base/include/modules/base/algorithm/randomutils.h
```diff
@@ -222,15 +272,10 @@ template <typename T, typename Rand = std::mt19937,
...
+std::shared_ptr<Layer> randomLayer(size2_t dims, Rand& randomNumberGenerator = Rand(),
                                    Dist& distribution = Dist(0, 1)) {
-    std::shared_ptr<Image> img = std::make_shared<Image>(dims, DataFormat<T>::get());
-    auto ram = static_cast<LayerRAMPrecision<T>*>(
-        img->getColorLayer()->getEditableRepresentation<LayerRAM>());
-
-    randomSequence(ram->getDataTyped(), dims.x * dims.y, randomNumberGenerator, distribution);
-
-    return img;
+    return detail::createFromRAMRepresentation<Layer>(
```
petersteneteg: The data range here will depend in the dist used.

petersteneteg: used https://en.cppreference.com/w/cpp/numeric/random/uniform_real_distribution/min / max
seems to exist for all dists

modules/base/include/modules/base/algorithm/randomutils.h
```diff
@@ -248,53 +293,53 @@ template <typename T, typename Rand = std::mt19937,
...
-std::shared_ptr<Image> perlinNoise(size2_t dims, float persistence, size_t startLevel,
+std::shared_ptr<Layer> perlinNoise(size2_t dims, float persistence, size_t startLevel,
                                    size_t endLevel, Rand& randomNumberGenerator = Rand()) {
-    std::shared_ptr<Image> img = std::make_shared<Image>(dims, DataFloat32::get());
-    auto ram = static_cast<LayerRAMPrecision<float>*>(
-        img->getColorLayer()->getEditableRepresentation<LayerRAM>());
-
-    auto size = nextPow2(std::max(dims.x, dims.y));
-    std::vector<std::shared_ptr<Image>> levels;
+    const auto size = nextPow2(std::max(dims.x, dims.y));
```
petersteneteg: use std::bit_ceil from <bit>

modules/base/include/modules/base/algorithm/randomutils.h
```diff
@@ -248,53 +293,53 @@ template <typename T, typename Rand = std::mt19937,
...
         currentPersistance *= persistence;
     }
 
-    auto data = ram->getDataTyped();
+    auto layerRam = std::make_shared<LayerRAMPrecision<float>>(LayerReprConfig{
+        .dimensions = dims,
+        .swizzleMask = swizzlemasks::defaultData(1),
+    });
+
+    auto data = layerRam->getDataTyped();
```
petersteneteg: getView()

modules/base/include/modules/base/algorithm/randomutils.h
```diff
@@ -309,25 +354,22 @@ std::shared_ptr<Image> perlinNoise(size2_t dims, float persistence, size_t start
             data[index(x, dims.y - 1 - y)] = glm::clamp(v, 0.0f, 1.0f);
         }
     }
-
-    return img;
+    return detail::createFromRAMRepresentation<Layer>(layerRam);
```
petersteneteg: use the Layer constructor. Maybe Layer(std::shared_ptr<LayerRepresentation>) should be
Layer(std::shared_ptr<LayerRepresentation>, LayerConfig = {})?
question is what to do if if the config also has values in is in the LayerReprConfig?

petersteneteg: But I guess that would only set the "default" values in the Layer then. so it should be fine.. but maybe unexpected?

modules/base/include/modules/base/algorithm/randomutils.h
```diff
@@ -376,6 +413,12 @@ std::shared_ptr<Image> poissonDisk(size2_t dims, size_t poissonDotsAlongX, size_
 
     int someNumber = 30;
 
+    auto layerRam = std::make_shared<LayerRAMPrecision<float>>(LayerReprConfig{
+        .dimensions = dims,
+        .swizzleMask = swizzlemasks::defaultData(1),
+    });
+    auto layerData = layerRam->getDataTyped();
```
petersteneteg: getView

modules/base/include/modules/base/algorithm/volume/volumegeneration.h
```diff
@@ -80,9 +80,9 @@ std::unique_ptr<Volume> generateVolume(const size3_t& dimensions, const mat3& ba
     auto volume = std::make_unique<Volume>(ram);
```
petersteneteg, line 80: create Volume(std::shared_ptr<VolumeRepresentation>, VolumeConfig = {})? ;

modules/base/src/algorithm/volume/volumegradient.cpp
```diff
@@ -66,8 +66,8 @@ std::shared_ptr<Volume> gradientVolume(std::shared_ptr<const Volume> volume, int
     auto newVolume = std::make_unique<Volume>(*volume, noData);
     auto newVolumeRep = std::make_shared<VolumeRAMPrecision<vec3>>(volume->getDimensions());
     newVolume->addRepresentation(newVolumeRep);
-    newVolume->dataMap_.valueAxis.name = "gradient";
-    newVolume->dataMap_.valueAxis.unit = volume->dataMap_.valueAxis.unit / volume->axes[0].unit;
+    newVolume->dataMap.valueAxis.name = "gradient";
```
petersteneteg: Could use the VolumeConfig in many of these cases if you want

modules/basegl/src/processors/layerprocessing/layergradient.cpp
petersteneteg: to calculate the resulting dataRange of the gradient

petersteneteg: This will be an upper bound

modules/vectorfieldvisualizationgl/src/processors/datageneration/vectorfieldgenerator2d.cpp
```diff
@@ -58,51 +47,90 @@ const ProcessorInfo VectorFieldGenerator2D::processorInfo_{
...
 }
 
 void VectorFieldGenerator2D::process() {
+    bool reattach = false;
+    const LayerConfig newConfig{.dimensions = size_.get(),
+                                .format = DataVec2Float32::get(),
+                                .swizzleMask = swizzlemasks::defaultData(0)};
+    if (!layer_ || config != newConfig) {
+        config = newConfig;
+        layer_ = std::make_shared<Layer>(config);
```
petersteneteg, line 102: this could use the same cache technique as in LayerGLProcessor

modules/vectorfieldvisualizationgl/tests/regression/lic/LIC.png
petersteneteg, line 1: seems the old one looked better? I get a lot of artifacts in the new one?

modules/vectorfieldvisualizationgl/tests/regression/lic_rk4/LIC.png
petersteneteg, line 1: same here?

petersteneteg (changes requested):

modules/base/src/processors/imagedistancetransform.cpp
martinfalk: Negative distances are not supported by our implementation since it is purely based on Euclidean distances.

modules/basegl/glsl/img_binary.frag
martinfalk: True. Note that this PR changes also the data format of the result to single channel and respective swizzlemask `swizzlemasks::luminance` (red, red, red, one). Hence, the alpha channel in the shader is just for show.

modules/basegl/glsl/img_normalize.frag
martinfalk: Also, we cannot use the regular mechanism in `ImageParameter` and `getNormalizedTexel()` since float values will also be normalized which we don't want here.

modules/basegl/src/processors/layerprocessing/layermapping.cpp
martinfalk: Renamed `LayerMapping` processor and files to `LayerColorMapping`.

modules/vectorfieldvisualizationgl/glsl/vector2dcurl.frag
martinfalk: moved `partialDiff()` to `utils/gradients.glsl`

modules/basegl/src/processors/imageprocessing/imagebinary.cpp
ResearchDaniel: The output is binary 0/1, but we do not support images with bool data type. The best we could do is then 8-bit uint.
It does not make much sense to force the data to be represented in the same data type as the input.

modules/basegl/src/processors/imageprocessing/imagegradient.cpp
martinfalk: Should be the data range for the calculation.

modules/vectorfieldvisualizationgl/src/processors/datageneration/vectorfieldgenerator2d.cpp
martinfalk: future PR

modules/vectorfieldvisualizationgl/tests/regression/lic/LIC.png
martinfalk: Caused by oversampling the vector field. Adjusted the dimensions of the vector field.

modules/vectorfieldvisualization/src/processors/2d/seedpointgenerator2d.cpp
```diff
@@ -123,22 +123,23 @@ SeedPointGenerator2D::SeedPointGenerator2D()
...
             case Generator::HaltonSequence: {
-                auto x = util::haltonSequence<float>(haltonXBase_.get(), numPoints_);
-                auto y = util::haltonSequence<float>(haltonYBase_.get(), numPoints_);
-                for (auto&& it : util::zip(x, y)) {
-                    seeds->emplace_back(get<0>(it), get<1>(it), 0.0f);
-                }
+                std::ranges::generate_n(seeds->begin(), numPoints_,
+                                        [i = 1, baseX = haltonXBase_.get(),
+                                         baseY = haltonYBase_.get()]() mutable -> vec3 {
+                                            return vec3{util::haltonSequence<float>(i++, baseX),
```
petersteneteg: Don't think you want to update i twice?

petersteneteg: Also the other of the update of I here is implementation defined.. might be better to update I first...

modules/vectorfieldvisualization/src/processors/2d/seedpointgenerator2d.cpp
```diff
@@ -123,22 +123,23 @@ SeedPointGenerator2D::SeedPointGenerator2D()
...
+                                                        util::haltonSequence<float>(i++, baseY),
+                                                        0.0f};
+                                        });
```
petersteneteg:
```suggestion
                std::ranges::copy(
                    std::views::iota(1, seeds->size()) 
                        | std::views::transform([baseX = haltonXBase_.get(), baseY = haltonYBase_.get()](int i){
                            return vec3{util::haltonSequence<float>(i, baseX),
                                                 util::haltonSequence<float>(i, baseY),
                                                 0.0f};
                             }), seeds->begin());
```

Might be an alternative that avoids the mutable state... 
to bad that generate does not call the functor with an index :/

petersteneteg (changes requested):

modules/basegl/src/processors/layerprocessing/layerbinary.cpp
ResearchDaniel: Why not? You would not use the input format when computing distances for example. I think that the same reasoning would apply here.

modules/basegl/src/processors/imageprocessing/imagegradient.cpp
ResearchDaniel: I don't see why, could you elaborate? The dataRange has no physical meaning.

modules/basegl/tests/regression/layer_operations/ImageBinaryCanvas.png
ResearchDaniel, line 1: Border is colored now, black before. Should be black and white=

modules/basegl/tests/regression/layer_operations/LayerBinaryCanvas.png
ResearchDaniel, line 1: Border also seems off here (before too though)

modules/vectorfieldvisualizationgl/glsl/vectorfieldgenerator2d.frag
```diff
@@ -32,23 +32,17 @@ in vec3 texCoord_;
...
+	return vec2(pos.x * (xRange.y - xRange.x) + xRange.x,
+				pos.y * (yRange.y - yRange.x) + yRange.x);
 }
 
 void main() {
-    vec4 pos = getPos();
-    vec4 value = vec4(0,0,0,1);
-    value.x = X_VALUE(pos.x,pos.y);
-    value.y = Y_VALUE(pos.x,pos.y);
+	vec2 worldPos = getWorldPosition(texCoord_.xy);
```
ResearchDaniel: Incorrect indentation

modules/basegl/tests/regression/layer_operations/ImageBinaryCanvas.png
petersteneteg: What do you mean? there is only white at the border. The red/green color is from the GitHub GUI

petersteneteg (dismissed):

petersteneteg (dismissed):

petersteneteg (approved):

PR #1509, opened 2024-03-15
Vcpkg features

petersteneteg:
Vcpkg: added dicom feature for the modules dicom/gdcm module. Also made ffmpeg hdf5 info features that are on by default. turn off by setting VCPKG_MANIFEST_NO_DEFAULT_FEATURES to ON

### DIFF
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -23,13 +23,6 @@
     "glfw3",
     "glm",
     "gtest",
-    {
-      "name": "hdf5",
-      "features": [
-        "cpp",
-        "zlib"
-      ]
-    },
     "libjpeg-turbo",
     "libpng",
     "minizip",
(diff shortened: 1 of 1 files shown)


vcpkg.json
```diff
@@ -84,9 +93,13 @@
         }
       ]
     },
+    "dicom" : {
+      "description": "Grassroots DICOM library",
```
ResearchDaniel, line 97:
```suggestion
      "description": "Support DICOM (.dcm) files, see https://github.com/inviwo/modules/tree/master/medvis/dicom",
```

vcpkg.json
```diff
@@ -47,11 +40,27 @@
     "zlib",
     "pegtl",
     "inja",
-    "ffmpeg",
     "warn",
     "zlib"
   ],
+  "default-features": [
+    "hdf5",
```
ResearchDaniel, line 47: The HDF5 module is not on by default, so it should not be added here?

ResearchDaniel (dismissed):

martinfalk (approved):

vcpkg.json
petersteneteg: It used to be on, I will keep it on

vcpkg.json
petersteneteg: I want to keep this short.

PR #1520, opened 2024-04-15
Feature/lightvolumeraycaster

martinfalk:
Changes proposed in this PR:
 * added a raycaster component and a template raycasting processor for light volumes
 * move illumination from `RaycastingComponent` to `LightComponent` to improve composition


modules/opengl/glsl/utils/structs.glsl
```diff
@@ -120,6 +120,19 @@ struct LightParameters {
     float specularExponent;
 };
 
+struct Material {
+    vec3 ambient;
+    vec3 diffuse;
+    vec3 specular;
```
petersteneteg, line 126: that about the exponent?

petersteneteg (dismissed):

modules/opengl/glsl/utils/structs.glsl
martinfalk: Can add that one. Didn't want to have a duplicate of the `LightParameters` at the moment.

petersteneteg: Could also do that later I guess...

petersteneteg: No point adding it if we aren't setting/using it

ResearchDaniel: Perhaps this is MaterialColors and then there are other structs for each material model. Exponent is not used in all material models

petersteneteg (dismissed):

modules/basegl/include/modules/basegl/processors/lightvolumeraycaster.h
ResearchDaniel, line 1: Suggestion: Call this something like this to make it clearer that it volume raycasting with global lighting
VolumeLightingRaycaster
VolumeLightRaycaster
VolumeRaycasterLighting

modules/basegl/include/modules/basegl/processors/lightvolumeraycaster.h
```diff
@@ -0,0 +1,67 @@
...
+#include <modules/basegl/shadercomponents/isotfcomponent.h>
+#include <modules/basegl/shadercomponents/positionindicatorcomponent.h>
+#include <modules/basegl/shadercomponents/raycastingcomponent.h>
+#include <modules/basegl/shadercomponents/sampletransformcomponent.h>
+#include <modules/basegl/shadercomponents/volumecomponent.h>
+#include <modules/basegl/shadercomponents/lightvolumecomponent.h>
+
+namespace inviwo {
+
+class IVW_MODULE_BASEGL_API LightVolumeRaycaster : public VolumeRaycasterBase {
```
ResearchDaniel, line 47: I think that it would be nicer if the illumination was optionally included in the regular raycaster without additional inports

modules/basegl/include/modules/basegl/shadercomponents/lightvolumecomponent.h
```diff
@@ -0,0 +1,95 @@
...
+    virtual std::vector<std::tuple<Inport*, std::string>> getInports() override;
+
+    virtual std::vector<Property*> getProperties() override;
+
+    virtual std::vector<Segment> getSegments() override;
+
+    bool setUsedChannels(size_t channels);
+
+private:
+    VolumeInport lightVolume_;
```
ResearchDaniel, line 79: I think that it would be nicer if the computation of the light volume would be integrated in the component. Now it just exposes the light volume rather “being” one

modules/basegl/glsl/meshrendering.frag
```diff
@@ -53,8 +53,9 @@ void main() {
     vec4 fragColor = color_;
     vec3 toCameraDir_ = camera.position - worldPosition_.xyz;
 
-    fragColor.rgb = APPLY_LIGHTING(lighting, color_.rgb, color_.rgb, vec3(1.0f), worldPosition_.xyz,
-                                   normalize(normal_), normalize(toCameraDir_));
+    ShadingParameters shadingParams = shading(color_.rgb, normal_, worldPosition_.xyz);
```
ResearchDaniel: No need to normalize the normal here? It was done before

modules/basegl/src/shadercomponents/lightvolumecomponent.cpp
```diff
@@ -0,0 +1,189 @@
...
+#include <fmt/format.h>
+
+namespace inviwo {
+
+LightVolumeComponent::LightVolumeComponent(Processor& processor, std::string_view volumeName,
+                                           std::string_view gradientName)
+    : ShaderComponent{}
+    , lightVolume_{"lightVolume"}
+    , lightSource_{"lightSource"}
+    , scaling_{"lightVolumeScaling", "Light Volume Scaling", util::ordinalScale(1.0f, 2.0f)}
```
ResearchDaniel: Light Volume Scaling could be confused with scaling the size of the light volume
```suggestion
    , scaling_{"lightIntensityScaling", "Light Intensity Scaling", util::ordinalScale(1.0f, 2.0f)}
```

modules/basegl/src/shadercomponents/lightvolumecomponent.cpp
```diff
@@ -0,0 +1,189 @@
...
+    lightVolume_.onDisconnect([&]() { processor.invalidate(InvalidationLevel::InvalidResources); });
+}
+
+std::string_view LightVolumeComponent::getName() const { return lightVolume_.getIdentifier(); }
+
+void LightVolumeComponent::process(Shader& shader, TextureUnitContainer& cont) {
+    using namespace fmt::literals;
+
+    if (auto type = lightSource_.getData()->getLightSourceType(); type != LightSourceType::Point) {
+        throw Exception(IVW_CONTEXT, "unsupported light source '{}', expected '{}'", type,
```
ResearchDaniel, line 89: Don't we support directional light?

modules/basegl/src/shadercomponents/raycastingcomponent.cpp
```diff
@@ -127,21 +127,27 @@ result = drawISO(result, {isoparams}, {volume}Voxel[{channel}], {volume}VoxelPre
...
-    #if defined(SHADING_ENABLED) && defined(GRADIENTS_ENABLED)
-    vec3 worldSpacePosition = ({volume}Parameters.textureToWorld * vec4(samplePosition, 1.0)).xyz;
-    color{color}.rgb = APPLY_LIGHTING(lighting, color{color}.rgb, color{color}.rgb, vec3(1.0),
-                                      worldSpacePosition, -{gradient}, cameraDir);
+    shadingParams.colors = defaultMaterialColors(color{color}.rgb);
+    #if (defined(SHADING_ENABLED) && defined(GRADIENTS_ENABLED))
+    shadingParams.normal = -{gradient};
+    shadingParams.worldPosition = ({volume}Parameters.textureToWorld * vec4(samplePosition, 1.0)).xyz;
     #endif
+    color{color}.rgb = APPLY_LIGHTING_FUNC(lighting, shadingParams, cameraDir);
```
ResearchDaniel, line 149: Should this not be within the `#if (defined(SHADING_ENABLED)`?

modules/basegl/src/shadercomponents/raycastingcomponent.cpp
```diff
@@ -157,7 +163,9 @@ auto RaycastingComponent::getSegments() -> std::vector<Segment> {
     std::vector<Segment> segments{
         {std::string(R"(#include "raycasting/iso.glsl")"), placeholder::include, 1100},
         {std::string(R"(#include "utils/compositing.glsl")"), placeholder::include, 1100},
-        {std::string(R"(uniform int channel = 0;)"), placeholder::uniform, 1100}};
+        {std::string(R"(uniform int channel = 0;)"), placeholder::uniform, 1100},
+        {std::string{init}, placeholder::first, 600},
```
ResearchDaniel, line 167: Place it first among the segments as well?

modules/basegl/src/shadercomponents/raycastingcomponent.cpp
```diff
@@ -259,6 +270,7 @@ auto MultiRaycastingComponent::getSegments() -> std::vector<Segment> {
     std::vector<Segment> segments{
         {std::string(R"(#include "raycasting/iso.glsl")"), placeholder::include, 1100},
         {std::string(R"(#include "utils/compositing.glsl")"), placeholder::include, 1100},
+        {std::string{init}, placeholder::first, 600},
```
ResearchDaniel, line 273: Place first

modules/basegl/src/shadercomponents/volumecomponent.cpp
```diff
@@ -77,9 +77,10 @@ constexpr std::string_view voxel = util::trim(R"(
 )");
 
 constexpr std::string_view gradientFirst = util::trim(R"(
-#if defined(GRADIENTS_ENABLED)
 vec3 {0}GradientPrev = vec3(0);
-vec3 {0}Gradient = useSurfaceNormals ? -texture(surfaceNormal, texCoords).xyz :
+vec3 {0}Gradient = vec3(0);
```
ResearchDaniel, line 81: Why should they be declared if gradients are disabled?

modules/basegl/src/shadercomponents/volumecomponent.cpp
```diff
@@ -98,10 +99,11 @@ constexpr std::string_view gradient = util::trim(R"(
 )");
 
 constexpr std::string_view allGradientsFirst = util::trim(R"(
-#if defined(GRADIENTS_ENABLED)
 mat4x3 {0}AllGradientsPrev = mat4x3(0);
+mat4x3 {0}AllGradients = mat4x3(0);
```
ResearchDaniel, line 103: Why declared?

modules/opengl/glsl/utils/shading.glsl
```diff
@@ -29,8 +29,44 @@
...
+#if !defined(APPLY_LIGHTING)
+fall-back to no shading in case APPLY_LIGHTING is not defined
+#  define APPLY_LIGHTING(lighting, materialAmbientColor, materialDiffuseColor, \
+    materialSpecularColor, position, normal, toCameraDir) \
+    materialAmbientColor
+#endif
+
 #include "utils/structs.glsl" //! #include "./structs.glsl"
 
+// default material based only on a diffuse color as used for volume rendering
```
ResearchDaniel:
```suggestion
// default material uses the supplied diffuse color also for the ambient term and white color for the specular term (as used for volume rendering).
```

ResearchDaniel (changes requested):
Some minor issues and some thoughts about the added feature :)

modules/basegl/glsl/meshrendering.frag
martinfalk: good catch.

modules/basegl/src/shadercomponents/lightvolumecomponent.cpp
martinfalk: In the light volume yes, but not in any other part relying on the SimpleLightingProperty (mesh rendering, volume rendering, etc.).

modules/basegl/src/shadercomponents/raycastingcomponent.cpp
martinfalk: No, since this would prevent this function from being called entirely. In case of the light volume calculations, these are encapsulated in `APPLY_LIGHTING_FUNC`.

modules/basegl/src/shadercomponents/raycastingcomponent.cpp
martinfalk: The order of the segments is `include`, `uniform`, `first`, ... The numbers given indicate the priority within each segment block.

modules/basegl/src/shadercomponents/volumecomponent.cpp
martinfalk: Since these variables are potentially used for lighting calculation, even though no gradients are present. Otherwise you end up with a non-compiling shader instead of just diffuse color without illumination.

ResearchDaniel (dismissed):

modules/basegl/include/modules/basegl/processors/lightvolumeraycaster.h
petersteneteg: This is a different kind of ray caster with very different cost / results. should be it's own processor.

modules/basegl/include/modules/basegl/shadercomponents/lightvolumecomponent.h
petersteneteg: would one ever want to reuse the light volume?

modules/basegl/src/shadercomponents/raycastingcomponent.cpp
petersteneteg: "first" represents the first iteration, not at the top of the template.

petersteneteg (dismissed):

petersteneteg (approved):

PR #1521, opened 2024-04-15
Core,Qt: commandline arg to reset Qt window state

martinfalk:
Fixes 
* ignores the previously stored window state and geometry of the Inviwo main window on startup. This should fix broken toolbars.


ResearchDaniel (dismissed):
Is there a way to automatically detect that it is broken and either do it or advice the user to do it?

martinfalk:
> Is there a way to automatically detect that it is broken and either do it or advice the user to do it?

Other than looking at a totally cramped toolbar, don't think so. Once Inviwo is running, it would be quite tricky to purge the broken state as all UI elements have a potentially broken state that is synced automatically.

petersteneteg (changes requested):
Move to inviwomainwindow

petersteneteg (approved):

petersteneteg:
> Is there a way to automatically detect that it is broken and either do it or advice the user to do it?

We have no way of knowing that we know of...

PR #1528, opened 2024-04-25
Core/Various: Create processors as shared_ptrs

petersteneteg:
The processor factory will now create processors as shared_ptr so begin with. 
Before we only created the shared_ptr later in the processes. That menat that if you tried to use shared_from_this in the constructor that would fail and throw an exception. This will now work.

### DIFF
--- a/include/inviwo/core/processors/processorfactoryobject.h
+++ b/include/inviwo/core/processors/processorfactoryobject.h
@@ -51,7 +51,7 @@ public:
         : info_(std::move(info)), typeName_{typeName} {}
     virtual ~ProcessorFactoryObject() = default;
 
-    virtual std::unique_ptr<Processor> create(InviwoApplication* app) = 0;
+    virtual std::shared_ptr<Processor> create(InviwoApplication* app) const = 0;
 
     const ProcessorInfo& getProcessorInfo() const { return info_; }
     const std::string& getClassIdentifier() const { return info_.classIdentifier; }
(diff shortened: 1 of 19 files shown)


include/inviwo/core/processors/processorfactoryobject.h
```diff
@@ -51,7 +51,7 @@ class IVW_CORE_API ProcessorFactoryObject {
         : info_(std::move(info)), typeName_{typeName} {}
     virtual ~ProcessorFactoryObject() = default;
 
-    virtual std::unique_ptr<Processor> create(InviwoApplication* app) = 0;
+    virtual std::shared_ptr<Processor> create(InviwoApplication* app) const = 0;
```
ResearchDaniel, line 54: In the other parts you added a new function called createShared. 
do the same here to be consistent?

src/core/processors/processorfactory.cpp
```diff
@@ -74,10 +74,26 @@ bool ProcessorFactory::registerObject(ProcessorFactoryObject* processor) {
     return true;
 }
 
-std::unique_ptr<Processor> ProcessorFactory::create(std::string_view key) const {
-    return Parent::create(key, app_);
+std::unique_ptr<Processor> ProcessorFactory::create(std::string_view, InviwoApplication*) const {
```
ResearchDaniel, line 77: Is this just for debugging?
Function could be removed?

petersteneteg: We need to implement this since it is part of the factory interface

include/inviwo/core/processors/processorfactoryobject.h
petersteneteg: in the other case we have both create and createShared, for unique_ptr and shared_ptr, here is only one. so create is enough, and its what we usually use in the factory objects.

martinfalk (approved):

PR #1536, opened 2024-05-08
Feature/meshcreator fix

martinfalk:
Changes proposed in this PR:
 * `MeshCreator` sphere created with radius 1 and scaling applied as model matrix. This fixes issues with scaled texture coords.
 * refactored glyph ray casting using `applyLighting()`


petersteneteg (approved):
Why the the shading in the regression change so much? why did you have to change the regression workspaces?

martinfalk:
> Why the the shading in the regression change so much? why did you have to change the regression workspaces?

Previously, the sphere's radius was 0.5 times a scaling factor. Colors and texture coords used the position direction and were thus quickly saturated. Now the radius set to 1 and the scaling is applied using the model matrix leaving colors and tex coords within [0,1].
The test updates were necessary since the spheres had to be shrunk to 50% of their size.

PR #1537, opened 2024-05-13
Core,OpenGL: two-sided illumination

martinfalk:
The current shading only considers surfaces facing toward the camera. This leads to unwanted results for thin meshes like stream ribbons or isosurfaces where the "front" side is shaded properly while the backside is only shaded with the ambient term.

Changes proposed in this PR:
* two-sided illumination is now default
* added options for front- and backside only

image

---


include/inviwo/core/datastructures/light/lightingstate.h
```diff
@@ -43,6 +43,10 @@ enum class ShadingMode : int {
     Specular,
     BlinnPhong,
     Phong,
+    BlinnPhongFront,
```
ResearchDaniel, line 46: Would it not make more sense to have a separate option with shading of front/back/both sides. This explodes a bit now and it is not entirely clear that for example Phong is both sides (unless one sees the other options)

petersteneteg (dismissed):

modules/opengl/src/shader/shaderutils.cpp
```diff
@@ -144,11 +144,24 @@ void addShaderDefines(Shader& shader, const ShadingMode& mode) {
                 return "shadeSpecular(lighting, materialSpecularColor, position, normal, "
```
ResearchDaniel, line 144: If you would have the normal manipulation outside of this code snippet you could also have front/back diffuse/specular shading.

modules/opengl/glsl/utils/shading.glsl
```diff
@@ -140,6 +140,24 @@ vec3 shadePhong(LightParameters light_, vec3 materialAmbientColor, vec3 material
     return resAmb + resDiff + resSpec;
 }
 
+vec3 shadeBlinnPhongTwoSided(LightParameters light_, vec3 materialAmbientColor, vec3 materialDiffuseColor,
+                     vec3 materialSpecularColor, vec3 position, vec3 normal, vec3 toCameraDir) {
+    if (dot(toCameraDir, normal) < 0) {
```
ResearchDaniel: normal = faceforward(normal, toCameraDir, normal)

ResearchDaniel (dismissed):
I would argue for incorporating a new enum for front/back/both-sided shading. The manipulation of the normal could be moved outside of the shadeXYZ to also accommodate front/back-face shading for diffuse/specular shading.

include/inviwo/core/datastructures/light/lightingstate.h
petersteneteg: since it is in the same dropdown in the gui this way it is quite hard to miss. I see not need of a separate option for this.

modules/opengl/src/shader/shaderutils.cpp
petersteneteg: if there ever is a demand for selecting two/front/back sided diffuse only shading we could add that then.
No point adding settings that will never be used anyway.

martinfalk: I would rather remove 'diffuse' and 'specular' only (maybe even 'ambient'). Reason: you can always get the same result by setting the other properties to black / zero.

petersteneteg (dismissed):

petersteneteg (dismissed):

petersteneteg (approved):

PR #1543, opened 2024-05-23
Feature/qtwidgets filepathfix

martinfalk:
Fixes
* fixed some corner cases in the `FilePathLineEditQt` widget related to the status icon
* better handling of std::filesystem exceptions


modules/qtwidgets/src/filepathlineeditqt.cpp
```diff
@@ -90,6 +90,9 @@ FilePathLineEditQt::FilePathLineEditQt(QWidget* parent)
     connect(this, &QLineEdit::editingFinished, [this]() {
         setCursorToEnd();
         const auto path = utilqt::toPath(text().trimmed());
+
+        // FIXME: handle case where new path is not valid. Should the internal path be updated? Does
+        // the line edit need updating?
```
petersteneteg: I think the actually error handing should happen in the property not in the widget. So I don't think we should do anything here

petersteneteg (dismissed):

petersteneteg (approved):

PR #1561, opened 2024-07-01
Feature/minmax property

martinfalk:
Fixes
* clamp values _before_ updating the `OrdinalMinMaxPropertyWidgetQt`
* consistent state for `RangeSliderQt`
* MinMaxRangeProperty slider snaps to int positions

### DIFF
--- a/modules/qtwidgets/src/rangesliderqt.cpp
+++ b/modules/qtwidgets/src/rangesliderqt.cpp
@@ -283,7 +283,7 @@ void RangeSliderQt::moveMiddle(int delta) {
 
 bool RangeSliderQt::eventFilter(QObject* obj, QEvent* event) {
     if (event->type() == QEvent::ToolTip) {
-        const QHelpEvent* helpEvent = static_cast<QHelpEvent*>(event);
+        const auto* helpEvent = dynamic_cast<QHelpEvent*>(event);
         if (obj == handle(1)) {
             QToolTip::showText(helpEvent->globalPos(),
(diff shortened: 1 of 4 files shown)


modules/qtwidgets/src/rangesliderqt.cpp
```diff
@@ -260,14 +274,16 @@ void RangeSliderQt::moveMiddle(int delta) {
...
+    emit valuesChanged(value_.x, value_.y);
+
     parentWidget()->repaint();
     repaint();
 }
 
 bool RangeSliderQt::eventFilter(QObject* obj, QEvent* event) {
     if (event->type() == QEvent::ToolTip) {
-        QHelpEvent* helpEvent = static_cast<QHelpEvent*>(event);
+        const auto* helpEvent = dynamic_cast<QHelpEvent*>(event);
```
petersteneteg: I know that ClangTidy thinks one should use dynamic casts here. But Im not sure we should follow. We to check the event type(), so we should know the object type and static casting should be fine. And since this might trigger a lot... dynamic cast might be a bit more expensive? Not that I have tested that in the last 5+ years...

petersteneteg (dismissed):

modules/qtwidgets/src/rangesliderqt.cpp
martinfalk: I'll revert the changes.

petersteneteg (approved):
Nice!... that is really tricky code..

Might add some nolint about the dynamic_casts here...

PR #1562, opened 2024-07-01
Core,Qt: prevent processor animation on paste and replace

### LINKED ISSUE #1560: [BUG]: animation for newly added processors
### Required prerequisites

### Problem description

Processors are added first to the network, then moved into position. For some reason this involves an animation for the transition.
For example, when adding an `Image Canvas` via the "add visualizer" menu, the canvas processor is added somewhere and then moved to its final position below the invocation.

https://github.com/inviwo/inviwo/assets/9251300/0cd30027-8498-4360-8ee5-427a2083e9dc

### Inviwo version

7df6155da530f94e2ad813bdf3efa97c84ea85ab

### Operating system?

Windows

### Compiler / Version

_No response_

### Qt Version

_No response_

### Reproducible example code

_No response_

### Inviwo log

_No response_

martinfalk:
Fixes #1560 

Changes proposed in this PR:
* apply offsets to processors before adding them to the network to prevent animations added in 8b886ed
* initially position visualizer networks at the outport they were invoked from instead of (0, 0)

### DIFF
--- a/include/inviwo/core/processors/processorutils.h
+++ b/include/inviwo/core/processors/processorutils.h
@@ -87,7 +87,7 @@ struct IVW_CORE_API GridPos {
     explicit GridPos(ivec2 pos) : pos_{pos} {}
 
     ivec2 operator+(const ivec2& rhs) const { return operator ivec2() + rhs; }
-    explicit operator ivec2() const { return pos_ * ivec2{25, 25}; }
+    operator ivec2() const { return pos_ * ivec2{25, 25}; }
 
 private:
     ivec2 pos_;
--- a/include/inviwo/core/processors/processorutils.h
+++ b/include/inviwo/core/processors/processorutils.h
@@ -86,6 +86,7 @@ struct IVW_CORE_API GridPos {
     GridPos(int x, int y) : pos_{x, y} {};
     explicit GridPos(ivec2 pos) : pos_{pos} {}
 
+    ivec2 operator+(const ivec2& rhs) { return ivec2(*this) + rhs; }
     operator ivec2() const { return pos_ * ivec2{25, 25}; }
 
 private:
--- a/include/inviwo/core/processors/processorutils.h
+++ b/include/inviwo/core/processors/processorutils.h
@@ -86,8 +86,8 @@ struct IVW_CORE_API GridPos {
     GridPos(int x, int y) : pos_{x, y} {};
     explicit GridPos(ivec2 pos) : pos_{pos} {}
 
-    ivec2 operator+(const ivec2& rhs) { return ivec2(*this) + rhs; }
-    operator ivec2() const { return pos_ * ivec2{25, 25}; }
+    ivec2 operator+(const ivec2& rhs) const { return operator ivec2() + rhs; }
+    explicit operator ivec2() const { return pos_ * ivec2{25, 25}; }
 
 private:
     ivec2 pos_;
--- a/include/inviwo/core/rendering/datavisualizer.h
+++ b/include/inviwo/core/rendering/datavisualizer.h
@@ -98,13 +98,13 @@ public:
      * @param filename The file to load in the source processor. The extension of the filename
      *        should be in the list of extensions returned by getSupportedFileExtensions.
      * @param network The network to add the processor to.
-     * @param initialPos the initial position for the processor
+     * @param origin the initial position for the processor
      * @return The added source processor and the outport in the source processor
      *         containing data from the given file.
      */
     virtual std::pair<Processor*, Outport*> addSourceProcessor(
         const std::filesystem::path& filename, ProcessorNetwork* network,
-        const ivec2& initialPos) const = 0;
+        const ivec2& origin) const = 0;
     /**
      * Adds a set of processors visualizing the data in the given outport.
      * Nothing will be added to the network if outport is not supported (isOutportSupported returns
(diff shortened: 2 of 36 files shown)


include/inviwo/core/processors/processorutils.h
```diff
@@ -86,7 +86,8 @@ struct IVW_CORE_API GridPos {
     GridPos(int x, int y) : pos_{x, y} {};
     explicit GridPos(ivec2 pos) : pos_{pos} {}
 
-    operator ivec2() const { return pos_ * ivec2{25, 25}; }
```
petersteneteg, line 89: don't we want this to be implicit?

modules/base/src/datavisualizer/imagetolayervisualizer.cpp
```diff
@@ -65,28 +65,23 @@ bool ImageToLayerVisualizer::hasSourceProcessor() const { return false; }
...
-    const std::filesystem::path& filename, ProcessorNetwork* net) const {
-
-    auto sourceAndOutport = addSourceProcessor(filename, net);
-    auto processors = addVisualizerNetwork(sourceAndOutport.second, net);
-
-    processors.push_back(sourceAndOutport.first);
-
-    return processors;
+    const std::filesystem::path&, ProcessorNetwork*, const ivec2&) const {
+    return {nullptr};
```
petersteneteg: why nullptr? should just return an empty vector

include/inviwo/core/rendering/datavisualizer.h
```diff
@@ -98,11 +98,13 @@ class IVW_CORE_API DataVisualizer {
      * @param filename The file to load in the source processor. The extension of the filename
      *        should be in the list of extensions returned by getSupportedFileExtensions.
      * @param network The network to add the processor to.
+     * @param initialPos the initial position for the processor
```
petersteneteg: Why "initial"pos? I would "pos" of "offset" not be better?

include/inviwo/core/rendering/datavisualizer.h
```diff
@@ -122,10 +124,12 @@ class IVW_CORE_API DataVisualizer {
      * @param filename The file to load in the source processor. The extension of the filename
      *        should be in the list of extensions returned by getSupportedFileExtensions.
      * @param network The network to add the processor to.
+     * @param initialPos the initial position for the visualizer network
```
petersteneteg: same here, I would just "pos" or "origin"

petersteneteg: initial implies that you will change it later?

include/inviwo/core/rendering/datavisualizer.h
petersteneteg: did you consider using a setOrigin method on the visualizer?

modules/base/src/datavisualizer/imageinformationvisualizer.cpp
```diff
@@ -76,26 +76,28 @@ bool ImageInformationVisualizer::hasSourceProcessor() const { return true; }
 bool ImageInformationVisualizer::hasVisualizerNetwork() const { return true; }
 
 std::pair<Processor*, Outport*> ImageInformationVisualizer::addSourceProcessor(
-    const std::filesystem::path& filename, ProcessorNetwork* net) const {
+    const std::filesystem::path& filename, ProcessorNetwork* net, const ivec2& initialPos) const {
 
-    auto source = net->addProcessor(util::makeProcessor<ImageSource>(GP{0, 0}, app_, filename));
```
petersteneteg, line 81: This line will no longer compile since you now require explicit cast of GP...

petersteneteg (changes requested):

include/inviwo/core/rendering/datavisualizer.h
martinfalk: > initial implies that you will change it later?

This is the case. The initial pos puts the processors relative to the currently selected outport (for port visualizers) or the bounding box of the network (data/source visualizer).

https://github.com/inviwo/inviwo/blob/0148bf08fbbd8d64fe05bb64e089ba84445a53f7/src/qt/editor/dataopener.cpp#L123-L133

petersteneteg: yes, but there is nothing "initial" about that?

include/inviwo/core/processors/processorutils.h
martinfalk: I would have liked to, but the linter insisted.

image

modules/base/src/datavisualizer/imageinformationvisualizer.cpp
martinfalk: I am aware of this. Complain to the linter.

include/inviwo/core/processors/processorutils.h
petersteneteg: Yes, but you can override the linter when needed. It is just guidelines, not rules...

include/inviwo/core/rendering/datavisualizer.h
martinfalk: The processors are positioned at the initialPos and then moved after adding all of them by `util::offsetPosition()`. This results in an animation of the processors moving from the current position to the final one (instead of starting at (0,0)).

include/inviwo/core/processors/processorutils.h
petersteneteg: explicit(false) might work, have not tested that my self. or you have to add //NOLINT comments... https://clang.llvm.org/extra/clang-tidy/#suppressing-undesired-diagnostics

modules/base/src/datavisualizer/imageinformationvisualizer.cpp
petersteneteg: one can overrule the linter...

include/inviwo/core/rendering/datavisualizer.h
martinfalk: > did you consider using a setOrigin method on the visualizer?

Yes, I did. Though `addVisualizerNetwork()` and `addSourceProcessor()/addSourceAndVisualizerNetwork()` differentiate in the positioning. `addVisualizerNetwork()` adds stuff relative to the outport, i.e. the initialPos is computed on the fly in each visualizer.

See for example https://github.com/inviwo/inviwo/blob/0148bf08fbbd8d64fe05bb64e089ba84445a53f7/modules/dataframeqt/src/datavisualizer/dataframetablevisualizer.cpp#L87-L91

petersteneteg (approved):

PR #1564, opened 2024-07-03
Feature/packaging

### LINKED ISSUE #1563: [BUG]: packaging of external dependencies broken
### Required prerequisites

- [x] Consider asking first in [Slack](https://join.slack.com/t/inviwo/shared_invite/enQtNTc2Nzc2NDQwNzIxLTRiMWM1ZWJiYjljZjkwNWE3OTk3MzYxODZlMDUyMzRmZjUzMzBiZjVhNTM3NWUyNzU1MjI4OWJjMzdkODViMzM) or in a [Discussion](https://github.com/inviwo/inviwo/discussionss).

### Problem description

Packaging does not include external libraries while Python is not enabled (`IVW_ENABLE_PYTHON` set to `OFF).

### Inviwo version

7df6155

### Operating system?

Windows

### Compiler / Version

_No response_

### Qt Version

_No response_

### Reproducible example code

_No response_

### Inviwo log

_No response_

martinfalk:
Fixes #1563 

Changes proposed in this PR:
 * `ivw_vcpkg_install()` uses find_package(Python3), if Python3 was not found yet
 * more verbose error handling in `ivw_add_py_wrapper()`

### DIFF
--- a/cmake/vcpkghelpers.cmake
+++ b/cmake/vcpkghelpers.cmake
@@ -109,6 +109,7 @@ endfunction()
 # corresponding folders by globbing the vcpkg package folders. 
 # It will also try and install any transitive dependencies automatically 
 # We also automatically register the license file using the metadata found in vcpkg
+# Requires a Python3 interpreter.
 # Parameters:
 #  * OUT_COPYRIGHT retrieve the path the the COPYRIGHT file
 #  * OUT_VERSION get the package version from the vcpkg metadata
@@ -120,7 +121,15 @@ function(ivw_vcpkg_install name)
     endif()
 
     if(NOT Python3_Interpreter_FOUND)
-        return()
+        find_package(Python3 COMPONENTS Interpreter)
+        if(NOT Python3_FOUND OR 
+           NOT Python3_Interpreter_FOUND)
+            message(ERROR "Python3 not available.\n"
+                "Please install Python3. If you have a Python installation that is not found, consider "
+                "setting Python3_ROOT_DIR to the root directory of your Python 3 installation."
+            )
+            return ()
+        endif()
     endif()
 
     set(options EXT)
--- a/cmake/vcpkghelpers.cmake
+++ b/cmake/vcpkghelpers.cmake
@@ -104,6 +104,8 @@ function(ivw_private_vcpkg_install_helper)
     endforeach()
 endfunction()
 
+# Reset global variable used to show Python warning in ivw_vcpkg_install only once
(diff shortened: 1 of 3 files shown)


cmake/vcpkghelpers.cmake
```diff
@@ -120,7 +121,15 @@ function(ivw_vcpkg_install name)
     endif()
 
     if(NOT Python3_Interpreter_FOUND)
-        return()
+        find_package(Python3 COMPONENTS Interpreter)
```
petersteneteg: I we should do a find_package here, it should be before the if on line 122...

cmake/vcpkghelpers.cmake
```diff
@@ -120,7 +121,15 @@ function(ivw_vcpkg_install name)
...
+        if(NOT Python3_FOUND OR 
+           NOT Python3_Interpreter_FOUND)
+            message(ERROR "Python3 not available.\n"
```
petersteneteg: I don't want to make this a hard error.  The information gathered here is only needed for the about dialog/license system and the installation. Hence this should not be a hard error. Most people building inviwo won't need this...

petersteneteg (changes requested):

cmake/vcpkghelpers.cmake
martinfalk: Good point.

cmake/vcpkghelpers.cmake
```diff
@@ -120,16 +120,13 @@ function(ivw_vcpkg_install name)
...
-        find_package(Python3 COMPONENTS Interpreter)
-        if(NOT Python3_FOUND OR 
-           NOT Python3_Interpreter_FOUND)
-            message(ERROR "Python3 not available.\n"
-                "Please install Python3. If you have a Python installation that is not found, consider "
-                "setting Python3_ROOT_DIR to the root directory of your Python 3 installation."
-            )
-            return ()
-        endif()
+        message(WARNING "Python3 not available.\n"
```
petersteneteg: We might print 100+ of these warnings now in cmake, one for each dependency... 

one might move the test and find package outside of the function definition. Then it should only run once. But you don't get the correct call stack I suppose...

petersteneteg (dismissed):

cmake/vcpkghelpers.cmake
martinfalk: Yes, that could get annoying. One could use a global variable to only issue the warning once.

petersteneteg: mmmm

petersteneteg (approved):

PR #1566, opened 2024-07-10
Feature/webbrowser refactoring

### LINKED ISSUE #1565: [BUG]: CEF key event handling broken
### Required prerequisites

### Problem description

Key events are not correctly forwarded to the Chromium Embedded Framework. Thus, any keyboard input does not appear in the browser.
Broken since 9d8d48f.

### Inviwo version

_No response_

### Operating system?

_No response_

### Compiler / Version

_No response_

### Qt Version

_No response_

### Reproducible example code

_No response_

### Inviwo log

_No response_

martinfalk:
Fixes #1565

Changes proposed in this PR:
 * refactored `WebBrowserProcessor` into a separate `WebBrowserBase` component
 * added refined example workspace for property syncing.


modules/webbrowser/include/modules/webbrowser/interaction/cefinteractionhandler.h
```diff
@@ -67,14 +67,12 @@ class IVW_MODULE_WEBBROWSER_API CEFInteractionHandler : public InteractionHandle
...
+    explicit CEFInteractionHandler(CefRefPtr<CefBrowserHost> host = nullptr);
     virtual ~CEFInteractionHandler() = default;
 
-    virtual std::string getClassIdentifier() const override {
-        return "org.inviwo.cefinteractionhandler";
-    }
+    std::string getClassIdentifier() const override { return "org.inviwo.cefinteractionhandler"; }
 
-    virtual void invokeEvent(Event* event) override;
+    void invokeEvent(Event* event) override;
```
petersteneteg: In inviwo we have kept both virtual and override. Mostly for legacy and old habits. I suppose, But if we decide on changing that we should try and do it in a coordinated way.

petersteneteg: I.e. don't remove them now

modules/webbrowser/include/modules/webbrowser/processors/webbrowserbase.h
```diff
@@ -0,0 +1,116 @@
...
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ *********************************************************************************/
+
+#pragma once
+
+#include <modules/webbrowser/webbrowsermoduledefine.h>
```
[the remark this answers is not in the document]
petersteneteg: you need to enable the web browser module in the clang-tidy CI build I think

modules/webbrowser/include/modules/webbrowser/processors/webbrowserbase.h
```diff
@@ -0,0 +1,116 @@
...
+class IVW_MODULE_WEBBROWSER_API WebBrowserBase : public CefLoadHandler {
+public:
+    WebBrowserBase(InviwoApplication* app, Processor* processor);
+    ~WebBrowserBase();
+
+    void reload();
+    bool isLoading() const;
+
+    void setSource(const std::filesystem::path& filename);
+    void setSource(std::string_view uri);
```
petersteneteg: What about passing in source as a string?

petersteneteg: I would rather name theses one something like "open" or "load" the setSource

modules/webbrowser/include/modules/webbrowser/processors/webbrowserbase.h
```diff
@@ -0,0 +1,116 @@
...
+
+    /**
+     * Set the zoom level of the browser to the absolute @p zoom factor.
+     * @param zoom   absolute zoom factor, i.e. 1.0 corresponds to a scaling of 100%
+     */
+    void setZoom(double zoom);
+
+    void render(ImageOutport& outport, ImageInport* background = nullptr);
+
+    void executeJavaScript(const std::string& javascript, int startLine = 1);
```
petersteneteg, line 102: string_view?

modules/webbrowser/include/modules/webbrowser/processors/webbrowserbase.h
```diff
@@ -0,0 +1,116 @@
...
+
+    InteractionHandler* getInteractionHandler();
+
+    /**
+     * Registers a callback function @p f that will be called whenever the web page has been loaded.
+     * @param f   callback function
+     * \see CefLoadHandler::OnLoadingStateChange
+     */
+    const BaseCallBack* addLoadingDoneCallback(std::function<void()> f);
```
petersteneteg: prefer the scoped ones for new code

modules/webbrowser/include/modules/webbrowser/processors/webbrowserbase.h
martinfalk: CEF expects a `CefString` which is implicitly convertible from std::string, but not std::string_view.
```c++
void CefFrameCToCpp::ExecuteJavaScript(const CefString& code,
                                       const CefString& script_url,
                                       int start_line);
```

modules/webbrowser/include/modules/webbrowser/interaction/cefinteractionhandler.h
martinfalk: Created issue #1569 on that.

modules/webbrowser/include/modules/webbrowser/processors/webbrowserbase.h
martinfalk: > What about passing in source as a string?

Do you mean source code of the webpage or just the URI/URL?

modules/webbrowser/include/modules/webbrowser/processors/webbrowserbase.h
petersteneteg: ok

modules/webbrowser/include/modules/webbrowser/processors/webbrowserbase.h
petersteneteg: the full source code

petersteneteg: think of having the html as a string property for example

modules/webbrowser/include/modules/webbrowser/processors/webbrowserbase.h
```diff
@@ -0,0 +1,117 @@
...
+
+    InteractionHandler* getInteractionHandler();
+
+    /**
+     * Registers a callback function @p f that will be called whenever the web page has been loaded.
+     * @param f   callback function
+     * @return RAII callback
+     * \see CefLoadHandler::OnLoadingStateChange
+     */
+    [[nodiscard]] std::shared_ptr<BaseCallBack> addLoadingDoneCallback(std::function<void()> f);
```
petersteneteg, line 112: I.e. just use the dispatcher directly instead of callbacklist...

petersteneteg: inviwo/core/util/dispatcher.h

modules/webbrowser/include/modules/webbrowser/processors/webbrowserprocessor.h
```diff
@@ -94,75 +74,41 @@ namespace inviwo {
...
     ImageInport background_;
     ImageOutport outport_;
 
+    OptionProperty<SourceType> sourceType_;
     FileProperty fileName_;
     BoolProperty autoReloadFile_;
-    StringProperty url_;     ///< Web page to show
-    ButtonProperty reload_;  ///< Force reload url
-
+    StringProperty url_;
```
petersteneteg, line 104: Add a StringProperty for the source?

modules/webbrowser/include/modules/webbrowser/processors/webbrowserprocessor.h
```diff
@@ -94,75 +74,41 @@ namespace inviwo {
...
-    PickingMapper picking_;
-    CefImageConverter cefToInviwoImageConverter_;
-    // create browser-window
-    CefRefPtr<RenderHandlerGL> renderHandler_;
-    CefRefPtr<CefBrowser> browser_;
-
     SingleFileObserver fileObserver_;
-
-    IMPLEMENT_REFCOUNTING(WebBrowserProcessor);
+    WebBrowserBase browser_;
```
petersteneteg: How does the ref counting of WebBrowserBase work? I don't know how IMPLEMENT_REFCOUNTING work, But I would expect that you would have to create this object on the heap? with some special pointer to hold it?

modules/webbrowser/include/modules/webbrowser/processors/webbrowserprocessor.h
```diff
@@ -94,75 +74,41 @@ namespace inviwo {
...
  */
-#include <warn/push>
-#include <warn/ignore/dll-interface-base>  // Fine if dependent libs use the same CEF lib binaries
-#include <warn/ignore/extra-semi>  // Due to IMPLEMENT_REFCOUNTING, remove when upgrading CEF
-class IVW_MODULE_WEBBROWSER_API WebBrowserProcessor : public Processor, public CefLoadHandler {
+class IVW_MODULE_WEBBROWSER_API WebBrowserProcessor : public Processor, public ProgressBarOwner {
 public:
-    WebBrowserProcessor(InviwoApplication* app);
-    virtual ~WebBrowserProcessor();
+    explicit WebBrowserProcessor(InviwoApplication* app);
```
petersteneteg, line 84: would be nice it this processor also had some import, maybe a data frame or json object that automatically got exposed in the webpage

modules/webbrowser/src/interaction/cefinteractionhandler.cpp
```diff
@@ -229,10 +229,9 @@ CefKeyEvent CEFInteractionHandler::mapKeyEvent(const KeyboardEvent* e) {
     cefEvent.type = e->state() & KeyState::Press ? KEYEVENT_KEYDOWN : KEYEVENT_KEYUP;
 
     // Convert event character from utf-8 to utf-16
-    auto text = utf8::utf8to16(e->text());
-
-    if (text.length() > 0) {
+    if (auto text = utf8::utf8to16(e->text()); text.length() > 0) {
```
petersteneteg: use empty?

modules/webbrowser/src/processors/webbrowserbase.cpp
```diff
@@ -0,0 +1,150 @@
...
+          app->getModuleByType<WebBrowserModule>()->getBrowserClient()->GetRenderHandler().get())} {
+
+    // Setup CEF browser
+    auto [windowInfo, browserSettings] = cefutil::getDefaultBrowserSettings();
+    auto browserClient = app->getModuleByType<WebBrowserModule>()->getBrowserClient();
+    // Note that browserClient_ outlives this class so make sure to remove
+    // renderHandler_ in
+    // destructor
+    browser_ = CefBrowserHost::CreateBrowserSync(
+        windowInfo, browserClient, std::string{getSource()}, browserSettings, nullptr, nullptr);
```
petersteneteg: getSource will always be empty in the constructor? Pass in source as an argument?

modules/webbrowser/src/processors/webbrowserbase.cpp
```diff
@@ -0,0 +1,150 @@
...
+}
+
+void WebBrowserBase::render(ImageOutport& outport, ImageInport* background) {
+    // TODO: needed here?
+    if (isLoading_) {
+        return;
+    }
+    // if (js_.isModified() && !js_.get().empty()) {
+    //     browser_->GetMainFrame()->ExecuteJavaScript(js_.get(), "", 1);
+    // }
```
petersteneteg: remove?

modules/webbrowser/include/modules/webbrowser/processors/webbrowserbase.h
```diff
@@ -0,0 +1,117 @@
...
+
+private:
+    std::string_view getSource() const;
```
petersteneteg: return a const std::string& to avoid the copies

modules/webbrowser/src/processors/webbrowserbase.cpp
```diff
@@ -0,0 +1,150 @@
...
+void WebBrowserBase::reload() {
+    isLoading_ = true;
+    browser_->GetMainFrame()->LoadURL(std::string{getSource()});
+}
+
+bool WebBrowserBase::isLoading() const { return isLoading_; }
+
+void WebBrowserBase::load(const std::filesystem::path& filename) {
+    if (!filename.empty()) {
+        source_ = fmt::format("file://{}", filename.string());
```
petersteneteg: rename to url or something like that

modules/webbrowser/include/modules/webbrowser/processors/webbrowserbase.h
petersteneteg: also rename to getURL or similar

modules/webbrowser/src/processors/webbrowserbase.cpp
```diff
@@ -0,0 +1,150 @@
...
+void WebBrowserBase::setZoom(double zoom) {
+    browser_->GetHost()->SetZoomLevel(cefutil::percentageToZoomLevel(zoom));
+}
+
+std::shared_ptr<BaseCallBack> WebBrowserBase::addLoadingDoneCallback(std::function<void()> f) {
+    return loadingDone_.addLambdaCallbackRaii(f);
+}
+
+std::string_view WebBrowserBase::getSource() const {
+#ifndef NDEBUG
```
petersteneteg: better handle this in reload where we call LoadURL

petersteneteg: should be some way to clear the page?

petersteneteg (changes requested):

modules/webbrowser/include/modules/webbrowser/processors/webbrowserbase.h
martinfalk: Would probably require a temporary file. The CEF interface only provides `LoadURL()` as far as I can see.
https://bitbucket.org/chromiumembedded/cef/src/68e1fcf3a4ff003b471f920f83008796095c6124/libcef_dll/ctocpp/frame_ctocpp.h#lines-50:54

modules/webbrowser/include/modules/webbrowser/processors/webbrowserprocessor.h
martinfalk: see https://github.com/inviwo/modules/pull/234

modules/webbrowser/include/modules/webbrowser/processors/webbrowserprocessor.h
martinfalk: Not directly supported by CEF.

modules/webbrowser/include/modules/webbrowser/processors/webbrowserprocessor.h
martinfalk: The macro extends to
```c++
///
/// Macro that provides a reference counting implementation for classes
/// extending CefBase.
///
#define IMPLEMENT_REFCOUNTING(ClassName)          \
 public:                                          \
  void AddRef() const override {                  \
    ref_count_.AddRef();                          \
  }                                               \
  bool Release() const override {                 \
    if (ref_count_.Release()) {                   \
      delete static_cast<const ClassName*>(this); \
      return true;                                \
    }                                             \
    return false;                                 \
  }                                               \
  bool HasOneRef() const override {               \
    return ref_count_.HasOneRef();                \
  }                                               \
  bool HasAtLeastOneRef() const override {        \
    return ref_count_.HasAtLeastOneRef();         \
  }                                               \
                                                  \
 private:                                         \
  CefRefCount ref_count_
```

petersteneteg: how does this interact with the destructor? seems shady as hell to have this as a member by value...
note the `delete static_cast<const ClassName*>(this);`call

petersteneteg: https://bitbucket.org/chromiumembedded/cef/wiki/GeneralUsage#markdown-header-reference-counting
need a CefRefPtr I think

modules/webbrowser/include/modules/webbrowser/processors/webbrowserprocessor.h
petersteneteg: see https://bitbucket.org/chromiumembedded/cef/src/68e1fcf3a4ff003b471f920f83008796095c6124/tests/cefclient/browser/test_runner.cc#lines-449

modules/webbrowser/include/modules/webbrowser/processors/webbrowserprocessor.h
petersteneteg: see https://bitbucket.org/chromiumembedded/cef/wiki/GeneralUsage#markdown-header-reference-counting
use CefRefPtr

modules/webbrowser/include/modules/webbrowser/processors/webbrowserprocessor.h
petersteneteg: I would prefer to make this more generic. I think we we add support for a string property for the html and a data frame / json port, There would be much less need for custom processor for each web thing

petersteneteg: Right now this processor is quite useless...

modules/webbrowser/src/processors/webbrowserbase.cpp
```diff
@@ -38,34 +39,36 @@
...
     browserClient->addLoadHandler(this);
-    // Inject events into CEF browser_
+    // Inject events into CEF browser
     cefInteractionHandler_.setHost(browser_->GetHost());
     cefInteractionHandler_.setRenderHandler(renderHandler_.get());
 }
 
 WebBrowserBase::~WebBrowserBase() {
-    auto client = static_cast<WebBrowserClient*>(browser_->GetHost()->GetClient().get());
+    auto* client = dynamic_cast<WebBrowserClient*>(browser_->GetHost()->GetClient().get());
```
petersteneteg: if you dynamic cast you also need to check the result...

modules/webbrowser/src/processors/webbrowserbase.cpp
```diff
@@ -93,44 +92,53 @@ void WebBrowserBase::executeJavaScript(const std::string& javascript, int startL
...
+    return loadingDone_.add(std::move(f));
 }
 
-std::string_view WebBrowserBase::getSource() const {
+const std::string& WebBrowserBase::getURL() const {
 #ifndef NDEBUG
     // CEF does not allow empty urls in debug mode
-    if (source_.empty()) {
-        return "https://www.inviwo.org";
+    if (url_.empty()) {
```
petersteneteg: can't return this temp as a reference? better handle this in reload. or check in the load functions instead to set the url_

petersteneteg (changes requested):

modules/webbrowser/include/modules/webbrowser/processors/webbrowserprocessor.h
martinfalk: Nice catch. Seems it was always "broken".

modules/webbrowser/include/modules/webbrowser/processors/webbrowserprocessor.h
martinfalk: I agree. This processor exists now purely for legacy reasons.

modules/webbrowser/include/modules/webbrowser/processors/webbrowserprocessor.h
petersteneteg: yes..

modules/webbrowser/src/processors/webbrowserbase.cpp
```diff
@@ -92,7 +99,18 @@ void WebBrowserBase::executeJavaScript(const std::string& javascript, int startL
 
 void WebBrowserBase::reload() {
     isLoading_ = true;
-    browser_->GetMainFrame()->LoadURL(getURL());
+
+#ifndef NDEBUG
```
petersteneteg, line 103: could just to this always...

modules/webbrowser/include/modules/webbrowser/processors/webbrowserbase.h
petersteneteg: can fix this later I guess

petersteneteg (approved):

PR #1587, opened 2024-09-10
Feature/linerenderer bnl

martinfalk:
Changes proposed in this PR:
 * extended `LineRenderer` processor with support for Brushing and Linking


modules/basegl/src/rendering/linerenderer.cpp
```diff
@@ -100,7 +123,7 @@ void LineRenderer::render(const Mesh& mesh, const Camera& camera, size2_t screen
             if (mesh.getIndexMeshInfo(i).dt != DrawType::Lines) continue;
             auto& shader = lineShaders_.getShader(mesh, mesh.getIndexMeshInfo(i));
             shader.activate();
-            setUniforms(shader, mesh, camera, screenDim);
+            setUniforms(shader, mesh, camera, screenDim, func);
```
petersteneteg, line 126: does this even compile? I would have expected you to call func(shader) directly.
Also maybe we should name func, something like setAdditinonalUniforms

petersteneteg: ahh there is a member function

petersteneteg (dismissed):

petersteneteg (approved):

PR #1598, opened 2024-10-11
Feature/int histogram

martinfalk:
Changes proposed in this PR:
 * refactored 1D histogram calculation 
 * made sure that the bin size is a whole number for integer ranges to prevent aliasing effects

### DIFF
--- a/include/inviwo/core/algorithm/histogram1d.h
+++ b/include/inviwo/core/algorithm/histogram1d.h
@@ -1,0 +1,189 @@
+/*********************************************************************************
+ *
+ * Inviwo - Interactive Visualization Workshop
+ *
+ * Copyright (c) 2024 Inviwo Foundation
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ *********************************************************************************/
+#pragma once
+
+#include <inviwo/core/common/inviwocoredefine.h>
+#include <inviwo/core/datastructures/histogram.h>
+#include <inviwo/core/util/glmutils.h>
+#include <inviwo/core/util/glmcomp.h>
+#include <inviwo/core/util/glmmatext.h>
+
+#include <glm/common.hpp>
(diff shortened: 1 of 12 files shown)


include/inviwo/core/algorithm/histogram1d.h
```diff
@@ -0,0 +1,200 @@
...
+    const auto range = (dataMap.dataRange.y - dataMap.dataRange.x);
+    if (range / static_cast<double>(bins) > 100.0) {  // case 1
+        // sufficient samples per bin
+        return {bins, range};
+    }
+    if constexpr (std::is_floating_point_v<util::value_type_t<T>>) {
+        if (range < static_cast<double>(bins)) {  // case 3.a
+            return {bins, range};
+        }
+        if (isNonFractional(dataMap.dataRange.x) && isNonFractional(dataMap.dataRange.y)) {
```
petersteneteg: What would happen if dataRange is like 1E-13 to 5E-13
Should probably also check that min - max >= 1?

include/inviwo/core/algorithm/histogram1d.h
```diff
@@ -0,0 +1,200 @@
...
+namespace inviwo::util {
+
+IVW_CORE_API std::vector<double> calculatePercentiles(const std::vector<size_t>& hist, dvec2 range,
+                                                      size_t sum);
+IVW_CORE_API Statistics calculateHistogramStats(const std::vector<size_t>& hist);
+
+namespace detail {
+
+/**
+ * Determine optimal number of bins of a histogram while considering \p bins as an upper bound
```
petersteneteg, line 55: Might write a line about why we need to do this... i.e that a fractional bin size, might align badly with integer data...

include/inviwo/core/algorithm/histogram1d.h
```diff
@@ -0,0 +1,200 @@
...
+    for (size_t channel = 0; channel < extent; ++channel) {
+        const auto maxBinCount = *std::ranges::max_element(hists[channel]);
+
+        histograms.push_back(Histogram1D{
+            .counts = hists[channel],
+            .totalCounts = count,
+            .maxCount = maxBinCount,
+            .dataMap = dataMap,
+            .effectiveDataRange = effectiveDataRange,
+            .effectiveValueRange = effectiveValueRange,
```
petersteneteg: Do we need to expose this externally?

petersteneteg: If so, you have to explain why and when they should be taken into account...

include/inviwo/core/datastructures/histogram.h
```diff
@@ -65,106 +57,13 @@ struct IVW_CORE_API Histogram1D {
     size_t totalCounts{0};
     size_t maxCount{0};
     DataMapper dataMap{};
+    dvec2 effectiveDataRange{0.0};
```
petersteneteg: As I stated above, do we need to expose there, and if so when should I use them?

include/inviwo/core/algorithm/histogram1d.h
martinfalk: That is a good point.

modules/volume/include/inviwo/volume/processors/histogramtodataframe.h
```diff
@@ -98,7 +100,7 @@ struct ProcessorTraits<HistogramToDataFrame<Volume>> {
 namespace detail {
 
 IVW_MODULE_VOLUME_API std::shared_ptr<DataFrame> createDataFrame(
-    const std::vector<Histogram1D>& histograms, HistogramMode mode);
+    const std::vector<Histogram1D>& histograms, HistogramMode mode, bool includeDataRangeValues);
```
petersteneteg: I don't understand what includeDataRangeValues it supposed do

petersteneteg (dismissed):

modules/volume/include/inviwo/volume/processors/histogramtodataframe.h
martinfalk: The returned DataFrame contains two colums: bin centers in ValueRange and bin counts. If true, this adds an additional column with the bin locations using the original, unmapped scalar values using DataRange.

include/inviwo/core/datastructures/histogram.h
martinfalk: Those are needed in case one wants to compute the bin centers in either DataRange or ValueRange as done in the HistogramToDataFrame processors.
https://github.com/inviwo/inviwo/blob/0a4d88da3abc8c091dd74a389973fdd2612e27c5/modules/volume/src/processors/histogramtodataframe.cpp#L76-L87

petersteneteg: But why it this not needed in the TF editor than? you did not change that right? 

Think you need to explain how this is related to just dividing the datarange in nbins steps.

modules/volume/include/inviwo/volume/processors/histogramtodataframe.h
petersteneteg: Will they be different? And if so by how much? Should be less then a bin then or?

petersteneteg: Is this needed for anything else then our debugging?

include/inviwo/core/algorithm/histogram1d.h
martinfalk: Removed those and adjusted the data map instead.

include/inviwo/core/datastructures/histogram.h
martinfalk: Adjusted the TF editor accordingly.

petersteneteg (approved):

petersteneteg (approved):

PR #1605, opened 2024-10-31
AnimationQt: Add support for using a presentation clicker

ResearchDaniel:
Clicking left/right on a presentation clicker animates to next (ControlTrack) keyframe. 
Double-clicking jumps to the next/prev (ControlTrack) keyframe

### DIFF
--- a/modules/animation/include/modules/animation/animationcontroller.h
+++ b/modules/animation/include/modules/animation/animationcontroller.h
@@ -85,6 +85,12 @@ public:
     // Pause and reset to start
     void stop();
 
+    void nextKeyframe();
+    void prevKeyframe();
+
+    void nextControlKeyframe();
+    void prevControlKeyframe();
+
     void setState(AnimationState newState);
 
     /// Advances the animation to the next time step in playing state.
--- a/modules/animation/include/modules/animation/animationcontroller.h
+++ b/modules/animation/include/modules/animation/animationcontroller.h
@@ -85,11 +85,31 @@ public:
     // Pause and reset to start
     void stop();
 
-    void nextKeyframe();
-    void prevKeyframe();
+    /*
+     * Set current time to that of the previous keyframe, if there is one.
+     * Equal to eval(getCurrentTime(), prevKeyframeTime).
+     * Does not alter current PlaybackDirection.
+     */
+    void jumpToPrevKeyframe();
+    /*
+     * Set current time to that of the next keyframe, if there is one.
+     * Equal to eval(getCurrentTime(), nextKeyframeTime).
+     * Does not alter current PlaybackDirection.
+     */
+    void jumpToNextKeyframe();
 
-    void nextControlKeyframe();
-    void prevControlKeyframe();
+    /*
+     * Set current time to that of the previous ControlTrack keyframe, if there is one.
+     * Equal to eval(getCurrentTime(), prevKeyframeTime).
+     * Does not alter current PlaybackDirection.
+     */
+    void jumpToPrevControlKeyframe();
+    /*
+     * Set current time to that of the next ControlTrack keyframe, if there is one.
+     * Equal to eval(getCurrentTime(), nextKeyframeTime).
+     * Does not alter current PlaybackDirection.
+     */
+    void jumpToNextControlKeyframe();
 
     void setState(AnimationState newState);
 
--- a/modules/animation/include/modules/animation/datastructures/animation.h
+++ b/modules/animation/include/modules/animation/datastructures/animation.h
@@ -41,6 +41,7 @@
 
 #include <cstddef>      // for size_t
 #include <memory>       // for unique_ptr
+#include <ranges>
 #include <string>       // for string
 #include <string_view>  // for string_view
 #include <vector>       // for vector<>::iterator, vect...
@@ -171,11 +172,18 @@ public:
     iterator findTrack(Property* withMe);
 
     /**
-     * Find Track of a given type.
-     * @return All tracks of the specified type.
+     * Check if any track of TrackType has been added.
+     * @return True if any of the tracks are of the provided type, false otherwise.
      */
     template <typename TrackType>
-    std::vector<TrackType*> getTracksOfType();
+    bool hasTrackType() const;
+
+    /**
+     * Find Track of a given type.
+     * @return View of all tracks of the specified type.
+     */
+    template <typename TrackType>
+    std::ranges::view auto getTracksOfType();
 
     /**
      * Remove all tracks. Calls TrackObserver::notifyTrackRemoved for each removed track.
@@ -197,6 +205,16 @@ public:
      * Return time of last Keyframe in all tracks, or 0 if no track exist.
      */
     Seconds getLastTime() const;
+
+    /**
+     * Return time of closest previous Keyframe in all tracks, or at if no previous keyframe exist.
+     */
+    Seconds getPrevTime(Seconds at) const;
+
+    /**
+     * Return time of closest next Keyframe in all tracks, or at if there is no next keyframe.
+     */
+    Seconds getNextTime(Seconds at) const;
 
     /**
      * Return the name of the Animation. Used for display in the GUI.
@@ -238,14 +256,14 @@ private:
 };
 
 template <typename TrackType>
-std::vector<TrackType*> Animation::getTracksOfType() {
-    std::vector<TrackType*> tracks;
-    for (auto& track : *this) {
-        if (auto tr = dynamic_cast<TrackType*>(&track)) {
-            tracks.push_back(tr);
-        }
-    }
-    return tracks;
+bool Animation::hasTrackType() const {
+    return end() !=
+           std::find_if(begin(), end(), [](const auto& t) { return dynamic_cast<const TrackType*>(&t) != nullptr; });
+}
+
+template <typename TrackType>
+std::ranges::view auto Animation::getTracksOfType() {
+    return *this | std::views::filter([](auto& t) { return dynamic_cast<TrackType*>(&t); });
 }
 
 }  // namespace animation
--- a/modules/animation/include/modules/animation/datastructures/animation.h
+++ b/modules/animation/include/modules/animation/datastructures/animation.h
@@ -171,6 +171,13 @@ public:
     iterator findTrack(Property* withMe);
 
     /**
+     * Find Track of a given type.
+     * @return All tracks of the specified type.
+     */
+    template<typename TrackType>
+    std::vector<TrackType*> getTracksOfType();
+
+    /**
      * Remove all tracks. Calls TrackObserver::notifyTrackRemoved for each removed track.
      */
     void clear();
@@ -229,6 +236,17 @@ private:
     std::string name_;
     AnimationManager* am_;
 };
+
+template <typename TrackType>
+std::vector<TrackType*> Animation::getTracksOfType() {
+    std::vector<TrackType*> tracks;
+    for (auto& track : *this) {
+        if (auto tr = dynamic_cast<TrackType*>(&track)) {
+            tracks.push_back(tr);
+        }
+    }
+    return tracks;
+}
 
 }  // namespace animation
 
--- a/modules/animation/include/modules/animation/datastructures/animation.h
+++ b/modules/animation/include/modules/animation/datastructures/animation.h
@@ -174,7 +174,7 @@ public:
      * Find Track of a given type.
      * @return All tracks of the specified type.
      */
-    template<typename TrackType>
+    template <typename TrackType>
     std::vector<TrackType*> getTracksOfType();
 
     /**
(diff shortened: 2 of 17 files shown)


modules/animation/include/modules/animation/animationcontroller.h
```diff
@@ -85,6 +85,12 @@ class IVW_MODULE_ANIMATION_API AnimationController : public AnimationControllerO
     // Pause and reset to start
     void stop();
 
+    void nextKeyframe();
```
petersteneteg: Think there should be named, playToNextkeyframe or jumpToNextKeyframe. 
Do they take the play direction into account?

modules/animation/include/modules/animation/animationcontroller.h
```diff
@@ -85,6 +85,12 @@ class IVW_MODULE_ANIMATION_API AnimationController : public AnimationControllerO
...
+    void prevKeyframe();
+
+    void nextControlKeyframe();
```
petersteneteg: Same here

modules/animation/include/modules/animation/datastructures/animation.h
```diff
@@ -170,6 +170,13 @@ class IVW_MODULE_ANIMATION_API Animation : public AnimationObservable,
      */
     iterator findTrack(Property* withMe);
 
+    /**
+     * Find Track of a given type.
+     * @return All tracks of the specified type.
+     */
+    template<typename TrackType>
+    std::vector<TrackType*> getTracksOfType();
```
petersteneteg: why not return a lazy range here instead?

modules/animation/src/animationcontroller.cpp
```diff
@@ -258,6 +258,50 @@ void AnimationController::stop() {
     eval(currentTime_, Seconds(0));
 }
 
+void AnimationController::nextKeyframe() {
```
petersteneteg: This functionality should be in animation, something like, getNextTime(time t). Maybe with a direction also.

petersteneteg: That would avoid the expensive getAllTimes.

petersteneteg: I.e. the functionallity the find the next time, not the evaluation

modules/animation/src/animationcontroller.cpp
```diff
@@ -258,6 +258,50 @@ void AnimationController::stop() {
...
+void AnimationController::prevKeyframe() {
+    auto times = animation_->getAllTimes();
+    auto it = std::lower_bound(times.begin(), times.end(), getCurrentTime());
+    if (it != times.begin()) {
+        eval(getCurrentTime(), *std::prev(it));
+    }
+}
+
+void AnimationController::nextControlKeyframe() {
+    std::vector<Seconds> times;
```
petersteneteg: Same here add a function to animation for finding the next time. Avoid allocation any vectors.

modules/animationqt/src/animationeditordockwidgetqt.cpp
```diff
@@ -312,18 +313,47 @@ AnimationEditorDockWidgetQt::AnimationEditorDockWidgetQt(
                 [&]() { controller_.eval(controller_.getCurrentTime(), Seconds(0.0)); });
     }
 
+    {
+        // For presentation, e.g, a clicker (Key_PageUp)
+        auto* prevControlKeyframe = mainWindow_->addAction("Prev Control Keyframe");
```
petersteneteg, line 327: Should this not also be in the menu to make it more discoverable?

modules/animationqt/src/animationeditordockwidgetqt.cpp
```diff
@@ -312,18 +313,47 @@ AnimationEditorDockWidgetQt::AnimationEditorDockWidgetQt(
...
+        prevControlKeyframe->setShortcut(Qt::Key_PageUp);  
+        prevControlKeyframe->setShortcutContext(Qt::WidgetWithChildrenShortcut);
+        prevControlKeyframe->setToolTip("Previous Control Keyframe");
+        connect(prevControlKeyframe, &QAction::triggered, [&]() {
+            if (isKeyDoublePressed()) {
+                if (controller_.getAnimation().getTracksOfType<ControlTrack>().empty()) {
```
petersteneteg: Add a hasTrackOfType, to avoid allocation extra vector

modules/animationqt/include/modules/animationqt/animationeditordockwidgetqt.h
```diff
@@ -97,6 +101,8 @@ class IVW_MODULE_ANIMATIONQT_API AnimationEditorDockWidgetQt : public InviwoDock
     TextLabelOverlay* overlay_;
     SequenceEditorPanel* sequenceEditorView_;
     QMainWindow* mainWindow_;
+    std::chrono::time_point<std::chrono::system_clock> lastKeyEventTimestamp_ =
```
petersteneteg: This state would be better to keep in the action lambda it self.

modules/animationqt/src/animationeditordockwidgetqt.cpp
```diff
@@ -360,6 +419,8 @@ AnimationEditorDockWidgetQt::AnimationEditorDockWidgetQt(
 
     toolBar->addSeparator();
 
+    animationView_->setFocus();
```
petersteneteg: Why?

petersteneteg (changes requested):

modules/animation/include/modules/animation/animationcontroller.h
ResearchDaniel: Fixed. 
They ignore play direction. Otherwise it gets very unintuitive to use them

modules/animation/include/modules/animation/datastructures/animation.h
ResearchDaniel: Fixed

modules/animation/src/animationcontroller.cpp
ResearchDaniel: Fixed, although I don't think there is any real performance gain here. The changes needed to be propagated to the tracks and sequences.

modules/animation/src/animationcontroller.cpp
ResearchDaniel: Fixed

modules/animationqt/src/animationeditordockwidgetqt.cpp
ResearchDaniel: Maybe, but would also make the toolbar cluttered with next/prev buttons. I think we need to design some nice buttons in that case. It is possible to use the corresponding track's prev/next buttons even though it is not as obvious

modules/animationqt/src/animationeditordockwidgetqt.cpp
ResearchDaniel: Fixed

modules/animationqt/src/animationeditordockwidgetqt.cpp
ResearchDaniel: It was needed to get the key events to go to the animation controller. Might be better ways of doing this, setTabOrder? I removed it for now. I think that the dock widgets do not get focus when they are shown right now? The focus is still on the network editor when I tested it. 
Is probably another PR

modules/animationqt/src/animationeditordockwidgetqt.cpp
petersteneteg: We can have items in the menu without having it in the toolbar. I would not put it in the toolbar either. But I'm thinking as a text entry in the menu.

modules/animation/include/modules/animation/datastructures/keyframesequence.h
```diff
@@ -74,6 +74,14 @@ class IVW_MODULE_ANIMATION_API KeyframeSequence : public Serializable,
 
     Seconds getFirstTime() const;
     Seconds getLastTime() const;
+    /*
+     * Return time of previous keyframe, or at if not found.
+     */
+    virtual Seconds getPrevTime(Seconds at) const = 0;
```
petersteneteg: would make more sense to return an std::optional<Seconds> here

modules/animation/src/datastructures/animation.cpp
```diff
@@ -266,6 +266,40 @@ Seconds Animation::getLastTime() const {
     return time;
 }
 
+Seconds Animation::getPrevTime(Seconds at) const {
+    auto prevKeyframeTime = at;
```
petersteneteg: better to use a optional here to track the "unset" state then checking if it equal to at... that also propagates better upwards....

petersteneteg (dismissed):
Nice! left some minor comments.

modules/animationqt/src/animationeditordockwidgetqt.cpp
ResearchDaniel: Fixed focus issue in latest commits

modules/animation/include/modules/animation/datastructures/keyframesequence.h
ResearchDaniel: Definitely, fixed

petersteneteg (approved):

petersteneteg:
@ResearchDaniel CEF really don't like you running multiple instances at the same time... Any easy fix for that?

ResearchDaniel:
> @ResearchDaniel CEF really don't like you running multiple instances at the same time... Any easy fix for that?

I think that this warning message might be the key:
"Please customize CefSettings.root_cache_path for your application"
https://www.magpcss.org/ceforum/viewtopic.php?f=6&t=19727
https://cef-builds.spotifycdn.com/docs/122.1/structcef__settings__t.html#a2e2be03f34ddd93de90e1cf196757a19

> Multiple application instances writing to the same root_cache_path directory could result in data corruption. A process singleton lock based on the root_cache_path value is therefore used to protect against this. This singleton behavior applies to all CEF-based applications using version 120 or newer. You should customize root_cache_path for your application and implement [CefBrowserProcessHandler](https://cef-builds.spotifycdn.com/docs/122.1/classCefBrowserProcessHandler.html):: OnAlreadyRunningAppRelaunch, which will then be called on any app relaunch with the same root_cache_path value.


> Failure to set the root_cache_path value correctly may result in startup crashes or other unexpected behaviors (for example, the sandbox blocking read/write access to certain files).

petersteneteg:
> > @ResearchDaniel CEF really don't like you running multiple instances at the same time... Any easy fix for that?
> 
> I think that this warning message might be the key: "Please customize CefSettings.root_cache_path for your application" https://www.magpcss.org/ceforum/viewtopic.php?f=6&t=19727 https://cef-builds.spotifycdn.com/docs/122.1/structcef__settings__t.html#a2e2be03f34ddd93de90e1cf196757a19
> 
> > Multiple application instances writing to the same root_cache_path directory could result in data corruption. A process singleton lock based on the root_cache_path value is therefore used to protect against this. This singleton behavior applies to all CEF-based applications using version 120 or newer. You should customize root_cache_path for your application and implement [CefBrowserProcessHandler](https://cef-builds.spotifycdn.com/docs/122.1/classCefBrowserProcessHandler.html):: OnAlreadyRunningAppRelaunch, which will then be called on any app relaunch with the same root_cache_path value.
> 
> > Failure to set the root_cache_path value correctly may result in startup crashes or other unexpected behaviors (for example, the sandbox blocking read/write access to certain files).

So should one give it a fresh dir on every start?

ResearchDaniel:
> > > @ResearchDaniel CEF really don't like you running multiple instances at the same time... Any easy fix for that?
> > 
> > 
> > I think that this warning message might be the key: "Please customize CefSettings.root_cache_path for your application" https://www.magpcss.org/ceforum/viewtopic.php?f=6&t=19727 https://cef-builds.spotifycdn.com/docs/122.1/structcef__settings__t.html#a2e2be03f34ddd93de90e1cf196757a19
> > > Multiple application instances writing to the same root_cache_path directory could result in data corruption. A process singleton lock based on the root_cache_path value is therefore used to protect against this. This singleton behavior applies to all CEF-based applications using version 120 or newer. You should customize root_cache_path for your application and implement [CefBrowserProcessHandler](https://cef-builds.spotifycdn.com/docs/122.1/classCefBrowserProcessHandler.html):: OnAlreadyRunningAppRelaunch, which will then be called on any app relaunch with the same root_cache_path value.
> > 
> > 
> > > Failure to set the root_cache_path value correctly may result in startup crashes or other unexpected behaviors (for example, the sandbox blocking read/write access to certain files).
> 
> So should one give it a fresh dir on every start?

Yes, unless we want to mimic the chrome behavior of having a single process with multiple windows. Seems like some people have solved it by appending the process id to the change path and then removing the directory after shutting down. The main downside of such an approach would be that data will not be cached between runs

> CEF will log this message if RootCachePath is not set, you should ensure that at a minimum you set CefSettings.RootCachePath to a unique folder per application to avoid unwanted calls to OnAlreadyRunningAppRelaunch.

> OnAlreadyRunningAppRelaunch can be used to achieve the same behaviour as Chromium. (Single process instance with multiple windows within the one process).
If you need multiple instances that are entirely isolated, then you will need to specify a unique RootCachePath for each.

petersteneteg:
using a unique path might be the best option...

PR #1620, opened 2024-11-27
Feature/dataminmaxgl

martinfalk:
Changes proposed in this PR:
 * symmetric and asymmetric normalization for signed fixed-point integers to match OpenGL behavior
 * added `utilgl::DataMinMaxGL` for determining min/max values of GL representations using compute shaders

This has been tested on: Windows, NVIDIA GPU

### DIFF
--- a/include/inviwo/core/datastructures/datamapper.h
+++ b/include/inviwo/core/datastructures/datamapper.h
@@ -51,15 +51,37 @@ class DataFormatBase;
  */
 class IVW_CORE_API DataMapper {
 public:
+    /**
+     * Signed fixed-point integers are normalized using either an asymmetric range
+     * [-2^(b-1), 2^(b-1) - 1] or a symmetric range [-2^(b-1) + 1, 2^(b-1) - 1], where b
+     * corresponds to the bit depth of the fixed-point integer.
+     *
+     * From version 4.2 and above, OpenGL uses a symmetric range to convert between normalized
+     * fixed-point integers and floating point values. For example, the range [-127, 127] is used
+     * for the normalization of 8bit signed integers instead of [-128, 127].
+     * In order to match the normalization of signed integer formats on the GPU, the symmetric
+     * normalization has to be used in the DataMapper for OpenGL >= 4.2.
+     *
+     * \see OpenGL 4.6 specification, Section 2.3.5 Fixed-Point Data Conversion
+     */
+    enum class SignedNormalization {
+        Asymmetric,  //!< uses an asymmetric range, i.e. [-2^(b-1), 2^(b-1) - 1]
+        Symmetric,   //!< uses a symmetric range, i.e. [-2^(b-1) + 1, 2^(b-1) - 1]
+    };
+
+    // asymetric
+
     DataMapper();
-    DataMapper(const DataFormatBase* format, Axis valueAxis = {});
+    DataMapper(const DataFormatBase* format,
+               SignedNormalization normalization = SignedNormalization::Asymmetric,
+               Axis valueAxis = {});
     explicit DataMapper(dvec2 dataRange, Axis valueAxis = {});
     DataMapper(dvec2 dataRange, dvec2 valueRange, Axis valueAxis = {});
     DataMapper(const DataMapper& rhs);
     DataMapper& operator=(const DataMapper& that);
 
     /**
-     * The ´dataRange´ is used to normalize the "raw" values to [0,1]
+     * The `dataRange` is used to normalize the "raw" values to [0,1]
      * Typically the minimum and maximum of the "raw" data is used. For 8 and 16 integer data that
      * is often the same or close to the same as the minimum and maximum of the data type. For data
      * where the minimum and maximum is far from the minimum and maximum of the data type, the
@@ -85,8 +107,11 @@ public:
      */
     Axis valueAxis;  ///< Name and Unit, i.e. "absorption", "Hounsfield".
 
-    static dvec2 defaultDataRangeFor(const DataFormatBase* format);
-    void initWithFormat(const DataFormatBase* format);
+    static dvec2 defaultDataRangeFor(
+        const DataFormatBase* format,
+        SignedNormalization normalization = SignedNormalization::Asymmetric);
+    void initWithFormat(const DataFormatBase* format,
+                        SignedNormalization normalization = SignedNormalization::Asymmetric);
 
     /**
      * Map from `dataRange` to the `valueRange`
@@ -116,12 +141,32 @@ public:
     }
 
     /**
+     * Map from `dataRange` to the Sign Normalized range [-1,1]
+     */
+    template <typename T>
+    util::same_extent_t<T, double> mapFromDataToSignNormalized(T val) const {
+        return util::linearMapToNormalized(static_cast<util::same_extent_t<T, double>>(val),
+                                           dataRange) *
+                   2.0 -
+               1.0;
+    }
+
+    /**
      * Map from Normalized range [0,1] to the `dataRange`
      */
     template <typename T>
     util::same_extent_t<T, double> mapFromNormalizedToData(T val) const {
         return util::linearMapFromNormalized(static_cast<util::same_extent_t<T, double>>(val),
                                              dataRange);
+    }
+
+    /**
+     * Map from Sign Normalized range [-1,1] to the `dataRange`
+     */
+    template <typename T>
+    util::same_extent_t<T, double> mapFromSignNormalizedToData(T val) const {
+        return util::linearMapFromNormalized(
+            (static_cast<util::same_extent_t<T, double>>(val) + 1.0) * 0.5, dataRange);
     }
 
     /**
--- a/include/inviwo/core/datastructures/datamapper.h
+++ b/include/inviwo/core/datastructures/datamapper.h
@@ -72,11 +72,11 @@ public:
     // asymetric
 
     DataMapper();
-    DataMapper(const DataFormatBase* format,
-               SignedNormalization normalization = SignedNormalization::Asymmetric,
-               Axis valueAxis = {});
-    explicit DataMapper(dvec2 dataRange, Axis valueAxis = {});
-    DataMapper(dvec2 dataRange, dvec2 valueRange, Axis valueAxis = {});
+    explicit DataMapper(const DataFormatBase* format,
+                        SignedNormalization normalization = SignedNormalization::Asymmetric,
+                        const Axis& valueAxis = {});
+    explicit DataMapper(dvec2 dataRange, const Axis& valueAxis = {});
+    DataMapper(dvec2 dataRange, dvec2 valueRange, const Axis& valueAxis = {});
     DataMapper(const DataMapper& rhs);
     DataMapper& operator=(const DataMapper& that);
 
--- a/include/inviwo/core/datastructures/datamapper.h
+++ b/include/inviwo/core/datastructures/datamapper.h
@@ -74,9 +74,9 @@ public:
     DataMapper();
     explicit DataMapper(const DataFormatBase* format,
                         SignedNormalization normalization = SignedNormalization::Asymmetric,
-                        const Axis& valueAxis = {});
-    explicit DataMapper(dvec2 dataRange, const Axis& valueAxis = {});
-    DataMapper(dvec2 dataRange, dvec2 valueRange, const Axis& valueAxis = {});
+                        Axis valueAxis = {});
+    explicit DataMapper(dvec2 dataRange, Axis valueAxis = {});
+    DataMapper(dvec2 dataRange, dvec2 valueRange, Axis valueAxis = {});
     DataMapper(const DataMapper& rhs);
     DataMapper& operator=(const DataMapper& that);
 
--- a/modules/basegl/CMakeLists.txt
+++ b/modules/basegl/CMakeLists.txt
@@ -3,6 +3,7 @@ ivw_module(BaseGL)
 
 # Add header files
 set(HEADER_FILES
+    include/modules/basegl/algorithm/dataminmaxgl.h
     include/modules/basegl/algorithm/entryexitpoints.h
     include/modules/basegl/algorithm/imageconvolution.h
     include/modules/basegl/algorithm/volumenormalization.h
@@ -133,6 +134,7 @@ ivw_group("Header Files" ${HEADER_FILES})
 
 # Add source files
 set(SOURCE_FILES
+    src/algorithm/dataminmaxgl.cpp
     src/algorithm/entryexitpoints.cpp
     src/algorithm/imageconvolution.cpp
     src/algorithm/volumenormalization.cpp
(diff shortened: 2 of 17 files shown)


modules/basegl/CMakeLists.txt
```diff
@@ -267,6 +269,10 @@ set(SHADER_FILES
     glsl/axisalignedcutplaneslice.frag
     glsl/axisalignedcutplaneslice.vert
     glsl/background.frag
+    glsl/compute/bufferminmax.comp
```
ResearchDaniel, line 272: We often organize these types of things into algorithm or buffer/layer/volume folder. Compute will be given by the file ending anyway

modules/basegl/glsl/compute/bufferminmax.comp
```diff
@@ -0,0 +1,90 @@
...
+shared vec4 workgroup_max[256 / 32];
+
+layout(local_size_x = 256, local_size_y = 1, local_size_z = 1) in;
+
+void main(){
+    uint threadId = gl_LocalInvocationIndex;
+    uint gridSize = gl_WorkGroupSize.x * gl_NumWorkGroups.x;
+    uint i = gl_WorkGroupID.x * gl_WorkGroupSize.x + threadId; 
+
+    vec4 minVal = vec4(1.0 / 0.0);
```
ResearchDaniel, line 65: This may not work on GLSL 4.0
https://stackoverflow.com/questions/10435253/glsl-infinity-constant

Can assume that first value exists?
```suggestion
    vec4 minVal = GetValue(source buffer, i);
    i += gridSize;
```

modules/basegl/glsl/compute/bufferminmax.comp
```diff
@@ -0,0 +1,90 @@
...
+    }
+    barrier();
+
+    if (gl_LocalInvocationIndex == 0) {
+        for (uint i = 0; i < gl_NumSubgroups; ++i) {
+            minVal = min(minVal, workgroup_min[i]);
+            maxVal = max(maxVal, workgroup_max[i]);
+        }
+        uint index = gl_WorkGroupID.x;
+        minmax[2 * index] = minVal;
```
ResearchDaniel, line 91: Nice!

modules/basegl/glsl/compute/bufferminmax.comp
```diff
@@ -0,0 +1,90 @@
...
+    minVal = subgroupMin(minVal);
+    maxVal = subgroupMax(maxVal);
+    if (subgroupElect()) {
+        workgroup_min[gl_SubgroupID] = minVal;
+        workgroup_max[gl_SubgroupID] = maxVal;
+    }
+    barrier();
+
+    if (gl_LocalInvocationIndex == 0) {
+        for (uint i = 0; i < gl_NumSubgroups; ++i) {
```
ResearchDaniel, line 86: Nit: One could add some explanation of what is going on here.

modules/basegl/glsl/compute/bufferminmax.comp
```diff
@@ -0,0 +1,90 @@
...
+    BUFFER_DATATYPE sourceBuffer[];
+};
+
+layout (std430, binding=1) buffer BufferObject {
+    vec4 minmax[];
+};
+
+uniform uint arrayLength;
+
+shared vec4 workgroup_min[256 / 32];
```
ResearchDaniel, line 55: Would be good with a comment about the reason for these values.

modules/basegl/glsl/compute/layerminmax.comp
```diff
@@ -0,0 +1,97 @@
...
+#else
+uniform layout(binding=0, IMAGE_FORMAT) readonly IMAGE_SOURCE sourceImage;
+#endif  // USE_IMAGE_SAMPLER
+
+layout (std430, binding=1) buffer BufferObject {
+    vec4 minmax[];
+};
+
+// 32x32 per workgroup
+shared vec4 workgroup_min[1024 / 32];
```
ResearchDaniel, line 62: Would be good with a comment about the reason for these values.

modules/basegl/glsl/compute/bufferminmax.comp
```diff
@@ -0,0 +1,90 @@
...
+// corresponding vec4 types when used with SSBOs, including arrays. To avoid padding, we instead
+// use linear arrays for 3 channels. For example `float sourceBuffer[]` instead of `vec3 sourceBuffer[]`. 
+// Note that this affects data access and indexing handled by GetValue()!
+//
+// See OpenGL 4.6 specification, Section 7.6.2.2 Standard Uniform Block Layout
+#if !defined BUFFER_DATATYPE
+#define BUFFER_DATATYPE vec4
+#endif
+#if !defined GetValue
+#define GetValue(value, index) value[index]
```
ResearchDaniel, line 40: Nit: lower case g
```suggestion
#define getValue(value, index) value[index]
```

modules/basegl/glsl/compute/layerminmax.comp
```diff
@@ -0,0 +1,97 @@
...
+#  define IMAGE_SOURCE image2D
+#endif
+
+// Need to use regular texture samplers for rgb formats, that is 3 channels, since there 
+// are no equivalent format/layout qualifiers and no support by imageLoad()/imageStore().
+//
+// See OpenGL 4.6 specification, Section 8.26 and Table 8.26
+// See https://www.khronos.org/opengl/wiki/Image_Load_Store#Format_qualifiers
+#if defined USE_IMAGE_SAMPLER
+#  define GetValue(gsampler, imageCoord) texelFetch(gsampler, imageCoord, 0)
```
ResearchDaniel, line 43:
```suggestion
#  define getValue(gsampler, imageCoord) texelFetch(gsampler, imageCoord, 0)
```

modules/basegl/glsl/compute/layerminmax.comp
```diff
@@ -0,0 +1,97 @@
...
+#  define GetSize(gsampler) textureSize(gsampler, 0)
```
ResearchDaniel, line 44:
```suggestion
#  define getSize(gsampler) textureSize(gsampler, 0)
```

modules/basegl/glsl/compute/layerminmax.comp
```diff
@@ -0,0 +1,97 @@
...
+
+    ivec2 layerIndex = ivec2(gl_GlobalInvocationID.x, gl_GlobalInvocationID.y);
+
+    if (all(lessThan(layerIndex, layerDims))) {
+        vec4 val = GetValue(sourceImage, layerIndex);
+        minVal = min(minVal, val);
+        maxVal = max(maxVal, val);
+    }
+
+    minVal = subgroupMin(minVal);
```
ResearchDaniel, line 81: Can this, and the code below be refactored into a function and reused across methods?

modules/basegl/src/algorithm/dataminmaxgl.cpp
```diff
@@ -0,0 +1,436 @@
...
+#include <modules/opengl/texture/texture3d.h>
+#include <modules/opengl/texture/textureunit.h>
+#include <modules/base/algorithm/dataminmax.h>
+
+#include <string_view>
+#include <array>
+
+namespace inviwo::utilgl {
+
+namespace detail {
```
petersteneteg: for funktions local to the TU, it is better to use a anonymous namespace to ensure the function gets internal linkage.

modules/basegl/src/algorithm/dataminmaxgl.cpp
```diff
@@ -0,0 +1,436 @@
...
+
+        minmaxGL[0] = dataMapper.mapFromSignNormalizedToData(minmaxGL[0]);
+        minmaxGL[1] = dataMapper.mapFromSignNormalizedToData(minmaxGL[1]);
+    }
+    return {minmaxGL[0], minmaxGL[1]};
+}
+
+[[nodiscard]] std::pair<dvec4, dvec4> reduce2D(Shader& shader, Shader& linearReduce, uvec2 dims,
+                                               const GLFormat& glFormat, const Texture& texture) {
+    const uvec3 groupSize{32, 32, 1};
```
petersteneteg, line 116: does this have to match the values in the shader? Might add a note about that.

modules/basegl/src/algorithm/dataminmaxgl.cpp
```diff
@@ -0,0 +1,436 @@
...
+std::pair<dvec4, dvec4> DataMinMaxGL::minMax(const Volume* volume) {
+    if (!gpuSupport_) {
+        return util::volumeMinMax(volume);
+    }
+    return minMax(volume->getRepresentation<VolumeGL>());
+}
+
+std::pair<dvec4, dvec4> DataMinMaxGL::minMax(const VolumeGL* volumeGL) {
+    if (!gpuSupport_) {
+        return util::volumeMinMax(volumeGL->getOwner());
```
petersteneteg: Owner can be nullptr

petersteneteg: Maybe it is better to throw an error here and have the Volume, overload handle the fallback... 
If you call it with a gl rep you are probably expecting it to use the gl rep also...

modules/basegl/src/algorithm/dataminmaxgl.cpp
```diff
@@ -0,0 +1,436 @@
...
+    , volumeMinMax_{{{ShaderType::Compute, "compute/volumeminmax.comp"}}, Shader::Build::No}
+    , layerMinMax_{{{ShaderType::Compute, "compute/layerminmax.comp"}}, Shader::Build::No}
+    , linearMinMax_{{{ShaderType::Compute, "compute/linearminmax.comp"}}, Shader::Build::No}
+    , bufferMinMax_{{{ShaderType::Compute, "compute/bufferminmax.comp"}}, Shader::Build::No} {}
+
+std::pair<dvec4, dvec4> DataMinMaxGL::minMax(const Volume* volume) {
+    if (!gpuSupport_) {
+        return util::volumeMinMax(volume);
+    }
+    return minMax(volume->getRepresentation<VolumeGL>());
```
petersteneteg: One could also consider what reps are currently available here?

include/inviwo/core/datastructures/datamapper.h
```diff
@@ -51,15 +51,37 @@ class DataFormatBase;
  */
 class IVW_CORE_API DataMapper {
 public:
+    /**
+     * Signed fixed-point integers are normalized using either an asymmetric range
+     * [-2^(b-1), 2^(b-1) - 1] or a symmetric range [-2^(b-1) + 1, 2^(b-1) - 1], where b
+     * corresponds to the bit depth of the fixed-point integer.
+     *
+     * From version 4.2 and above, OpenGL uses a symmetric range to convert between normalized
```
ResearchDaniel, line 59: Should/can the setting be retrieved from the application when one knows the OpenGL version?

modules/basegl/glsl/compute/bufferminmax.comp
```diff
@@ -0,0 +1,90 @@
...
+#endif
+
+layout (std430, binding=0) readonly buffer SourceBufferObject {
+    BUFFER_DATATYPE sourceBuffer[];
+};
+
+layout (std430, binding=1) buffer BufferObject {
```
ResearchDaniel:
```suggestion
layout (std430, binding=1)  restrict writeonly buffer BufferObject {
```

petersteneteg (dismissed):
Nice: looking really good!

modules/basegl/glsl/compute/layerminmax.comp
```diff
@@ -0,0 +1,97 @@
...
+#  define GetSize(gimage) imageSize(gimage)
+#endif  // USE_IMAGE_SAMPLER
+
+#if defined USE_IMAGE_SAMPLER
+uniform IMAGE_SOURCE sourceImage;
+#else
+uniform layout(binding=0, IMAGE_FORMAT) readonly IMAGE_SOURCE sourceImage;
+#endif  // USE_IMAGE_SAMPLER
+
+layout (std430, binding=1) buffer BufferObject {
```
ResearchDaniel:
```suggestion
layout (std430, binding=1)  restrict writeonly buffer BufferObject {
```

modules/basegl/glsl/compute/linearminmax.comp
ResearchDaniel, line 1: linear?

Is it not more about being inplace?

modules/basegl/src/algorithm/dataminmaxgl.cpp
ResearchDaniel: And potentially their size. Small ones will probably be faster to do on the CPU

ResearchDaniel (changes requested):
Nice! 
Some comments

modules/basegl/glsl/compute/bufferminmax.comp
petersteneteg: For a macro it should rather be all caps...

modules/basegl/glsl/compute/bufferminmax.comp
martinfalk: Yes, we can assume that. Compute shaders require OpenGL 4.3 or an extension, plus this solution requires additional extensions.
see https://www.khronos.org/opengl/wiki/Compute_Shader

modules/basegl/CMakeLists.txt
martinfalk: The separate category/folder for compute shaders was on purpose.

modules/basegl/src/algorithm/dataminmaxgl.cpp
martinfalk: @ResearchDaniel even when one has to download them first?

@petersteneteg Sure. That would be part of a future PR for example with minmax providers. Here, I want to use the GL version explicitly since you have to keep a `DataMinMaxGL` instance around anyways.

ResearchDaniel: > @ResearchDaniel even when one has to download them first?

No, I don't think so in that case

modules/basegl/glsl/compute/layerminmax.comp
martinfalk: no, it cannot.
> barrier may only be placed inside the function main()
> [OpenGL Reference: barrier](https://registry.khronos.org/OpenGL-Refpages/gl4/html/barrier.xhtml)

include/inviwo/core/datastructures/datamapper.h
martinfalk: See addition in OpenGLCapabilities.

https://github.com/inviwo/inviwo/blob/1315b02b2ed92bc21bf48310e6609f748e2272eb/modules/opengl/include/modules/opengl/openglcapabilities.h#L96-L104

modules/basegl/glsl/compute/linearminmax.comp
martinfalk: Linear as in linear reduction of a buffer instead of a 2D or 3D texture/image. And it is not generic as is bufferminmax.comp.

include/inviwo/core/datastructures/datamapper.h
martinfalk: The shaderutils responsible for data range scaling and offset use this setting to use the correct version in connection with the sign_normalized GL formats.

petersteneteg (approved):

ResearchDaniel (approved):

PR #1640, opened 2025-01-17
Base: FileCache

petersteneteg:
Fixes #...

Changes proposed in this PR:
 * ...
 * ...
 * ...

This has been tested on: ...

> 


modules/base/include/modules/base/processors/filecache.h
```diff
@@ -0,0 +1,254 @@
...
+    DirectoryProperty cacheDir_;
+    OptionProperty<FileExtension> extensions_;
+
+    const DataReaderFactory& rf;
+    const DataWriterFactory& wf;
+    struct Cache {
+        std::string key;
+        std::filesystem::path file;
+    };
+    std::pmr::string xml;
```
martinfalk: naming convention?
```suggestion
    const DataReaderFactory& rf_;
    const DataWriterFactory& wf_;
    struct Cache {
        std::string key;
        std::filesystem::path file;
    };
    std::pmr::string xml_;
```

modules/base/src/processors/filecache.cpp
```diff
@@ -0,0 +1,94 @@
...
+    std::pmr::unordered_set<Processor*> state(&mbr);
+    util::traverseNetwork<util::TraversalDirection::Up, util::VisitPattern::Post>(
+        state, processor, [&processors](Processor* p) { processors.push_back(p); });
+
+    // ensure a consistent order
+    std::ranges::sort(processors, std::less<>{},
+                      [](const Processor* p) { return p->getIdentifier(); });
+
+    Serializer s{"", SerializeConstants::InviwoWorkspace, &mbr};
+    s.setWorkspaceSaveMode(WorkspaceSaveMode::Undo);  // don't same any "metadata"
```
martinfalk, line 54:
```suggestion
    s.setWorkspaceSaveMode(WorkspaceSaveMode::Undo);  // don't save any "metadata"
```

martinfalk (changes requested):

CHANGELOG.md
```diff
@@ -1,5 +1,28 @@
...
+ * `DataFrame File Cache`
+
+They can be used to "cache" result from step within the network. Just drop a File Cache processor on a connection, and it will capture the state of all the processors above it in a hash and then save it's data to disk.
+If the same state happens at a later point the processor will load the data from disk, and the potentially expensive computation above can be avoided. 
+
+The the screenshot below show an example where we have cached the results of an "expensive"  `Volume Creator`. 
+Here we have reloaded the network and can note that the `Volume Creator` have not been run, since the `File Cache` was able to provide the data.
+![File cache example](resources/changelog/filecache.png)
+
+For the File Cache processor to work you have to provide an existing directory to store caches in. The processor will never remove any caches, so you might want to do that manually periodically.
```
martinfalk:
```suggestion
They can be used to "cache" results from steps within the network. Just drop a File Cache processor onto a connection, and it will capture the state of all the processors above it in a hash and then save its data to disk.
If the same state happens at a later point, the processor will load the data from disk, and the potentially expensive computation above can be avoided. 

The screenshot below shows an example where we have cached the results of an "expensive"  `Volume Creator`. 
Here we have reloaded the network and can note that the `Volume Creator` has not been run since the `File Cache` was able to provide the data.
![File cache example](resources/changelog/filecache.png)

For the File Cache processor to work you have to provide an existing directory to store caches in. The processor will never remove any caches, so you might want to do that manually periodically. 
```

martinfalk (changes requested):

martinfalk (approved):

PR #1647, opened 2025-02-06
Core: Make TFLookupTable::getRepresentation const to be consistent with Data::getRepresentation

ResearchDaniel:
Enables use of TFLookupTable in an inport.

### DIFF
--- a/include/inviwo/core/datastructures/tflookuptable.h
+++ b/include/inviwo/core/datastructures/tflookuptable.h
@@ -55,7 +55,7 @@ public:
     void setTransferFunction(TransferFunction& tf);
 
     template <typename T>
-    const T* getRepresentation() {
+    const T* getRepresentation() const {
         if (invalid_) {
             calc();
         }
@@ -70,13 +70,14 @@ protected:
     virtual void onTFMaskChanged(const TFPrimitiveSet& set, dvec2 mask) override;
 
 private:
-    void calc();
+    void calc() const;
 
     TransferFunction* tf_;  // Should not be null
     size_t size_;
-    bool invalid_;
-    std::shared_ptr<LayerRAMPrecision<vec4>> repr_;
-    std::unique_ptr<Layer> data_;
+    // Mutable due to modification in const context by getRepresentation()/calc()
+    mutable bool invalid_;
+    mutable std::shared_ptr<LayerRAMPrecision<vec4>> repr_;
+    mutable std::unique_ptr<Layer> data_;
 };
 
 }  // namespace inviwo
--- a/include/inviwo/core/properties/transferfunctionproperty.h
+++ b/include/inviwo/core/properties/transferfunctionproperty.h
@@ -241,7 +241,7 @@ public:
     const TFData& data() const { return data_; }
 
     template <typename T>
-    const T* getRepresentation() {
+    const T* getRepresentation() const {
         return lookup_.getRepresentation<T>();
     }
     size_t getLookUpTableSize() const { return lookup_.getSize(); }
--- a/src/core/datastructures/tflookuptable.cpp
+++ b/src/core/datastructures/tflookuptable.cpp
@@ -54,7 +54,7 @@ void TFLookupTable::setTransferFunction(TransferFunction& tf) {
     }
 }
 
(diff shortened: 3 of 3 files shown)


include/inviwo/core/datastructures/tflookuptable.h
```diff
@@ -55,9 +55,9 @@ class IVW_CORE_API TFLookupTable : public TFPrimitiveSetObserver {
     void setTransferFunction(TransferFunction& tf);
 
     template <typename T>
-    const T* getRepresentation() {
+    const T* getRepresentation() const {
         if (invalid_) {
-            calc();
+            const_cast<TFLookupTable*>(this)->calc();
```
martinfalk: `const_cast` should be avoided according to the C++ Core Guidelines.  See [es50-dont-cast-away-const](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es50-dont-cast-away-const). As outlined in one of the examples in the core guidelines ("Sometimes, “cast away const” is to allow the updating of some transient information of an otherwise immutable object. ..."), a potential alternative would be using `mutable` for the class members that need to be updated. Or use indirection.

martinfalk: Since you already use `mutable` simply mark the `calc()` function const as well and remove the const_cast.

martinfalk (changes requested):

include/inviwo/core/datastructures/tflookuptable.h
ResearchDaniel: Good catch

petersteneteg:
Why can't you just create a local TFLookupTable? Or a local TFProperty? Why are you sharing that over ports? 

In principle any use of mutable to avoid const also has to be protected by a mutex to avoid races.

petersteneteg:
The main reason it is in it's own class is to avoid having to what you just did here...

petersteneteg:
Data::getRepresentation uses mutexs. I still don't understand why we need this?

include/inviwo/core/datastructures/tflookuptable.h
```diff
@@ -70,13 +70,14 @@ class IVW_CORE_API TFLookupTable : public TFPrimitiveSetObserver {
...
-    void calc();
+    void calc() const;
 
     TransferFunction* tf_;  // Should not be null
     size_t size_;
-    bool invalid_;
-    std::shared_ptr<LayerRAMPrecision<vec4>> repr_;
-    std::unique_ptr<Layer> data_;
+    // Mutable due to modification in const context by getRepresentation()/calc()
+    mutable bool invalid_;
```
petersteneteg, line 78: Any mutable variable has to be protected by a mutex...

petersteneteg (changes requested):
Can't this use be avoided on from the caller?

ResearchDaniel:
> Why can't you just create a local TFLookupTable? Or a local TFProperty? Why are you sharing that over ports?

The implementation of the processor was done before the TF was refactored to exclude the data table.
Two main reasons for using ports for the TF in this case:
- Avoid recomputing the TF table
- Making the required TF input explicit (not possible with property linkage)

A local TFLookupTable would resolve the second point at least and the first point is not crucial in this case. I think that the workspace would not need to be updated either, which is convenient.

ResearchDaniel:
@petersteneteg On second thought it is not possible to do a local TFLookupTable as it does not accept a const Transerfunction as input

petersteneteg:
I you are taking the TF as input, you don't need to use the TFLookupTable at all. That is needed for observer the TF when used as a property. Here the property way would be to take a TF as input on a port, and just compute a local "TFLayer" using that input. No need for a caching and observing TFLookupTable.

You can then also decide if you want to take the TF as input or a "TFLayer".

petersteneteg:
the only thing you need is
```c++
tf_->interpolateAndStoreColors(repr_->getView());
```

petersteneteg:
@ResearchDaniel Status?

ResearchDaniel:
Have done the suggested workaround, but not tested. Aim to close the PR if it works.

PR #1650, opened 2025-02-18
Feature/ordered json

martinfalk:
Fixes
* converting from JSON to DataFrame no longer quotes strings

Changes proposed in this PR:
 * one common header for including nlohman/json and `using json = ::nlohman::json;` ([modules/json/include/modules/json/json.h](https://github.com/inviwo/inviwo/blob/8d83e71505b8a8b70c52bd37df4a065138d81540/modules/json/include/modules/json/json.h))
 * reworked DataFrame to JSON conversion to be more in line with Pandas `pandas.DataFrame.to_json(order='split')` (see [pandas docs](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_json.html))

Example JSON output:
```json
{
    "columns": [
        "index",
        "sepalLength",
        "petalLength",
        "petalWidth",
        "species"
    ],
    "data": [
        [
            5.1,
            1.4,
            0.2,
            "setosa"
        ],
        [
            4.9,
            1.4,
            0.2,
            "setosa"
        ]
    ],
    "index": [
        0,
        1
    ],
    "types": [
        "FLOAT64",
        "FLOAT64",
        "FLOAT64",
        "CATEGORICAL"
    ]
}
```


ResearchDaniel (changes requested):
By switching to ordered json we also move away from how json works in most other places. I think that we should only use this for parts where it is necessary and we are in full control of the serialization/deserialization. An alternative is to add the order in the serialization/deserialization step.

For ports/properties/networkcefsynchronizer there is no need to use the ordered json?

martinfalk:
> By switching to ordered json we also move away from how json works in most other places. I think that we should only use this for parts where it is necessary and we are in full control of the serialization/deserialization. An alternative is to add the order in the serialization/deserialization step.

You can always assign `nlohmann::json` to `nlohmann::ordered_json` and vice versa. However, loosing the given order in the process. "how json works in most other places" is a weird phrasing. Some browsers with the V8 engine might use alphanumerical ordering for object keys, but there is no guarantee (least from the specs). How would you add the order in the serialization/deserialization when working with a DataFrame? At the moment a DataFrame is converted to a plain JavaScript array. Thus, the column ordering is no longer guaranteed when converting to/from JSON. Changing that to an object containing an array will break existing code and web interfaces whereas switching to `ordered_json` does not.
For example
```json
[{"sepalLength":5.1,"petalLength":1.4,"petalWidth":0.2,"species":"setosa"},{"sepalLength":4.9,"petalLength":1.4,"petalWidth":0.2,"species":"setosa"}]
```
Another potential issue is the serialization and saving stuff to JSON in connection with the caching processors.

> For ports/properties/networkcefsynchronizer there is no need to use the ordered json?

This we can discuss. For now I aimed for consistency (see first part of my comment).

petersteneteg:
The main issue is that our conversion from data frame to json is not valid. Ideal solution would fix that, and maintain standard json semantics. What is the main problem with changing our dataframe to/from json conversions?

ResearchDaniel (dismissed):

modules/dataframe/include/inviwo/dataframe/io/json/dataframepropertyjsonconverter.h
```diff
@@ -31,7 +31,7 @@
 
 #include <inviwo/dataframe/dataframemoduledefine.h>  // for IVW_MODULE_DATAFRAME_API
 
-#include <nlohmann/json.hpp>  // for json
+#include <nlohmann/json.hpp>  // for ordered_json
```
petersteneteg: If you are changing this everywhere an we put this include and the new using line in it's own inviwo header, so we have it consistent everywhere... 
We don't want to end up pulling a headers with 
using json = nlohmann::json;
and 
using json = ::nlohmann::ordered_json;
at the same time....

modules/dataframe/src/jsondataframeconversion.cpp
```diff
@@ -93,7 +93,7 @@ void addDataFrameColumnHelper(json::value_t valueType, std::string header, DataF
             df.addColumn<uint32_t>(header, 0u);
             break;
         case json::value_t::number_float:  ///< number value (floating-point)
-            df.addColumn<float>(header, 0u);
+            df.addColumn<double>(header, 0u);
```
petersteneteg: Just not that this change might break cases where we later use the column in a plot for example.

ResearchDaniel:
Ok, I guess it is a good solution that will keep compatibility. "how json works in most places" refers to the specs that as you say do not guarantee order.

modules/dataframe/src/jsondataframeconversion.cpp
```diff
@@ -167,59 +222,8 @@ void from_json(const json& j, DataFrame& df) {
...
-                    "Value was discarded by the the parser callback function");
-                break;
-            case json::value_t::binary:
-                // Not supported
-                throw JSONConversionException("Binary elements is unsupported");
-                break;
-        }
-    }
+    // Extract header names and column types from first object/row
+    detail::extractColumnsFromRow(j, df);
```
petersteneteg: Is there any external formats that we need to take into account for this? or could we just change the json format to properly encode the dataframe? Are there any external example of this? like can one do a panda dataframe to json? 
https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_json.html
might be good to be compatible with one of those format? Is there a js dataframe library? might consider it's format?

petersteneteg (changes requested):

petersteneteg:
https://github.com/hosseinmoein/DataFrame also supports JSON

modules/dataframe/src/jsondataframeconversion.cpp
martinfalk: Will investigate.

martinfalk:
> What is the main problem with changing our dataframe to/from json conversions?

Mostly backward compatibility with existing code.

martinfalk:
> https://github.com/hosseinmoein/DataFrame also supports JSON

Maybe a deal breaker:
```
// To access a column, you must know its name (or index) and its type. In case of a "standard" DataFrame (not a view),
// the columns are returned as a reference to a std::vector of type of that column.
```
https://github.com/hosseinmoein/DataFrame/blob/88610283508948e29f10d56d24e7eaa40e227364/examples/hello_world.cc#L117-L124

modules/dataframe/include/inviwo/dataframe/io/json/dataframepropertyjsonconverter.h
martinfalk: Good suggestion.

modules/dataframe/src/jsondataframeconversion.cpp
martinfalk: Sure, we can also agree on a "common" format for floating points. The `CSVSource`, for example, has an option for double precision, which defaults to single precision.

petersteneteg:
about https://github.com/hosseinmoein/DataFrame I was not suggesting that we switch to that library, only that we look at their json format as a reference. Although using an external DataFrame library might also be one path to take. I think Apache Arrow might be of relevance for that.

martinfalk:
Updated DataFrame to JSON conversion.

modules/dataframe/src/jsondataframeconversion.cpp
```diff
@@ -106,18 +106,103 @@ void addDataFrameColumnHelper(json::value_t valueType, std::string header, DataF
...
+                ->dispatch<void, dispatching::filter::Scalars>([&](auto buffer) {
+                    using T = util::PrecisionValueType<decltype(buffer)>;
+                    auto transform = std::views::transform([colIndex](auto& row) {
+                        if (row[colIndex - 1].is_null()) {
+                            return std::numeric_limits<T>::quiet_NaN();
+                        } else {
+                            return row[colIndex - 1].template get<T>();
+                        }
+                    });
+                    buffer->getDataContainer() = data | transform | std::ranges::to<std::vector>();
```
petersteneteg, line 190: That is neat :)

petersteneteg (dismissed):
Nice! That seems a lot better, What about the webbrowser examples / tests?

petersteneteg (approved):

PR #1657, opened 2025-03-05
Added curl to support downloading files in source processors

petersteneteg:
Fixes #...

Changes proposed in this PR:
 * ...
 * ...
 * ...

This has been tested on: ...

> 

### DIFF
--- a/ext/CMakeLists.txt
+++ b/ext/CMakeLists.txt
@@ -151,6 +151,14 @@ ivw_vcpkg_install(llnl-units MODULE Core EXT)
 find_package(md4c CONFIG REQUIRED)
 ivw_vcpkg_install(md4c MODULE Core EXT)
 
+# FastFloat
+find_package(FastFloat CONFIG REQUIRED)
+ivw_vcpkg_install(fast-float MODULE Core EXT)
+
+# curlcpp
+find_package(curlcpp CONFIG REQUIRED)
+ivw_vcpkg_install(curlcpp MODULE Core EXT)
+
 if(WIN32) # Stackwalker for windows for stack traces in the log
     find_package(stackwalker CONFIG REQUIRED)
     ivw_vcpkg_install(stackwalker MODULE Core EXT)
--- a/include/inviwo/core/io/curlutils.h
+++ b/include/inviwo/core/io/curlutils.h
@@ -35,8 +35,7 @@
 
 namespace inviwo::net {
 
-enum class ResponseCode : long {  // NOLINT(performance-enum-size): want conversion from curl
-                                  // without narrowing
+enum class ResponseCode : long {  // NOLINT: want conversion from curl without narrowing
     Continue = 100,
     Switching_Protocols = 101,
     Processing = 102,
@@ -126,7 +125,6 @@ enum class ResponseCode : long {  // NOLINT(performance-enum-size): want convers
 
 IVW_CORE_API std::string_view format_as(ResponseCode rs);
 IVW_CORE_API std::string_view description(ResponseCode rs);
-
 
 IVW_CORE_API bool isUrl(const std::filesystem::path& path);
 IVW_CORE_API std::filesystem::path downloadAndCacheIfUrl(const std::filesystem::path& url);
--- a/include/inviwo/core/io/curlutils.h
+++ b/include/inviwo/core/io/curlutils.h
@@ -35,7 +35,7 @@
 
 namespace inviwo::net {
 
-enum class ResponseCode {
+enum class ResponseCode : long {
     Continue = 100,
     Switching_Protocols = 101,
     Processing = 102,
--- a/include/inviwo/core/io/curlutils.h
+++ b/include/inviwo/core/io/curlutils.h
@@ -1,0 +1,143 @@
+/*********************************************************************************
+ *
+ * Inviwo - Interactive Visualization Workshop
+ *
+ * Copyright (c) 2025 Inviwo Foundation
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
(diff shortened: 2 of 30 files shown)


include/inviwo/core/io/curlutils.h
```diff
@@ -0,0 +1,143 @@
...
+    A_Timeout_Occurred = 524,
+    SSL_Handshake_Failed = 525,
+    Invalid_SSL_Certificate = 526,
+    Railgun_Listener_to_Origin_Error = 527,
+    Origin_DNS_Error = 530,
+    Network_Read_Timeout_Error = 598,
+};
+
+template <typename Path>
+bool isUrl(const Path& path) {
```
ResearchDaniel: ftp should also be valid?

include/inviwo/core/util/pathtype.h
```diff
@@ -42,6 +42,7 @@ enum class PathType {
     Resources,          // /resources
     TransferFunctions,  // /data/transferfunctions
     Settings,           // path to the current users settings
+    Cache,              // cache folder in the current users settings
```
ResearchDaniel, line 47: I think that the fetched data should be in the downloads folder, at least on windows. This should in that case be “Downloads” instead
```suggestion
    Downloads,              // current user’s download directory 
```

modules/base/src/io/ivfvolumereader.cpp
```diff
@@ -64,9 +64,11 @@ IvfVolumeReader::IvfVolumeReader() : DataReaderType<Volume>() {
 
 IvfVolumeReader* IvfVolumeReader::clone() const { return new IvfVolumeReader(*this); }
 
-std::shared_ptr<Volume> IvfVolumeReader::readData(const std::filesystem::path& filePath) {
+std::shared_ptr<Volume> IvfVolumeReader::readData(const std::filesystem::path& file) {
+    const auto filePath = downloadAndCacheIfUrl(file);
+
     checkExists(filePath);
-    const auto fileDirectory = filePath.parent_path();
+    const auto fileDirectory = file.parent_path();
```
ResearchDaniel:
```suggestion
    const auto fileDirectory = localPath.parent_path();
```

ext/CMakeLists.txt
```diff
@@ -151,6 +151,14 @@ ivw_vcpkg_install(llnl-units MODULE Core EXT)
 find_package(md4c CONFIG REQUIRED)
 ivw_vcpkg_install(md4c MODULE Core EXT)
 
+# FastFloat
+find_package(FastFloat CONFIG REQUIRED)
```
ResearchDaniel, line 155: Needed for?

modules/base/src/io/ivfvolumereader.cpp
```diff
@@ -64,9 +64,11 @@ IvfVolumeReader::IvfVolumeReader() : DataReaderType<Volume>() {
 
 IvfVolumeReader* IvfVolumeReader::clone() const { return new IvfVolumeReader(*this); }
 
-std::shared_ptr<Volume> IvfVolumeReader::readData(const std::filesystem::path& filePath) {
+std::shared_ptr<Volume> IvfVolumeReader::readData(const std::filesystem::path& file) {
```
ResearchDaniel:
```suggestion
std::shared_ptr<Volume> IvfVolumeReader::readData(const std::filesystem::path& path) {
```

modules/base/src/io/ivfvolumereader.cpp
```diff
@@ -64,9 +64,11 @@ IvfVolumeReader::IvfVolumeReader() : DataReaderType<Volume>() {
...
+    const auto filePath = downloadAndCacheIfUrl(file);
```
ResearchDaniel:
```suggestion
    const auto filePath = downloadAndCacheIfUrl(path);
```

ResearchDaniel: A
```suggestion
    const auto localPath = downloadAndCacheIfUrl(path);
```

src/core/io/curlutils.cpp
```diff
@@ -0,0 +1,512 @@
...
+    return "Unknown response code";
+}
+
+std::filesystem::path downloadAndCacheIfUrl(const std::filesystem::path& url) {
+    if (!net::isUrl(url)) {
+        return url;
+    }
+
+    auto cacheDir = filesystem::getPath(PathType::Cache);
+    std::filesystem::create_directories(cacheDir);
```
ResearchDaniel, line 474: Would make sense to use the downloads directory instead. Aligns better with how most apps handle downloads.

ResearchDaniel (changes requested):
Nice! 
As commented I think that the downloads directory would make more sense instead of a folder in the user data directory. The user data folder is hidden away and difficult for a user to find. As the data could be large it would be nice to not hide it too much

include/inviwo/core/io/curlutils.h
petersteneteg: I guess... when was the last time you ever came across one?

include/inviwo/core/util/pathtype.h
petersteneteg: A application should **never** write in to a users files, be in downloads, documents etc, without explicitly asking about it each and every time. 

Caching like this is usually done to some "hidden" directly, consider how chrome does it caching, or vcpkg. or basically any other software. 

I write in the log that the file was cached and to where so, it is not impossible for a user to find. And if we wan we can add some option to clear the cache.

modules/base/src/io/ivfvolumereader.cpp
petersteneteg: No that should still be the remove path otherwise download of like dat and raw files will not work

ext/CMakeLists.txt
petersteneteg: I had forgotten to add it here previously, it is in src/core/CMakeLists.txt, it provides a from:chars impl when we don't have it in the stdlib.

src/core/io/curlutils.cpp
petersteneteg: no, see above

include/inviwo/core/util/pathtype.h
ResearchDaniel: I thought that it was more like downloading something from an application, but I don't have strong opinions about it.

ResearchDaniel (dismissed):

PR #1659, opened 2025-03-11
Feature/qt numberwidget

martinfalk:
Fixes
 * several issues related to the current `NumberLineEdit` like fixed decimals, out-of-sync state between UI and property, ...

Changes proposed in this PR:
 * new, simpler widget `NumberWidget` for editing numbers based on QLineEdit that is inspired by Blender's FloatWidgets
 * when not in focus, the values can be dragged or incremented/decremented
 * the widget also has prefix and suffix support
 * the input is validated on-the-fly and flagged when invalid 
image
--- 

*CameraProperty with updated number widgets*
image

--- 

*Value dragging & editing*
<video src='https://github.com/user-attachments/assets/cb14317f-2a7e-480e-beaa-fa7739702ea1' width=432/>


petersteneteg (dismissed):
Nice work!

petersteneteg (dismissed):

petersteneteg (approved):

PR #1664, opened 2025-03-20
Feature/numberwidget

martinfalk:
Fixes
 * increment issue for integral values with min increment = 1
 * underflow with negative increment for unsigned values


petersteneteg (dismissed):
Might look at the conversion warnings...

petersteneteg (approved):

PR #1667, opened 2025-03-21
Feature/ivf format compression

martinfalk:
Changes proposed in this PR:
 * added Serializer constructor with version number, thus not using the `InviwoWorkspaceVersion` number
 * added support to ivf (Inviwo Volume Format) and dat/raw readers for zipped data using bxzstr


include/inviwo/core/io/inviwofileformattypes.h
```diff
@@ -0,0 +1,47 @@
...
+
+#include <inviwo/core/common/inviwocoredefine.h>
+
+#include <string_view>
+#include <fmt/base.h>
+
+namespace inviwo::iff {
+
+enum class ByteOrder : int { LittleEndian, BigEndian };
+enum class Compression : int { Off, Enabled };
```
ResearchDaniel:
```suggestion
enum class Compress: int { Disabled, Enabled };
```

include/inviwo/core/io/inviwofileformattypes.h
```diff
@@ -0,0 +1,47 @@
...
+#pragma once
+
+#include <inviwo/core/common/inviwocoredefine.h>
+
+#include <string_view>
+#include <fmt/base.h>
+
+namespace inviwo::iff {
+
+enum class ByteOrder : int { LittleEndian, BigEndian };
```
ResearchDaniel:
```suggestion
enum class Endianness : int { LittleEndian, BigEndian };
```

include/inviwo/core/io/rawvolumeramloader.h
```diff
@@ -58,7 +60,8 @@ class IVW_CORE_API RawVolumeRAMLoader : public DiskRepresentationLoader<VolumeRe
 private:
     std::filesystem::path rawFile_;
     size_t offset_;
-    bool littleEndian_;
+    iff::ByteOrder byteOrder_;
```
ResearchDaniel:
```suggestion
    iff::Endianness endianness_;
```

include/inviwo/core/io/volumedatareaderdialog.h
```diff
@@ -48,16 +49,18 @@ class IVW_CORE_API VolumeDataReaderDialog : public Dialog {
     virtual const DataFormatBase* getFormat() const = 0;
     virtual uvec3 getDimensions() const = 0;
     virtual dvec3 getSpacing() const = 0;
-    virtual bool getEndianess() const = 0;
+    virtual iff::ByteOrder getByteOrder() const = 0;
```
ResearchDaniel:
```suggestion
    virtual iff::Endianness getEndianness() const = 0;
```

include/inviwo/core/io/volumedatareaderdialog.h
```diff
@@ -48,16 +49,18 @@ class IVW_CORE_API VolumeDataReaderDialog : public Dialog {
...
     virtual DataMapper getDataMapper() const = 0;
     virtual size_t getByteOffset() const = 0;
+    virtual iff::Compression getCompression() const = 0;
 
     virtual void setFormat(const DataFormatBase* format) = 0;
     virtual void setDimensions(uvec3 dim) = 0;
     virtual void setSpacing(dvec3 spacing) = 0;
-    virtual void setEndianess(bool endian) = 0;
+    virtual void setByteOrder(iff::ByteOrder byteOrder) = 0;
```
ResearchDaniel:
```suggestion
    virtual void setEndianness(iff::Endianness endianness) = 0;
```

modules/base/src/io/datvolumesequencereader.cpp
```diff
@@ -122,14 +124,25 @@ State parseDatFile(std::ifstream& f, const std::filesystem::path& filePath) {
...
+                 state.byteOrder = iff::ByteOrder::LittleEndian;
              } else {
                  ss.setstate(std::ios_base::failbit);
              }
          }},
         {"byteoffset", [](State& state, std::stringstream& ss) { ss >> state.byteOffset; }},
+        {"compressed",
+         [](State& state, std::stringstream& ss) {
+             const auto val = toLower(ss.str());
+             if (val == "enabled") {
```
ResearchDaniel: Would probably make more sense to have Yes/No

modules/base/src/io/datvolumewriter.cpp
```diff
@@ -93,24 +95,36 @@ void writeDatVolume(const Volume& data, const std::filesystem::path& filePath,
     glm::mat4 wtm = glm::transpose(data.getWorldMatrix());
 
     auto print = util::overloaded{
-        [&](std::string_view key, const std::string& val) { fmt::print(ss, "{}: {}\n", key, val); },
-        [&](std::string_view key, std::string_view val) { fmt::print(ss, "{}: {}\n", key, val); },
-        [&](std::string_view key, InterpolationType val) { fmt::print(ss, "{}: {}\n", key, val); },
-        [&](std::string_view key, const SwizzleMask& mask) {
+        [&ss](std::string_view key, const std::string& val) {
```
ResearchDaniel: Is `string&` needed when string_view is there?

modules/base/src/io/ivfvolumereader.cpp
```diff
@@ -56,8 +59,92 @@
...
+            }
+            root->InsertBeforeChild(metaDataMap, byteOrder);
+            metaDataMap->RemoveChild(littleEndianNode->Parent());
+        }
+    }
+}
+
+bool Converter::convert(TxElement* root) {
+    if (root->Value() != InviwoVolume) {
+        // ignore version number if the root node is not already an InviwoVolume, but for example an
```
ResearchDaniel, line 105:
```suggestion
        // XML-file is not an ivf-file. It could for example be an Inviwo workspace file, which has its own version element. In this case we ignore the version number.
```

modules/base/src/io/ivfvolumereader.cpp
```diff
@@ -56,8 +59,92 @@
...
+        // InviwoWorkspace
```
ResearchDaniel, line 106:
```suggestion
```

modules/base/src/io/ivfvolumereader.cpp
```diff
@@ -56,8 +59,92 @@
...
+        version_ = 1;
+    }
+
+    switch (version_) {
+        case 0:
+        case 1: {
+            auto metaDataMap = xml::getElement(root, "MetaDataMap");
+
+            if (metaDataMap) {
+                if (auto littleEndianNode =
```
ResearchDaniel: Use updateByteOrder?

modules/base/src/io/ivfvolumewriter.cpp
```diff
@@ -83,10 +87,16 @@ void writeIvfVolume(const Volume& data, const std::filesystem::path& filePath,
...
+    Serializer s{filePath, InviwoVolume, InviwoVolumeVersion, &mbr};
+    {
+        const auto nodeSwitch = s.switchToNewNode("RawFile");
+        s.serialize("content", fmt::format("{}.raw", fileName), SerializationTarget::Attribute);
+        data.getMetaDataMap()->serialize(s);
+    }
     s.serialize("Format", vr->getDataFormatString());
     s.serialize("ByteOffset", 0u);
+    s.serialize("ByteOrder", iff::ByteOrder::LittleEndian);
+    s.serialize("Compression", iff::Compression::Off);
```
ResearchDaniel: Would be better to compress by default?

modules/qtwidgets/include/modules/qtwidgets/rawdatareaderdialogqt.h
```diff
@@ -64,16 +65,18 @@ class IVW_MODULE_QTWIDGETS_API RawDataReaderDialogQt : public VolumeDataReaderDi
     virtual const DataFormatBase* getFormat() const override;
     virtual uvec3 getDimensions() const override;
     virtual dvec3 getSpacing() const override;
-    virtual bool getEndianess() const override;
+    virtual iff::ByteOrder getByteOrder() const override;
```
ResearchDaniel:
```suggestion
    virtual iff::Endianness getEndianness() const override;
```

modules/qtwidgets/src/rawdatareaderdialogqt.cpp
```diff
@@ -198,25 +205,34 @@ RawDataReaderDialogQt::RawDataReaderDialogQt() {
...
     timeStepOffset_->setSuffix(" Byte");
     */
-    QLabel* endianessLabel = new QLabel("Endianess");
-    endianess_ = new QComboBox();
-    endianess_->addItem("Little Endian");
-    endianess_->addItem("Big Endian");
+    byteOrder_ = new QComboBox();
+    byteOrder_->addItem("Little Endian", QVariant::fromValue(iff::ByteOrder::LittleEndian));
+    byteOrder_->addItem("Big Endian", QVariant::fromValue(iff::ByteOrder::BigEndian));
+    QLabel* endianessLabel = new QLabel("Byte order");
```
ResearchDaniel: I think that here it makes sense to say byte order

modules/qtwidgets/src/rawdatareaderdialogqt.cpp
```diff
@@ -198,25 +205,34 @@ RawDataReaderDialogQt::RawDataReaderDialogQt() {
...
     readOptionsLayout->addWidget(timeStepOffsetLabel, 1, 0);
     readOptionsLayout->addWidget(timeStepOffset_, 1, 1);
     */
-    readOptionsLayout->addWidget(endianessLabel, 2, 0);
-    readOptionsLayout->addWidget(endianess_, 2, 1);
+    readOptionsLayout->addWidget(endianessLabel, 1, 0);
+    readOptionsLayout->addWidget(byteOrder_, 1, 1);
+
+    useCompression_ = new QCheckBox{};
+    QLabel* useCompressionLabel = new QLabel("Compressed data (bzip, gz, lzma)");
```
ResearchDaniel: No need to be so specific with the compression algorithm. Suggest to simply specify zip. 
```suggestion
    QLabel* useCompressionLabel = new QLabel("Compress (zip)");
```

include/inviwo/core/io/inviwofileformattypes.h
ResearchDaniel: From a coder-perspective I think that it becomes more consistent and easieto use Endianness as the enum values are named like that. In the user interface I think it is good to name it Byte order as it is now

ResearchDaniel (changes requested):

include/inviwo/core/io/inviwofileformattypes.h
```diff
@@ -0,0 +1,47 @@
...
+ *
+ *********************************************************************************/
+#pragma once
+
+#include <inviwo/core/common/inviwocoredefine.h>
+
+#include <string_view>
+#include <fmt/base.h>
+
+namespace inviwo::iff {
```
petersteneteg: Do we have more then 1 ByteOrder enum in inviwo? Skip the extra namespace.

src/core/io/inviwofileformattypes.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+std::string_view enumToStr(ByteOrder byteOrder) {
+    switch (byteOrder) {
+        case ByteOrder::LittleEndian:
+            return "LittleEndian";
+        case ByteOrder::BigEndian:
+            return "BigEndian";
+    }
+    throw Exception{SourceContext{}, "Found invalid ByteOrder enum value '{}'",
+                    std::to_underlying(byteOrder)};
```
petersteneteg: to_underlying is not supported on clang

include/inviwo/core/io/rawvolumereader.h
```diff
@@ -54,25 +54,27 @@ class IVW_CORE_API RawVolumeReader : public DataReaderType<Volume> {
     virtual RawVolumeReader* clone() const override;
     virtual ~RawVolumeReader() = default;
 
-    virtual void setParameters(const DataFormatBase* format, ivec3 dimensions, bool littleEndian,
-                               DataMapper dataMapper, size_t byteOffset = 0u);
+    virtual void setParameters(const DataFormatBase* format, ivec3 dimensions,
```
petersteneteg: Generally all these parameters should be passed to the constructor.  I can't find any callers of this function let alone anyone overriding it...

modules/base/src/io/datvolumesequencereader.cpp
petersteneteg: Why not? 
use val == format_as(Compression::Enabled)

modules/base/src/io/ivfvolumereader.cpp
petersteneteg: no that is correct

modules/base/src/io/ivfvolumereader.cpp
```diff
@@ -56,8 +59,92 @@
...
+                        xml::getElement(metaDataMap, "MetaDataItem&key=LittleEndian/MetaData")) {
+                    // move endianness to own tag
+                    TxElement byteOrder{"ByteOrder"};
+
+                    auto content = littleEndianNode->Attribute("content").value_or("1");
+                    byteOrder.SetAttribute(
+                        "content",
+                        fmt::format("{}", content == "1"
+                                              ? std::to_underlying(iff::ByteOrder::LittleEndian)
```
petersteneteg: can not use std::to_underlying

modules/base/src/io/ivfvolumereader.cpp
petersteneteg: littleEndianNode make much more sense

modules/qtwidgets/src/rawdatareaderdialogqt.cpp
petersteneteg: why you want to know that formats that can be decompressed. And zip is not even really a valid one.

petersteneteg (changes requested):

modules/base/src/io/ivfvolumewriter.cpp
martinfalk: In principle yes :-)

src/core/io/inviwofileformattypes.cpp
martinfalk: Good to know.

modules/base/src/io/datvolumewriter.cpp
martinfalk: Doesn't seem to be the case anymore. Will remove.

modules/base/src/io/ivfvolumereader.cpp
martinfalk: @ResearchDaniel 

The reasoning here is that `InviwoWorkspace` has a different version number (3 at the moment) than the `InviwoVolume`. And for backwards compatibility InviwoWorkspace is still accepted for volumes and a warning is printed by the deserializer. However, in this case we cannot rely on the wrong version number in the converter.

include/inviwo/core/io/rawvolumereader.h
martinfalk: Will remove this function. It was already part of the initial v0.9.0 release.

modules/base/src/io/datvolumesequencereader.cpp
martinfalk: Changed flag to "compression" and the values to `format_as(Compression::Disabled)` / `format_as(Compression::Enabled)` using the enum directly, that is "Disabled" and "Enabled" respectively.

include/inviwo/core/io/inviwofileformattypes.h
martinfalk: Decided to stick with ByteOrder.

modules/base/src/io/ivfvolumereader.cpp
ResearchDaniel: I see. That was not obvious :)

modules/qtwidgets/src/rawdatareaderdialogqt.cpp
ResearchDaniel: Yeah, that was my thinking too. I put zip there as I thought there may have been a special reason to specify the compression method

petersteneteg (approved):

ResearchDaniel (dismissed):

petersteneteg (dismissed):

petersteneteg (dismissed):

include/inviwo/core/io/inviwofileformattypes.h
```diff
@@ -0,0 +1,58 @@
...
+inline std::string_view format_as(Compression compression) { return enumToStr(compression); }
+
+namespace util {
+
+/**
+ * Check whether compression with zlib, that is bxz::z, is available
+ *
+ * @return true if zlib compression is enabled
+ */
+IVW_CORE_API bool isCompressionSupported();
```
petersteneteg: is this needed? the dependency is not optional?

include/inviwo/core/io/inviwofileformattypes.h
```diff
@@ -0,0 +1,58 @@
...
+
+#include <inviwo/core/common/inviwocoredefine.h>
+
+#include <string_view>
+#include <cstdint>
+
+namespace inviwo {
+
+enum class ByteOrder : std::uint8_t { LittleEndian, BigEndian };
+enum class Compression : std::uint8_t { Disabled, Enabled };
```
petersteneteg, line 39: Do we want this to also contain the type of compression?

modules/base/src/io/ivfvolumereader.cpp
```diff
@@ -56,97 +59,253 @@
...
+    return dataMap;
+}
+
+std::array<Axis, 3> deserializeAxes(Deserializer& d) {
+    std::array<Axis, 3> axes;
+    std::string tmp;
+    d.deserialize("Axis1Unit", tmp);
+    if (!tmp.empty()) {
+        axes[0].unit = units::unit_from_string(tmp);
+    }
```
petersteneteg: this begs for a function... 
does unit_from_string not work with an empty string?

src/core/CMakeLists.txt
```diff
@@ -886,6 +890,9 @@ target_include_directories(inviwo-core PUBLIC
     $<BUILD_INTERFACE:${IVW_INCLUDE_DIR}>
 )
 
+find_path(BXZSTR_INCLUDE_DIRS "bxzstr.hpp" PATH_SUFFIXES "include/bxzstr")
+target_include_directories(inviwo-core PRIVATE ${BXZSTR_INCLUDE_DIRS})
```
petersteneteg: We might want to add a SYSTEM also

petersteneteg (dismissed):
Nice very comprehensive work!

include/inviwo/core/io/inviwofileformattypes.h
martinfalk: Future PR

include/inviwo/core/io/inviwofileformattypes.h
martinfalk: removed

petersteneteg (approved):

PR #1680, opened 2025-04-10
Feature/port datasequence

martinfalk:
Changes proposed in this PR:
 * `DataSequence<Data>` outports can be connected to multi-inports of type `Data`
 * the `Surface Extraction` processor now outputs a `DataSequence<Mesh>` instead of a `std::vector<std::shared_ptr<Mesh>>`

This has been tested on: lates MSVC

### DIFF
--- a/include/inviwo/core/datastructures/datasequence.h
+++ b/include/inviwo/core/datastructures/datasequence.h
@@ -47,31 +47,32 @@ class DataSequence {
     using DataVariant = std::variant<std::shared_ptr<const Data>, std::shared_ptr<Data>>;
     std::vector<DataVariant> data_;
 
-    static constexpr auto getConstData =
-        [](const DataVariant& data) -> std::shared_ptr<const Data> {
-        return std::visit([](auto& d) -> std::shared_ptr<const Data> { return d; }, data);
+    struct ConstDataAccess {
+        std::shared_ptr<const Data> operator()(const DataVariant& data) const {
+            return std::visit([](auto& d) -> std::shared_ptr<const Data> { return d; }, data);
+        }
     };
-
-    static constexpr auto getData = [](DataVariant& data) -> std::shared_ptr<Data> {
-        if (std::holds_alternative<std::shared_ptr<Data>>(data)) {
-            return std::get<std::shared_ptr<Data>>(data);
-        } else {
-            return nullptr;
+    struct DataAccess {
+        std::shared_ptr<Data> operator()(DataVariant& data) const {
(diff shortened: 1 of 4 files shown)


ResearchDaniel (dismissed):

modules/base/src/processors/surfaceextractionprocessor.cpp
```diff
@@ -211,16 +211,17 @@ void SurfaceExtraction::process() {
...
+                jobs.push_back(changeColor(getColor(i), (*meshes_)[i]));
                 inds.push_back(i);
             }
         }
         if (!jobs.empty()) {
             dispatchMany(jobs, [this, inds](std::vector<std::shared_ptr<Mesh>> results) {
                 for (auto [i, result] : util::zip(inds, results)) {
-                    meshes_[i] = result;
+                    auto it = meshes_->begin() + i;
+                    meshes_->replace(it, result);
```
petersteneteg: It would be better to output a new DataSequence each time. It can still reuse the unchanged meshes from the old one.

petersteneteg: We should not modify old data unless we have a very good reason to...

include/inviwo/core/ports/outportiterable.h
```diff
@@ -297,6 +298,42 @@ class OutportIterableWrapper<std::vector<std::unique_ptr<T>, Alloc>> {
     bool end_;
 };
 
+// Specialization for DataSequence of data
+template <typename T>
+class OutportIterableWrapper<DataSequence<T>> {
```
petersteneteg, line 315: Nice! did not consider this when adding the DataSequence.

include/inviwo/core/datastructures/datasequence.h
```diff
@@ -125,6 +125,10 @@ class DataSequence {
 
     void erase(iterator pos) { data_.erase(pos.base()); }
     void erase(const_iterator begin, const_iterator end) { data_.erase(begin.base(), end.base()); }
+
```
petersteneteg, line 129: I wonder if the decltype(getData) above is the reason for all the warnings about internal linkage below...
The lambda might have unique type in each TU. which would be bad.
Maybe you can try and make those into a named struct with a call operator?

petersteneteg (changes requested):
Nice!! I had missed all about the multiports when refactoring the DataSequence. Do we have any other cases that still output std::vector<...>?

petersteneteg (approved):

PR #1688, opened 2025-05-06
Feature/various fixes

martinfalk:
Changes proposed in this PR:
 * various minor fixes


petersteneteg (dismissed):

modules/base/src/properties/valueaxisproperty.cpp
```diff
@@ -0,0 +1,152 @@
...
+namespace inviwo {
+
+std::string_view ValueAxisProperty::getClassIdentifier() const { return classIdentifier; }
+
+ValueAxisProperty::ValueAxisProperty(std::string_view identifier, std::string_view displayName,
+                                     bool customAxis, InvalidationLevel invalidationLevel,
+                                     PropertySemantics semantics)
+    : CompositeProperty{identifier, displayName, invalidationLevel, semantics}
+    , customAxis_{customAxis}
+    , valueName_{"valueName", "Value name", ""}
```
ResearchDaniel, line 43: Might be enough with only Name/Unit due to the composite property name?
```suggestion
    , valueName_{"valueName", "Name", ""}
```

ResearchDaniel (dismissed):
Is nifti handled by vcpkg now?

petersteneteg:
nitti has been in handled by vcpkg for some time. I just missed removing the embedded sources.

modules/base/src/properties/valueaxisproperty.cpp
martinfalk: I would stick with "Value name" to be explicit and in line with `DataRangeProperty`.

image

petersteneteg (approved):

PR #1692, opened 2025-05-12
Feature/plotcamera

petersteneteg:
https://github.com/user-attachments/assets/a415c738-d234-43c2-9f71-3a170b4e7b99

Core: Added a PlotCamera, similar to a orthographic camera but where one can change width and height independently. i.e. by ignoring the aspect ratio
Also refactors zooming, allowing independent zoom in x and y, and enabling mouse centered zoom.

### DIFF
--- a/modules/plottinggl/src/processors/orthographicaxis2d.cpp
+++ b/modules/plottinggl/src/processors/orthographicaxis2d.cpp
@@ -52,14 +52,17 @@ const ProcessorInfo& OrthographicAxis2D::getProcessorInfo() const { return proce
 OrthographicAxis2D::OrthographicAxis2D()
     : Processor{}
     , inport_{"inport"}
-    , mesh_{"mesh"}
+    , mesh_{"mesh", "A mesh to extract a World to Model matrix from"_help}
     , outport_{"outport"}
     , style_{"style", "Global Style"}
     , backgroundColor_{"backgroundColor", "Background Color",
                        util::ordinalColor(1.f, 1.f, 1.f, 1.f)}
-    , clipContent_{"clipContent", "Clip Content", true}
-    , axis1_{"axis1", "Axis 1"}
-    , axis2_{"axis2", "Axis 2", plot::AxisProperty::Orientation::Vertical}
+    , clipContent_{"clipContent", "Clip Content",
+                   "The input image will only be shown inside of the margins when enabled"_help,
+                   true}
+    , axis1_{"axis1", "Axis 1", "Horizontal axis aligned to the bottom of the view"_help}
+    , axis2_{"axis2", "Axis 2", "Vertical axis aligned to the left of the view"_help,
+             plot::AxisProperty::Orientation::Vertical}
     , margins_{"margins", "Margins", 5.0f, 5.0f, 55.0f, 65.0f}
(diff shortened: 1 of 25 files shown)


modules/plottinggl/src/processors/orthographicaxis2d.cpp
```diff
@@ -89,14 +97,34 @@ OrthographicAxis2D::OrthographicAxis2D()
...
     const auto dims = outport_.getDimensions();
-    const size2_t lowerLeft(margins_.getLeft(), margins_.getBottom());
-    const size2_t upperRight(dims.x - 1 - margins_.getRight(), dims.y - 1 - margins_.getTop());
 
     const auto padding = axisMargin_.get();
 
+    const utilgl::ClearColor clearColor{backgroundColor_.get()};
+    utilgl::activateAndClearTarget(outport_);
+    if (auto source = inport_.getData()) {
+        const auto scissors =
```
martinfalk, line 111: The scissor test is only enabled for this if statement and not the subsequent rendering. Is this intended? Maybe put in a comment that you want to clip the input image/layer to the plot area/margins.
Also, the scissor test only applies to rendered fragments, so in case `copyRepresentationsTo()` uses a pixel transfer or directly copies the textures it wont have an effect.

src/core/interaction/trackball.cpp
```diff
@@ -840,7 +865,8 @@ vec3 Trackball::getBoundedTranslation(const vec3& /*lookFrom*/, const vec3& look
     // simply disallow movements that would cause the lookTo point to
     // go out of bounds
     const auto newPos(lookTo - translation);
-    // auto clampedPos = glm::clamp(newPos, dvec3(getLookToMinValue()), dvec3(getLookToMaxValue()));
+    // auto clampedPos = glm::clamp(newPos, dvec3(getLookToMinValue()),
+    // dvec3(getLookToMaxValue()));
```
martinfalk, line 869: Remove?

src/core/interaction/trackball.cpp
```diff
@@ -853,7 +879,8 @@ vec3 Trackball::getBoundedTranslation(const vec3& /*lookFrom*/, const vec3& look
     // Clamping does not work when movement is restricted along horizontal or vertical axes.
     // else {
     //
-    //    translation = glm::clamp(translation, dvec3(-axesMinDistance), dvec3(axesMinDistance));
+    //    translation = glm::clamp(translation, dvec3(-axesMinDistance),
+    //    dvec3(axesMinDistance));
```
martinfalk, line 883: remove?

martinfalk (changes requested):
Nice addition.

src/core/interaction/trackball.cpp
petersteneteg: different PR

src/core/interaction/trackball.cpp
petersteneteg: different PR

modules/plottinggl/src/processors/orthographicaxis2d.cpp
petersteneteg: fixed

martinfalk (approved):

PR #1710, opened 2025-06-24
Feature/parallel coords blending and picking fix

martinfalk:
Fixes
 * blending and picking issues in the Parallel Coordinates Plot
 * when using additive or subtractive blend functions, these also applied to the picking buffer. Thus breaking brushing & linking.

Changes proposed in this PR:
 * use separate blend function for picking buffer attachment where available (OpenGL >4.0 or `ARB_draw_buffers_blend`)
 * added example workspace for additive blending PCP
 * regression test for PCP blend modes

---
**Result of additive blending**

**Picking buffer with incorrect, additive blending**

**Fixed picking buffer**


ResearchDaniel (commented):
Are additive and subtractive blending options needed? We could also remove them I think

martinfalk:
> Are additive and subtractive blending options needed? We could also remove them I think

I strongly support those blend modes. The results look much nicer when you have a large amount of lines. It also aids in following individual lines as in the example given above.

petersteneteg (approved):
Nice catch!

petersteneteg:
Wonder if we mess that up in other places?

martinfalk:
> Wonder if we mess that up in other places?

Had a brief look and could not find other instances using different blend functions.

PR #1728, opened 2025-09-05
Core: case insensitive find

martinfalk:
Changes proposed in this PR:
 * added case insensitive `util::iCaseFind()` and `util::iCaseContains()` for std::string_view


petersteneteg (dismissed):
could also consider using https://en.cppreference.com/w/cpp/algorithm/ranges/search.html

petersteneteg (approved):

PR #1739, opened 2025-09-23
OpactOpt: Fix linking error on Linux/Mac

arrebarritra:
I noticed there was a linking error on Linux/Mac during the the integration tests. This should fix it.


petersteneteg (approved):
Guess one could also make that static constexpr...

PR #1758, opened 2025-10-05
OpactOpt: line renderer working again

arrebarritra:
After the latest refactoring the line renderer stopped working, here is the fix


petersteneteg (approved):
Nice, good catch

PR #1777, opened 2025-10-24
Feature/misc

martinfalk:
Fixes 
 * zero increment in `LayerInformationProperty`

Changes proposed in this PR:
 * disable float normalization when computing the 2D magnitude
 * expose `swizzlemasks::defaultColor()` and `swizzlemasks::defaultData()` functions in Python


modules/vectorfieldvisualizationgl/src/processors/2d/vector2dmagnitude.cpp
```diff
@@ -59,4 +59,13 @@ LayerConfig Vector2DMagnitude::outputConfig(const Layer& input) const {
                                       .valueRange = dvec2{0.0, conservativeMax}});
 }
 
+void Vector2DMagnitude::preProcess(TextureUnitContainer&, const Layer& input, Layer&) {
```
petersteneteg: I understand that the processor is "wrong", but I don't think this is the correct solution...
And I don't think this "fix" solves it for none float data.

petersteneteg (changes requested):

modules/vectorfieldvisualizationgl/src/processors/2d/vector2dmagnitude.cpp
petersteneteg: I don't think getNormalizedTexel is appropriate here. its main purpose for tf lookups. 
We should probably add some more functions for this

Also have to think about how the result is converted "back" to what ever format it used...
What if we write to a uint16 texture? What should we put in FragData0?

martinfalk: OpenGL takes care of converting the float value in [0,1] back to the underlying range of the data type. Need to check signed types.

petersteneteg: Yes, it will convert [0, 1] to the range of the destination data type. But if you are not working with normalized data in the shader, the we need to handle that...

modules/opengl/include/modules/opengl/shader/shaderutils.h
```diff
@@ -78,11 +78,27 @@ class TemplateProperty;
...
 namespace utilgl {
+struct GLFormatConversion;
 
 /**
- * Set the shader uniforms of \p shader related to format scaling based on \p dataMapper and the
- * corresponding data format \p format  using \p name as a base name, that is `<name>.formatScaling`
+ * Set the shader uniforms of @p shader related to format conversions from normalized values to
+ * value space and to the OpenGL output range.
+ *
+ scaling based on @p dataMapper and the corresponding
```
petersteneteg:
```suggestion
* scaling based on @p dataMapper and the corresponding
```

modules/opengl/glsl/utils/conversion.glsl
```diff
@@ -0,0 +1,80 @@
...
+float mapFromValueToGLOutput(in float value, in GLFormatConversion conversion) {
+    return ((value - conversion.outputValueOffset) * conversion.outputScaling) + conversion.outputOffset;
+}
+vec2 mapFromValueToGLOutput(in vec2 value, in GLFormatConversion conversion) {
+    return ((value - conversion.outputValueOffset) * conversion.outputScaling) + conversion.outputOffset;
+}
+vec3 mapFromValueToGLOutput(in vec3 value, in GLFormatConversion conversion) {
+    return ((value - conversion.outputValueOffset) * conversion.outputScaling) + conversion.outputOffset;
+}
+vec4 mapFromValueToGLOutput(in vec4 value, in GLFormatConversion conversion) {
```
petersteneteg: Why not add a getValueTexel texture sample function. to go directly from texture to value? Is there any reason to do `mapFromNormalizedGLToValue(getNormalizedTexel(..),..)`? Does not `ImageParameters` and `GLFormatConversion` have very similar parameters? At least for the input part?

modules/opengl/glsl/utils/structs.glsl
```diff
@@ -147,4 +147,19 @@ struct StipplingParameters {
     float worldScale;
 };
 
+// Conversion factors for conversions from [0,1] obtained from getNormalizedTexel/Voxel() to 
+// value space and to the OpenGL output range
+// see glformatsutil.h
+struct GLFormatConversion {
```
petersteneteg: GL seems very redundant in the name here. 

I would also like the member to have more symmetric naming.

modules/opengl/include/modules/opengl/util/glformatutils.h
```diff
@@ -0,0 +1,157 @@
...
+    double outputScaling = 1.0;
+    double outputOffset = 0.0;
+};
+
+struct IVW_MODULE_OPENGL_API GLFormatRenormalization {
+    double formatScaling = 1.0;
+    double signedFormatScaling = 1.0;
+    double formatOffset = 0.0;
+    double signedFormatOffset = 0.0;
+};
```
petersteneteg: Do we need both of these? Can we not merge then and make the names more consistent? 

```
struct IVW_MODULE_OPENGL_API ConversionMap {
    double scale;
    double offset;
};

struct IVW_MODULE_OPENGL_API FormatConversion {
    ConversionMap texToNormalized;
    ConversionMap texToSignNormalized;
    ConversionMap texToValue;

    ConversionMap valueToTex; 
};
```

petersteneteg (changes requested):

tests/integrationtests/glformatconversion-test.cpp
```diff
@@ -0,0 +1,662 @@
...
+
+#include <ranges>
+#include <numeric>
+
+#include <warn/push>
+#include <warn/ignore/all>
+#include <gtest/gtest.h>
+#include <warn/pop>
+
+#pragma optimize("", off)
```
petersteneteg: Might remove this I guess?

tests/integrationtests/glformatconversion-test.cpp
```diff
@@ -0,0 +1,662 @@
...
+
+    // for (auto&& [index, expected] : std::views::zip(std::views::iota(0), expectedValues)) {
+    //     const auto result = layerRam->getAsDouble(size2_t{index, 0});
+    //     if constexpr (std::is_floating_point_v<U>) {
+    //         EXPECT_DOUBLE_EQ(result, expected);
+    //     } else {
+    //         const auto res = static_cast<U>(result);
+    //         EXPECT_EQ(res, expected);
+    //     }
+    // }
```
petersteneteg:
```suggestion
```

petersteneteg (dismissed):
NIce work! that is a very comprehensive refactoring!

tests/integrationtests/glformatconversion-test.cpp
martinfalk: sure

petersteneteg (dismissed):

petersteneteg (dismissed):

petersteneteg (dismissed):

petersteneteg:
let me just double check on my mac before merging.

petersteneteg:
> let me just double check on my mac before merging.

Seems to work fine!

petersteneteg (approved):

PR #1803, opened 2025-11-20
Feature/pythonbindings

martinfalk:
Changes proposed in this PR:
 * expose `SpatialEntity` and `StructuredGridEntity<N>` to Python
 * expose properties of the `Base` module

### DIFF
--- a/modules/python3/bindings/src/pymesh.cpp
+++ b/modules/python3/bindings/src/pymesh.cpp
@@ -93,7 +93,7 @@ void exposeMesh(pybind11::module& m) {
         return list;
     };
 
-    py::classh<Mesh, SpatialEntity>(m, "Mesh")
+    py::classh<Mesh, SpatialEntity>(m, "Mesh", py::multiple_inheritance{})
         .def(py::init<>())
         .def(py::init<DrawType, ConnectivityType>(), py::arg("dt") = DrawType::Points,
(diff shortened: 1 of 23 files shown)


modules/python3/bindings/src/pymesh.cpp
```diff
@@ -93,18 +93,13 @@ void exposeMesh(pybind11::module& m) {
         return list;
     };
 
-    py::classh<Mesh>(m, "Mesh")
+    py::classh<Mesh, SpatialEntity>(m, "Mesh")
```
petersteneteg: add the multiple_inheritance here to

modules/python3/bindings/src/pyproperties.cpp
```diff
@@ -88,6 +88,10 @@ void exposeProperties(py::module& m) {
         .value("All", PropertySerializationMode::All)
         .value("Nothing", PropertySerializationMode::None);
 
+    py::enum_<util::OverwriteState>(m, "OverwriteState")
```
petersteneteg: might use native_enum instead, 
see https://pybind11.readthedocs.io/en/stable/classes.html#enumerations-and-internal-types

petersteneteg (changes requested):
Nice work getting all that into python!

petersteneteg (dismissed):

petersteneteg (approved):

PR #1807, opened 2025-11-25
Feature/instancing performance

martinfalk:
Changes proposed in this PR:
 * removed bounding box calculations in CameraProperty::zoom() as those can be very expensive, especially when used with the `InstanceRenderer` processor
 * fixed transform feedback OpenGL performance warning in `InstanceRenderer`

### DIFF
--- a/modules/basegl/src/processors/instancerenderer.cpp
+++ b/modules/basegl/src/processors/instancerenderer.cpp
@@ -794,7 +794,7 @@ std::optional<mat4> InstanceRenderer::render(bool enableBoundingBoxCalc) {
     if (enableBoundingBoxCalc) {
         const auto nVertices = nInstances * numberOfVertices(meshGL);
         BufferObject feedbackBuffer{nVertices * sizeof(vec4), GLFormats::get(DataVec4Float32::id()),
-                                    GL_STATIC_READ, GL_TRANSFORM_FEEDBACK_BUFFER};
+                                    GL_STATIC_COPY, GL_TRANSFORM_FEEDBACK_BUFFER};
         size_t byteOffset = 0;
         forEach(meshGL, shader_, vecPorts_, nInstances,
                 [&](size_t size, Mesh::MeshInfo mi, const auto& draw) {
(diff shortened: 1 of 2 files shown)


modules/basegl/src/processors/instancerenderer.cpp
```diff
@@ -794,7 +794,7 @@ std::optional<mat4> InstanceRenderer::render(bool enableBoundingBoxCalc) {
     if (enableBoundingBoxCalc) {
         const auto nVertices = nInstances * numberOfVertices(meshGL);
         BufferObject feedbackBuffer{nVertices * sizeof(vec4), GLFormats::get(DataVec4Float32::id()),
-                                    GL_STATIC_READ, GL_TRANSFORM_FEEDBACK_BUFFER};
+                                    GL_STATIC_COPY, GL_TRANSFORM_FEEDBACK_BUFFER};
```
petersteneteg, line 797: This makes no sense, explain why? the docs align exactly to the read case? If anything it should be dynamic_read.

martinfalk: My OpenGL driver has a different opinion.

> Buffer performance warning: Buffer object 24 (bound to GL_TRANSFORM_FEEDBACK_BUFFER_NV (0), and GL_TRANSFORM_FEEDBACK_BUFFER_NV, usage hint is GL_STATIC_READ) is being copied/moved from VIDEO memory to HOST memory.
> [Severity: Medium, Type: Performance, Source: Api, Id: 131186, Context:  DefaultContext (000001F1BB82A940)]

martinfalk: It might be connected to `glMapBuffer()` and accessing the data from the CPU.

petersteneteg: but that should be exactly what READ mean? 
> READ
> The data store contents are modified by reading data from the GL, and used to return that data when queried by the application.

https://docs.gl/es3/glBufferData#:~:text=The%20data%20store%20contents%20are%20modified%20by%20reading%20data%20from%20the%20GL%2C%20and%20used%20to%20return%20that%20data%20when%20queried%20by%20the%20application.

Or how should I interpret that?

>COPY
> The data store contents are modified by reading data from the GL, and used as the source for GL drawing and image specification commands.

We are definitely __not__ using it as a source for GL drawing.

PR #1827, opened 2025-12-16
Feature/convexhull

martinfalk:
Changes proposed in this PR:
 * refactored convex hull algorithm to accept spans
 * added allocation-free implementation

### DIFF
--- a/modules/base/include/modules/base/algorithm/convexhull.h
+++ b/modules/base/include/modules/base/algorithm/convexhull.h
@@ -129,7 +129,7 @@ template <RangeOfVec<2> C>
 std::vector<std::ranges::range_value_t<C>> convexHull(const C& points) {
     using E = std::ranges::range_value_t<C>;
     // sort points according to x coordinate, if equal chose lower y coordinate
-    std::vector<E> p = points;
+    std::vector<E> p(points.begin(), points.end());
     auto compare = [](const E& a, const E& b) {
         return (a.x < b.x) || ((a.x == b.x) && (a.y < b.y));
     };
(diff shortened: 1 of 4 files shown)


modules/base/include/modules/base/algorithm/convexhull.h
```diff
@@ -167,14 +174,70 @@ std::vector<T> convexHull(const std::vector<T>& points) {
...
+            // -> remove last hull point
+            --k;
+        }
+        hull[k++] = p[i];
+    }
+    // build upper hull
+    const std::size_t lastIndex = k + 1;
+    for (int i = static_cast<int>(N) - 2; i >= 0; --i) {
+        while ((k >= lastIndex) &&
+               (cross2D(E(hull[k - 1] - hull[k - 2]), E(p[i] - hull[k - 2])) <= 0)) {
```
petersteneteg:
```suggestion
               (cross2D(hull[k - 1] - hull[k - 2], p[i] - hull[k - 2]) <= 0)) {
```

modules/base/include/modules/base/algorithm/convexhull.h
```diff
@@ -167,14 +174,70 @@ std::vector<T> convexHull(const std::vector<T>& points) {
...
+ * This version avoids unnecessary allocations by passing in the container @p hull where the result
+ * is written. Note that the hull is of size 2*N.
+ *
+ * @param points   set of 2D points
+ * @param hull     resulting convex hull of the input points, holding twice the number of points in
+ *                 @p points
+ * @return span of the actual hull pointing into @p hull
+ */
+template <size_t N, RangeOfVec<2> C, typename E = std::ranges::range_value_t<C>>
+std::span<const E> convexHull(const C& points,
```
petersteneteg: We have to ensure that points.size == N at compile time. 

I would probably try to keep is simple and do
```
template <size_t Size, typename T>
std::span<const glm::vec<2, T>> convexHull(
    std::span<const glm::vec<2, T>, Size> points,
    std::span<glm::vec<2, T>, 2 * Size> hull) {
    
...
}
```

can we actually get to the case where we add each point more then 1 time? Like what happens when we pass in points that are all on a line?

petersteneteg: does not seem to happen
https://array.godbolt.org/z/x7Ee6381c

petersteneteg (changes requested):

petersteneteg (approved):

PR #1878, opened 2026-02-20
Feature/pythonbindings

martinfalk:
Changes proposed in this PR:
 * exposed more enums to Python
 * extended `Unit` constructor in line with C++ constructor

### DIFF
--- a/modules/python3/bindings/src/pybuffer.cpp
+++ b/modules/python3/bindings/src/pybuffer.cpp
@@ -113,6 +113,11 @@ void exposeBuffer(pybind11::module& m) {
         .value("Data", BufferTarget::Data)
         .value("Index", BufferTarget::Index);
 
+    py::enum_<CartesianCoordinateAxis>(m, "CartesianCoordinateAxis")
+        .value("X", CartesianCoordinateAxis::X)
+        .value("Y", CartesianCoordinateAxis::Y)
+        .value("Z", CartesianCoordinateAxis::Z);
+
     py::enum_<DrawType>(m, "DrawType")
         .value("NotSpecified", DrawType::NotSpecified)
         .value("Points", DrawType::Points)
--- a/modules/python3/bindings/src/pydatamapper.cpp
+++ b/modules/python3/bindings/src/pydatamapper.cpp
@@ -42,7 +42,7 @@
 #include <modules/python3/opaquetypes.h>
 #include <modules/python3/polymorphictypehooks.h>
 
-#include <fmt/core.h>
+#include <fmt/format.h>
 
 namespace py = pybind11;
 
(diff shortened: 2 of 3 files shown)


modules/python3/bindings/src/pybuffer.cpp
```diff
@@ -113,6 +113,17 @@ void exposeBuffer(pybind11::module& m) {
         .value("Data", BufferTarget::Data)
         .value("Index", BufferTarget::Index);
 
+    py::enum_<CoordinatePlane>(m, "CoordinatePlane")
+        .value("XY", CoordinatePlane::XY)
+        .value("XZ", CoordinatePlane::XZ)
+        .value("YZ", CoordinatePlane::YZ)
```
petersteneteg: Why do we have both YZ and ZY? how do they differ? Are the normal flipped? And If so, why don't we have that for the other 2?

modules/python3/bindings/src/pydatamapper.cpp
```diff
@@ -52,10 +52,14 @@ void exposeDataMapper(py::module& m) {
 
     py::classh<Unit>(m, "Unit")
         .def(py::init([](std::string unit) { return units::unit_from_string(std::move(unit)); }))
-        .def("to_string",
-             [](const Unit& unit, std::string_view format = "{}") {
-                 return fmt::format(fmt::runtime(format), unit);
-             })
+        .def(py::init([](double multiplier, const Unit& unit) { return Unit(multiplier, unit); }),
+             py::arg("multiplier"), py::arg("unit"))
+        .def(
```
petersteneteg, line 93: can we make this work with python format strings?

petersteneteg (changes requested):

modules/python3/bindings/src/pybuffer.cpp
martinfalk: will remove `CoordinatePlane`

petersteneteg (dismissed):

petersteneteg (approved):

modules/python3/bindings/src/pydatamapper.cpp
```diff
@@ -42,21 +42,63 @@
...
-        .def("__repr__", [](const Unit& unit) { return fmt::format("{}", unit); });
+        .def(py::init([](double multiplier, const Unit& unit) { return Unit(multiplier, unit); }),
+             py::arg("multiplier"), py::arg("unit"))
+        .def(
+            "__format__",
+            [](const Unit& unit, std::string_view spec) {
+                const auto format = fmt::format("{{:{}}}", spec);
+                return fmt::format(fmt::runtime(format), unit);
+            },
+            py::arg("format") = "{}", py::doc(formatDoc.data()))
```
petersteneteg: can't use string_view.data() here

petersteneteg (changes requested):

petersteneteg (approved):

PR #1916, opened 2026-03-17
Feature/axisrangeevent

martinfalk:
Changes proposed in this PR:
 * refactored box selection, particularly `BoxSelectionInteractionHandler` and associated callbacks
 * added `AxisRangeEvent` to propagate range selection
 * selection & filtering support in `DataFrameToMesh` processor


include/inviwo/core/properties/eventproperty.h
```diff
@@ -48,7 +48,23 @@ namespace inviwo {
  */
 class IVW_CORE_API EventProperty : public Property {
 public:
-    using Action = std::function<void(Event*)>;
+    enum class State : std::uint8_t { Active, Finished };
```
petersteneteg: Do we want to track Start here to? to make it more similar to picking etc?

modules/dataframe/src/processors/dataframetomesh.cpp
```diff
@@ -337,4 +351,57 @@ void DataFrameToMesh::picking(PickingEvent* event) {
     }
 }
 
+void DataFrameToMesh::axisInteraction(AxisRangeEvent* event) {
+    const auto data = dataFrame_.getData();
+
+    std::array<std::shared_ptr<const Column>, 3> axisCols;
```
petersteneteg, line 357: pos.sources.size() == 3?

modules/plottinggl/src/processors/scatterplotprocessor.cpp
```diff
@@ -107,35 +107,38 @@ ScatterPlotProcessor::ScatterPlotProcessor()
             if (auto data = dataFramePort_.getData()) {
                 BitSet indices;
                 auto iCol = data->getIndexColumn();
-                auto& indexCol = iCol->getTypedBuffer()->getRAMRepresentation()->getDataContainer();
+                const auto& indexCol = iCol->getTypedBuffer()->getRAMRepresentation()->getDataContainer();
                 for (auto idx : highlighted) {
                     indices.add(indexCol[idx]);
                 }
                 brushingPort_.highlight(indices);
```
petersteneteg, line 114: Will the the port action not trigger an invalidation anyway?

petersteneteg (dismissed):
Nice work!

modules/plottinggl/src/processors/scatterplotprocessor.cpp
martinfalk: Not if the bnl state does not change, which might be the case.

martinfalk: And the invalidation is needed to redraw/remove the selection rectangle.

modules/dataframe/src/processors/dataframetomesh.cpp
martinfalk: yes, as defined here:
https://github.com/inviwo/inviwo/blob/402e43c46eeeb1051e9d4cf434b5235409fc4e3a/modules/dataframe/include/inviwo/dataframe/processors/dataframetomesh.h#L123-L128

Probably better for `Type` to also store `N`, so it can be accessed as `SpatialType::size` or similar.

petersteneteg (approved):

PR #1940, opened 2026-04-14
Feature/plotting

martinfalk:
Fixes 
 * division by zero in PlotCamera

Changes proposed in this PR:
 * optional ports in `OrthographicAxis2D` processor to also connect Volume and Layer ports

This has been tested on: Windows, Linux


modules/plottinggl/src/processors/orthographicaxis2d.cpp
```diff
@@ -32,13 +32,20 @@
...
 #include <modules/opengl/image/imagegl.h>
 
 #include <inviwo/core/interaction/events/axisrangeevent.h>
 
+#include <algorithm>
+#include <initializer_list>
+
 namespace inviwo {
 
+#pragma optimize("", off)
```
petersteneteg: Remove
```suggestion
```

modules/plottinggl/src/processors/orthographicaxis2d.cpp
```diff
@@ -135,6 +167,24 @@ OrthographicAxis2D::OrthographicAxis2D()
 }
 
 void OrthographicAxis2D::process() {
+    auto [w2m, axes] = [this]() -> std::pair<mat4, std::array<const Axis*, 2>> {
```
petersteneteg: should this not return a dmat4?

modules/plottinggl/src/processors/orthographicaxis2d.cpp
```diff
@@ -175,20 +225,15 @@ void OrthographicAxis2D::process() {
         2.0f * vec3{dvec2{xStart.x, yStart.y} / dvec2{dims}, 0.5} - vec3{1.0, 1.0, 1.0};
     const auto end = 2.0f * vec3{dvec2{xEnd.x, yEnd.y} / dvec2{dims}, 0.5} - vec3{1.0, 1.0, 1.0};
 
-    const auto data = mesh_.getData();
-
-    auto w2m = data->getCoordinateTransformer().getWorldToModelMatrix();
     const auto wStart =
         w2m * vec4{camera_.get().getWorldPosFromNormalizedDeviceCoords(start), 1.0f};
```
petersteneteg: use doubles here too

modules/plottinggl/src/processors/orthographicaxis2d.cpp
```diff
@@ -175,20 +225,15 @@ void OrthographicAxis2D::process() {
...
     const auto wEnd = w2m * vec4{camera_.get().getWorldPosFromNormalizedDeviceCoords(end), 1.0f};
 
     axis1_.setRange(dvec2{wStart.x, wEnd.x});
     axis2_.setRange(dvec2{wStart.y, wEnd.y});
 
-    axis1_.captionSettings_.title_.set(
-        fmt::format("{}{: [}", data->getAxis(0)->name, data->getAxis(0)->unit));
-    axis2_.captionSettings_.title_.set(
-        fmt::format("{}{: [}", data->getAxis(1)->name, data->getAxis(1)->unit));
+    axis1_.captionSettings_.title_.set(fmt::format("{}{: [}", axes[0]->name, axes[0]->unit));
```
petersteneteg, line 233: This could use the same stuff as in Axis3DProcessorHelper, but maybe that is a different PR

petersteneteg (changes requested):

modules/plottinggl/src/processors/orthographicaxis2d.cpp
martinfalk: Yes, good idea. Though this will need some work since the Axis2DProcessorHelper always uses world coordinates and not screen coords.

petersteneteg (dismissed):

modules/plottinggl/src/processors/orthographicaxis2d.cpp
petersteneteg: Was mostly thinking the axis and unit stuff we added. Don't think we need to use the Axis2DProcessorHelper

PR #1975, opened 2026-05-25
Feature/networkeval

martinfalk:
Fixes 
 * infinite network evaluations caused by exceptions thrown in process()

Changes proposed in this PR:
 * set processors to valid after handling the exceptions


petersteneteg (changes requested):
Check the unit tests and if possible, add some check that uses this new code?

PR #1977, opened 2026-05-25
Feature/plotting fixes

martinfalk:
Changes proposed in this PR:
 * enable blending in `OrthoGraphicAxis2D` to get correct blending for the plot contents and background
 * parallel coords selection behaving similar to scatter plot (additive selection with <Shift>)

### DIFF
--- a/modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
+++ b/modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
@@ -709,7 +709,7 @@ void ParallelCoordinates::drawLines(size2_t size) {
 }
 
 void ParallelCoordinates::linePicked(PickingEvent* p) {
-    const KeyModifier appendModifier = KeyModifier::Shift;
+    const KeyModifier appendModifier = KeyModifier::Control;
 
     if (auto df = dataFrame_.getData()) {
         // Show tooltip about current line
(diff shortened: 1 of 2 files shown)


modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
```diff
@@ -709,6 +709,8 @@ void ParallelCoordinates::drawLines(size2_t size) {
 }
 
 void ParallelCoordinates::linePicked(PickingEvent* p) {
+    const KeyModifier appendModifier = KeyModifier::Shift;
```
petersteneteg: Should this not be Ctrl? shift is usually for selecting a range?

petersteneteg (dismissed):

modules/plottinggl/src/processors/parallelcoordinates/parallelcoordinates.cpp
martinfalk: This is in line with `BoxSelection`. The reason was that `OrthographicAxis2D` uses left drag for interacting with the axes and Ctrl+left for box selection. Hence, Shift for adding.

https://github.com/inviwo/inviwo/blob/dc97abf08a1b23ac365bc9bb47c0515f218d99c5/modules/plotting/src/interaction/boxselection.cpp#L85-L88

petersteneteg (approved):

PR #1980, opened 2026-05-27
Feature/cropwidget

martinfalk:
Fixes
 * several issues with `CropWidget` like oblique volume bases and rendering artifacts

Changes proposed in this PR:
 * alignment and orientation of interaction handles based on `Axis3DProcessorHelper`. Handles can now be overlaid onto volume axes.
 * fixed invalidation problem by removing "output" ranges

**Before / After**
cropimage


include/inviwo/core/util/glm.h
```diff
@@ -279,6 +279,28 @@ glm::vec<3, T> transformNormal(const glm::mat<4, 4, T>& modelViewMatrix,
     return glm::transpose(glm::inverse(Matrix<3, T>{modelViewMatrix})) * normal;
 }
 
+template <typename T>
+glm::mat<4, 4, T> rotate(glm::vec<3, T> from, glm::vec<3, T> to) {
```
petersteneteg: it this different from glm::rotation?

petersteneteg (dismissed):

include/inviwo/core/util/glm.h
martinfalk: Ah, only looked for `glm::rotate` and missed `glm::rotation`, which returns a quaternion. Will revert the change.

petersteneteg: mmm finding the "right" function in glm is hard...

PR #1984, opened 2026-05-29
Feature/canvas fixes

### LINKED ISSUE #1983: [BUG]: Fullscreen canvas issues
`Canvas With Properties` has a `BoolProperty` to toggle full-screen mode. However, when the network editor and canvas are shown on the same screen, there is no way to leave the full-screen mode once entered. The regular `Canvas` provides a shortcut (<shift>+f) for toggling full screen.
 
Additionally, a full-screen canvas is always shown on top which prevents context menus from being visible. This also affects context menus of properties in the property pane of  `Canvas With Properties`. The context menu is only shown directly after the canvas widget regained focus.

The canvas also turns black on the next update after the widget regained focus, for example camera movement.

### Inviwo version

master

martinfalk:
Fixes #1983 

Changes proposed in this PR:
 * added event properties toggling full-screen mode to `CanvasWithProperties`, partial fix for #1983
 * added event property to leave full-screen mode with <Esc>
 * issue where `HiddenCanvas` initially propagates an incorrect window size

This has been tested on: Windows, Qt 6.10.3

### DIFF
--- a/modules/openglqt/src/processors/canvaswithpropertiesprocessor.cpp
+++ b/modules/openglqt/src/processors/canvaswithpropertiesprocessor.cpp
@@ -253,7 +253,12 @@ void CanvasWithPropertiesProcessor::setProcessorWidget(
 }
 
 void CanvasWithPropertiesProcessor::propagateEvent(Event* event, Outport*) {
+    if (event->hasVisitedProcessor(this)) return;
     event->markAsVisited(this);
+    
+    invokeEvent(event);
+    if (event->hasBeenUsed()) return;
+
     inport_.propagateEvent(event);
 }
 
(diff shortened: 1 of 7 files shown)


modules/openglqt/src/processors/canvaswithpropertiesprocessor.cpp
```diff
@@ -127,10 +127,30 @@ CanvasWithPropertiesProcessor::CanvasWithPropertiesProcessor()
...
+                       "Full Screen",
+                       "Shortcut to toggle full-screen mode"_help,
+                       [this](Event* event) {
+                           event->markAsUsed();
+                           fullScreen_.set(!fullScreen_);
+                       },
+                       IvwKey::F,
+                       KeyState::Press,
+                       KeyModifier::Shift}
+    , leaveFullScreenEvent_{"leaveFullScreenEvent",
```
petersteneteg: Why is this needed you can just use the fullscreen event again to toggle the state?

petersteneteg (approved):

petersteneteg (approved):

PR #1989, opened 2026-06-02
Feature/measuredistance

martinfalk:
Changes proposed in this PR:
 * added a processor to perform interactive length measurements and annotations with the mouse in world space

https://github.com/user-attachments/assets/e7cbe35f-76a4-46cd-929e-eaba8568fe30


modules/plottinggl/include/modules/plottinggl/processors/measuredistance3d.h
```diff
@@ -0,0 +1,114 @@
...
+    AxisProperty measurementAxis_;
+    CameraProperty camera_;
+
+    EventProperty placeLocatorEvent_;
+    EventProperty cancelMeasurementEvent_;
+
+    AxisRenderer3D axisRenderer_;
+    algorithm::LineRenderer lineRenderer_;
+
+    // MeasurementState measurementState_;
```
petersteneteg: Remove?

modules/plottinggl/include/modules/plottinggl/processors/measuredistance3d.h
```diff
@@ -0,0 +1,114 @@
...
+                              ConnectivityType::None,
+                              {
+                                  {{-1.0f, 0.0f, 0.0f}},
+                                  {{1.0f, 0.0f, 0.0f}},
+                                  {{0.0f, -1.0f, 0.0f}},
+                                  {{0.0f, 1.0f, 0.0f}},
+                                  {{0.0f, 0.0f, -1.0f}},
+                                  {{0.0f, 0.0f, 1.0f}},
+                              },
+                              {0, 1, 2, 3, 4, 5}};
```
petersteneteg: I would keep the initialization to the cpp, to avoid changing the header etc...

modules/plottinggl/src/processors/measuredistance3d.cpp
```diff
@@ -0,0 +1,455 @@
...
+    isReady_.setUpdate([this]() -> ProcessorStatus {
+        if (auto connectedPorts = std::ranges::fold_left(
+                std::initializer_list<const Inport*>{&mesh_, &layer_, &volume_}, 0,
+                [](int count, const Inport* port) {
+                    if (port->isConnected()) {
+                        return count + 1;
+                    }
+                    return count;
+                });
+            connectedPorts > 1) {
```
petersteneteg, line 227: Nice, might even use a transform and std::plus :) 
```suggestion
        if (auto connectedPorts = std::ranges::fold_left(
                std::to_array<const Inport*>{&mesh_, &layer_, &volume_} | std::ranges::transform{[](auto* p){return p->isConnected()?1:0;}}, 0,
                std::plus<>{});
            connectedPorts > 1) {
```
think to_array should work here to? I prefer to avoid initializer_list since that one has some quite odd properties....

modules/plottinggl/src/processors/measuredistance3d.cpp
```diff
@@ -0,0 +1,455 @@
...
+                   util::ordinalSymmetricVector(dvec3{0.0})
+                       .set("First location of the measurement"_help)},
+                  {"endPosition", "End Position",
+                   util::ordinalSymmetricVector(dvec3{0.0})
+                       .set("Second location of the measurement"_help)}}}
+    , distance_{"distance", "Distance",
+                util::ordinalLength(0.0)
+                    .setInc(0.0001)
+                    .set(InvalidationLevel::Valid)
+                    .set(ReadOnly::Yes)}
```
petersteneteg, line 173: might disable serialization for distance too

modules/plottinggl/src/processors/measuredistance3d.cpp
```diff
@@ -0,0 +1,455 @@
...
+                mousePosNdc.z =
+                    camera_.getNormalizedDeviceFromNormalizedScreenAtFocusPointDepth({0.0, 0.0}).z;
+            }
+            const auto mousePos{camera_.getWorldPosFromNormalizedDeviceCoords(mousePosNdc)};
+
+            switch (mouseEvent->state()) {
+                case MouseState::Press:
+                    break;
+                case MouseState::Move:
+                    sm->sm.process_event(MoveLocator{.pos = mousePos});
```
petersteneteg, line 430: Seems like using the SM worked well! the code does not have any of the usual nasty if chains that we usually get for this kind of event based code...

petersteneteg (commented):
Nice work! Very clean

petersteneteg (approved):

PR #1999, opened 2026-06-12
Feature/heightfield

martinfalk:
Fixes
 * various issues with the `HeightField` processor like two-sided illum, data range, etc.
 * flipped texture coords of sphere mesh created by the `Mesh Creator` processor 

Changes proposed in this PR:
 * `HeightField` uses a Layer as height input, utilizing its value range
 * added a shader property to directly modify the heightfield mapping in the vertex shader


modules/basegl/include/modules/basegl/processors/heightfieldprocessor.h
```diff
@@ -60,29 +64,30 @@ class IVW_MODULE_BASEGL_API HeightFieldProcessor : public Processor {
...
+    static const ProcessorInfo processorInfo_;
+
 private:
     MeshFlatMultiInport inport_;
-    ImageInport inportHeightfield_;  //!< inport for the 2D heightfield texture
-    ImageInport inportTexture_;      //!< inport for the 2D color texture (optional)
-    ImageInport inportNormalMap_;    //!< inport for the 2D normal map texture (optional)
+    LayerInport inportHeightfield_;
+    ImageInport inportTexture_;
+    ImageInport inportNormalMap_;
```
petersteneteg: Why not use layers here to?

petersteneteg (dismissed):
Nice, we could add the "Sphere" as an additional example?

petersteneteg (approved):

petersteneteg (approved):

PR #2012, opened 2026-06-25
Feature/splatting fixes

martinfalk:
Fixes
 * splatting assumes world positions and does not apply model/world transformations of the input Mesh

Changes proposed in this PR:
 * Mesh positions can be transformed prior splatting to consider model and world transformations
 * Expose the mesh bounding box in data, model, and world space in the `MeshInformationProperty`
 * added regression test for Mesh splatting


petersteneteg (dismissed):
Nice fix, forgot about that transform. And good with a regression test!

petersteneteg (dismissed):

petersteneteg (approved):

PR #2013, opened 2026-06-25
Feature/inputselector

martinfalk:
Changes proposed in this PR:
 * optional "empty inport" option in `Input Selector` to toggle between forwarding the selected inport or no data.
 * adjusted mesh rendering processors to deal with the cases where the Mesh inport has no data.

This allows for more flexible networks where it is possible to toggle parts of the visualization, like the volume bounding box, without breaking the rest of the network.

image

### DIFF
--- a/modules/basegl/src/processors/layerrenderer.cpp
+++ b/modules/basegl/src/processors/layerrenderer.cpp
@@ -63,8 +63,9 @@ LayerRenderer::LayerRenderer()
     , shader_{"layerrendering.vert", "layerrendering.frag", Shader::Build::No}
     , mesh_{DrawType::Triangles, ConnectivityType::Strip} {
 
-    background_.setOptional(true);
-    addPorts(inport_, background_, outport_);
+    addPort(inport_).setOptional(true);
+    addPort(background_).setOptional(true);
+    addPort(outport_);
     addProperties(camera_, trackball_);
 
     auto verticesBuffer =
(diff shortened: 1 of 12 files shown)


modules/basegl/src/processors/layerrenderer.cpp
```diff
@@ -87,17 +89,19 @@ void LayerRenderer::process() {
...
 
     utilgl::setUniforms(shader_, camera_);
     MeshDrawerGL::DrawObject drawer{mesh_.getRepresentation<MeshGL>(), mesh_.getDefaultMeshInfo()};
 
-    TextureUnit unit;
+    const TextureUnit unit;
     shader_.setUniform("colorTex", unit);
 
     for (const auto& layer : inport_) {
+        if (!layer) continue;
```
petersteneteg: do you get nullptrs here? thought the iterators would remove those?

modules/basegl/src/processors/linerendererprocessor.cpp
```diff
@@ -104,6 +104,8 @@ void LineRendererProcessor::process() {
 
 void LineRendererProcessor::drawMeshes() {
     for (const auto& mesh : inport_) {
+        if (!mesh) continue;
```
petersteneteg: Same here? is this check needed?

petersteneteg (dismissed):

modules/basegl/src/processors/layerrenderer.cpp
martinfalk: The port iterators iterate over the connected outports of a `LayerMultiInport`, i.e. `DataInport<Layer, 0>`. There is no check whether the connected outport is valid/has data.

https://github.com/inviwo/inviwo/blob/e1b54fc4fd51c42baa9059e6ffa87e5ce1f7b506/include/inviwo/core/ports/inportiterable.h#L307-L312

PR #2021, opened 2026-07-08
Feature/networkeditor fixes

martinfalk:
Fixes 
 * ensure that the network search widget no longer moves with the graphics scene
 * changed modifier to `<ctrl>` for adding multiple processor properties to the `PropertyListWidget`, `<shift>` and `<alt>` are already used for selecting predecessors and successors

Changes proposed in this PR:
 * changed network search highlight to distinguish it from a regular selection

**Network Search before/after**
imageimage

This has been tested on: Windows, Qt 6.10.3

### DIFF
--- a/src/qt/editor/processorgraphicsitem.cpp
+++ b/src/qt/editor/processorgraphicsitem.cpp
@@ -56,6 +56,7 @@
 #include <QApplication>
 #include <QGraphicsDropShadowEffect>
 #include <QPainter>
+#include <QGradient>
 #include <QStyleOptionGraphicsItem>
 #include <QVector2D>
 #include <QTextCursor>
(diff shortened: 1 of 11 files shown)


src/qt/editor/processorgraphicsitem.cpp
```diff
@@ -527,13 +538,13 @@ void ProcessorGraphicsItem::updateWidgets() {
     if (isSelected()) {
         setZValue(depth::processorSelected);
         if (!highlight_) {
```
petersteneteg: This test should not be needed now?

petersteneteg (changes requested):
There is also a use of setHighlight in networkautomation.cpp

Not sure about the gradient effect, have you tried with a red border? Maybe I just have to get used to that.

martinfalk:
> Not sure about the gradient effect, have you tried with a red border? Maybe I just have to get used to that.

Red border is too subtle to notice.

image

petersteneteg (approved):

petersteneteg (changes requested):
need to fix updateProcessorTarget in networkautomation.cpp also

petersteneteg (approved):

PR #2054, opened 2026-09-08
Feature/tfdialog

martinfalk:
Fixes several TF dialog issues
 * unhandled exceptions for invalid positions updates
 * read-only state when no points are selected

Changes proposed in this PR:
 * TF dialog using `NumberWidget`

### DIFF
--- a/modules/qtwidgets/src/tf/tfpropertydialog.cpp
+++ b/modules/qtwidgets/src/tf/tfpropertydialog.cpp
@@ -46,6 +46,7 @@
 #include <inviwo/core/properties/propertyowner.h>
 #include <inviwo/core/properties/transferfunctionproperty.h>
 #include <inviwo/core/util/glmvec.h>
+// #include <inviwo/core/util/logcentral.h>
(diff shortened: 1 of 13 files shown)


modules/qtwidgets/src/tf/tfpropertydialog.cpp
```diff
@@ -46,6 +46,7 @@
 #include <inviwo/core/properties/propertyowner.h>
 #include <inviwo/core/properties/transferfunctionproperty.h>
 #include <inviwo/core/util/glmvec.h>
+// #include <inviwo/core/util/logcentral.h>
```
petersteneteg: remove?

petersteneteg (dismissed):

modules/qtwidgets/src/tf/tfpropertydialog.cpp
martinfalk: fixed, that include should actually be in there for the guarded exception

petersteneteg (approved):

